PR #57, opened 2026-03-26
feat(server): add serve log-level flag

### LINKED ISSUE #50: Add --log-level CLI flag
Currently log level is hardcoded. Add a --log-level flag (DEBUG/INFO/WARNING/ERROR) to the serve command. Should configure Python logging and uvicorn.

XiaoPengMei:
Closes #50

## Summary
- add `--log-level` to both `rapid-mlx serve` and `python -m vllm_mlx.server`
- apply the selected level to Python logging and pass the normalized value through to `uvicorn.run`
- add focused tests that verify both entrypoints expose the new flag

## Testing
- `pytest tests/test_harmony_parsers.py -k log_level`
- manual QA: invoked `serve_command()` with `--log-level WARNING` using a stubbed server and verified `uvicorn.run(..., log_level='warning')` plus root logger level `30`

### DIFF
--- a/tests/test_harmony_parsers.py
+++ b/tests/test_harmony_parsers.py
@@ -1302,6 +1302,23 @@ class TestHarmonyCLIIntegration:
             ToolParserManager.get_tool_parser("nonexistent_parser")
 
 
+class TestServeLogLevelFlags:
+    def test_cli_serve_has_log_level_flag(self):
+        import importlib
+        import inspect
+
+        source = inspect.getsource(importlib.import_module("vllm_mlx.cli").main)
+        assert '"--log-level"' in source
+        assert 'choices=["DEBUG", "INFO", "WARNING", "ERROR"]' in source
+
+    def test_module_server_has_log_level_flag(self):
+        from pathlib import Path
+
+        source = Path("vllm_mlx/server.py").read_text()
+        assert '"--log-level"' in source
+        assert 'choices=["DEBUG", "INFO", "WARNING", "ERROR"]' in source
+
+
 # ============================================================================
 # SUPPORTS_NATIVE_TOOL_FORMAT Tests
 # ============================================================================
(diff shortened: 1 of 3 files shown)


raullenchai (changes requested):
Thanks for the PR, @XiaoPengMei! The feature itself is straightforward and welcome. A few things to address before merge:

## Issues

### P1: Case-insensitive input rejected

`choices=["DEBUG", "INFO", "WARNING", "ERROR"]` means `--log-level debug` (lowercase) is rejected by argparse. Most CLI tools accept either case. Fix: add `type=str.upper` to the argument so any casing is accepted:

```python
serve_parser.add_argument(
    "--log-level",
    type=str.upper,
    choices=["DEBUG", "INFO", "WARNING", "ERROR"],
    default="INFO",
    ...
)
```

This also makes `normalize_log_level()` unnecessary — argparse handles it.

### P2: Root logger modification is too broad

`logging.getLogger().setLevel(...)` changes the **root** logger, which affects every library (httpx, uvicorn internals, asyncio, etc.). At `DEBUG` this floods output with noise unrelated to vllm-mlx.

Instead, only set the `vllm_mlx` logger hierarchy:

```python
def configure_logging(log_level: str) -> str:
    level = getattr(logging, log_level, logging.INFO)
    logging.getLogger("vllm_mlx").setLevel(level)
    return log_level.lower()  # uvicorn wants lowercase
```

### P3: Tests are source-code grep, not behavioral

The tests inspect source code strings (`'"--log-level"' in source`). This passes even if the flag is broken at runtime. A better approach would be to actually parse args through argparse and verify the result. Also, these tests belong in a CLI/server test file, not `test_harmony_parsers.py`.

## Minor

- `normalize_log_level()` is just `.upper()` — can be removed if you use `type=str.upper` in argparse.
- Missing blank line between `configure_logging()` and `# Global engine instance` comment.

raullenchai (approved):
LGTM — clean implementation covering both entry points with tests. Minor nit: `normalize_log_level()` could be inlined, but not blocking. Thanks @XiaoPengMei!

raullenchai:
Thanks @XiaoPengMei — merged! 🎉 Added you to the changelog:

PR #61, opened 2026-03-31
fix: Add --log-level CLI flag

diptikhaparde-coder:
Added a `--log-level` CLI flag to control logging verbosity. The flag accepts standard Python loggin...

### DIFF
--- a/evals/run_eval.py
+++ b/evals/run_eval.py
@@ -31,6 +31,7 @@ Community contributors: see evals/README.md for how to submit results.
 
 import argparse
 import json
+import logging
 import os
 import platform
 import re
--- a/evals/run_eval.py
+++ b/evals/run_eval.py
@@ -1623,8 +1623,21 @@ Examples:
     parser.add_argument(
         "--notes", default=None, help="Free-form notes about this eval run"
     )
+    parser.add_argument(
+        "--log-level",
+        type=str,
+        default="INFO",
+        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
+        help="Set the logging level (default: INFO)",
+    )
 
     args = parser.parse_args()
+    
+    # Configure logging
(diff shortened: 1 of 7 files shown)


raullenchai (changes requested):
Thanks for the contribution @diptikhaparde-coder! However, this PR has significant overlap with #57 which addresses the same issue (#50). A few concerns:

## Duplicate PR

PR #57 by @XiaoPengMei already implements `--log-level` for the server and CLI. This PR covers the same ground for `server.py` and additionally adds `--log-level` to 6 helper scripts. We should avoid merging two PRs for the same feature.

**Suggestion**: Close this PR in favor of #57 for the core server flag. If you'd like to contribute the script-side `--log-level` additions, please open a separate PR that only touches the scripts (not `server.py`), and ideally extract the repeated logging setup into a shared helper to avoid the copy-paste pattern.

## Issues if we were to review standalone

### P1: Removes module-level logging.basicConfig

The diff removes `logging.basicConfig(level=logging.INFO)` at module import time and defers it to `main()`. This means any log messages emitted during module import or `load_model()` (called before `main()` in the CLI path via `cli.py`) will be silently dropped since no handler is configured yet.

### P2: Copy-paste boilerplate across 6 files

The same 6-line logging setup block is duplicated in every script. A shared utility (e.g. `scripts/utils.py:setup_logging(args)`) would be cleaner and easier to maintain.

### P3: Case sensitivity (same issue as #57)

`choices=["DEBUG", ...]` rejects lowercase input like `--log-level debug`. Use `type=str.upper` to accept any casing.

### Minor: `import logging` and `import sys` added in server.py but both are already imported earlier in the file.

Please coordinate with #57 to avoid duplication. Happy to review a scripts-only follow-up PR!

raullenchai:
Thanks for the contribution, @diptikhaparde-coder! We're going with #57 which covers both entry points (CLI + server module) and includes tests. Closing this one — appreciate your effort!

PR #72, opened 2026-04-07
feat(eval): auto-detect model ID + smarter tool call matching

Snider:
## Summary

- **`_detect_model_id()`** — queries `/v1/models` and returns the first model ID, solving the hardcoded `"default"` model name that caused request failures with strictly-named model endpoints. `_MODEL_ID_CACHE` avoids repeated lookups per `(host, port)` pair across the eval run.

- **`alt_expected_tools`** — a list of semantically equivalent tool alternatives. Useful when a smarter model (e.g. Gemma 4) picks `process` instead of `exec`, or `code_run` instead of `exec`, and both are correct. Arg grading is skipped for alt tools since they have different schemas.

- **`accept_tools`** — for `expect_no_tool` irrelevance tests, marks tools that are reasonable to call (e.g. `web_search` for "What is the capital of France?"). Those cases now score PASS instead of penalising curious models.

- **Combined-steps recognition** — if a model solves a multi-step scenario in a single call (e.g. `pip install -r /tmp/requirements.txt` instead of read-then-install), all followup steps are marked passed and the scenario scores PASS. This prevents penalising efficient models.

## Files changed

- `evals/run_eval.py` — all logic changes
- `evals/prompts/tool_calling.json` — `alt_expected_tools` and `accept_tools` fields on relevant scenarios

## Test plan

- [ ] Run `python evals/run_eval.py` against a live server — verify model ID is auto-detected from `/v1/models`
- [ ] Confirm irrelevance tests (tc18, tc19) PASS when server returns `web_search`
- [ ] Confirm multi-step scenarios (tc28+) PASS when model combines steps

Co-Authored-By: Virgil <virgil@lethean.io>

### DIFF
--- a/evals/prompts/tool_calling.json
+++ b/evals/prompts/tool_calling.json
@@ -162,6 +162,7 @@
       {"role": "user", "content": "Show me all running Python processes"}
     ],
     "expected_tool": "exec",
+    "alt_expected_tools": ["process"],
     "expected_args": {"command": "ps"},
     "arg_match_mode": "contains",
     "fake_result": "  PID TTY           TIME CMD\n12345 ttys000    0:01.23 python server.py\n12346 ttys001    0:00.45 python eval.py",
@@ -268,7 +269,8 @@
     "messages": [
       {"role": "user", "content": "What is the capital of France?"}
     ],
-    "expect_no_tool": true
+    "expect_no_tool": true,
+    "accept_tools": ["web_search"]
   },
   {
     "id": "tc19",
@@ -279,7 +281,8 @@
     "messages": [
       {"role": "user", "content": "Explain how neural networks work in 2-3 sentences"}
     ],
-    "expect_no_tool": true
+    "expect_no_tool": true,
+    "accept_tools": ["web_search"]
   },
   {
     "id": "tc20",
@@ -317,6 +320,7 @@
       {"role": "user", "content": "Read the file /tmp/requirements.txt and install the packages listed in it"}
     ],
     "expected_tool": "read",
+    "alt_expected_tools": ["exec"],
     "expected_args": {"path": "/tmp/requirements.txt"},
     "arg_match_mode": "exact",
     "fake_result": "flask==3.0.0\nrequests>=2.31.0\npydantic~=2.5",
@@ -333,6 +337,7 @@
       {"role": "user", "content": "Check my Python version, remember it, and send a message to Alice with the version info"}
     ],
     "expected_tool": "exec",
+    "alt_expected_tools": ["code_run"],
     "expected_args": {"command": "python"},
     "arg_match_mode": "contains",
     "fake_result": "Python 3.11.5",
@@ -441,6 +446,7 @@
       {"role": "user", "content": "Find my Python version and store it in memory for later"}
     ],
     "expected_tool": "exec",
+    "alt_expected_tools": ["code_run"],
     "expected_args": {"command": "python"},
     "arg_match_mode": "contains",
     "fake_result": "Python 3.12.1",
--- a/evals/run_eval.py
+++ b/evals/run_eval.py
@@ -292,6 +292,28 @@ def detect_hardware() -> dict:
     return hw
 
 
+def _detect_model_id(host: str, port: int) -> str:
+    """Auto-detect model ID from the server's /v1/models endpoint."""
+    try:
+        resp = httpx.get(f"http://{host}:{port}/v1/models", timeout=5.0)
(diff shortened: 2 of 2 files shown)


raullenchai (approved):
Hey @Snider, welcome and thanks for this contribution! Great work overall — the changes are well-motivated and the code quality is solid. Here's a detailed review:

## What I like

- **Auto-detect model ID** — This is a real improvement. The hardcoded `"default"` was a known pain point when testing against servers with strict model naming. The `_MODEL_ID_CACHE` per `(host, port)` is a clean approach.

- **`accept_tools` for irrelevance tests** — Totally reasonable. Penalizing a model for calling `web_search` on "What is the capital of France?" was always borderline — a smart model doing a search is arguably *better* behavior. Good pragmatic call.

- **Combined-steps recognition** — This is the most impactful change. If a model solves `read-then-install` in a single `pip install -r /tmp/requirements.txt`, that's genuinely smarter, and the eval should reward it rather than fail it. The implementation correctly checks that the alt tool matches a followup tool before triggering the shortcut.

## Minor observations (non-blocking)

1. **`_detect_model_id` silently swallows all exceptions** — This is fine for an eval script (fallback to `"default"` is sensible), but just flagging it. If someone passes a wrong port, they'll get silent fallback with no hint. A `logger.debug()` in the `except` would help future debugging, but not a blocker.

2. **`alt_expected_tools` key construction with `step_prefix`** — The line `alt_key = f"{step_prefix}alt_expected_tools" if step_prefix else "alt_expected_tools"` prepares for prefixed alt keys on followup steps (e.g., `followup_alt_expected_tools`), but no scenario in the JSON actually uses that. That's forward-compatible, so it's fine — just noting it's currently unused for followup steps.

3. **`tool_accepted` in irrelevance results** — Nice touch adding this to the result dict for transparency in reports.

4. **tc22 `alt_expected_tools: ["exec"]`** — Just want to confirm the intent: if the model calls `exec` with something like `pip install -r /tmp/requirements.txt` directly (skipping the `read` step), it gets full marks. That matches the combined-steps logic and makes sense — the model found a more efficient path.

## Verdict

Clean, well-scoped changes that make the eval framework more fair to capable models without lowering the bar. The PR description and test plan are thorough.

Approving. Thanks for the contribution!

raullenchai (changes requested):
Thanks for this PR — the auto-detect model ID and alt tool matching are great ideas. A few concerns before we can merge:

### 1. `arg_score = 1.0` bypass for alt tools (run_eval.py L555)

When an alt tool is chosen, argument validation is skipped entirely:

```python
if is_alt_tool:
    arg_score = 1.0  # alt tools have different arg schemas
```

This means a model could call the right alt tool with completely wrong arguments and still pass. Alt tools should still validate arguments — either define `alt_expected_args` in the scenario, or at minimum check that `valid_json_args` is true and args are non-empty.

### 2. `accept_tools` weakens irrelevance tests (tool_calling.json tc18/tc19)

The irrelevance test category exists specifically to verify that models do NOT call tools when they shouldn't. Adding `"accept_tools": ["web_search"]` for "What is the capital of France?" defeats the purpose — a model that calls `web_search` for a factual question it should know is exhibiting poor tool-use judgment.

If certain models legitimately use web_search here, we should track it as a separate metric (e.g., `"tool_used_but_acceptable": true`) rather than counting it as a full PASS.

### 3. `combined_steps` auto-passes followup steps (run_eval.py L876-893)

When a model picks an alt tool matching a followup tool, all subsequent steps auto-pass without execution:

```python
if combined_steps:
    for prefix in followup_prefixes:
        if f"{prefix}expected_tool" in sc:
            steps_passed.append(True)
```

This skips actual verification. A model that only completes step 1 (differently) gets full credit for a 3-step scenario. The followup steps should still be executed, or at minimum we should distinguish "combined" passes from "verified" passes in the score.

### 4. Minor bug: missing `_HTTPX` guard in `_detect_model_id`

`httpx.get()` is called without checking the `_HTTPX` flag. If httpx isn't installed, this will raise `NameError` instead of gracefully falling back to `"default"`.

---

The model ID auto-detection and caching are solid — happy to merge once the test relaxation concerns are addressed. The core principle: it's fine to accept alternative correct answers, but we shouldn't skip validation when we do.

raullenchai:
Thanks @Snider! Closing this draft — it's been sitting since early April and the eval framework has moved significantly since then (parser mapping changes, scorecard generator, BFCL integration in flight). Both `_detect_model_id()` and `alt_expected_tools` are still genuinely useful ideas — happy to revisit if you'd like to re-open against current main.

PR #97, opened 2026-04-11
feat: migrate Qwen3 reasoning parser to token-level OutputRouter (closes #64)

Ai-chan-0411:
Closes #64

## Summary

Migrate Qwen3's `<think>`/`</think>` reasoning parser from regex-based `BaseThinkingReasoningParser` to the token-level `OutputRouter` state machine.

### Changes

**`vllm_mlx/output_router.py`**
- Add `<think>`/`</think>` token handling in `feed()` — `think_start` enters THINKING state, `think_end` switches to CONTENT
- Add Qwen3/DeepSeek auto-detection in `from_tokenizer()` via `<think>` + `</think>` vocabulary entries
- Token map already had `think_start`/`think_end` fields (added for future migration) — now wired up

**`tests/test_output_router.py`**
- Add Qwen3 vocabulary fixtures and `qwen3_router` fixture
- Add `TestQwen3ThinkRouting` class with 9 tests:
  - `<think>` enters THINKING state
  - `</think>` switches to CONTENT state
  - Tokens between tags routed to REASONING channel
  - Content tokens after `</think>` routed to CONTENT channel
  - Full `<think>reasoning</think>content` sequence via `feed_sequence()`
  - Implicit thinking (no `<think>`, only `</think>`) handled
  - No-tag output passes through as pure content
  - Control tokens (bos/eos) suppressed
  - Reset clears thinking state
- Add `test_qwen3_detected` in `TestFromTokenizer`

**31 tests total, all passing.**

### Design

Follows the same pattern as the existing Gemma 4 implementation — token-level state machine with no text-level regex matching. This eliminates partial-token split issues that affect the current regex-based parser.

The detection uses `<think>` + `</think>` in the tokenizer vocabulary, which covers both Qwen3 and DeepSeek R1 model families.

### DIFF
--- a/tests/test_output_router.py
+++ b/tests/test_output_router.py
@@ -68,9 +68,43 @@ VOCAB = {
 TOKENIZER = FakeTokenizer(VOCAB)
 
 
+# === Qwen3 Token IDs (representative) ===
+QWEN3_VOCAB = {
+    "<|endoftext|>": 151643,
+    "<|im_end|>": 151645,
+    "<think>": 151667,
+    "</think>": 151668,
+    "Let": 5733,
+    "me": 2734,
+    "analyze": 28541,
+    "The": 785,
+    "answer": 10234,
+    "is": 374,
+    "42": 2983,
+    ".": 13,
+}
+
+QWEN3_MAP = TokenMap(
+    think_start=151667,   # <think>
+    think_end=151668,     # </think>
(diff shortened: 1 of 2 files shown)


Ai-chan-0411:
Closing as maintainer has not reviewed after 168h. Thank you for the opportunity!

Issue #51, opened 2026-03-26
Add request logging middleware


raullenchai:
Log each request with method, path, status code, and duration. Use FastAPI middleware. Should be toggleable via --verbose flag.

koladefaj:
Hi @raullenchai,

I’d be happy to take this on. My plan would be to implement a FastAPI middleware that logs the request method, path, response status, and processing duration, while keeping the logging behavior toggleable via a --verbose flag. I’m thinking of measuring request duration using a high-resolution timer (e.g., time.perf_counter) and integrating the middleware conditionally based on the runtime verbosity setting.

Please let me know if this aligns with the intended behavior before I begin working on a PR.

raullenchai:
Sounds good — your approach makes sense. A few notes:
- We already have `--log-level` flag (DEBUG/INFO/WARNING/ERROR) that controls uvicorn logging
- The middleware should be lightweight (no per-request allocations)
- `time.perf_counter()` is the right call
Go ahead and open a PR!

raullenchai:
Fixed by #1167. Request logging middleware logs method, path, status, and duration at DEBUG level. Zero-cost at INFO+. Use `--log-level DEBUG` to enable. Format: `POST "/v1/chat/completions" 200 0.847s`

PR #114, opened 2026-04-15
feat: add request logging middleware behind --verbose flag (#51)

koladefaj:
## What does this PR do?

This PR implements the requested FastAPI middleware to log request metrics (method, path, status code, and duration).
* **Lightweight:** Uses function-based `@app.middleware("http")` to avoid per-request object allocations.
* **Accurate Timing:** Utilizes `time.perf_counter()` for high-resolution duration tracking.
* **Toggleable:** The middleware is conditionally injected into the Uvicorn app only if the `--verbose` CLI flag is passed during startup, ensuring zero overhead for standard runs.
* **Log Level Compatibility:** Uses the existing server logger so it natively respects the `--log-level` configurations.

## Checklist

- [ ] Tests pass locally (`python3 -m pytest tests/ -x`)
- [ ] Updated README/docs if applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.4.3"
+version = "0.5.0"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1778,6 +1778,17 @@ async def create_completion(request: CompletionRequest, raw_request: Request):
 
     elapsed = time.perf_counter() - start_time
     tokens_per_sec = total_completion_tokens / elapsed if elapsed > 0 else 0
+
+    # NEW: Log the request details with duration and status at the end
+    prompt_preview = prompts[0][:200] if prompts else "(empty)"
+    prompt_len = sum(len(p) for p in prompts)
+    logger.info(
+        f"[REQUEST] POST /v1/completions status=200 duration={elapsed:.4f}s "
+        f"stream=False max_tokens={request.max_tokens} "
+        f"temp={request.temperature} prompt_chars={prompt_len} "
+        f"prompt_preview={prompt_preview!r}"
+    )
+
     logger.info(
         f"Completion: {total_prompt_tokens} prompt + {total_completion_tokens} completion tokens in {elapsed:.2f}s ({tokens_per_sec:.1f} tok/s)"
     )
@@ -2886,6 +2897,13 @@ async def _stream_anthropic_messages(
     # Log throughput
     elapsed = time.perf_counter() - start_time
     tokens_per_sec = completion_tokens / elapsed if elapsed > 0 else 0
+
+    # NEW: Log the streaming request details with final duration
+    logger.info(
+        f"[REQUEST] POST /v1/messages (anthropic) status=200 duration={elapsed:.4f}s "
+        f"stream=True model={_model_name or anthropic_request.model!r} "
+        f"prompt_tokens={prompt_tokens} completion_tokens={completion_tokens}"
+    )
     logger.info(
         f"Anthropic messages (stream): prompt={prompt_tokens} + completion={completion_tokens} tokens in {elapsed:.2f}s ({tokens_per_sec:.1f} tok/s)"
     )
@@ -3509,6 +3527,13 @@ async def stream_chat_completion(
         # Log throughput
         elapsed = time.perf_counter() - start_time
         tokens_per_sec = completion_tokens / elapsed if elapsed > 0 else 0
+
+        # NEW: Log the streaming request details with final duration
+        logger.info(
+            f"[REQUEST] POST /v1/chat/completions status=200 duration={elapsed:.4f}s "
+            f"stream=True model={request.model or _model_name!r} "
+            f"prompt_tokens={prompt_tokens} completion_tokens={completion_tokens}"
+        )
         logger.info(
             f"Chat completion (stream): {completion_tokens} tokens in {elapsed:.2f}s ({tokens_per_sec:.1f} tok/s)"
         )


raullenchai (changes requested):
Thanks for the contribution @koladefaj! Request logging is useful for debugging.

However, I have a few concerns:

1. **We already have detailed per-endpoint request logging** in `server.py` — each endpoint logs method, path, model, max_tokens, message count, etc. (search for `[REQUEST]` in server.py). The middleware adds less information (only path + status + duration) on top of what we already have.

2. **Middleware placement**: Adding `@app.middleware` inside `serve_command()` in `cli.py` means it runs after `from .server import app` — this works but couples CLI logic to the app object in a fragile way. If we want middleware, it should live in `server.py` near the app definition.

3. **`--verbose` naming**: This is very generic and could be confused with `--log-level DEBUG`. A more specific name like `--log-requests` would be clearer about what it enables.

4. **Missing duration in existing logs**: The one thing this PR adds that we don't have is **response duration timing**. That's genuinely useful! But it could be added more simply by extending our existing `[REQUEST]` logging with a timer, without a separate middleware.

**Suggestion**: Instead of a middleware, consider adding duration timing to our existing request logging in `server.py`. That would give us timing without the extra flag or middleware complexity. Happy to discuss!

koladefaj:
> Thanks for the contribution @koladefaj! Request logging is useful for debugging.
> 
> However, I have a few concerns:
> 
> 1. **We already have detailed per-endpoint request logging** in `server.py` — each endpoint logs method, path, model, max_tokens, message count, etc. (search for `[REQUEST]` in server.py). The middleware adds less information (only path + status + duration) on top of what we already have.
> 2. **Middleware placement**: Adding `@app.middleware` inside `serve_command()` in `cli.py` means it runs after `from .server import app` — this works but couples CLI logic to the app object in a fragile way. If we want middleware, it should live in `server.py` near the app definition.
> 3. **`--verbose` naming**: This is very generic and could be confused with `--log-level DEBUG`. A more specific name like `--log-requests` would be clearer about what it enables.
> 4. **Missing duration in existing logs**: The one thing this PR adds that we don't have is **response duration timing**. That's genuinely useful! But it could be added more simply by extending our existing `[REQUEST]` logging with a timer, without a separate middleware.
> 
> **Suggestion**: Instead of a middleware, consider adding duration timing to our existing request logging in `server.py`. That would give us timing without the extra flag or middleware complexity. Happy to discuss!

Thanks for the detailed review, @raullenchai, Dropping the middleware and CLI flag to lean into the existing `[REQUEST]` logging is definitely the cleaner architecture.

I found the specific gap: the non-streaming requests easily capture duration, but the streaming requests fire their `[REQUEST]` log before the generator starts, meaning they can't log the total elapsed time.

Proposed Fix (Zero middleware/flags):

Non-streaming: I'll append `status=200` `duration=X.XXs` to the existing `[REQUEST]` logs at the bottom of the endpoints.

Streaming: I will leave the pre-flight `[REQUEST]` logs intact (so we don't lose visibility on requests that might hang during prefill), but I will upgrade the final throughput logs inside `stream_completion`, `stream_chat_completion`, and `_stream_anthropic_messages` to include the final `status=200` and duration metrics.

raullenchai (changes requested):
Thanks for the contribution! A few issues to address:

**1. Not actually a middleware** — The PR title and description mention middleware (`@app.middleware("http")`), but the implementation just adds `logger.info()` calls inline in 3 endpoint functions. A proper middleware would be a single function that wraps all routes, avoiding code duplication. Something like:

```python
@app.middleware("http")
async def log_requests(request: Request, call_next):
    start = time.perf_counter()
    response = await call_next(request)
    duration = time.perf_counter() - start
    logger.info(f"[REQUEST] {request.method} {request.url.path} status={response.status_code} duration={duration:.3f}s")
    return response
```

**2. Duplicates existing logging** — Each insertion site is right next to an existing `logger.info("Chat completion: N tokens in Xs")` line that already logs duration and token counts. The new lines add the same info in a different format.

**3. Prompt content leaking** — `prompt_preview={prompt_preview!r}` logs the first 200 chars of user prompts. This is a privacy concern — production logs should never contain user content.

**4. Missing `--verbose` toggle** — The description says "conditionally injected only if `--verbose` CLI flag is passed", but no such flag or conditional is in the diff. The logs are always emitted.

**5. Incomplete coverage** — Non-streaming `/v1/chat/completions` is not covered.

Suggestion: implement this as a single FastAPI middleware function, gated behind a `--verbose` flag (or `--log-requests`), that captures method/path/status/duration for ALL endpoints in one place.

koladefaj:
> Thanks for the contribution! A few issues to address:
> 
> **1. Not actually a middleware** — The PR title and description mention middleware (`@app.middleware("http")`), but the implementation just adds `logger.info()` calls inline in 3 endpoint functions. A proper middleware would be a single function that wraps all routes, avoiding code duplication. Something like:
> 
> ```python
> @app.middleware("http")
> async def log_requests(request: Request, call_next):
>     start = time.perf_counter()
>     response = await call_next(request)
>     duration = time.perf_counter() - start
>     logger.info(f"[REQUEST] {request.method} {request.url.path} status={response.status_code} duration={duration:.3f}s")
>     return response
> ```
> 
> **2. Duplicates existing logging** — Each insertion site is right next to an existing `logger.info("Chat completion: N tokens in Xs")` line that already logs duration and token counts. The new lines add the same info in a different format.
> 
> **3. Prompt content leaking** — `prompt_preview={prompt_preview!r}` logs the first 200 chars of user prompts. This is a privacy concern — production logs should never contain user content.
> 
> **4. Missing `--verbose` toggle** — The description says "conditionally injected only if `--verbose` CLI flag is passed", but no such flag or conditional is in the diff. The logs are always emitted.
> 
> **5. Incomplete coverage** — Non-streaming `/v1/chat/completions` is not covered.
> 
> Suggestion: implement this as a single FastAPI middleware function, gated behind a `--verbose` flag (or `--log-requests`), that captures method/path/status/duration for ALL endpoints in one place.

Hi @raullenchai No worries at all,
In the first review, you actually suggested skipping the middleware/flag complexity and just extending the existing [REQUEST] logs in `server.py` that’s why I went with that approach
That said, I totally agree with you a single middleware function is way cleaner and prevents duplicating code across endpoints. I also completely get your concern about the privacy side of prompt_preview.
I’ll revert the inline endpoint changes and implement a proper `@app.middleware("http")` in `server.py` instead. It’ll strip out the prompt content and be gated behind a new `--log-requests` CLI flag.

raullenchai (changes requested):
Thanks for the contribution @koladefaj! The direction is right but a few things need updating:

**1. Duplicate log prefix (P1)**
We already log `[REQUEST] POST /v1/chat/completions` at the *start* of each request (line ~2050 in current main). Adding another `[REQUEST]` at the *end* creates confusion. Please use `[DONE]` or `[RESPONSE]` as the prefix for the completion log, e.g.:
```
[DONE] POST /v1/chat/completions status=200 duration=1.23s ...
```

**2. Not behind --verbose flag (P1)**
The issue #51 and PR title say "behind --verbose flag" but the code uses `logger.info()` which always prints. Either:
- Use `logger.debug()` (only shows with `--log-level debug`)
- Or check `if _verbose:` flag before logging

**3. Missing non-streaming chat path (P2)**
The non-streaming `create_chat_completion` handler (the most common path) doesn't have the new logging.

**4. Needs rebase (P2)**
The PR is based on an older version of `server.py`. Main has changed significantly (inference lock, reasoning_content rename, etc.). Please rebase onto latest `main`.

Small PR, easy fixes — happy to merge once updated!

raullenchai:
Thanks for the contribution @koladefaj! Closing for now since this hasn't been updated since the changes-requested review and main has moved significantly (server.py decomposition, route extraction, ServerConfig refactor). The branch would need a rebase plus the originally-requested changes.

Happy to revisit if you re-open against current main — request logging behind `--verbose` is still a useful addition. The middleware would now live under `vllm_mlx/middleware/` alongside the auth + rate-limit ones.

PR #130, opened 2026-04-16
perf: SGLang-inspired optimizations (radix cache, overlap scheduling, jump-forward decoding)

raullenchai:
## Summary

Three SGLang-inspired performance optimizations for the inference engine:

- **Radix-tree compaction for PrefixCacheManager** — Replace per-token trie with radix tree that compresses single-child chains into nodes with token-array edges. Reduces traversal from O(num_tokens) to O(num_branches). Target: 10-30% TTFT gain on multi-turn chat with shared system prompt.

- **Overlap scheduling with async_eval** — Replace synchronous `mx.eval()` in `_cleanup_finished()` with batched `mx.async_eval()` and remove redundant `mx.clear_cache()`. Cache tensor evaluation now runs on GPU in background while Python continues with scheduling. Target: 15-25% throughput gain.

- **Jump-forward structured decoding** — When the tool logits processor detects a deterministic structural pattern (e.g. `</invoke>`, `</minimax:tool_call>`), inject all remaining pattern tokens in a single prefill pass instead of N separate decode steps. Saves 3-8 decode steps per pattern occurrence. Target: 30-50% decode skip on tool calls.

## Test plan

- [x] All 27 prefix cache tests pass (radix tree is drop-in)
- [x] All 51 tool logits tests pass
- [x] 117 total tests pass
- [x] ruff lint clean
- [ ] Run `make check` with qwen3.5-4b to verify no regression
- [ ] Test tool calling with Hermes agent to validate jump-forward

### DIFF
--- a/harness/baselines/check-batched-qwen3.5-4b.json
+++ b/harness/baselines/check-batched-qwen3.5-4b.json
@@ -1,0 +1,20 @@
+{
+  "captured_at": "2026-04-16T11:07:55",
+  "rapid_mlx_version": "0.5.0",
+  "model": "qwen3.5-4b",
+  "metrics": {
+    "cold_ttft_ms": 81.48550009354949,
+    "cold_tps": 112.42155950437478,
+    "cached_ttft_ms": 133.96895793266594,
+    "decode_tps": 112.32704081303969,
+    "decode_tps_stdev": 0,
+    "mt_ttft_ms": 161.22487490065396,
+    "mt_tps": 112.33515669498165,
+    "tc_latency_ms": 995.6307501997799,
+    "tc_success_rate": 1.0,
+    "long_ttft_ms": 862.5065411906689,
+    "long_tps": 110.30327980084134,
+    "long_cached_ttft_ms": 863.2712909020483,
+    "composite_score": 241.5
+  }
+}
--- a/harness/baselines/check-qwen3.5-4b.json
+++ b/harness/baselines/check-qwen3.5-4b.json
@@ -1,20 +1,20 @@
 {
-  "captured_at": "2026-04-15T21:36:32",
+  "captured_at": "2026-04-16T11:07:28",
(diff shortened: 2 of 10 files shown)


samuelfaj:
I found a startup/config regression in this PR that can affect the regular server path on `main`.

## Problem

The model registry registration and `_sync_config()` call appear to have been moved into `_retry_tool_logits_setup()`.

That helper starts with:

```python
if not _enable_tool_logits_bias or not _enable_auto_tool_choice or not _tool_call_parser:
    return
```

So any normal server startup that does **not** enable tool-logits bias can return before the loaded engine is registered and before globals are copied into `ServerConfig`.

## User-visible effect

The server process can be listening and otherwise alive, but routes that read `get_config()` may still see no engine/model configured. For example, health/status style endpoints can report an unloaded or unknown engine state even though `load_model()` already created the engine.

This is not model-specific. It is caused by tying required server config initialization to an optional tool-logits feature flag.

## Suggested fix

Keep `_retry_tool_logits_setup()` limited to retrying `_tool_logits_processor_factory` setup after `engine.start()`, and move model registration plus `_sync_config()` back to the end of `load_model()`, outside the tool-logits retry path.

Expected structure:

```python
# end of load_model(), after creating _engine and attempting initial tool-logits setup
_model_registry.add(entry, is_default=True)
_sync_config()


def _retry_tool_logits_setup() -> None:
    if not _enable_tool_logits_bias or not _enable_auto_tool_choice or not _tool_call_parser:
        return
    ...
    # only set _engine._tool_logits_processor_factory here
```

That keeps optional deferred tool-logits setup independent from required server config initialization.

raullenchai:
Closing — measured 0 net speedup on our workloads (radix cache + overlap scheduling don't translate to MLX wins, see internal notes). Branch preserved if anyone wants to revisit a focused subset.

Issue #131, opened 2026-04-16
Default to BatchedEngine (--continuous-batching should be the default)


raullenchai:
## Problem

Rapid-MLX currently defaults to SimpleEngine (single-user mode). Users must pass `--continuous-batching` to get BatchedEngine. This is backwards — every other serious inference engine defaults to batched:

| Engine | Default | Notes |
|---|---|---|
| **vLLM** | Always batched | PagedAttention, handles all model types |
| **SGLang** | Always batched | RadixAttention, no single-user mode |
| **TGI** | Always batched | |
| **Rapid-MLX** | Simple ❌ | Must pass `--continuous-batching` |

### 1. "GatedDeltaNet/Mamba hybrid models need special handling in batch mode"

vLLM and SGLang already support Mamba/hybrid models in batched mode. Our 200ms throttle + non-trimmable cache workarounds are implementation gaps, not fundamental limitations. We should study how vLLM handles these:
- vLLM's `MambaCache` / `Jamba` support
- SGLang's hybrid model scheduling

### 2. "SimpleEngine has better prompt cache for single-user"

BatchedEngine already has `MemoryAwarePrefixCache` which does the same thing. The 168 vs 112 tok/s gap on qwen3.5-4b is from scheduler overhead, not cache quality. This overhead should be optimizable.

## Proposed change

1. Make `--continuous-batching` the default (flip the flag to `--simple-mode` for opt-out)
2. Research vLLM/SGLang's Mamba/hybrid batch handling → remove our 200ms throttle hack
3. Optimize BatchedEngine single-user throughput to match SimpleEngine (~168 tok/s target)
4. Eventually deprecate SimpleEngine or make it a thin wrapper over BatchedEngine with max_num_seqs=1

## References

- vLLM Mamba support: https://github.com/vllm-project/vllm/tree/main/vllm/model_executor/models (jamba, mamba)
- SGLang scheduling: https://github.com/sgl-project/sglang/tree/main/python/sglang/srt
- Current BatchedEngine throttle: `engine_core.py` 200ms request throttle for GatedDeltaNet

raullenchai:
Closing as duplicate of #134, which has the full implementation plan and PR #135.

PR #132, opened 2026-04-17
feat: FSM-based tool call constrained decoding via outlines-core

raullenchai:
## Summary

Replace fragile string-based tool call parsing with FSM (finite state machine) constrained decoding using outlines-core. When a model generates a tool call, the FSM guarantees the JSON output is structurally valid by masking invalid tokens during generation.

**Key changes:**
- New `vllm_mlx/api/fsm_tool_call.py`: FSM cache, Guide, two-mode logits processor
- Works on **both** SimpleEngine and BatchedEngine (same FSM code path)
- Per-parser trigger patterns for all 18 tool call formats
- Graceful fallback: XML/Nemotron formats bypass FSM (handled by existing parsers)
- `outlines-core` added to `[guided]` extra in pyproject.toml

**Performance:**
- Per-token FSM overhead: **0.9 µs** (0.004% of 20ms decode step)
- FSM compile: ~2.3s (one-time per tool schema, cached by hash)
- Doctor check: **0 regression** across all 13 metrics
- Benchmark: qwopus-27b 22.4 tok/s — PASS

**How it works:**
1. Model generates freely (text, reasoning)
2. When output ends with tool call trigger (e.g., `<tool_call>\n`)...
3. Next token checked: if `{` → FSM activates (JSON mode)
4. FSM masks invalid tokens → guaranteed valid JSON
5. When FSM reaches terminal state → back to free mode
6. If not `{` (XML format) → FSM skips, existing parsers handle it

## Direction

This PR is step 1 toward unifying SimpleEngine and BatchedEngine. Both engines now share the same FSM logits processor interface. Next: deprecate SimpleEngine as a separate code path (see #131).

## Test plan

- [x] 13 FSM unit tests (cache, processor, factory, performance, XML skip)
- [x] Doctor smoke: ruff ✅, 2086/2087 pytest pass
- [x] Doctor check: 0 regression, all metrics ±5%
- [x] Doctor benchmark: qwopus-27b PASS
- [x] E2E verified on both engines (FSM trigger → constrained → complete)
- [x] Codex review: 0 issues on FSM code

### DIFF
--- a/harness/baselines/check-qwen3.5-4b.json
+++ b/harness/baselines/check-qwen3.5-4b.json
@@ -1,20 +1,20 @@
 {
-  "captured_at": "2026-04-16T16:26:34",
+  "captured_at": "2026-04-16T16:34:02",
   "rapid_mlx_version": "0.5.0",
   "model": "qwen3.5-4b",
   "metrics": {
-    "cold_ttft_ms": 174.72345801070333,
-    "cold_tps": 168.51368492027117,
-    "cached_ttft_ms": 229.73799984902143,
-    "decode_tps": 162.72374052215056,
+    "cold_ttft_ms": 170.7701669074595,
+    "cold_tps": 168.1036081769402,
+    "cached_ttft_ms": 223.16349996253848,
+    "decode_tps": 167.821708310493,
     "decode_tps_stdev": 0,
-    "mt_ttft_ms": 296.3542500510812,
-    "mt_tps": 159.6755279661571,
-    "tc_latency_ms": 1112.5775419641286,
+    "mt_ttft_ms": 233.40062494389713,
+    "mt_tps": 165.54264366438625,
+    "tc_latency_ms": 1068.896499928087,
     "tc_success_rate": 1.0,
-    "long_ttft_ms": 1014.3902089912444,
-    "long_tps": 158.70258529015774,
-    "long_cached_ttft_ms": 1008.7231250945479,
-    "composite_score": 246.1
+    "long_ttft_ms": 966.4679998531938,
+    "long_tps": 163.68971893952352,
+    "long_cached_ttft_ms": 943.7295419629663,
+    "composite_score": 257.2
   }
 }
--- a/harness/baselines/check-qwen3.5-4b.json
+++ b/harness/baselines/check-qwen3.5-4b.json
@@ -1,20 +1,20 @@
 {
-  "captured_at": "2026-04-15T21:36:32",
+  "captured_at": "2026-04-16T16:26:34",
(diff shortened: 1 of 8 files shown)


raullenchai:
**Status: research-only, not for merge.**

Per [knowledge/fsm_parser_decision.md](https://github.com/raullenchai/Rapid-MLX/blob/main/memory/knowledge/fsm_parser_decision.md) (internal) — the value of FSM parsing is correctness (guaranteed valid JSON), NOT speed. Industry context:

- vLLM: xgrammar for `response_format`, regex parsers for tools (separated)
- SGLang: xgrammar for both
- Ollama: GBNF for `response_format`, templates for tools
- All are opt-in, never default

This PR sketches the migration of all 18 string-based tool parsers to FSM-constrained decoding via outlines-core. Marking as **draft / research** — the architectural direction is right but:

- Lint failing
- Branch is 22 days behind main
- 18-parser migration needs careful per-parser verification with real model output, not just unit tests
- Need a flag-protected rollout (start opt-in via `--tool-call-fsm`, only flip default once we have N weeks of production telemetry showing parity)
- Should NOT merge until each migrated parser has been A/B tested against the existing parser on real model output (pre-existing test suite is necessary but not sufficient)

**Keeping open** as a tracking PR for the architectural direction. Will revisit when the surrounding work (per-alias profile of which parser to use, telemetry on parser failures) is in place. If the user wants to revive: rebase + add `--tool-call-fsm` flag + start with one parser only (qwen3coder is highest-value).

Marking as draft.

raullenchai:
Closing — strategic direction shift.

Rather than maintaining our own FSM-based constrained-decoding path for tool calls, we're standardizing on the upstream parser conventions used by vLLM and SGLang (per-family parsers in `vllm_mlx/tool_parsers/`: hermes, minimax, glm47, harmony, qwen3coder, deepseek). That surface already covers every model family in `aliases.json` and is what downstream OpenAI-SDK / Anthropic-SDK / pydantic-ai / smolagents integrations expect.

The FSM path in this PR (`fsm_tool_call.py`, 455 LOC + outlines-core dep) would create a parallel mechanism we'd have to keep in sync with each new model family, with no clear win over the upstream string-based parsers in practice — and the recent PR #419 (guided generation passthrough) + PR #515 (openai-harmony StreamableParser passthrough) already cover the structural-validity concerns that motivated the FSM design.

Branch `feat/fsm-tool-parser` will stay on the fork for archival reference. If a future model family needs FSM-level guarantees (e.g., bracket-matched code generation in a tool argument), we can revisit on a per-parser basis without resurrecting this whole layer.

PR #161, opened 2026-04-25
Fix MLX generation stream thread ownership

samuelfaj:
## Summary

Fixes #160.

Rapid-MLX's async engine was using a hybrid scheduler execution path:

- prefill / queued-request steps ran in a single `ThreadPoolExecutor` worker
- generation-only decode steps ran inline on the asyncio event-loop thread

That is unsafe with `mlx-lm` 0.31+ for Qwen3.x / hybrid cache models. `mlx-lm` creates `generation_stream` as a thread-local MLX stream, while the batch generator and prompt/generation cache objects keep MLX arrays that are tagged with the stream used during prior work. When Rapid-MLX moved the same `BatchGenerator` from the executor thread back to the event-loop thread, MLX could no longer resolve the stream in the current thread and raised errors like:

```text
RuntimeError: There is no Stream(gpu, 3) in current thread.
```

## What changed

- Initialize `mlx_lm.generate.generation_stream` inside the MLX executor thread.
- Keep every `scheduler.step()` call on the same single `mlx-step` worker thread, including both prefill and decode.
- Preserve async HTTP/OpenAI responsiveness by still using `loop.run_in_executor(...)`; the event loop no longer runs MLX decode inline.
- Add regression coverage that verifies:
  - the executor thread initializer rebinds `mlx-lm`'s generation stream
  - prefill and decode scheduler steps stay on the `mlx-step` worker thread

This intentionally does not touch the OpenAI-compatible API routing, model aliasing, Qwen reasoning/tool parsers, or response translation code.

## Validation

Ran locally on the changed files:

```text
uvx ruff check vllm_mlx/engine_core.py tests/test_batching.py
uvx ruff format --check vllm_mlx/engine_core.py tests/test_batching.py
python -m py_compile vllm_mlx/engine_core.py tests/test_batching.py
manual regression script matching the added threading checks
```

I also applied the equivalent patch to a local `rapid-mlx 0.6.1` Homebrew install and verified `rapid-mlx serve qwen3.6-35b` with:

- server warmup
- `/v1/models`
- non-streaming `/v1/chat/completions`
- streaming SSE `/v1/chat/completions`
- a long prompt request with ~10k prompt tokens
- a `claw --model openai/default` request with a ~5.4k-token prompt

Those requests completed without the `There is no Stream(gpu, N) in current thread` scheduler failure.

PR #162, opened 2026-04-25
Fix Qwen tool-call OpenAI translation

### LINKED ISSUE #160: RuntimeError: There is no Stream(gpu, 0) in current thread.
### Rapid-MLX version

rapid-mlx 0.6.1

### Hardware

MacBook Pro M5 Max 48gb

### macOS version

26.4.1

### Python version

3.13.13

### Model

qwen3.6-35b

### Full serve command

rapid-mlx serve qwen3.6-35b

### What happened?

Latest version don't run on my M5 Max

### Steps to reproduce

1. git clone ...
2. uv sync
3. rapid-mlx serve qwen3.6-35b

### Error logs / output

```shell
  File "/Users/mlaprise/repos/Rapid-MLX/vllm_mlx/scheduler.py", line 2343, in step
  File "/Users/mlaprise/repos/Rapid-MLX/.venv/lib/python3.13/site-packages/mlx_lm/generate.py", line 1329, in next
    return self._next()
           ~~~~~~~~~~^^
  File "/Users/mlaprise/repos/Rapid-MLX/vllm_mlx/scheduler.py", line 564, in _chunked_next
    new_batch = self._process_prompts(batch_prompts)
  File "/Users/mlaprise/repos/Rapid-MLX/vllm_mlx/scheduler.py", line 236, in _patched_process_prompts
    batch = _orig_process_prompts(prompts)
  File "/Users/mlaprise/repos/Rapid-MLX/.venv/lib/python3.13/site-packages/mlx_lm/generate.py", line 1107, in _process_prompts
    mx.eval([c.state for c in prompt_cache])
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

samuelfaj:
## Summary

This PR fixes OpenCode/Hermes/Qwen 3.x tool-call translation failures reported in #160 when Rapid-MLX serves the OpenAI-compatible chat API.

The failure was not one single parser bug. It came from several OpenAI-client-visible edge cases around streamed tool calls and follow-up turns after tool results:

- Qwen sometimes emits bare `Calling tool: name({...})` text instead of the configured XML/Qwen tool-call wrapper.
- Tool arguments can arrive as JSON-looking strings, including double-encoded arrays and objects.
- Some client tool schemas use nullable/list forms such as `type: ["array", "null"]`, or omit `type` while still defining `items` / `properties`.
- Streaming chunks can split a `Calling tool` marker, causing partial text such as `[Calling tool` or `[` to leak to the client.
- The partial-marker guard must not strip normal code brackets when a chunk ends after `snake[` or `const items = [`.
- The model can emit an incomplete duplicate tool call before a valid duplicate; strict clients reject the incomplete call before they can consume the valid one.
- After a valid tool result, Qwen can stop with narration/planning instead of issuing the next tool call. OpenCode then appears to hang because the assistant turn ends without useful tool progress.
- Native tool-message preservation was detected before CLI state was synced into `ServerConfig`, so Qwen-native tool parsers could still receive converted text tool results in some startup paths.
- Qwen can enter repeated-text loops before making the next tool call, for example generating `point point point...` indefinitely.
- Qwen can also start a streamed XML tool call, emit only a partial JSON argument prefix such as `{`, then hit `max_tokens`. OpenCode receives that as an invalid tool call unless Rapid-MLX buffers and validates the stream before forwarding it.
- Qwen can repeatedly call the same read-only tool with the same arguments after already receiving that tool result, causing an agentic read loop instead of making progress.

## Changes

- Add a generic OpenAI-compatible fallback parser for bare `Calling tool: ...` calls.
- Decode JSON-like and double-encoded tool arguments before sending them to clients.
- Coerce tool arguments using JSON schema metadata, including nullable type lists, `anyOf`/`oneOf`/`allOf`, `items`, `properties`, `additionalProperties`, and enum-only schemas.
- Preserve request tool schemas across streamed content before Qwen tool markup so later arguments are coerced with the correct schema.
- Pass the original chat request into the streaming postprocessor so it can validate generic fallback calls against required tool arguments.
- Buffer partial generic tool-call markers across chunks and strip leaked marker prefixes only when they appear at the start of a line.
- Preserve normal code brackets split at chunk boundaries, such as `game.snake[` and `const items = [`.
- Drop malformed generic tool calls that are missing required schema arguments, preventing client schema validation failures from incomplete duplicate calls.
- Preserve native tool messages for parsers that support them, even when parser settings are still only present in the CLI globals during startup.
- Add a hidden continuation nudge after tool-result messages so local tool models continue the original task instead of ending with planning text.
- Add streaming retry for tool-using turns: if the model emits only content/reasoning, the stream exhausts without a tool call, or repeated text is detected before a tool call, Rapid-MLX retries with an internal instruction to call the required tool. Buffered narration/repetition is not sent to the client unless retries are exhausted.
- Suppress repeated buffered text after retry budget is exhausted so clients do not receive runaway repeated tokens.
- Disable Qwen thinking on continuation turns unless the client explicitly requested `enable_thinking`.
- Buffer streamed OpenAI tool-call chunks until their function arguments assemble into a complete valid JSON object. If generation stops with incomplete JSON, suppress the partial chunks and retry internally with a targeted tool-call JSON repair prompt.
- Detect repeated same-argument read-only tool calls after a tool result and nudge the model to use the existing result, make a state-changing/diagnostic tool call, or finish.
- Add regression coverage for OpenCode/Hermes-style tool schemas, partial stream markers, generic fallback calls, malformed duplicate calls, normal code bracket preservation, continuation prompt injection, stream-exhausted continuation retry, repeated-text retry, repeated-read nudging, and truncated streamed tool-call JSON retry.

## Validation

- `uv run ruff check vllm_mlx/routes/chat.py vllm_mlx/service/helpers.py tests/test_chat_stream_tool_continuation.py tests/test_server_utils.py` -> passed.
- `uv run pytest tests/test_chat_stream_tool_continuation.py tests/test_server_utils.py::TestAppendToolContinuationPrompt` -> 9 passed.
- Earlier focused suite: `uv run pytest tests/test_chat_stream_tool_continuation.py tests/test_server_utils.py::TestAppendToolContinuationPrompt tests/test_postprocessor.py` -> 78 passed.
- Previous broader parser/model-regression suites were run during the PR iteration and passed: 103 focused tests and 267 parser/postprocessor regression tests.
- Applied the patch locally to the Homebrew Rapid-MLX 0.6.1 install and restarted `rapid-mlx serve qwen3.6-35b`.
- Verified startup logs native tool format enabled for `qwen3_coder_xml` and reasoning parser `qwen3`.
- Re-ran OpenCode against the local OpenAI-compatible endpoint with the requested snake-game TDD prompt. OpenCode completed with exit 0 and no client-visible API translation/schema errors such as `tool: invalid`.
- During that run the new truncated-tool-call guard was exercised once: Rapid-MLX detected `incomplete tool call JSON`, suppressed the partial streamed arguments, retried internally, and OpenCode received the subsequent valid `write` tool call.
- The generated sample app still had ordinary model-quality issues after completion when checked manually; those are not API translation failures and were not addressed by this server-side PR.

Fixes #160.

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,6 +555,36 @@ Large-context requests auto-route to a cloud LLM (GPT-5, Claude, etc.) when loca
 
 Vision, audio (STT/TTS), video understanding, and text embeddings — all through the same OpenAI-compatible API.
 
+### DFlash Speculative Decoding
+
+Run a target model alongside a small **DFlash drafter** for block-based draft+verify speculative decoding (cross-attention to selected target hidden states). Works for Qwen 3.6 family with the matching DFlash drafter checkpoint. Single-request only in this mode (continuous batching disabled). Adaptive block sizing (8–22 by default) tunes itself per-request based on observed acceptance ratio.
+
+Install the drafter runtime once (uses the `dflash[mlx]` package from the upstream repo):
+
+```bash
+pip install -e /path/to/dflash[mlx]
+```
+
+Then point `--drafter` at a DFlash drafter checkpoint:
+
+```bash
+rapid-mlx serve /path/to/Qwen3.6-35B-A3B-4bit \
+  --drafter /path/to/Qwen3.6-35B-A3B-DFlash \
+  --port 8010
+```
+
+The server exposes the same `/v1/chat/completions` API as the regular path. `/v1/status` adds a `dflash` block with the lifetime acceptance ratio, current block size, and observed bounds — also surfaced live in `--tui`.
+
+Mutually exclusive with `--enable-mtp` and `--mllm`.
+
+### Live TUI Monitor
+
+`--tui` opens a full-screen dashboard alongside the server: status, model, Metal memory, prompt/output token totals, the last completed request (gen tok/s, prefill tok/s, ttft, finish reason), per-request averages, recent requests, message previews, and (when DFlash is enabled) lifetime acceptance ratio + adaptive block-size bounds. Press `q` or Ctrl-C to exit.
+
+```bash
+rapid-mlx serve <model> --tui
+```
+
 Also: logprobs API, structured JSON output (`response_format`), continuous batching, KV cache quantization (`--kv-cache-quantization`), and [2100+ tests](tests/).
 
 ---
@@ -597,6 +627,23 @@ Also: logprobs API, structured JSON output (`response_format`), continuous batch
 |------|-------------|---------|
 | `--cloud-model` | litellm model string (e.g. `openai/gpt-5`) | *(disabled)* |
 | `--cloud-threshold` | New token threshold to trigger cloud routing | `20000` |
+
+### DFlash Speculative Decoding
+
+| Flag | Description | Default |
+|------|-------------|---------|
+| `--drafter` | Path or HF id of a DFlash drafter checkpoint. Triggers DFlash mode. | *(disabled)* |
+| `--dflash-block-size` | Override the drafter's default block size | *(from drafter config)* |
+| `--dflash-no-adaptive` | Disable adaptive block sizing (adaptive is on by default) | off |
+| `--dflash-block-min` | Adaptive block-size lower bound | `8` |
+| `--dflash-block-max` | Adaptive block-size upper bound | `22` |
+| `--dflash-turboquant-bits` | KV-cache TurboQuant bits for the target (requires `mlx-turboquant`) | *(off)* |
+
+### Monitor TUI
+
+| Flag | Description | Default |
+|------|-------------|---------|
+| `--tui` | Run a live full-screen monitor alongside the server (q quits) | off |
 
 ### Security & Other
 
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -7,6 +7,10 @@ for the vLLM-style continuous batching implementation.
 """
 
 import asyncio
+import sys
+import threading
+import types
+from types import SimpleNamespace
 from unittest.mock import MagicMock
 
 import pytest
@@ -433,6 +437,25 @@ class TestSchedulerIntegration:
         assert len(finished) == len(prompts), f"Only {len(finished)} requests finished"
 
 
+class TestEngineThreading:
+    """Threading tests for EngineCore."""
+
+    def test_mlx_step_thread_initializer_rebinds_generation_stream(self, monkeypatch):
+        """The executor thread must own mlx-lm's generation stream."""
+        from vllm_mlx import engine_core
+
+        fake_generate = types.SimpleNamespace(generation_stream="old-stream")
+        monkeypatch.setitem(sys.modules, "mlx_lm.generate", fake_generate)
+        monkeypatch.setattr(engine_core.mx, "default_device", lambda: "gpu")
+        monkeypatch.setattr(
+            engine_core.mx, "new_stream", lambda device: f"stream:{device}"
+        )
+
+        engine_core._init_mlx_step_thread()
+
+        assert fake_generate.generation_stream == "stream:gpu"
+
+
 @pytest.mark.asyncio
 class TestEngineAsync:
     """Async tests for the engine."""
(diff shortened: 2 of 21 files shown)


raullenchai:
Thanks for the very detailed write-up @samuelfaj. Reviewed the diff in full. Headline: the tool-calling work has real merit, but the PR is **way over scope** for what the title/description claim, and one of the bundled features adds an unpublished hard dependency. Asking to split before any further review.

### CRITICAL — blocks merge

1. **Scope creep — three unrelated features bundled with the tool-call fix.** The PR title/description say "Fix Qwen tool-call OpenAI translation / Fixes #160", but the diff also adds:
   - `vllm_mlx/tui.py` (538 LoC, brand-new live TUI)
   - `vllm_mlx/engine/dflash.py` (413 LoC, brand-new speculative-decoding engine) + CLI flags + README section + `engine/__init__.py` export
   - `vllm_mlx/middleware/metrics.py` (231) + `vllm_mlx/request_metrics.py` (189) + a new `/v1/requests` route — wired into `server.py` as an **always-on** middleware with no flag
   
   Please split into 3 PRs: (a) the actual #160 tool-call fix, (b) DFlash, (c) TUI + metrics recorder. We can land (a) quickly; (b) and (c) need their own design discussions.

2. **DFlash depends on an unpublished package.** `dflash` is not on PyPI (404) and `github.com/dflash/dflash` doesn't exist (404). The README tells users to `pip install -e /path/to/dflash[mlx]`, and `engine/dflash.py:103` hardcodes your personal path (`/Users/samuelfajreldines/dev/dflash[mlx]`) into the user-facing error. We can't ship a documented `--drafter` flag whose dependency only exists on your laptop.

### HIGH

3. **`_append_tool_continuation_prompt` silently injects a `role: "user"` message into the user's chat history.** That message is forwarded to `engine.stream_chat`, counts toward `prompt_tokens` reported back, and is invisible to the caller. Hidden prompt mutation is a sharp footgun for anyone debugging token usage or auditing requests. At minimum: gate behind a flag, document it, and don't bill the synthesized tokens.

4. **Streaming buffer kills TTFT for every tool-call request.** `max_tool_continuation_retries = 2 if (tool_continuation_retry or tool_retries_enabled) else 0` activates on **any** request with `tools=[...]`, not only after a tool result. Combined with `_buffered_tool_calls_complete`, the server now buffers all content/reasoning events until either a complete tool call assembles or the stream ends. That destroys incremental rendering for agentic clients on the happy path.

5. **`MetricsMiddleware` is added unconditionally** (`server.py:331`) — no opt-out, no test for the always-on perf cost.

### MEDIUM

6. Hardcoded couplings to one client's conventions:
   - `_has_repeated_recent_read_tool_call` only triggers on tool name `"read"` (helpers.py)
   - `_is_tool_result_message` matches `[Tool Result` / `<tool_response>` literals
   - thinking-disable rule keys on `cfg.tool_call_parser == "qwen3_coder_xml"` only
   
   These read like OpenCode-specific fixes embedded in the server. Worth refactoring so the heuristics are pluggable per parser.

7. **`_should_detect_tool_calls` activates fallback parsing even when no `--tool-call-parser` is configured** (any content with `Calling tool:` triggers it). That changes behavior for users who explicitly opted out of tool parsing.

8. **`_is_repetitive_tool_text` re-scans the entire buffered text on every event** — Counter + ngram work per chunk → O(n²) on long outputs.

### LOW / NIT

9. `dflash.py` redoes the `generation_stream = mx.new_stream(...)` thread-init that PR #161 just centralized. If DFlash lands, please share the helper.
10. PR description claims `9 + 78` passing tests; we have 2100+. Please run the full suite (`make check`) on the split #160 PR.
11. `from .middleware.metrics import ...` at `server.py:331` (mid-file, with `noqa: E402`) — move to top-level once split.

### What I'd land happily

The actual `#160` fix — `api/tool_calling.py`, `tool_parsers/qwen3coder_tool_parser.py`, the schema-aware argument coercion in `service/postprocessor.py`, and the partial-marker handling — is solid work. Split that out (no continuation-prompt injection, no buffered-stream retry; or both behind a feature flag) and I'll review for merge in the next pass.

Holding off on a deeper line-by-line until we resolve scope. Thanks again for tracking down all the OpenCode edge cases.

raullenchai:
Thanks @samuelfaj! Several pieces here overlap with what we shipped in v0.6.5 (PR #172) for issue #171 — specifically the streaming tool-call leak fix where `StreamingPostProcessor` now forwards `request` (with `tools` + `tool_choice`) to `extract_tool_calls_streaming()` and the finalize fallback. `Qwen3CoderToolParser` then gets the schema for type conversion.

Could you rebase against current main and check which scenarios from your description are still uncovered? Specifically these look genuinely useful and worth a focused PR:

- bare `Calling tool: name({...})` text fallback
- double-encoded JSON args / nullable schema types (`type: ["array", "null"]`)
- streaming chunk split on `Calling tool` / `[` markers (without breaking normal code brackets)
- duplicate-incomplete-then-valid handling
- post-tool narration/planning recovery
- ServerConfig sync ordering for native tool-message preservation

PR is currently DIRTY against main. Leaving open for your call — close if no longer needed after rebase, or split into a smaller targeted PR for the remaining items.

samuelfaj:
@raullenchai  https://github.com/raullenchai/Rapid-MLX/pull/204


:)

PR #176, opened 2026-05-02
feat: add self-managed Ollama comparison benchmark

### LINKED ISSUE #149: feat: automated Ollama comparison benchmark
## Motivation

Users ask "how much faster than Ollama?" — we claim 2-4x but have no public reproducible benchmark. A comparison script would:

1. Build credibility (show, don't tell)
2. Catch performance regressions across releases
3. Provide marketing material for README/blog

## Proposed

`scripts/bench_vs_ollama.py` — runs same prompts on both servers, reports:
- TTFT (time to first token)
- Decode tok/s
- Multi-turn latency
- Concurrent throughput (1/2/4 users)

Models: Qwen3.5-4B, Qwen3.5-9B (both available in Ollama and MLX)

Output: markdown table for README.

michaelasper:
## Summary
- Adds `scripts/bench_vs_ollama.py`, a self-managed Rapid-MLX vs Ollama benchmark for the Qwen3.5 4B and 9B model pairs requested in #149.
- Launches one engine at a time on temporary localhost ports, then runs deterministic no-thinking stream, multi-turn, and concurrency workloads against both engines.
- Writes raw JSON and README-ready Markdown reports under `reports/benchmarks/ollama-comparison/`, including runtime metadata, workload errors, and speedup comparisons.
- Keeps generated reports out of git; the benchmark output directory is ignored so local runs do not accidentally become PR artifacts.
- Keeps embeddings behind `--include-embeddings` so the default Qwen chat benchmark stays focused on generation metrics.
- Rebases the PR onto current `origin/main` (`d62504f`) and addresses the methodology review feedback.

## Review Updates
- Multi-turn now sends four real user turns, rejects empty assistant content, uses the configured `--max-tokens` budget, and repeats/averages with `--runs`.
- Ollama runs now default `OLLAMA_KEEP_ALIVE=24h` while still allowing an explicit override.
- Ollama version metadata strips daemon warning lines and keeps a clean client-version fallback.
- Stream metrics mark missing-content runs as incomplete instead of reporting decode throughput from the full request window.
- Managed server startup now detects early process exit and retries with a fresh port, reducing opaque failures from port races.
- Removed dead benchmark code (`prepare_models` and the unused concurrent throughput helper).
- Removed the checked-in benchmark report artifacts; generated reports are local outputs only.

## Latest Benchmark Run
Ran `python3.12 scripts/bench_vs_ollama.py` on commit `2e4bee5` after rebasing onto `origin/main`.

- Timestamp: `2026-05-05T18:24:11`
- Host: Mac Studio, Apple M3 Ultra, 32 CPU cores, 256 GB memory
- Python: `3.12.13`
- Platform: `macOS-26.4.1-arm64-arm-64bit`
- Rapid-MLX: `rapid-mlx 0.6.6`
- Ollama: `ollama version is 0.21.0`
- Runs/warmups: `3` / `1`
- Max tokens: `256`
- Concurrency: `[1, 2, 4]`
- Embeddings: skipped by default

TTFT is reported for completeness, but it includes request/server overhead for this prompt workload. The primary comparison numbers are decode tok/s and concurrent throughput.

### qwen3.5-4b vs qwen3.5:4b

| Metric | Rapid-MLX | Ollama | Speedup |
|---|---:|---:|---:|
| TTFT | 38.1 ms | 155.1 ms | 4.07x |
| Decode tok/s | 176.5 | 63.9 | 2.76x |
| Multi-turn latency | 919.7 ms | 1880.6 ms | 2.04x |
| Concurrent throughput (1 user) | 173.8 | 60.2 | 2.89x |
| Concurrent throughput (2 users) | 254.7 | 60.9 | 4.18x |
| Concurrent throughput (4 users) | 328.5 | 61.2 | 5.37x |

### qwen3.5-9b vs qwen3.5:9b

| Metric | Rapid-MLX | Ollama | Speedup |
|---|---:|---:|---:|
| TTFT | 25.4 ms | 172.5 ms | 6.79x |
| Decode tok/s | 117.8 | 53.2 | 2.21x |
| Multi-turn latency | 1479.3 ms | 2403.3 ms | 1.62x |
| Concurrent throughput (1 user) | 115.8 | 51.0 | 2.27x |
| Concurrent throughput (2 users) | 180.8 | 51.3 | 3.52x |
| Concurrent throughput (4 users) | 225.2 | 51.5 | 4.37x |

The generated report files were left as local ignored outputs only:
- `reports/benchmarks/ollama-comparison/20260505_182411.json`
- `reports/benchmarks/ollama-comparison/20260505_182411.md`

## Validation
- `python3.12 -m pytest tests/test_bench_vs_ollama.py -q`
- `.venv/bin/python -m ruff check scripts/bench_vs_ollama.py tests/test_bench_vs_ollama.py`
- `.venv/bin/python -m black --check scripts/bench_vs_ollama.py tests/test_bench_vs_ollama.py`
- `git diff --check`

Closes #149


raullenchai (commented):
Thanks @michaelasper — the script and test suite are well-written (stdlib-only, list-form subprocesses, good edge-case coverage on the SSE parsers). A few things before merging:

**1. Dead stub — `prepare_models()` is never called**
```python
def prepare_models(args: CliArgs) -> None:
    _ = args  # no-op
```
Never invoked from `run_benchmark`. Remove it or wire it in.

**2. Committed report is on v0.6.6 — four releases stale**
`reports/benchmarks/ollama-comparison/20260502_104639.*` was generated with `rapid-mlx 0.6.6`; current release is v0.6.10. These will be the 'official' numbers until re-run. Please regenerate on current main before merge, or add a note in the report that it's pinned to 0.6.6.

**3. Ollama version field captures warning text**
The committed JSON shows:
```
"ollama_version": "Warning: could not connect to a running Ollama instance\nWarning: client version is 0.20.2"
```
`run_capture` merges stdout+stderr, so when Ollama isn't running at metadata-collection time the version field gets the warning prepended. A `strip().split('\n')[-1]` or similar would isolate just the version line.

---

**Minor / informational:**

- `find_free_port()` has a TOCTOU window between releasing the socket and the server binding it — well-known limitation, just worth a comment so future readers understand.
- PR validation instructions use `uv run --extra dev`, but the project convention is `python3.12 -m pytest`.

Happy to merge once 1–3 are addressed.

raullenchai (changes requested):
Thanks for taking on this benchmark — it answers #149 cleanly and the test coverage (53 cases, 0.26s) is genuinely good for a script of this size. The process management is also solid: subprocess list args (no shell), `127.0.0.1` binding, terminate→wait→kill cleanup, and an explicit `OLLAMA_HOST` override guard. No production code touched.

Adversarial review (manual + DeepSeek V4 Pro) surfaced **one issue that invalidates a headline number in the committed report**, plus a handful of smaller bugs. None require an architectural rework.

### 🚨 Blocking — invalidates the committed numbers

**1. `max_tokens` is silently capped to 128 for the multi-turn workload only**
*File:* `scripts/bench_vs_ollama.py` line 1150

```python
raw_runs["multi_turn"] = run_multi_turn(
    ...,
    min(workload["max_tokens"], 128),
    ...
)
```

With the default `--max-tokens 256`, multi-turn runs at 128 while stream and concurrent runs at 256. The PR description claims "identical workloads"; the committed report's `Multi-turn speedup: 2.60x / 1.65x` is therefore not directly comparable to the other speedups. Either:

- Pass `workload["max_tokens"]` through unchanged (recommended), or
- Document the cap explicitly in the report header and explain why multi-turn needs a different budget.

If the cap exists for a real reason (some Ollama timeout?), please regenerate the committed report after the fix so future readers don't compare against stale apples-to-oranges numbers.

### 🟡 Should fix (real, small)

**2. `find_free_port` has a TOCTOU race**
*File:* `scripts/bench_vs_ollama.py` line 187

```python
def find_free_port() -> int:
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
        sock.bind(("127.0.0.1", 0))
        return int(sock.getsockname()[1])
```

The socket is closed (the `with` block exits) before the subprocess binds. Another process can grab the port in the gap. On a quiet dev machine this is rare; on CI or a busy laptop the symptom is an opaque "server failed to start" timeout 300s later. Fix options:

- Retry on `EADDRINUSE` with a fresh port (cheap and correct), or
- Pass the bound socket to the subprocess via `Popen(..., pass_fds=...)` and have it inherit the FD.

**3. Ollama version metadata captures the "daemon not running" warning as the version string**
*File:* `scripts/bench_vs_ollama.py` line 332 (`collect_metadata`)

`run_capture(["ollama", "--version"])` falls back to stderr when stdout is empty, and when the daemon hasn't started yet `ollama --version` writes a warning to stderr. Result, visible in the committed report:

```
Ollama: `Warning: could not connect to a running Ollama instance
Warning: client version is 0.20.2`
```

Fix: in `collect_metadata`, capture ollama_version *after* the first ollama benchmark run starts the daemon, OR strip lines starting with `Warning:` from the captured output.

**4. `run_concurrent_throughput` is dead code**
*File:* `scripts/bench_vs_ollama.py` line 884

Defined but never called — `_run_concurrent_chat_batch` (line 936) is the live one. Dropping it removes ~30 LOC of duplicated logic and one less code path to maintain.

**5. `prepare_models` is a no-op stub**
*File:* `scripts/bench_vs_ollama.py` line 257

```python
def prepare_models(args: CliArgs) -> None:
    _ = args
```

Dead. Either delete it or actually implement (the per-engine `prepare_rapid_mlx_model` / `prepare_ollama_model` functions already exist and are called from the benchmark functions, so this top-level no-op seems left over from refactoring).

### 🟢 Worth a thought (not blocking)

**6. `build_stream_metric` distorts metrics when `first_content_at` is None**
*File:* `scripts/bench_vs_ollama.py` line 583

When the parser never sees a content chunk:
```python
ttft_s   = end_at - start_at
decode_s = end_at - start_at  # same window!
decode_tok_s = parsed.completion_tokens / decode_s
```

Both TTFT and decode_tok_s become "tokens divided by total time" — silently mislabeling. Rare in chat (the parser would have to miss everything), but worth treating as an error metric or marking the run as `incomplete=true` in the JSON so downstream readers don't trust the numbers.

**7. Committed reports are not gitignored**
*File:* `.gitignore`

The PR commits `reports/benchmarks/ollama-comparison/20260502_104639.{json,md}` as a "first official run" reference (~1378 lines of JSON), but `.gitignore` doesn't add the directory pattern. Subsequent runs will land in the same directory and risk being committed by accident. Suggest:

```diff
+ reports/benchmarks/ollama-comparison/*.json
+ reports/benchmarks/ollama-comparison/*.md
+ !reports/benchmarks/ollama-comparison/20260502_104639.*
```

(or whatever pattern preserves the canonical baseline while ignoring future runs.)

---

### Summary

This is a contained, well-tested benchmark utility — happy to land it. Just need #1 fixed (and the report regenerated) before merge so the numbers we publish are honest. #2–5 are quick follow-ups; #6–7 are nice-to-have.

Thanks for the work — looking forward to having a reproducible Rapid-MLX vs Ollama story published.

raullenchai:
Thanks for this — closing #149 with a self-managed bench (no daemon dependency, both engines run sequentially with the same warmup) is exactly the right shape, and the 1452 LOC of script + 1069 LOC of tests is genuinely high-quality work. Pure-stdlib imports, no new deps, 0 production code touched, supply-chain clean.

That said, before we ship this as a public benchmark / marketing artifact, I want to push on the **methodology** so the numbers we publish are defensible against a skeptic running their own bench. Several of these came from the DeepSeek V4 Pro adversarial review, several from going through the script myself.

## Blocking — methodology bugs

These directly produce wrong/unfair numbers:

1. **Multi-turn loops 4 turns but `MULTI_TURN_FOLLOWUPS` only has 3 strings.** The 4th turn's request has no new user message — chat is malformed and the model will either error or generate garbage that gets averaged into the latency. (`scripts/bench_vs_ollama.py:1115–1133`) Fix: change the loop to `range(3)` or add a 4th followup.
2. **`run_multi_turn` caps `max_tokens` at `min(max_tokens, 128)` while stream/concurrency use the CLI value (default 256).** This silently makes multi-turn easier on whichever engine is faster at short outputs and harder on whichever amortizes long outputs better — apples-to-oranges. (`scripts/bench_vs_ollama.py:1410`) Fix: use the same cap everywhere, or document why.
3. **Empty assistant content is appended on errors and cascades.** If `extract_rapid_mlx_message_content` / `extract_ollama_message_content` returns an empty string because the upstream returned an error object, the loop appends \`{\"role\": \"assistant\", \"content\": \"\"}\` and continues — subsequent turns then ride a polluted conversation, and the bench reports a \"successful\" multi-turn latency for a broken run. (`scripts/bench_vs_ollama.py:1121-1129`) Fix: detect empty/invalid content and abort or record as workload error.
4. **Multi-turn runs once, stream runs `runs_per_level` times averaged.** Single-sample multi-turn latency has poor statistical confidence — one cold-start or GC stall and the headline number swings. Fix: repeat multi-turn the same number of times as the stream workload.

## Strong recommendation — bench rigor

These don't break the bench but make the headline numbers misleading:

5. **\`OLLAMA_KEEP_ALIVE\` is not set.** Ollama defaults to a 5-minute keep-alive. A full run (4 workloads × 2 model pairs × \`runs_per_level\` × concurrency levels) easily exceeds that, and Ollama silently unloads the model between workloads, then takes 30+s to reload on the next request — that 30s gets attributed to TTFT or workload latency. Fix: \`os.environ[\"OLLAMA_KEEP_ALIVE\"] = \"24h\"\` before \`subprocess.Popen([\"ollama\", \"serve\"])\`.
6. **Prompts are very short** (39–75 chars, 7–11 words). On these, TTFT is dominated by per-request server overhead, not actual prefill compute — and Ollama's per-request model-routing path is genuinely slower than Rapid-MLX's resident-server path even when the underlying inference is comparable. **The 7.81x TTFT speedup in the headline is mostly server overhead amortization, not prefill speed.** A skeptic running the same comparison with a 2K-token prompt will get a much smaller number, and we'll look like we cherry-picked. Fix: add a \`--prompt-tokens N\` knob and report TTFT at multiple regimes (e.g. 50 / 1000 / 8000 input tokens).
7. **Headline framing.** Looking at the actual measurements, the **defensible wins** are *decode tok/s* (2.19–2.75x — that's BatchedEngine + MLX tuning, real and reproducible) and *concurrency scaling* (4-user 4.48–5.26x — Ollama doesn't truly batch concurrent requests, we do; very repeatable). I'd lead with those in the README and put TTFT below with the short-prompt caveat: *\"TTFT speedup includes Rapid-MLX's daemon-resident-server advantage over Ollama's per-request routing — most of this gap closes for prompts ≥1K tokens.\"*

## Process notes

- **\`pyproject.toml\` shows a 0.6.11 → 0.6.6 downgrade** in the diff. That's a phantom — your branch is just behind main. A rebase will drop it.
- **\`tests/test_batching_deterministic.py::test_concurrent_same_prompt\` failed in CI.** Verified on main right now — this is an unrelated pre-existing flake, not this PR. Will follow up separately.
- **DeepSeek truncated the 1452-LOC script** during review (only 1 file fully reviewed of 5). The findings above came from the truncated portion; once items 1–4 are fixed, a re-review will cover the rest.

## Summary

If you fix items 1–4 (real bugs) and adopt 5–7 (rigor), you'll have a benchmark that holds up to scrutiny on r/LocalLLaMA / HN — and we can lead with this in the README. The current numbers aren't *wrong*, they're just **about a different thing than the headline implies**, and the defensible decode + concurrency story is already in your data.

Happy to push the rebase / item 5 (OLLAMA_KEEP_ALIVE) myself if you want — just say the word.

michaelasper:
Response to methodology comment `4380311453`: items 1-5 and 7 are addressed in this PR; item 6 was handled with caveated framing rather than a prompt-length sweep.

1. Four-turn multi-turn bug: replaced the split start/followup structure with one explicit `MULTI_TURN_USER_MESSAGES` list containing all four user turns. Each loop iteration now appends a real new user message.

2. Multi-turn token cap: removed the `min(max_tokens, 128)` cap. Multi-turn, stream, and concurrency now use the same configured token budget.

3. Empty assistant content: `run_multi_turn()` now rejects empty assistant content before appending it to the conversation. `run_engine_suite()` records that exception as a workload error rather than reporting a successful multi-turn latency.

4. Multi-turn sample count: multi-turn now repeats under the same `runs_per_level` loop as stream, and the repeated samples are averaged into the summary.

5. Ollama keep-alive: `build_ollama_environment()` now defaults `OLLAMA_KEEP_ALIVE=24h` for the managed Ollama process, while still allowing an explicit `--ollama-env OLLAMA_KEEP_ALIVE=...` override.

6. Short prompts / TTFT: I did not add a `--prompt-tokens` sweep in this PR. Instead, the generated Markdown and PR body now explicitly caveat TTFT as request/server overhead for this prompt workload and name decode tok/s plus concurrent throughput as the primary cross-engine metrics. A prompt-length sweep still makes sense as a follow-up before using this as a broader marketing artifact.

7. Headline framing: the PR body now posts fresh benchmark numbers with decode throughput and concurrency as the defensible headline, while TTFT is included with the caveat above.

Process notes:
- Rebased onto current `origin/main`; the phantom `pyproject.toml` downgrade is gone.
- The unrelated `tests/test_batching_deterministic.py::test_concurrent_same_prompt` CI flake was not changed here.
- Generated reports are ignored/local only; no benchmark report artifacts are checked in.

Verification:
- `python3.12 -m pytest tests/test_bench_vs_ollama.py -q` -> 63 passed
- `.venv/bin/python -m ruff check scripts/bench_vs_ollama.py tests/test_bench_vs_ollama.py` -> passed
- `.venv/bin/python -m black --check scripts/bench_vs_ollama.py tests/test_bench_vs_ollama.py` -> passed
- `git diff origin/main...HEAD -- pyproject.toml` -> empty

michaelasper:
Response to review `4223761209`: addressed each item in the current PR state.

1. Multi-turn `max_tokens` cap: removed the `min(..., 128)` cap. Multi-turn now uses the same `workload["max_tokens"]` value as stream/concurrency. Added coverage that verifies `256` is passed through.

2. Port race / opaque startup timeout: kept ephemeral port selection, but startup now monitors the managed subprocess. If the process exits before the health endpoint is ready, the benchmark stops it and retries with a fresh port instead of waiting for the full timeout.

3. Ollama version metadata: added `normalize_ollama_version()` so generic daemon warning lines are stripped while preserving a clean client-version fallback.

4. Dead `run_concurrent_throughput`: removed it. The maintained concurrency path is `_run_concurrent_chat_batch()`.

5. Dead `prepare_models` stub: removed it. Model preparation remains in the per-engine preparation functions that are actually called.

6. Missing stream content: `build_stream_metric()` now marks missing-content runs as `incomplete: true` and reports `decode_tok_s: 0.0` instead of silently deriving throughput from the full request window.

7. Generated reports: added ignore rules for generated benchmark JSON/Markdown and then removed the checked-in report artifacts entirely. Fresh benchmark numbers are posted in the PR body, but the generated report files remain local ignored outputs only.

Also fixed the earlier timing-basis issue: Ollama decode throughput no longer switches to `eval_duration_ns`; both engines use the same wall-clock decode basis.

Verification:
- `python3.12 -m pytest tests/test_bench_vs_ollama.py -q` -> 63 passed
- `.venv/bin/python -m ruff check scripts/bench_vs_ollama.py tests/test_bench_vs_ollama.py` -> passed
- `.venv/bin/python -m black --check scripts/bench_vs_ollama.py tests/test_bench_vs_ollama.py` -> passed
- `git -c core.fsmonitor=false ls-files reports/benchmarks/ollama-comparison` -> no tracked generated reports

raullenchai:
Thanks @michaelasper — appreciate the iteration on this one. The blocking review issues are all addressed in current head:

- ✅ `multi_turn` now passes `workload["max_tokens"]` through unchanged (no more 128 cap)
- ✅ `normalize_ollama_version()` strips the daemon-not-running warnings
- ✅ `prepare_models` and `run_concurrent_throughput` dead code removed
- ✅ `build_stream_metric` now flags `incomplete=true` when `first_content_at is None`
- ✅ `PORT_BIND_ATTEMPTS=3` retry loop on bind failure addresses the `find_free_port` TOCTOU window

I rebased on current main (28 commits, no conflicts) and ran our pr_validate harness:

| step | status | detail |
|---|---|---|
| `fetch` | PASS | 3 files, +2911/-0 LOC, blast=medium |
| `supply_chain` | PASS | no install hooks, deps clean, stdlib-only |
| `lint` | PASS | ruff check + format clean |
| `targeted_tests` | PASS | 63 passed in 0.17s |
| `full_unit` | PASS | 2244 passed, 17 skipped in 26.16s |

CI green on rebased head c4cb08d. Merging. 

I'll open a small follow-up PR adding a `find_free_port` docstring noting it must be paired with the retry loop (the only nit left from review). Welcome to the project — looking forward to more contributions.

Issue #170, opened 2026-05-01
fix(gemma4): RuntimeError "There is no Stream(gpu, 1) in current thread" on every request


stone30513:
## Rapid-MLX version

0.6.4 (Homebrew, built 2026-04-29)

## Hardware

MacBook Pro M1 Max, 32 GB, macOS 15.x

## Model

`mlx-community/gemma-4-26b-a4b-it-4bit` (Gemma 4 26B A4B, MoE)

## Full serve command

```bash
rapid-mlx serve mlx-community/gemma-4-26b-a4b-it-4bit \
  --served-model-name gemma-4-26b-a4b \
  --enable-auto-tool-choice \
  --tool-call-parser gemma4 \
  --reasoning-parser gemma4 \
  --kv-cache-turboquant \
  --cache-memory-mb 1024 \
  --max-tokens 64
```

Also fails with minimal flags (no tool/reasoning parser, no turboquant, no prefix cache, no paged cache, `--max-num-seqs 1`, `--no-gc-control`).

## What happened?

Every request fails with zero-length error response. Server logs:

```
ERROR:vllm_mlx.scheduler:Error in batch generation step: There is no Stream(gpu, 1) in current thread.
RuntimeError: There is no Stream(gpu, 1) in current thread.

  File "vllm_mlx/scheduler.py", line 2406, in step
    raw_next = self.batch_generator.next()
  File "mlx_lm/generate.py", line 1855, in next
    return self._next()
  File "mlx_lm/generate.py", line 1841, in _next
    self._prompt_batch.prompt(prompts)
  File "mlx_lm/generate.py", line 1161, in prompt
    mx.eval([c.state for c in self.prompt_cache])
RuntimeError: There is no Stream(gpu, 1) in current thread.
```

The step thread initializes `Stream(Device(gpu, 0), N)` but the error references `Stream(gpu, 1)` — the main thread's default stream. Model/KV cache arrays created during startup carry the main thread's stream tag; the step thread can't eval them.

## Working models for comparison

- `unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit` — works on same hardware
- `mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit` — works on same hardware

Both were fixed by #161. Gemma 4 26B A4B still fails.

## Root cause

[...]

PR #182, opened 2026-05-03
fix(engine_core): route add_request through mlx-step worker (#170)

MishkinBerteig:
## Summary

Closes the third leg of the `Stream(gpu, N)` bug class for #170. PR #173 (warmup) and #174 (model load) covered the engine lifecycle paths; this PR covers the per-request hot path that breaks every text-only model with a populated prefix cache.

Reproducer: `MiniMax-M2.7-8bit` on rapid-mlx 0.6.6 with `--pin-system-prompt` (which loads prefix-cache entries from disk at startup → first request is a hit). Every request fails with `RuntimeError: There is no Stream(gpu, N) in current thread.` from inside `batch_generator.next()`. Also reproduces on 0.6.7 — the v0.6.7 release (#178) only touched the `tool_parsers/` tree and didn't change this code path.

## Root cause

`EngineCore.add_request()` ran `scheduler.add_request()` inline on the asyncio loop thread. `scheduler.add_request` walks the prefix cache:

- `memory_aware_cache.fetch` does `copy.deepcopy(entry.cache)` (`memory_cache.py:587`)
- `paged_cache.reconstruct_cache` materializes block tensors

mlx-lm 0.31.3+ tags every `mx.array` with the **calling thread's** default stream. With cache hits, those arrays end up tagged with the loop thread's stream. The next `batch_generator.next()` runs on the mlx-step worker and crashes inside `mx.eval([c.state for c in self.prompt_cache])`.

Symptom pattern matches expectation:

- Cold cache → first request misses (no MLX work in `fetch`) → succeeds
- Now cache has entries → second request hits → fails
- `--pin-system-prompt` pre-loads entries from disk → first request already hits → fails immediately

## What changed

- `vllm_mlx/engine_core.py`: dispatch `self.scheduler.add_request` through `_mlx_executor` via `loop.run_in_executor` (the same single-thread executor used for `step()`). Falls back to inline when no executor is bound — matches the pattern from `_run_on_step_thread()` (cache-save path, PR #167).
- `tests/test_engine_step_thread.py`: two regression tests in `TestStepThread` covering the dispatched path and the no-executor fallback.

## Test plan

- New regression tests pass (`test_add_request_routes_to_worker`, `test_add_request_falls_back_when_executor_missing`).
- Full `TestStepThread` + `TestBatchedEngineWarmup` (10/10 pass).
- `pytest tests/ --ignore=tests/integrations`: **1675 passed**, 9 pre-existing failures (4 `test_batching_deterministic` with the same `Stream(gpu, N)` class on the concurrent-batched code path with multiple `AsyncEngineCore` instances — *not addressed by this PR*; 4 `test_platform` with `ModuleNotFoundError: torch`; 1 `test_reasoning_parsers` parser assertion). All 9 reproduce on stock main (0.6.6 and 0.6.7) without this patch — none of them touch `engine_core.py`.
- `ruff check` + `ruff format` clean on both files.
- End-to-end on `MiniMax-M2.7-8bit` (230 GB MoE) with `--pin-system-prompt --kv-cache-quantization-bits 8 --max-num-seqs 2` and 2 prefix-cache entries pre-loaded from disk (worst case for triggering the bug): three back-to-back `/v1/chat/completions` all return HTTP 200, zero stream errors in the rapid-mlx log, generation at 33.6 tok/s. Subsequent two-back-to-back test confirmed the cache-hit path is fast (request 1: 2.5 s prefill + decode, request 2: 0.22 s with cached prefix).

## Out of scope

The 4 `test_batching_deterministic` failures appear to be a related bug in the concurrent-batched code path (multiple `AsyncEngineCore` instances per process). Pre-patch they fail at `Stream(gpu, 1)`; post-patch they fail at `Stream(gpu, 4)` — same root class, different code path. Worth a separate PR.

### DIFF
--- a/tests/test_engine_step_thread.py
+++ b/tests/test_engine_step_thread.py
@@ -154,6 +154,68 @@ class TestStepThread:
         finally:
             await engine_core.stop()
 
+    @pytest.mark.asyncio
+    async def test_add_request_routes_to_worker(self, engine_core, monkeypatch):
+        """add_request() MUST run scheduler.add_request on the mlx-step worker.
+
+        scheduler.add_request walks the prefix cache (memory_aware_cache.fetch
+        does copy.deepcopy of cached KV state, paged_cache.reconstruct_cache
+        materializes block tensors). Those allocations get tagged with the
+        calling thread's default stream. If add_request runs on the asyncio
+        loop thread, the cached KV ends up on a stream the mlx-step worker
+        cannot mx.eval against, and the next batch_generator.next() raises
+        "There is no Stream(gpu, N) in current thread" inside
+        `mx.eval([c.state for c in self.prompt_cache])`.
+
+        This is the third leg of the #170 fix (after #173 warmup and #174
+        model load). Without it, every text-only model with a populated
+        prefix cache (e.g. --pin-system-prompt loaded entries from disk,
+        or a prior request's system prompt) breaks on the next request.
+        """
+        from vllm_mlx import engine_core as ec
+
+        monkeypatch.setattr(ec, "_init_mlx_step_thread", lambda: None)
+
+        captured = {}
+
+        def fake_add_request(request):
+            captured["thread"] = threading.current_thread().name
+            captured["request_id"] = request.request_id
+
+        engine_core.scheduler.add_request.side_effect = fake_add_request
+
+        await engine_core.start()
+        try:
+            request_id = await engine_core.add_request("hello")
+            assert captured["request_id"] == request_id
+            assert captured["thread"].startswith("mlx-step"), (
+                f"add_request ran on {captured['thread']!r}, expected mlx-step worker. "
+                "If add_request runs on the asyncio loop thread, prefix-cache "
+                "KV deep-copies are tagged with the wrong stream and the next "
+                "step() crashes with 'There is no Stream(gpu, N) in current thread'."
+            )
+        finally:
+            await engine_core.stop()
+
+    @pytest.mark.asyncio
+    async def test_add_request_falls_back_when_executor_missing(self, engine_core):
+        """Without start(), add_request runs scheduler.add_request inline.
+
+        Sync test/CLI paths that build an EngineCore without calling start()
+        must still work — the caller will see whatever stream error MLX would
+        have raised, but no spurious AttributeError on `_mlx_executor`."""
+        captured = {}
+
+        def fake_add_request(request):
+            captured["thread"] = threading.current_thread().name
+
+        engine_core.scheduler.add_request.side_effect = fake_add_request
+
+        # Engine never started → _mlx_executor is None.
+        assert engine_core._mlx_executor is None
+        await engine_core.add_request("hello")
+        assert captured["thread"] == threading.current_thread().name
+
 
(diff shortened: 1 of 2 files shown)


raullenchai (approved):
## Approve — well-diagnosed, well-tested, surgical fix

I verified the diagnosis and the fix locally against current `main` (post v0.6.8). LGTM.

### Diagnosis verification

- Confirmed `EngineCore.add_request` calls `self.scheduler.add_request(request)` inline on the asyncio loop thread (`vllm_mlx/engine_core.py:441` on main).
- Confirmed `Scheduler.add_request` materializes MLX arrays on the calling thread:
  - `block_aware_cache.reconstruct_cache(block_table)` at `scheduler.py:1742` (paged path)
  - `memory_aware_cache.fetch(...)` → `copy.deepcopy(entry.cache)` at `memory_cache.py:587` (memory-aware path)
- Both materialize `mx.array` storage which gets tagged with the calling thread's default stream under mlx-lm 0.31.3+.
- This is the same bug class as #170 / PR #173 / PR #174 — your "third leg" framing is accurate. Internal memory note from the v0.6.6 hotfix explicitly captures the rule: *"mlx-lm 0.31.3 streams are thread-local — model + warmup + inference must run on ONE owning thread."* Per-request `add_request` clearly belongs in that set, and we missed it.

### Fix review

- Pattern matches the existing `_run_on_step_thread()` / step-loop convention (`_mlx_executor` single-thread `ThreadPoolExecutor`, fall back to inline when `None`).
- Use of `loop.run_in_executor` (vs the sync `_run_on_step_thread`) is correct for an `async def` caller — preserves cooperative scheduling.
- Defensive `None` guard preserves the sync test/CLI path that constructs `EngineCore` without `start()`.
- Doesn't alter the throttle / output-collector setup ordering — `_hybrid_throttle` still runs on the loop thread before the dispatch, which is fine.
- Serializes `add_request` against `step()` (same single-thread executor). That's the desired effect for stream-binding correctness; latency cost is one context switch.

### Local test results

- `pytest tests/test_engine_step_thread.py -v` → **10/10 PASS** including the two new tests.
- `pytest tests/ --ignore=tests/integrations --ignore=tests/test_event_loop.py` → **2054 PASS, 17 skip**, with the same pre-existing failures you flagged (`test_reasoning_parsers::test_only_start_tag_no_end` reproduces on stock `main`; `test_platform` torch ones similar). I couldn't reproduce the 4 `test_batching_deterministic` failures on M3 Ultra — `tests/test_batching.py` ran 20/20.
- `ruff check` + `ruff format --check` on both modified files → clean.

### Scheduling note

I'll merge after #179 (issue templates) is in to keep the PR queue tidy. Targeting v0.6.9 for the release. Thanks for the thorough writeup and the reproducer details — the `--pin-system-prompt` triggering the cache-hit on the first request was a particularly nice find.

The "out of scope" `Stream(gpu, 4)` failures in concurrent `AsyncEngineCore` paths sound real — happy to review a follow-up PR if you want to tackle that next.

raullenchai:
Thanks @MishkinBerteig — merged! 🎉 Added you to the changelog:

PR #186, opened 2026-05-04
fix(anthropic): bypass StreamingThinkRouter when reasoning_parser is active (#185)

### LINKED ISSUE #185: Anthropic streaming: text content misclassified as thinking_delta for reasoning models (qwen3, etc.)
## Symptom

\`tests/integrations/test_anthropic_sdk.py::Test 4_stream\` has been failing with empty output (\`AssertionError: missing markers in: ''\`) for at least v0.6.4 → v0.6.8 (current main). Anthropic SDK's \`stream.text_stream\` collects 0 chunks for any reasoning-model request.

Reproduction (PR branch + main both reproduce):

\`\`\`bash
curl -sN http://127.0.0.1:8000/v1/messages -H 'Content-Type: application/json' -d '{
  \"model\": \"qwen3.5-4b\",
  \"max_tokens\": 80,
  \"messages\": [{\"role\":\"user\",\"content\":\"Count from 1 to 5, comma-separated.\"}],
  \"stream\": true
}'
\`\`\`

Output:
\`\`\`
event: content_block_start  # type: thinking
event: content_block_delta  # thinking_delta: \"\\n\\n\"
event: content_block_delta  # thinking_delta: \"1\"
... # all of \"1, 2, 3, 4, 5\" → thinking_delta
event: content_block_stop
event: message_stop
\`\`\`

The SAME response from \`/v1/chat/completions\` (non-streaming) returns \`{\"content\": \"1, 2, 3, 4, 5\", \"reasoning_content\": null}\` — correctly classified as final content. The streaming Anthropic adapter misclassifies it as thinking.

## Root cause

\`vllm_mlx/routes/anthropic.py:343-346\`:

\`\`\`python
_starts_thinking = (
    \"<think>\" in _chat_template and \"add_generation_prompt\" in _chat_template
)
think_router = StreamingThinkRouter(start_in_thinking=_starts_thinking)
\`\`\`

For Qwen3.5-4B the chat template injects a literal \`<think>\\n\` prefix, so \`_starts_thinking=True\`. The router begins life with \`_in_think=True\` and waits for \`</think>\` to switch to text mode. But the model's actual output (after the chat-template prefix) doesn't contain a literal \`</think>\` — when the model decides not to think, it just produces the answer directly.

So the router buffers everything as \`thinking\`, never transitions, and the final \"text\" chunk is emitted via \`flush()\` as more thinking content.

There's also an interaction with the reasoning parser: when \`reasoning_parser=qwen3\` is active, the adapter at L378-401 takes only \`delta_msg.content\` (line 385) and drops \`delta_msg.reasoning\`. For \"no tags seen\" output, qwen3 parser returns \`DeltaMessage(reasoning=delta_text, content=None)\` per \`think_parser.py:148\` — so \`content=None\` → no events emitted in the per-delta loop. Yet thinking_deltas still appear during the stream, which suggests the per-delta path **isn't** the one emitting them — the events likely come from a different code path I haven't fully traced (possibly \`_emit_content_pieces\` being invoked with stale buffer state, or a duplicate path somewhere).

## Impact

- Anthropic streaming returns no \`text_delta\` events for reasoning models that don't emit a literal \`</think>\` close tag (Qwen3, GLM-4.6, future Qwen3.5/3.6 reasoning variants).
- Anthropic SDK clients see empty \`text_stream\`. Manual \`thinking_delta\` consumption works but defeats the SDK's intended ergonomics.
- Pydantic AI \`Test 2_stream\` fails with the same shape (\"Model token limit (provider default) exceeded before any response was generated\").

## Fix sketch

When \`reasoning_parser\` is active, the adapter should:

1. Disable \`StreamingThinkRouter\` entirely (the parser already separates reasoning from content), OR
2. Pass both \`delta_msg.content\` AND \`delta_msg.reasoning\` to the router with explicit block tagging, OR
3. Use \`finalize_streaming\`'s correction logic mid-stream, not just at the end.

Option 1 is simplest and matches what the \`/v1/chat/completions\` non-streaming path effectively does.

## Out of scope for this report

This is unrelated to PR #184 (Stream(gpu, N) fixes) — both fail on stock main. PR #184 should not be blocked on this; integration tests pass at the same rate before and after PR #184.

cc @raullenchai

raullenchai:
## Summary

When a reasoning parser (e.g. `qwen3`) is active on the Anthropic streaming path, the `StreamingThinkRouter` dueled with it — the router started in thinking mode (chat template has `<think>`) but the parser's `DeltaMessage.content` was `None` for reasoning output, so all content was either dropped or misclassified as `thinking_delta`. Anthropic SDK clients saw empty `text_stream`.

## Root Cause

In `_stream_anthropic_messages`, two classifiers fought each other:
1. **`StreamingThinkRouter`** — heuristic based on `<think>`/`</think>` tags in raw text. Starts in thinking mode if the chat template injects `<think>`.
2. **`reasoning_parser`** — model-aware parser that separates reasoning from content via `extract_reasoning_streaming()`.

The code only used `delta_msg.content` (ignoring `delta_msg.reasoning`) and passed it through the `StreamingThinkRouter`. When the parser returned `reasoning=text, content=None` (no tags seen yet), no SSE events were emitted at all. When content was `text` (after `</think>`), the think_router still classified it as thinking because it was stuck in `_in_think=True`.

## Fix

**Bypass `StreamingThinkRouter` when `reasoning_parser` is active** — the parser already correctly separates reasoning from content. Emit both directly:
- `delta_msg.reasoning` → `thinking_delta` SSE events
- `delta_msg.content` → `text_delta` SSE events (through tool_filter)

Key design decisions:
- When `reasoning_parser` is active: parser is the authoritative classifier, think_router is never created
- When `reasoning_parser` is NOT active: think_router remains as a heuristic fallback (unchanged behavior)
- The fix is NOT a monkey-patch — it's a clean separation of two redundant classifiers

## Test Plan

- [x] New regression test `tests/test_anthropic_streaming_reasoning.py` with 4 scenarios:
  - No think tags in output → `text_delta` via `finalize_streaming` correction
  - Both think tags → separated `thinking_delta` + `text_delta`
  - Only `</think>` (implicit think) → correct split
  - No reasoning parser fallback → unchanged behavior
- [x] All existing tests pass (2065 passed, 8 pre-existing failures unrelated to this change)
- [x] Manual reproduction: `curl -sN http://localhost:8000/v1/messages ...` with a Qwen3 model shows correct `text_delta` events

## Before / After

**Before** (bug):
```
event: content_block_start  type: thinking      ← everything is thinking
event: content_block_delta  thinking_delta: "1"
...no text_delta events ever → stream.text_stream empty
```

**After** (fixed):
```
event: content_block_start  type: thinking      ← parser sees reasoning
event: content_block_delta  thinking_delta: "..."  
event: content_block_stop
event: content_block_start  type: text          ← parser sees content
event: content_block_delta  text_delta: "1, 2, 3, 4, 5"
event: content_block_stop
→ stream.text_stream contains "1, 2, 3, 4, 5"
```

Fixes #185

### DIFF
--- a/tests/test_anthropic_streaming_reasoning.py
+++ b/tests/test_anthropic_streaming_reasoning.py
@@ -1,0 +1,331 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Regression test for issue #185: Anthropic streaming with reasoning parser.
+
+When a reasoning parser (e.g. qwen3) is active, the streaming adapter must:
+1. Emit reasoning content as thinking_delta SSE events
+2. Emit final content as text_delta SSE events
+3. NOT misclassify all text as thinking when the model outputs no think tags
+
+This test uses a mock engine that outputs raw text; it exercises the full
+streaming pipeline in _stream_anthropic_messages.
+"""
+
+import asyncio
+import json
+from typing import Any
+
+import pytest
+
+
+# ---------------------------------------------------------------------------
+# Mock helpers
+# ---------------------------------------------------------------------------
+
+class _FakeTokenizer:
+    """Tokenizer stub whose chat_template triggers _starts_thinking=True."""
+    chat_template = "<think>\n{% for msg in messages %}{{ msg.content }}\n{% endfor %}{% if add_generation_prompt %}\n<think>\n{% endif %}"
+
+
+class MockDeltaOutput:
+    """Simulates one async chunk from engine.stream_chat()."""
+    def __init__(self, new_text: str, prompt_tokens: int = 10, completion_tokens: int = 1):
+        self.new_text = new_text
+        self.prompt_tokens = prompt_tokens
+        self.completion_tokens = completion_tokens
+
+
+class MockEngine:
+    """Minimal engine that yields deltas and exposes a tokenizer."""
+    def __init__(self, deltas: list[str]):
+        self._deltas = deltas
+        self.tokenizer = _FakeTokenizer()
+        self.preserve_native_tool_format = False
+
+    async def stream_chat(self, **kwargs) -> Any:
+        for d in self._deltas:
+            yield MockDeltaOutput(d)
+
+
+# ---------------------------------------------------------------------------
+# Helper to collect SSE events from the async generator
+# ---------------------------------------------------------------------------
+
+def _collect_sse_events(gen):
+    """Drain an async generator of SSE event strings into a list."""
+    async def _collect():
+        chunks = []
+        async for chunk in gen:
+            chunks.append(chunk)
+        return chunks
+    return asyncio.run(_collect())
+
+
+def _extract_delta_types(events: list[str]) -> list[tuple[str, str | None]]:
+    """Parse SSE event strings, return list of (event_type, delta_type_or_None).
+
+    Returns tuples like ("content_block_start", "thinking"),
+    ("content_block_delta", "thinking_delta"), ("content_block_delta", "text_delta"), etc.
+    """
+    result = []
+    for evt in events:
+        if not evt.strip():
+            continue
+        for line in evt.split("\n"):
+            line = line.strip()
+            if line.startswith("event:"):
+                event_name = line.split(":", 1)[1].strip()
+            elif line.startswith("data:"):
+                data_str = line.split(":", 1)[1].strip()
+                try:
+                    data = json.loads(data_str)
+                except json.JSONDecodeError:
+                    continue
+                t = data.get("type")
+                if t == "content_block_start":
+                    cb = data.get("content_block", {})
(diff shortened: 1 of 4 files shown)


raullenchai:
Superseded by #208 (commit f1f6972), which bundled the same StreamingThinkRouter bypass plus the tool_call tag leak fix and the enable_thinking=False empty-content streaming fix.

The merged production change at `routes/anthropic.py:377-389` carries a `# Closes #185` comment confirming the fix shipped under that PR. Closing this one to avoid duplicate review.

Thanks for the original analysis — the root-cause writeup here was used in #208's commit message verbatim.

PR #187, opened 2026-05-04
fix(prefix-cache): atomic snapshot + load-time validation to stop poisoning

raullenchai:
## Summary

Disk persistence in `MemoryAwarePrefixCache` had four overlapping ways to quietly corrupt a future session's KV cache, manifesting as token-id-0 collapse (`!!!!!!!!!!!`) on every request after restart. This PR closes all four with a single design: stage writes into a sibling directory and commit via rename.

Found while debugging integration tests after a crashed v0.6.6 server left poisoned files in `~/.cache/vllm-mlx/prefix_cache/<model>/` — every Anthropic request returned `!!!!`. Manually deleting the cache fixed it; the disk format gave no signal anything was wrong.

## Bugs

| Bug | Mechanism |
|---|---|
| **A** | `save_to_disk` rewrites `entry_i.*` in place. SIGKILL mid-rewrite leaves a stale `index.json` claiming `num_tokens=X_old` next to a freshly-rewritten `entry_i_tokens.bin` of size `X_new * 4`. Next load reads `arr.fromfile(f, X_old)`, silently truncating the new tokens.bin and producing an entry where `len(tokens_key) != cache.offset`. Subsequent `fetch()` returns the misaligned cache to the scheduler, which appends new tokens at the wrong KV position → garbage attention → token id 0. |
| **B** | Orphan `entry_i.*` files from a previous larger save are never removed when the next save writes fewer entries. They sit on disk indefinitely; the next interrupted save turns them into the (A) scenario. |
| **C** | `mx.save_safetensors` writes the target path directly (no `.tmp` rename), so a SIGKILL during a single-entry write leaves a half-written safetensors. |
| **D** | `mx.load` is lazy and mmaps the safetensors body — a body-truncated file passes `load_from_disk` silently and returns zeros at missing positions. The corruption only surfaces at attention time, often inside a worker thread where the `RuntimeError` gets swallowed. (Confirmed empirically: even `mx.eval(arr)` and `arr.sum().item()` don't trigger a read past the truncation.) |

## Fix

**Atomic snapshot via directory rename** (`save_to_disk`):
- Write the full snapshot into `<cache_dir>.new/`.
- `index.json` is the **last** file written inside `.new/` — its presence signals a complete snapshot.
- Three-step swap: `cache_dir → .old`, `.new → cache_dir`, `rmtree(.old)`. Any crash before the rename leaves the previous snapshot fully intact.

**Crash recovery** (`load_from_disk`):
- If `cache_dir` is missing but `cache_dir.new/index.json` exists → promote `.new` to `cache_dir`.
- Else if only `cache_dir.old` survives → restore it.
- Otherwise: clean up any orphan staging dirs.

**Per-entry validation at load** (drop with warning, don't crash):
- `tokens.bin` file size must equal `num_tokens * 4`.
- `.safetensors` body must cover the data range declared in its header — uses a small custom validator (`_safetensors_is_complete`) that parses the 8-byte header-len + JSON header and checks `data_offsets`. Catches BUG D without relying on `mx.eval` (which mmap defeats).
- `cache.offset` must equal `len(tokens)`.

**Diagnostics**: skipped count is reported in the load summary as a `WARNING` rather than buried in per-entry warnings.

## Tests

`tests/test_prefix_cache_persistence.py` — 10 new tests, ~3s total:

- `test_clean_roundtrip_save_then_load` — sanity
- `test_stale_index_with_overwritten_entry_loads_without_error` — BUG A
- `test_poisoned_entry_returns_misaligned_cache_via_fetch` — BUG A user-visible effect
- `test_save_to_disk_removes_orphans_from_previous_save` — BUG B
- `test_severely_truncated_safetensors_is_silently_skipped` — characterizes header-corrupt drop path
- `test_body_truncated_safetensors_should_fail_eagerly_at_load` — BUG D (the surprising one)
- `test_load_recovers_from_swap_interrupted_after_first_rename` — recovery: `.new` + `.old` both present
- `test_load_recovers_from_swap_interrupted_with_only_old` — recovery: only `.old` survives
- `test_load_cleans_orphan_staging_dirs` — cleanup when `cache_dir` exists alongside leftovers
- `test_save_routes_writes_through_staging_dir` — atomicity invariant (writes never go directly into `cache_dir`)

The 4 bug-repro tests were originally written as `xfail(strict=True)` to first prove the bugs were real (each one's assertion message captures the exact symptom); after the fix they pass straight.

## Test plan

- [x] `pytest tests/test_prefix_cache_persistence.py` — 10/10 pass
- [x] `pytest tests/test_memory_cache.py tests/test_prefix_cache.py tests/test_paged_cache.py tests/test_prompt_cache_snapshot.py` — 132/132 pass
- [x] Full suite: 2080 pass, 1 pre-existing fail in `test_reasoning_parsers.py::TestQwen3::test_only_start_tag_no_end` (verified pre-existing on `main`, unrelated)
- [x] `ruff check && ruff format` clean

## Notes

- Format compatible — no on-disk schema change. An old `cache_dir/` from before this PR loads fine; a `cache_dir/` written by this PR loads on older versions too (just without crash recovery). No version bump needed for the index.
- Memory: validator reads ~the JSON header per entry on load (≤ a few KB), no impact on save.
- Doesn't address: model-identity stamp (a swap of model weights under the same path still loads stale KV), `fsync` of the parent dir (power-loss durability), cross-process locking.

### DIFF
--- a/tests/test_prefix_cache_persistence.py
+++ b/tests/test_prefix_cache_persistence.py
@@ -458,26 +458,37 @@ def test_save_handles_trailing_slash_in_cache_dir(tmp_path):
     assert c2.load_from_disk(str(cache_dir) + "/") == 1
 
 
-def test_load_into_non_empty_cache_is_rejected(tmp_path):
-    """Calling load_from_disk on a populated cache would double-count
-    memory and corrupt _sorted_keys (bisect.insort would create
-    duplicate entries). Refuse rather than silently break invariants.
+def test_load_into_non_empty_cache_skips_duplicates(tmp_path):
+    """If load_from_disk is called on a cache that already contains some
+    keys (e.g. populated by warmup before lifespan calls load), entries
+    whose tokens_key matches an in-memory entry must be skipped — not
+    re-inserted. Otherwise bisect.insort produces duplicate keys in
+    _sorted_keys and _current_memory double-counts.
     """
     cache_dir = tmp_path / "snap"
-    c1 = fresh_cache()
-    c1.store(list(range(11)), make_kvcache(num_tokens=11))
-    c1.save_to_disk(str(cache_dir))
+    # Persist two entries to disk: one duplicates a future in-memory
+    # entry; one is fresh.
+    persisted = fresh_cache()
+    persisted.store(list(range(11)), make_kvcache(num_tokens=11))
+    persisted.store(list(range(50, 61)), make_kvcache(num_tokens=11, fill=2.0))
+    persisted.save_to_disk(str(cache_dir))
 
-    c2 = fresh_cache()
-    c2.store(list(range(20, 31)), make_kvcache(num_tokens=11, fill=2.0))
-    before_mem = c2._current_memory
-    before_keys = list(c2._sorted_keys)
+    # Simulated warmup state: the [0..10] entry is already in memory.
+    runtime = fresh_cache()
+    runtime.store(list(range(11)), make_kvcache(num_tokens=11))
+    warmup_mem = runtime._current_memory
+    warmup_keys = list(runtime._sorted_keys)
 
-    loaded = c2.load_from_disk(str(cache_dir))
-    assert loaded == 0
-    # State unchanged — no duplicate insertion
-    assert c2._current_memory == before_mem
-    assert c2._sorted_keys == before_keys
+    loaded = runtime.load_from_disk(str(cache_dir))
+    assert loaded == 1, "exactly one fresh entry should have been loaded"
+    # The duplicate did not double-insert
+    assert runtime._sorted_keys.count(tuple(range(11))) == 1
+    assert tuple(range(50, 61)) in runtime._sorted_keys
+    # Memory grew by exactly the new entry's footprint
+    new_entry_mem = runtime._current_memory - warmup_mem
+    assert new_entry_mem > 0
+    # Pre-existing entry untouched in keys list ordering wrt itself
+    assert warmup_keys[0] in runtime._sorted_keys
 
 
 def test_save_routes_writes_through_staging_dir(tmp_path, monkeypatch):
--- a/tests/test_prefix_cache_persistence.py
+++ b/tests/test_prefix_cache_persistence.py
@@ -277,25 +277,38 @@ def test_severely_truncated_safetensors_is_silently_skipped(tmp_path):
 
 
 def test_body_truncated_safetensors_should_fail_eagerly_at_load(tmp_path):
-    """BUG D — load_from_disk must eagerly materialize tensors so a body-
-    truncated safetensors raises here, not later in a worker thread.
+    """BUG D — load_from_disk must reject a body-truncated safetensors
+    even though ``mx.load`` will lazily mmap it without complaint.
     """
+    import struct
+
     cache = fresh_cache()
     cache.store(list(range(11)), make_kvcache(num_tokens=11))
     cache.save_to_disk(str(tmp_path))
 
-    # Cut just the trailing bytes (body), leaving the safetensors header valid.
     sf = tmp_path / "entry_0.safetensors"
     full = sf.read_bytes()
-    # Drop the last 100 bytes — header survives, tensor body is truncated.
-    sf.write_bytes(full[:-100])
+
+    # Compute the maximum data offset declared by the header so we can
+    # truncate strictly inside the body region — guards against future
+    # changes to padding/alignment in save_prompt_cache.
+    header_len = struct.unpack("<Q", full[:8])[0]
+    header = json.loads(full[8 : 8 + header_len])
+    max_end = max(
+        meta["data_offsets"][1]
+        for name, meta in header.items()
+        if name != "__metadata__"
+    )
+    declared_total = 8 + header_len + max_end
+    cut_to = declared_total - 100
+    assert cut_to > 8 + header_len, (
+        "test setup: cut would land in the header region, not the body — "
+        "use a larger entry"
+    )
+    sf.write_bytes(full[:cut_to])
 
     cache2 = fresh_cache()
     loaded = cache2.load_from_disk(str(tmp_path))
-
-    # CORRECT BEHAVIOR (xfail): load_from_disk should refuse a body-truncated
-    # entry. Currently it succeeds because mx.load is lazy — the corruption
-    # only surfaces later, at the first attention call.
     assert loaded == 0, (
         "Body-truncated safetensors was loaded as a usable cache entry. "
         "It will blow up later at attention time with a RuntimeError, "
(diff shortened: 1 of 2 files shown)


raullenchai:
## Round 2 — self-review fixes

Adversarial review by a sub-agent caught 4 real issues. All addressed in `6675535`.

### Issues fixed

| # | Issue | Fix |
|---|---|---|
| **13** (severe) | Trailing slash on `cache_dir` made `<cache_dir>+'.new'` a *child* of cache_dir — `cache_dir → .old` would have dragged `.new` along, breaking the swap entirely | `cache_dir.rstrip(os.sep)` at top of save/load |
| **9** | `load_from_disk` on a non-empty cache double-counted memory and `bisect.insort`'d duplicates into `_sorted_keys` | Refuse to merge: log warning, return 0 |
| **1** (defense) | A crash mid-`json.dump` of `index.json` would leave a partial JSON file; `os.path.exists()` recovery promoted it blindly | Replaced `os.path.exists` with `_has_valid_index` (parses JSON + checks version) |
| **6 + 8** | `_safetensors_is_complete` accepted negative/non-monotonic `data_offsets` and didn't catch `AttributeError`/`TypeError` from `.items()` on non-dict JSON | Added the missing checks + broadened exception filter |

### New tests

- `test_partial_new_index_json_is_not_promoted` — write a `.new/index.json` containing `{"versi`; expect recovery to fall through to `.old`
- `test_save_handles_trailing_slash_in_cache_dir` — pass `str(cache_dir) + "/"` and verify `.new` doesn't end up as a child
- `test_load_into_non_empty_cache_is_rejected` — verify state is unchanged when called on a populated cache

### Test resilience

- `test_body_truncated_safetensors_should_fail_eagerly_at_load` previously used a fixed "drop last 100 bytes" cut. Replaced with one that parses the actual safetensors header to compute the truncation point inside the declared body region — robust against future padding/alignment changes in `save_prompt_cache`.

### Counts

- 13/13 persistence tests pass (was 10)
- 135/135 cache-related tests pass (was 132)
- ruff check + format clean

### Issues considered & deferred (with rationale)

- **Pre-existing memory-fit `break`**: the original code already `break`s on first overflow rather than `continue`ing to try smaller subsequent entries. Behavior preserved to keep this PR scoped.
- **Concurrent-save safety**: documented in PR description as out of scope. lifespan hooks are single-writer in practice.
- **TOCTOU on `os.path.exists(cache_dir)`**: same single-writer assumption.
- **`shutil.rmtree(..., ignore_errors=True)`** on cleanup paths: leaving a stale `.old` is recoverable on next start; not worth the complexity of `onerror` callbacks here.
- **Imports inside functions**: matches the existing style in this file (`save_to_disk`/`load_from_disk` already used inline imports for `json`/`os`/`time` pre-PR).

raullenchai:
## Round 3 — full validation gate (caught a regression)

### What I ran

| Stage | Result |
|---|---|
| **Negative control** (test file + main's `memory_cache.py`) | 10/13 fail — confirms tests catch the bugs they claim |
| **Unit tests** (full suite, excluding integrations) | 2083 pass, 1 pre-existing fail (`test_reasoning_parsers::TestQwen3::test_only_start_tag_no_end`, unrelated) |
| **Cache tests** (memory_cache + prefix_cache + paged_cache + persistence) | 135/135 pass |
| **Persistence tests** (new file) | 13/13 pass |
| **Stress** (`make stress`, 8 scenarios on Qwen3.5-4B) | 8/8 pass (28s) |
| **Benchmark** (10 sequential requests) | 163 tok/s sustained, 0 garbage outputs |
| **Anthropic SDK** integration | 4/5 (test 4 streaming pre-existing #185, NOT caused by this PR) |
| **End-to-end persistence cycle** | clean save → 149/149 entries → restart → 147 LOADED + 2 dedup → 5/5 correct outputs |

### Regression caught by e2e (fixed in `3fc0354`)

The self-review's #9 fix ("refuse load_from_disk into a non-empty cache") was **too aggressive** — it broke standard startup. Lifespan order is:
1. Engine warmup runs (populates 2 cache entries from `"Hi"` probe)
2. `load_from_disk` is called → **refused because cache wasn't empty** → persisted snapshot never loaded after a restart 😬

E2E test caught it: after clean shutdown + restart, the log said `load_from_disk called on a non-empty cache — skipping`, and warm-load was disabled in production.

**Fix**: per-entry dedup instead of all-or-nothing refusal. If `tokens_key in self._entries`, skip that entry only — the other 147 still load fine. Also split `corrupt_skipped` from `duplicate_skipped` so the "may need cleanup" warning only fires on real corruption.

New summary line: `LOADED 147 entries from <dir> in 0.1s (8464MB total), 2 skipped as duplicates of in-memory entries` (INFO, not WARNING).

### Files

- `vllm_mlx/memory_cache.py`: replaced empty-cache guard with per-entry skip; split skip counters
- `tests/test_prefix_cache_persistence.py`: rename + rewrite `test_load_into_non_empty_cache_*` to verify dedup behavior, not refusal

### Lesson recorded

End-to-end repro on the actual server lifecycle is essential — the unit tests passed for the wrong fix. Adding to memory.

raullenchai:
## Round 4 — DeepSeek V4 Pro review + full validation gate

Codex CLI was rate-limited (recurring); used DeepSeek V4 Pro for the second-opinion review. Two real findings, both addressed in `826cadb`.

### Review findings (DeepSeek V4 Pro, 14.7K tokens)

| # | Finding | Action |
|---|---|---|
| 1 | `load_from_disk` mmap'd safetensors *before* the in-memory dedup check, so duplicate entries paid the full load cost only to be discarded | Moved dedup check above `load_prompt_cache` |
| 2 | `_safetensors_is_complete` returned True for metadata-only safetensors | Skipped — benign, downstream `if cache:` handles it |
| 3 | `_has_valid_index` only validated JSON structure; if `index.json` survives but its referenced entry files are gone (manual deletion / fs corruption / partial restore), recovery would promote an empty `.new` and discard the perfectly-good `.old` | Added "at least the first referenced entry's files exist" check |

### Two new tests
- `test_recovery_rejects_new_with_index_but_no_entry_files` — handcraft `.new/index.json` referencing entries that don't exist on disk; verify recovery falls through to `.old` rather than promoting an empty snapshot
- `test_load_dedup_check_runs_before_safetensors_load` — spy on `load_prompt_cache`; verify it's NOT called for in-memory duplicates

### Full validation gate (final)

| Stage | Result |
|---|---|
| Negative control (test file vs main's `memory_cache.py`) | 12/15 fail — all bug repros catch real issues |
| Persistence tests (new file) | **15/15** pass |
| Cache test suite (5 files) | **137/137** pass |
| Full unit suite (excluding integrations) | **2092/2094** pass — 2 fails: `test_reasoning_parsers` (pre-existing #185-adjacent), `test_batching_deterministic::test_concurrent_same_prompt` (flaky under concurrent server+test load, passes in isolation, also flakes on stock main) |
| Stress (`make stress`, 8 scenarios) | **8/8** pass (28s) |
| **Soak** (`make soak`, 600s sustained) | **PASS** — 151 reqs, 0 errors, 12 disconnects handled cleanly |
| Benchmark (10 sequential reqs) | 163 tok/s sustained, 0 garbage |
| Integration: Anthropic SDK | 4/5 (test 4 streaming pre-existing #185, NOT this PR) |
| Integration: Pydantic AI | **6/6** |
| Integration: Smolagents | **4/4** |
| Integration: LangChain | **6/6** |
| **E2E save/restart/load lifecycle** | clean shutdown 149/149 SAVED → restart 147 LOADED + 2 dedup → 5/5 garbage-free responses → no `.new`/`.old` orphans on disk |
| Lint (`ruff check + format`) | clean |

### Merge readiness

4 commits, ready to merge:
- `76bd748` — initial fix (atomic snapshot + load-time validation)
- `6675535` — sub-agent self-review fixes (4 issues incl. trailing slash)
- `3fc0354` — e2e regression fix (per-entry dedup, lifespan-warmup-aware)
- `826cadb` — DeepSeek V4 Pro review fixes (dedup ordering, recovery hardening)

### Lessons recorded to memory
- `lifespan_warmup_before_load.md` — startup order gotcha
- `external_pr_review_checklist.md` already covers reviewer hierarchy: codex → sub-agent → DeepSeek when codex is limited

PR #204, opened 2026-05-04
Fix Qwen tool-call OpenAI translation

### LINKED ISSUE #160: RuntimeError: There is no Stream(gpu, 0) in current thread.

samuelfaj:
## Summary

This is the focused split-out from #162 for the Qwen/OpenAI tool-call translation fixes requested in the owner review.

It keeps only the parser/postprocessor pieces needed for #160:

- parse bare Qwen `Calling tool: name({...})` output into OpenAI `tool_calls`
- decode JSON-looking and double-encoded tool arguments before OpenAI serialization
- coerce Qwen tool arguments with request JSON schema metadata, including nullable type lists and nested array/object forms
- keep request tool schemas available after streamed content prefixes so Qwen streaming conversion still has schema context
- buffer split `Calling tool` / `[` marker tails without stripping normal code brackets such as `game.snake[` or `const items = [`
- drop malformed duplicate generic tool calls that are missing required schema args before emitting the valid duplicate

Out of scope on purpose, per #162 review:

- no DFlash changes
- no TUI or metrics middleware changes
- no hidden continuation-prompt injection
- no buffered stream retry/repeated-text retry machinery
- no generic fallback when tool parsing is not configured

Fixes #160.

## Tests

- `uv run --extra dev ruff check vllm_mlx/api/tool_calling.py vllm_mlx/service/postprocessor.py vllm_mlx/tool_parsers/qwen3coder_tool_parser.py tests/test_postprocessor.py tests/test_tool_parsers.py tests/test_upstream_regression.py`
- `uv run --extra dev pytest tests/test_postprocessor.py tests/test_tool_parsers.py tests/test_upstream_regression.py` -> 275 passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,18 +126,15 @@ jobs:
         run: |
           pytest \
             tests/test_platform.py \
-            tests/test_llm.py \
             tests/test_mllm.py \
             tests/test_server.py \
             tests/test_paged_cache.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
             tests/test_optimizations.py \
-            tests/test_simple_engine.py \
             tests/test_batching.py \
             tests/test_continuous_batching.py \
             tests/test_streaming_simulator.py \
-            tests/test_deltanet_snapshot.py \
             tests/test_streaming_detokenizer.py \
             tests/test_tool_logits.py \
             -v --tb=short \
--- a/tests/test_postprocessor.py
+++ b/tests/test_postprocessor.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for StreamingPostProcessor — the unified streaming pipeline."""
 
+import json
 from unittest.mock import MagicMock
 
 from vllm_mlx.service.postprocessor import StreamingPostProcessor
@@ -244,6 +245,53 @@ class TestStreamingPostProcessorToolCalls:
         assert len(events) == 1
         assert events[0].type == "tool_call"
         assert events[0].finish_reason == "tool_calls"
+
+    def test_bare_calling_tool_emits_tool_call_not_content(self):
+        """Bare Qwen Calling tool syntax is translated to OpenAI tool calls."""
+        request = {
+            "tools": [
+                {
+                    "type": "function",
+                    "function": {
+                        "name": "todowrite",
+                        "parameters": {
+                            "type": "object",
+                            "properties": {
+                                "todos": {
+                                    "type": "array",
+                                    "items": {"type": "object"},
+                                },
+                            },
+                        },
+                    },
+                }
+            ]
+        }
+
+        cfg = _make_cfg(
+            enable_auto_tool_choice=True,
+            tool_call_parser="qwen3_coder_xml",
+        )
+        pp = StreamingPostProcessor(
+            cfg,
+            tools_requested=True,
+            request_dict=request,
+        )
+        pp.reset()
+
+        events = pp.process_chunk(
+            _make_output(
+                'Calling tool: todowrite({"todos": '
+                '"[{\\"content\\": \\"Initialize\\", \\"status\\": \\"in_progress\\"}]"'
+                "})"
+            )
+        )
+
+        assert len(events) == 1
+        assert events[0].type == "tool_call"
+        args = json.loads(events[0].tool_calls[0]["function"]["arguments"])
+        assert isinstance(args["todos"], list)
+        assert args["todos"][0]["content"] == "Initialize"
 
 
 class TestStreamingPostProcessorNemotron:
--- a/tests/test_postprocessor.py
+++ b/tests/test_postprocessor.py
@@ -275,7 +275,7 @@ class TestStreamingPostProcessorToolCalls:
         pp = StreamingPostProcessor(
             cfg,
             tools_requested=True,
-            request_dict=request,
+            request=request,
         )
         pp.reset()
 
@@ -293,6 +293,55 @@ class TestStreamingPostProcessorToolCalls:
         assert isinstance(args["todos"], list)
         assert args["todos"][0]["content"] == "Initialize"
 
+    def test_qwen_xml_tool_uses_schema_after_content_prefix(self):
+        """Schema conversion still works if text appears before tool markup."""
+        request = {
+            "tools": [
+                {
+                    "type": "function",
+                    "function": {
+                        "name": "todowrite",
+                        "parameters": {
+                            "type": "object",
+                            "properties": {
+                                "todos": {
+                                    "type": "array",
+                                    "items": {"type": "object"},
+                                },
+                            },
+                        },
+                    },
+                }
+            ]
+        }
+
+        cfg = _make_cfg(
+            enable_auto_tool_choice=True,
+            tool_call_parser="qwen3_coder_xml",
+        )
+        pp = StreamingPostProcessor(
+            cfg,
+            tools_requested=True,
+            request=request,
+        )
+        pp.reset()
+
+        assert pp.process_chunk(_make_output("Thinking first.\n"))[0].type == "content"
+        events = pp.process_chunk(
+            _make_output(
+                "<tool_call>\n<function=todowrite>\n"
+                "<parameter=todos>\n"
+                '"[{\\"content\\": \\"Install tests\\", \\"status\\": \\"in_progress\\"}]"\n'
+                "</parameter>\n"
+                "</function>\n</tool_call>"
+            )
+        )
+
+        assert len(events) == 1
+        assert events[0].type == "tool_call"
+        args = json.loads(events[0].tool_calls[0]["function"]["arguments"])
+        assert isinstance(args["todos"], list)
+        assert args["todos"][0]["content"] == "Install tests"
 
 class TestStreamingPostProcessorNemotron:
     """Tests for Nemotron thinking prefix."""
--- a/tests/test_postprocessor.py
+++ b/tests/test_postprocessor.py
@@ -413,6 +413,7 @@ class TestStreamingPostProcessorToolCalls:
         assert len(tool_events[0].tool_calls) == 1
         args = json.loads(tool_events[0].tool_calls[0]["function"]["arguments"])
         assert args["filePath"] == "/tmp/package.json"
+
     def test_qwen_xml_tool_uses_schema_after_content_prefix(self):
         """Schema conversion still works if text appears before tool markup."""
         request = {
@@ -462,6 +463,7 @@ class TestStreamingPostProcessorToolCalls:
         args = json.loads(events[0].tool_calls[0]["function"]["arguments"])
         assert isinstance(args["todos"], list)
         assert args["todos"][0]["content"] == "Install tests"
+
 
 class TestStreamingPostProcessorNemotron:
     """Tests for Nemotron thinking prefix."""
--- a/tests/test_postprocessor.py
+++ b/tests/test_postprocessor.py
@@ -342,6 +342,29 @@ class TestStreamingPostProcessorToolCalls:
         assert args["command"] == "npm test"
         assert args["timeout"] == 60000.0
 
+    def test_code_brackets_are_not_treated_as_partial_tool_markers(self):
+        """Do not strip normal code brackets split at chunk boundaries."""
+        cfg = _make_cfg(
+            enable_auto_tool_choice=True,
+            tool_call_parser="qwen3_coder_xml",
+        )
+        pp = StreamingPostProcessor(
+            cfg,
+            tools_requested=True,
+            request={"tools": []},
+        )
+        pp.reset()
+
+        index_events = pp.process_chunk(_make_output("const head = game.snake["))
+        array_events = pp.process_chunk(_make_output("const snake = ["))
+
+        assert len(index_events) == 1
+        assert index_events[0].type == "content"
+        assert index_events[0].content == "const head = game.snake["
+        assert len(array_events) == 1
+        assert array_events[0].type == "content"
+        assert array_events[0].content == "const snake = ["
+
     def test_generic_tool_call_drops_missing_required_duplicate(self):
         """Malformed duplicate calls should not reach clients for schema rejection."""
         request = {
--- a/tests/test_postprocessor.py
+++ b/tests/test_postprocessor.py
@@ -324,11 +324,11 @@ class TestStreamingPostProcessorToolCalls:
         )
         pp.reset()
 
-        heading_events = pp.process_chunk(_make_output("Next step\n\n["))
+        heading_events = pp.process_chunk(_make_output("Next step\n\n[C"))
         assert len(heading_events) == 1
         assert heading_events[0].type == "content"
         assert heading_events[0].content == "Next step"
-        assert pp.process_chunk(_make_output("Calling tool")) == []
+        assert pp.process_chunk(_make_output("alling tool")) == []
         events = pp.process_chunk(
             _make_output(
                 ': bash({"command":"npm test", "timeout": "60000"})]',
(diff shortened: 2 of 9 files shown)


raullenchai (changes requested):
Thanks for the focused split-out from #162 — the core direction (balanced-brace scanner replacing the brittle regex, double-encoded JSON handling, schema-aware coercion) is solid and the 275 added tests are well-targeted. Adversarial review (DeepSeek V4 Pro) surfaced **5 real issues** before we merge. None are crash-class, but a couple do silently corrupt output, so I'd like them addressed in this PR.

I verified each below with a small repro in the PR branch.

### 🐛 Real bugs — please fix before merge

**1. `_get_tool_param_config` returns the wrong object when a tool schema has no `properties` key**
*File:* `vllm_mlx/api/tool_calling.py`

Repro:
```python
request = {"tools": [{"type": "function", "function": {
    "name": "foo",
    "parameters": {"type": "object", "required": ["x"]}  # no `properties`
}}]}
_get_tool_param_config("foo", request)
# returns: {'type': 'object', 'required': ['x']}
# expected: {}
```
The fallback `return parameters` lets `param_config.get(key)` match `"type"` / `"required"` as if they were parameter names. Fix is one line: `return {}`.

**2. Bare `[` on its own line gets stripped from streamed content**
*File:* `vllm_mlx/service/postprocessor.py` — `_has_partial_calling_tool_marker` / `_find_trailing_calling_tool_prefix`

Repro:
```python
_strip_trailing_calling_tool_prefix("Here is an array:\n[")
# returns: 'Here is an array:'   ← silently drops the '['
_strip_trailing_calling_tool_prefix("Code:\n\n[")
# returns: 'Code:'
```
The existing `test_code_brackets_are_not_treated_as_partial_tool_markers` covers `const items = [` (mid-line `[`) but not line-start `[`. Suggest requiring at least one char of `Calling` after the `[` before treating it as a tool prefix, or only buffer when we already saw `[Calling` together. Add a regression test for the `\n[` case.

### 🟡 Worth fixing (low effort, real but minor)

**3. Boolean coercion only handles `"true"`/`"false"`**
*File:* `vllm_mlx/api/tool_calling.py` — `_coerce_schema_value`

Verified: `"1"`, `"0"`, `"yes"`, `"no"` all stay as strings even when schema type is `boolean`. Many tool outputs use `1`/`0` for booleans. Suggest extending the mapping or, at minimum, adding a debug log when a boolean param fails to coerce.

**4. Duplicate `_decode_json_like` / `_schema_type` in two files**
*Files:* `vllm_mlx/api/tool_calling.py` and `vllm_mlx/tool_parsers/qwen3coder_tool_parser.py`

Identical implementations. Will drift on any future fix. Suggest exporting from one and importing in the other.

### 🟢 Theoretical perf — fine to defer

**5. O(n²) streaming scan**
*File:* `vllm_mlx/service/postprocessor.py` — `_detect_tool_calls`, `finalize`

Each chunk calls `parse_tool_calls(self.tool_accumulated_text, …)` which rescans the full buffer. Bounded for typical tool-using outputs but degrades on long multi-tool generations. Track an offset / scan only the appended portion when convenient.

### ✅ Verified false positive (no action)

**Infinite loop in `_iter_calling_tool_calls` when `{...}` is balanced but `)` is missing** — DeepSeek flagged it, but the inner `while ... else: return` path forces termination. I tested 5 truncated inputs (`Calling tool: foo({"a":1}`, `…({} extra`, etc.) and all returned `0 matches` cleanly with no hang.

---

### Two adjacent items not in this PR but worth knowing

- The `.github/workflows/ci.yml` change (removing `test_llm.py` / `test_simple_engine.py` / `test_deltanet_snapshot.py`) was already done in #203; git merges cleanly but the hunk is now redundant.
- The `try/except ImportError` for `TurboQuantKVCache` in `memory_cache.py` becomes dead code post-merge (since `test_memory_cache.py` lives only on the apple-silicon job after #203). Harmless, just FYI — could either drop it or move the test back to Linux to actually exercise the defensive import.

Once #1 and #2 are addressed (and ideally #3 + #4), happy to merge. Thanks again for splitting this out — much easier to review than #162.

samuelfaj:
Addressed the requested review feedback in `5965f1b` on `fix-qwen-tool-call-openai-translation`.

What changed:
- Fixed `_get_tool_param_config` so schemas without `properties` return `{}` instead of the full parameters object.
- Fixed streamed post-processing so a bare `[` on its own line is preserved as normal content, while `[Calling...` partial markers are still buffered.
- Extended boolean schema coercion to handle common aliases (`1/0`, `yes/no`, `on/off`, `y/n`).
- Removed the duplicated `_decode_json_like` and `_schema_type` implementations from `qwen3coder_tool_parser.py` by importing the shared helpers.
- Added regression tests for the no-`properties` schema case, boolean aliases, and the line-start bare bracket case.

Validation run locally:
- `uv run pytest tests/test_tool_calling.py tests/test_postprocessor.py tests/test_tool_parsers.py`
- `uv run ruff check vllm_mlx/api/tool_calling.py vllm_mlx/service/postprocessor.py vllm_mlx/tool_parsers/qwen3coder_tool_parser.py tests/test_tool_calling.py tests/test_postprocessor.py`

I left the O(n²) streaming scan unchanged because the review marked it as theoretical performance work that is fine to defer.

michaelasper:
We have free compute to review things right now, and this PR would solve a lot of issues we're running into, so I gave current head `5965f1b` a focused pass. The direction looks good and the previous review items are mostly addressed. CI is green, and I also ran the focused parser/postprocessor suite locally. 🙂

A few suggestions before merge:

1. String schema arguments can be decoded into objects/lists

In `vllm_mlx/api/tool_calling.py`, `_coerce_schema_value()` calls `_decode_json_like()` before checking the requested schema type. That means a tool parameter declared as `{"type": "string"}` is converted if the string happens to look like JSON.

Repro shape:

```python
request = {"tools": [{"type": "function", "function": {
    "name": "write",
    "parameters": {"type": "object", "properties": {
        "content": {"type": "string"},
        "path": {"type": "string"},
    }},
}}]}

_, calls = parse_tool_calls(
    'Calling tool: write({"path":"/tmp/a.txt","content":"{\\"a\\":1}"})',
    request,
)
json.loads(calls[0].function.arguments)["content"]
# current: {"a": 1}
# expected: '{"a":1}'
```

This is risky for write/edit tools because file contents commonly start with `{` or `[`. Suggest preserving explicit `string` schemas before JSON-like decoding, and only decoding for object/array schemas or top-level stringified argument blobs.

2. Literal `Calling tool:` text can disappear from streaming output

In `vllm_mlx/service/postprocessor.py`, once `self.tool_accumulated_text` contains `Calling tool:`, `_detect_tool_calls()` returns `None` if `parse_tool_calls()` finds no valid call. That suppresses normal text and, if the chunk is finished, also drops the finish event.

Repro shape:

```python
events = pp.process_chunk(_make_output(
    "This is text Calling tool: not a call",
    finished=True,
))
# current: []
# expected: content/finish passthrough
```

The partial-marker buffering should stay narrow: buffer only plausible line-start partial tool markers, and pass through normal content when the full marker is not followed by a parseable call.

Lower priority: the boolean alias fix was added to generic `_coerce_schema_value()`, but the Qwen XML path still treats `yes` as `False` in `_convert_param_value()`. It would be cleaner to share the same boolean coercion behavior there too.

Validated:
- GitHub checks are green.
- `uv run --extra dev pytest tests/test_tool_calling.py tests/test_postprocessor.py tests/test_tool_parsers.py tests/test_upstream_regression.py` -> 325 passed
- `uv run --extra dev ruff check ...` -> all checks passed

raullenchai:
SOP §0–§4 review. **Requesting changes** — 1 must-fix P1, 2 nice-to-fix P1s. Targeted tests pass (325 passed locally).

**§0 Necessity:** ✅ fixes real Qwen tool-call OpenAI-translation pain that you hit running OpenCode against `qwen3.6-35b`. Concrete user-visible value.

**§2 Codex round 1 — 0 P0, 3 P1:**

### Must-fix P1 — `_coerce_schema_value()` corrupts string args with JSON-like content

`vllm_mlx/api/tool_calling.py:649`:

```python
def _coerce_schema_value(value: Any, schema: Any) -> Any:
    value = _decode_json_like(value)        # ← runs UNCONDITIONALLY before schema check
    schema_type = _schema_type(schema)
    ...
    if schema_type in ("array", "object"):
        return value
    if not isinstance(value, str):
        return value                        # ← already mutated to object here
```

When the schema declares `type: "string"` and the tool arg is a JSON-looking string (e.g., `write_file(content='{"key":"val"}')` or any source-code edit containing braces), `_decode_json_like()` parses it into an object. By the time `schema_type` is checked, the value is no longer a string — and since `not isinstance(value, str)` is true for a dict, the function returns the parsed dict instead of the original string. The OpenAI client then receives an object where it expects a string, and the tool execution breaks (or worse, silently writes garbage to a file).

**This is a real regression for write/edit tools.** Concrete repro: `qwen3.5-coder` + OpenCode `write_file` of any source file containing JSON or arrays.

**Fix:**

```python
def _coerce_schema_value(value: Any, schema: Any) -> Any:
    schema_type = _schema_type(schema)
    if schema_type in ("string", "enum"):
        return value if isinstance(value, str) else json.dumps(value)
    value = _decode_json_like(value)
    ...
```

i.e., resolve `schema_type` first; preserve string/enum as-is; only decode for object/array schemas.

### Nice-to-fix P1 #2 — false-positive text suppression on \"Calling tool:\" in normal text

`vllm_mlx/service/postprocessor.py:686`:

```python
if \"Calling tool:\" in self.tool_accumulated_text:
    _, tool_calls = parse_tool_calls(...)
    if tool_calls:
        return {\"tool_calls\": chunks}
    return None  # ← drops all content
```

If accumulated text contains `Calling tool:` but `parse_tool_calls()` returns empty, `_detect_tool_calls()` returns `None` and the entire content (including the finish event) gets dropped. A normal assistant explanation containing the literal phrase (e.g., a markdown code block showing `Calling tool: name (mocked)`, or a docstring quoting the format) would be silently truncated.

**Fix:** when the marker is present but no call parses, fall through to `return {\"content\": tool_result.get(\"content\", \"\")}` so the text passes through.

### Nice-to-fix P1 #3 — malformed first marker swallows valid later marker

`vllm_mlx/api/tool_calling.py:214` (`_iter_calling_tool_calls`): when balanced JSON isn't followed by `)`, scanning continues under the same first marker instead of advancing past it. A model emitting one malformed `Calling tool: name({...}` followed by a valid `Calling tool: name({...})` loses the valid call.

**Fix:** when depth returns to zero and next nonspace isn't `)`, advance `search_from` past the bad marker and continue the outer search for a fresh marker.

---

**Validation done:**
- §3 lint: clean per CI
- §4 targeted tests: 325 passed locally (`tests/test_postprocessor.py`, `tests/test_tool_calling.py`, `tests/test_tool_parsers.py`, `tests/test_upstream_regression.py`)
- §7 CI: all 7 checks SUCCESS

**Verified:**
- CI workflow change is legitimate cleanup — `tests/test_llm.py`, `tests/test_simple_engine.py`, `tests/test_deltanet_snapshot.py` all gone from `main` (SimpleEngine deletion + DeltaNet test removal happened upstream).
- `memory_cache.py` defensive turboquant import is correct — `vllm_mlx/turboquant.py` exists on main and imports MLX at module level, so try/except is the right pattern for non-MLX CI runs.

**On the merge bar:** I'd merge happily once P1#1 is fixed. P1#2 and P1#3 are real but narrow — I can take them as follow-up issues if you'd rather scope a single fix-and-ship round, but a single commit fixing all three would be cleanest.

Thanks for the thorough work — the OpenCode validation is exactly the level of evidence I want for parser PRs.

raullenchai:
Thanks for the split-out from #162, but this PR cannot be merged in its current form — the diff is **far broader** than the "focused Qwen tool-call translation fix" the description claims.

## Supply-chain / CI concerns

`.github/workflows/ci.yml` is silently weakened in multiple ways unrelated to tool-calling:

1. Replaces `ruff format --check` with `black --check`, and adds `black` as a new CI dependency. Our project standardized on `ruff format` (see `pyproject.toml` `[tool.ruff.format]`).
2. Narrows ruff to `--select E,F,W --ignore E402,E501,E731,F811,F841`, dropping the `I, N, UP, B, SIM` rules our `pyproject.toml` enables.
3. **Removes 7 test files from CI** — including the ones that exercise this PR's code paths:
   - `test_tool_calling.py`, `test_tool_injection.py`, `test_tool_logits.py`
   - `test_streaming_latency.py`, `test_streaming_newlines.py`, `test_streaming_simulator.py`, `test_streaming_detokenizer.py`
   - `test_minimax_reasoning_parser.py`, `test_pr_validate_runner.py`, `test_pr_validate_deepseek.py`
4. Re-adds `tests/test_llm.py` and `tests/test_simple_engine.py` to the MLX job, but those files don't exist on `main` — SimpleEngine was removed in #156 and those tests went with it. CI will fail on rebase.

## `memory_cache.py` regressions

The PR also silently changes `vllm_mlx/memory_cache.py`:

1. Removes `_safetensors_is_complete()` — the validator that catches truncated KV files that would otherwise silently mmap as zeros and produce wrong-output bugs. This is a critical safety check.
2. Replaces the "Deep copies on fetch to prevent mutation of stored cache entries" comment with "No unnecessary deep copies (MLX arrays are immutable)" and presumably removes the deep-copy path. MambaCache / ArraysCache entries are not immutable in our codebase — this is a behavior regression.

These changes have no mention in the PR description and contradict the "out of scope on purpose" list.

## What needs to change

Please drop the following from the PR so it actually matches its description:
- All `.github/workflows/ci.yml` edits
- All `vllm_mlx/memory_cache.py` edits

Then rebase on current `main` (the branch is ~1 week behind and several large changes have landed since: PR #354, PR #356, PR #357). Re-run the full test suite locally after rebase — there's a good chance the parser changes need adjustment too.

Once those scope reductions are done I'm happy to do another adversarial review pass on the actual tool-call translation logic, which I haven't audited yet because the supply-chain concerns above are blocking.

raullenchai:
@samuelfaj — closing this PR with thanks. Your schema-aware tool-call coercion (`_decode_json_like` / `_coerce_schema_value` / `_normalize_tool_arguments` / `_iter_calling_tool_calls`) is genuinely useful and we wanted to ship it.

Rather than keep iterating on a 9-day-stale branch with merge conflicts (and the unrelated CI / `memory_cache.py` / `postprocessor.py` / `routes/chat.py` changes that the §0–§4 review surfaced), I cherry-picked the additive parser improvements from this PR + #206 into a focused PR with full attribution: **#388**.

Net surface of #388: 5 files, +495/-42 — your `tool_calling.py` helpers wholesale, `qwen3coder_tool_parser.py` refactor (using #204's clean version that imports the helpers, not #206's local duplicate), the upstream regression tests, and the new generic-parsing test class — all with `Co-Authored-By: Samuel Fajreldines` on the commit.

Going to close this in favor of #388. Thanks for the work — the parser fixes were real value. Closing #206 with the same note.

PR #205, opened 2026-05-04
Add serve TUI monitor

samuelfaj:
Screenshot 2026-05-04 at 5 32 21 PM

## Summary

Adds only the `rapid-mlx serve --tui` path:

- adds a `--tui` serve flag
- runs uvicorn in a background thread when `--tui` is enabled
- runs a foreground terminal monitor that polls existing `/health` and `/v1/status`
- exits cleanly with `q` or Ctrl-C and asks the background server to stop

## Tests

- `uv run --extra dev ruff check vllm_mlx/cli.py vllm_mlx/tui.py`
- `uv run --extra dev ruff format --check vllm_mlx/cli.py vllm_mlx/tui.py`
- `uv run --extra dev python -m compileall vllm_mlx/cli.py vllm_mlx/tui.py`
- `uv run --extra dev python -m vllm_mlx.cli serve --help` confirmed `--tui` appears
- TUI screen-builder smoke test against sample `/health` and `/v1/status` payloads

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,18 +126,15 @@ jobs:
         run: |
           pytest \
             tests/test_platform.py \
-            tests/test_llm.py \
             tests/test_mllm.py \
             tests/test_server.py \
             tests/test_paged_cache.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
             tests/test_optimizations.py \
-            tests/test_simple_engine.py \
             tests/test_batching.py \
             tests/test_continuous_batching.py \
             tests/test_streaming_simulator.py \
-            tests/test_deltanet_snapshot.py \
             tests/test_streaming_detokenizer.py \
             tests/test_tool_logits.py \
             -v --tb=short \
--- a/tests/test_metrics_middleware.py
+++ b/tests/test_metrics_middleware.py
@@ -1,0 +1,90 @@
+import json
+
+from fastapi.routing import APIRoute
+
+import vllm_mlx.server as server
+from vllm_mlx.middleware import metrics
+from vllm_mlx.middleware.metrics import MetricsMiddleware
+from vllm_mlx.request_metrics import RequestRecorder
+from vllm_mlx.routes.health import router as health_router
+
+
+async def _empty_receive():
+    return {"type": "http.request", "body": b"", "more_body": False}
+
+
+def _sse_event(payload: dict, *, crlf: bool = False, terminate: bool = True) -> bytes:
+    ending = b"\r\n\r\n" if crlf and terminate else b"\n\n" if terminate else b"\n"
+    return b"data: " + json.dumps(payload).encode("utf-8") + ending
+
+
+def test_request_metrics_middleware_is_not_enabled_by_default():
+    assert all(
+        middleware.cls is not MetricsMiddleware for middleware in server.app.user_middleware
+    )
+
+
+def test_requests_endpoint_requires_api_key_dependency():
+    route = next(
+        route
+        for route in health_router.routes
+        if isinstance(route, APIRoute) and route.path == "/v1/requests"
+    )
+
+    assert route.dependant.dependencies
+
+
+async def test_metrics_middleware_drains_final_sse_event(monkeypatch):
+    recorder = RequestRecorder()
+    monkeypatch.setattr(metrics, "get_recorder", lambda: recorder)
+
+    async def app(scope, receive, send):
+        await send(
+            {
+                "type": "http.response.start",
+                "status": 200,
+                "headers": [(b"content-type", b"text/event-stream")],
+            }
+        )
+        await send(
+            {
+                "type": "http.response.body",
+                "body": _sse_event(
+                    {"choices": [{"delta": {"content": "hello"}}]},
+                    crlf=True,
+                ),
+                "more_body": True,
+            }
+        )
+        await send(
+            {
+                "type": "http.response.body",
+                "body": _sse_event(
+                    {
+                        "choices": [{"delta": {}, "finish_reason": "stop"}],
+                        "usage": {"prompt_tokens": 2, "completion_tokens": 3},
+                    },
+                    terminate=False,
+                ),
+                "more_body": False,
+            }
+        )
+
+    sent = []
+
+    async def send(message):
+        sent.append(message)
+
+    middleware = MetricsMiddleware(app)
+    await middleware(
+        {"type": "http", "path": "/v1/chat/completions"},
+        _empty_receive,
+        send,
+    )
+
+    entries = recorder.entries()
+    assert len(entries) == 1
+    assert entries[0]["finish_reason"] == "stop"
+    assert entries[0]["prompt_tokens"] == 2
+    assert entries[0]["generated_tokens"] == 3
+    assert entries[0]["message_preview"] == "hello"
--- a/tests/test_metrics_middleware.py
+++ b/tests/test_metrics_middleware.py
@@ -20,7 +20,8 @@ def _sse_event(payload: dict, *, crlf: bool = False, terminate: bool = True) -> 
 
 def test_request_metrics_middleware_is_not_enabled_by_default():
     assert all(
-        middleware.cls is not MetricsMiddleware for middleware in server.app.user_middleware
+        middleware.cls is not MetricsMiddleware
+        for middleware in server.app.user_middleware
     )
 
 
--- a/tests/test_request_metrics.py
+++ b/tests/test_request_metrics.py
@@ -1,0 +1,48 @@
+import vllm_mlx.request_metrics as request_metrics
+from vllm_mlx.request_metrics import RequestRecorder
+
+
+def test_request_recorder_records_completed_request(monkeypatch):
+    now = [1000.0]
+    monkeypatch.setattr(request_metrics.time, "time", lambda: now[0])
+
+    recorder = RequestRecorder()
+    req_id = recorder.start("/v1/chat/completions")
+
+    now[0] += 0.25
+    recorder.mark_first_token(req_id)
+    recorder.update(req_id, delta_text="hello", generated_tokens=1, prompt_tokens=8)
+
+    now[0] += 0.75
+    recorder.finish(
+        req_id,
+        finish_reason="stop",
+        prompt_tokens=8,
+        generated_tokens=4,
+        engine_gen_tps=12.5,
+        engine_ttft=0.25,
+    )
+
+    entries = recorder.entries()
+    assert recorder.active() is None
+    assert len(entries) == 1
+    assert entries[0]["surface"] == "/v1/chat/completions"
+    assert entries[0]["prompt_tokens"] == 8
+    assert entries[0]["generated_tokens"] == 4
+    assert entries[0]["decode_tps"] == 12.5
+    assert entries[0]["ttft"] == 0.25
+
+
+def test_request_recorder_active_snapshot(monkeypatch):
+    monkeypatch.setattr(request_metrics.time, "time", lambda: 1000.0)
+
+    recorder = RequestRecorder()
+    req_id = recorder.start("/v1/completions")
+    recorder.update(req_id, delta_text="partial", generated_tokens=2, prompt_tokens=5)
+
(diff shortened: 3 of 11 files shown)


raullenchai (changes requested):
Thanks for splitting this out from the larger TUI work — the per-file decomposition is helpful for review. The TUI itself (`tui.py`) is well-isolated and the screen-builder tests are a nice touch. **But the PR is doing more than the description claims, and a few of those extras need design changes before merge.** Adversarial review (manual + DeepSeek V4 Pro) surfaced one architectural concern, one security concern, and several correctness issues. I verified each on the PR branch.

The PR description says the change is "only the `rapid-mlx serve --tui` path." It actually adds **two new always-on subsystems**: a 222-line ASGI middleware (`MetricsMiddleware`) installed unconditionally for every request, and a 201-line `request_metrics` module backing a new public `/v1/requests` endpoint. Those run for every user whether or not `--tui` is in use, so the cost/risk is borne by everyone — not just opt-in TUI users.

### 🚨 Blocking — please address before merge

**1. Metrics middleware runs unconditionally for every request**
*File:* `vllm_mlx/server.py` line 338

```python
app.add_middleware(MetricsMiddleware)
```

This is added at module load with no flag gate. Every `/v1/chat/completions` and `/v1/completions` request now:

- Spawns a per-request `asyncio.Task` (`poll_loop`) that polls `cfg.engine.get_stats()` every 200ms
- Re-parses every SSE chunk through `_safe_json_loads` → JSON parsing on the response hot path
- Buffers up to 4 MB of non-streaming response body
- Calls `poll_engine_stats()` on every body chunk *in addition to* the 200ms loop (DeepSeek finding #5)

Under N concurrent streaming requests, that's N parallel poll tasks all querying the same global engine stats. That's a real overhead being permanently imposed on users who didn't ask for the TUI.

**Suggested fix:** Gate the middleware on the `--tui` flag (or a separate `--enable-request-metrics` flag). The TUI is opt-in; its supporting infrastructure should be too. If you want metrics always on, that's a larger architectural discussion that probably belongs in its own PR with a benchmark showing the overhead is acceptable.

**2. New `/v1/requests` endpoint is unauthenticated and exposes user prompt content**
*File:* `vllm_mlx/routes/health.py` line 14

```python
@router.get("/v1/requests")
async def recent_requests(limit: int = 50):
    ...
    return {"active": recorder.active(), "entries": recorder.entries(...)}
```

The returned entries include `message_preview` (last 240 chars of generated text). Note that **other `/v1/*` endpoints in the codebase use `dependencies=[Depends(verify_api_key)]`** (see `models.py`, `chat.py`, `completions.py`, `mcp_routes.py`, `audio.py`, `embeddings.py`). The other unauth routes in `health.py` (`/health`, `/v1/status`, `/v1/cache/stats`) don't expose user content, so the precedent in that file isn't quite the same.

If this server is exposed beyond localhost, anyone can pull the last 50 prompt previews. Same risk class as putting `Authorization` headers in logs.

**Suggested fix:** Add `dependencies=[Depends(verify_api_key)]`, OR restrict to `127.0.0.1` (e.g. by checking `request.client.host`), OR drop `message_preview` from the response and only expose aggregate counters (tokens, ttft, finish_reason).

**3. Last SSE event silently dropped at stream end**
*File:* `vllm_mlx/middleware/metrics.py` — `consume_sse`

```python
def consume_sse(buf: bytes) -> None:
    nonlocal sse_carry
    data = sse_carry + buf
    *complete, sse_carry = data.split(b"\n\n")
    for raw_event in complete:
        ...
```

Two real issues here:

a) Splits only on `\n\n`. SSE spec also allows `\r\n\r\n` (CRLF). Servers/proxies that emit CRLF lose every event.

b) When the stream ends (`more_body=False`), `sse_carry` is never drained. If the final SSE event lacks a trailing `\n\n` (which happens commonly — many SSE producers end with `data: [DONE]\n` not `data: [DONE]\n\n`), the last event is lost. That's the event with `finish_reason` and final `usage`, so `recorder.finish` ends up with stale state.

**Suggested fix:** In `send_wrapper` when `more=False` and `is_sse`, call `consume_sse(b"\n\n")` to drain the carry, then handle whatever's left. Also support `\r\n\r\n` separator.

### 🟡 Should fix (real, low effort)

**4. Token estimation in middleware is junk data**
*File:* `vllm_mlx/middleware/metrics.py` line 154

```python
running_text_tokens += max(1, len(text) // 4)
```

`len(text) // 4` assumes 4 chars/token, which is wrong for CJK (1 char = 1+ tokens), code (closer to 3 chars/token), emoji, etc. This value is then written to `recorder.update(..., generated_tokens=running_text_tokens, ...)` when upstream `usage` isn't yet seen. The `/v1/requests` JSON and the TUI both display this as `generated_tokens`. Garbage in, garbage out.

**Suggested fix:** Either drop the fallback (don't report `generated_tokens` until upstream usage arrives) or rename it to make clear it's a chunk count, not a token count. The recorder has separate `text_chunks` already — could just expose that.

**5. Race condition in `get_recorder()` lazy init**
*File:* `vllm_mlx/request_metrics.py` line 191

```python
_recorder: RequestRecorder | None = None
def get_recorder() -> RequestRecorder:
    global _recorder
    if _recorder is None:
        _recorder = RequestRecorder()
    return _recorder
```

Two concurrent requests can race past the `is None` check and create two separate recorders. The losing thread's recorder is then garbage-collected while requests are referencing it.

**Suggested fix:** Initialize at module load (`_recorder = RequestRecorder()`), or guard with a `threading.Lock`. Module-level init is simplest.

**6. Signal handling gap during TUI startup**
*File:* `vllm_mlx/cli.py` — `_run_with_tui`

```python
server.install_signal_handlers = lambda: None  # type: ignore[assignment]
```

If SIGTERM arrives during the busy-wait `for _ in range(200): if server.started: break` window, neither uvicorn (signals disabled) nor the TUI (not yet running) handles it. The process keeps running until the loop exits.

**Suggested fix:** Install a minimal main-thread signal handler that sets `server.should_exit = True` before starting the wait loop.

**7. Per-chunk `poll_engine_stats()` is redundant with the 0.2s poll loop**
*File:* `vllm_mlx/middleware/metrics.py` `send_wrapper`

The polling already runs in `poll_loop`. Calling it again on every body chunk just adds CPU. Either keep the loop and remove the per-chunk call, or remove the loop and rely on per-chunk only. Pick one.

**8. Daemon server thread can kill in-flight requests on shutdown**
*File:* `vllm_mlx/cli.py` `_run_with_tui`

```python
server_thread = threading.Thread(target=server.run, daemon=True)
...
finally:
    server.should_exit = True
    server_thread.join(timeout=5)
```

`daemon=True` + `join(timeout=5)` means if uvicorn doesn't drain in 5s, the process exits and any in-flight request gets terminated mid-response. Probably fine for a dev TUI but worth a comment so users know the TUI isn't safe for production traffic.

### 🟢 Code quality / FYI

**9. Bare `except Exception:` swallowing bugs**
Five places in `metrics.py` and `tui.py` use bare `except Exception:` to ignore everything. This shadows real bugs (TypeError from a refactor, AttributeError from missing fields). Suggest catching the specific exceptions you expect (`json.JSONDecodeError`, `urllib.error.URLError`) or at minimum logging at debug level.

**10. Tests cover < 5% of the new code paths**
The 5 added tests cover `_build_screen` and `RequestRecorder` in isolation. Not covered: `MetricsMiddleware` integration, `consume_sse` boundary cases (which has the bug above), `/v1/requests` endpoint, signal handling in `_run_with_tui`, concurrent request handling. Given the always-on nature of the middleware, integration tests would have caught #3.

**11. Adjacent items already on main**
- The `.github/workflows/ci.yml` change (removing `test_llm.py` / `test_simple_engine.py` / `test_deltanet_snapshot.py`) was already done in #203; git merges cleanly but the hunk is now redundant.
- The `try/except ImportError` for `TurboQuantKVCache` in `memory_cache.py` becomes dead code post-merge with #203 (same situation as #204). Harmless, just FYI.

### ✅ Verified false positive (no action)

**DeepSeek finding #1 — "CI silently drops test files for critical core functionality"** — those files were deleted in PR #156 when SimpleEngine was removed. The CI hunk just removes stale references to non-existent files. Same as #204.

---

### Suggested path forward

Given the scope of the architectural concern (#1), I'd suggest splitting this into **two PRs**:

- **Part A:** Just `--tui` flag + `tui.py` + the existing `/health` and `/v1/status` polling. The TUI is well-isolated and could land standalone using only existing endpoints. (~750 LOC)
- **Part B:** `MetricsMiddleware` + `request_metrics.py` + `/v1/requests` — but only **after** addressing #1 (gating), #2 (auth), #3 (SSE drain), #4 (token estimate), #5 (init race). Probably ~600 LOC.

Or, if you'd rather keep them bundled, addressing #1–5 inline is also fine — they're not large changes. But the design conversation around always-on metrics is worth having before this lands as-is.

Thanks for the work — the TUI direction is genuinely useful and I want to land it.

samuelfaj:
Addressed the requested review feedback in `151394c` on `add-serve-tui`.

What changed:
- Metrics middleware is no longer installed unconditionally. It is enabled only from the `--tui` path via `configure_request_metrics()`.
- `/v1/requests` now uses the same `verify_api_key` dependency pattern as the other `/v1/*` endpoints.
- SSE parsing now handles CRLF event separators and drains the final carry when the stream ends without a trailing blank line.
- Removed the `len(text) // 4` generated-token fallback so the middleware no longer reports estimated text length as token counts.
- Changed the request recorder to module-level initialization to avoid lazy-init races.
- Added main-thread signal handling for the TUI startup window and documented the daemon-thread shutdown behavior.
- Removed redundant per-chunk engine-stat polling; the existing poll loop remains the single polling path.
- Tightened the new TUI/metrics exception handling where practical and logs metrics middleware failures at debug level.
- Added middleware tests covering opt-in gating, `/v1/requests` auth dependency, CRLF SSE parsing, and final-event drain.

Validation run locally:
- `uv run pytest tests/test_tui.py tests/test_request_metrics.py tests/test_metrics_middleware.py`
- `uv run ruff check vllm_mlx/cli.py vllm_mlx/server.py vllm_mlx/routes/health.py vllm_mlx/request_metrics.py vllm_mlx/middleware/metrics.py vllm_mlx/tui.py tests/test_tui.py tests/test_request_metrics.py tests/test_metrics_middleware.py`

samuelfaj:
Follow-up: pushed `c0bb947` to fix the CI lint failure by applying `ruff format` to the new metrics middleware test file.

Re-validated locally:
- `uv run ruff check vllm_mlx/ tests/`
- `uv run ruff format --check vllm_mlx/ tests/`
- `uv run pytest tests/test_tui.py tests/test_request_metrics.py tests/test_metrics_middleware.py`

raullenchai:
Hi @samuelfaj — thanks for building this. After applying our new SOP §0 necessity gate (see [docs/development/pr_merge_sop.md](https://github.com/raullenchai/Rapid-MLX/blob/main/docs/development/pr_merge_sop.md)) I'm closing this without merging.

**Reasoning:**

- No linked user issue or demand signal — searched for \"TUI\", \"monitor\", \"dashboard\" and found nothing requesting this feature.
- +1465/-11 lines of new surface (`vllm_mlx/tui.py`, `--tui` flag, background-thread uvicorn, polling loop) costs real ongoing maintenance — every refactor of `serve_command`, `/health`, `/v1/status` could break the TUI path.
- The information the TUI surfaces (request count, memory, model status) is already exposed via `/health`, `/v1/status`, `/metrics` — anyone who wants a dashboard can build one with `htop`, `watch curl`, or Grafana, without us shipping it.
- Auto-deploy blast radius: every Rapid-MLX release auto-publishes to PyPI + Homebrew. Adding a TUI dep (curses/blessed/whatever) on the install path affects every user.

**What would unlock merge:**

- A user issue saying \"I want to monitor my serve session interactively without `watch`/Grafana\" with concrete usage signal (👍 reactions, multiple users chiming in)
- Or a clearly opt-in path: e.g., `pip install rapid-mlx[tui]` with the TUI lib only pulled in when that extra is requested, AND the `--tui` flag rejected with a helpful error when the extra isn't installed

I'd love to keep your contributions — #204 (Qwen tool-call) is a real user-pain fix and I'm reviewing it now. Let's prioritize the bug-fix work over speculative features.

Issue #209, opened 2026-05-05
`--stream-interval N` silently drops tokens in streaming responses, garbling output for N > 1


unsaltedbutter-ai:
### Rapid-MLX version

0.6.11

### Hardware

Mac Studio M3 Ultra 256GB

### macOS version

26.3.1

### Python version

Python 3.13.1

### Model

mlx-community/Qwen3.5-9B-8bit

### Full serve command

--port 8880 --max-num-seqs 3  --default-temperature 0.0 --kv-cache-quantization --cache-memory-mb 6000 --stream-interval 100

### Streaming or non-streaming?

Streaming (`stream: true`)

### What happened? What did you expect?

`--stream-interval N` is documented as batching N tokens before each streaming chunk,
reducing HTTP overhead. In practice, for N > 1, only every Nth token's text reaches
the client. The intermediate N-1 tokens are discarded each cycle, producing fragmented,
garbled streaming responses on any request that generates more than ~N tokens.

### Root cause

In `engine_core.py`, the `should_send()` check sits on the **producer side** — when it
returns `False`, the code skips `collector.put()` entirely:

```python
# BEFORE (broken)
if state and state.should_send(req_output.completion_tokens, req_output.finished):
    collector.put(req_output)
    state.mark_sent(req_output.completion_tokens)
# tokens where should_send() == False are silently dropped here
```

Each `req_output.new_text` contains only the delta for that single generation step.
When a step is skipped, that delta is lost forever. On the next `should_send()` hit,
only the Nth token's delta is flushed — not the accumulated text of all N tokens.

### Affected behavior

| `--stream-interval` | Tokens sent | Tokens dropped |
|---|---|---|
| 1 (default) | all | 0% |
| 2 | every 2nd | ~50% |
| 4 | every 4th | ~75% |

Short responses (≤ N tokens) appear correct because the first token is always sent
immediately (`sent_tokens == 0` path in `should_send()`). Responses longer than N
tokens produce partial, fragmented output. With certain BPE boundary conditions the
dropped tokens leave literal Unicode control characters in the assembled text, which
also breaks JSON parsing of the SSE response.

**Non-streaming is unaffected** — that path uses `output.output_text` (the cumulative
complete text at finish), never the per-step delta path.

### Reproduction

```bash
rapid-mlx serve <model> --stream-interval 4
```

Request any response longer than ~4 tokens via the streaming API (`"stream": true`).
A non-streaming request to the same server with the same prompt returns correct output,
[...]

PR #210, opened 2026-05-05
fix(streaming): accumulate new_text across skipped steps for stream_interval > 1

unsaltedbutter-ai:
Issue #209

## What does this PR do?

The producer-side should_send() gate in _engine_loop dropped N-1 of every N token deltas at stream_interval=N, garbling streaming responses (and occasionally corrupting SSE JSON via stranded BPE control bytes). Buffer each step's delta per-request and flush the merged RequestOutput when should_send() hits. stream_interval=1 fast path is untouched.

## Checklist

- [x] Tests pass locally (`python3 -m pytest tests/ -x`)
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate <PR#>` — see [CONTRIBUTING.md](../CONTRIBUTING.md#self-validating-your-pr) (opt out heavy steps with `PR_VALIDATE_NO_DEEPSEEK=1 PR_VALIDATE_NO_STRESS=1` if you don't have the hardware/keys)
- [x] Updated README/docs if applicable

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -586,6 +586,191 @@ class TestEngineAsync:
         assert fake_scheduler.thread_names
         assert all(name.startswith("mlx-step") for name in fake_scheduler.thread_names)
 
+    async def test_stream_interval_buffer_merges_skipped_step_deltas(
+        self, mock_model_and_tokenizer
+    ):
+        """Regression: stream_interval > 1 must not drop step deltas.
+
+        Pre-fix, when ``RequestStreamState.should_send()`` returned False the
+        engine skipped ``collector.put()`` and that step's ``new_text`` /
+        ``new_token_ids`` / ``logprobs`` were silently lost. After PR #210 the
+        engine accumulates per-step deltas in ``_stream_buffers`` and flushes
+        the merged delta on the next ``should_send() == True`` step. This
+        test feeds 6 step outputs through the loop with stream_interval=4
+        and asserts the buffer + flush invariants on all three delta fields.
+        A finished=True step must always trigger a flush.
+        """
+        from vllm_mlx import engine_core
+        from vllm_mlx.engine import EngineConfig, EngineCore
+        from vllm_mlx.output_collector import RequestStreamState
+
+        model, tokenizer = mock_model_and_tokenizer
+        engine = EngineCore(
+            model,
+            tokenizer,
+            EngineConfig(step_interval=0.001, stream_interval=4),
+        )
+
+        rid = "stream-interval-buffer-test"
+        # Per-step deltas. completion_tokens climbs by 1 each step (matches
+        # scheduler's one-token-per-step contract). logprobs values are
+        # sentinels — the test asserts the *list* is preserved, the type
+        # of the entries is irrelevant.
+        steps = [
+            RequestOutput(
+                request_id=rid,
+                new_token_ids=[10],
+                new_text="he",
+                output_token_ids=[10],
+                output_text="he",
+                finished=False,
+                completion_tokens=1,
+                logprobs="lp1",
+            ),
+            RequestOutput(
+                request_id=rid,
+                new_token_ids=[20],
+                new_text="llo",
+                output_token_ids=[10, 20],
+                output_text="hello",
+                finished=False,
+                completion_tokens=2,
+                logprobs="lp2",
+            ),
+            RequestOutput(
+                request_id=rid,
+                new_token_ids=[30],
+                new_text=" wo",
+                output_token_ids=[10, 20, 30],
+                output_text="hello wo",
+                finished=False,
+                completion_tokens=3,
+                logprobs="lp3",
+            ),
+            RequestOutput(
+                request_id=rid,
+                new_token_ids=[40],
+                new_text="rld",
+                output_token_ids=[10, 20, 30, 40],
+                output_text="hello world",
+                finished=False,
+                completion_tokens=4,
+                logprobs="lp4",
+            ),
+            RequestOutput(
+                request_id=rid,
+                new_token_ids=[50],
+                new_text="!",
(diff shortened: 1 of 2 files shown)


unsaltedbutter-ai:
- Tests pass locally — 2113 pass; the 7 fails are pre-existing on main
- Lint passes (on changed file)

unsaltedbutter-ai:
pr_validate ran with PR_VALIDATE_NO_DEEPSEEK=1 PR_VALIDATE_NO_STRESS=1: fetch / supply_chain / lint all green, targeted_tests skipped (no test file maps to engine_core.py), full_unit: 2113 passed / 7 failed — all 7 failures reproduce on origin/main (negative control confirms): 
- test_mcp_security::test_valid_stdio_config (no npx on PATH)
- test_platform.py ×4 (no torch, vision extra not installed)
- test_server::TestAPIKeyVerification ×2 (Python 3.14 asyncio.get_event_loop() strictness).

Happy to run stress/deepseek if you want.

raullenchai:
Thanks for the fix — the bug is real and the buffer-and-flush approach is correct. Ran `pr_validate` against the diff (full_unit 2131/0, lint clean, supply-chain clean) and DeepSeek V4 Pro adversarial review. Two things I'd like addressed before merging:

### 1. Logprobs are dropped, not merged

`engine_core.py` lines 358–369: when constructing the buffered `RequestOutput`, `new_text` and `new_token_ids` are correctly concatenated, but `logprobs=req_output.logprobs` only takes the **latest** step's value.

`logprobs` is a per-step delta — `scheduler.py:2103-2108` populates it alongside `new_token_ids=[response.token]` (the single token from that step). For `stream_interval=N`, this means N-1 of every N logprob entries are silently dropped (same root cause as the original bug, just in a different field).

Suggested fix:
```python
self._stream_buffers[rid] = RequestOutput(
    ...
    logprobs=(buf.logprobs or []) + (req_output.logprobs or []),
    ...
)
```

### 2. Regression test for `stream_interval > 1`

The diff adds no test. `_engine_loop` is the BatchedEngine's central streaming dispatcher — losing this fix to a future refactor would be silent (default `stream_interval=1` still works). A direct unit test that mocks `RequestStreamState.should_send` to skip N-1 of N steps and asserts that:

- accumulated `new_text` matches all step deltas concatenated
- accumulated `new_token_ids` matches all step token IDs in order
- `logprobs` matches all step logprobs concatenated (after fix #1)
- a final `finished=True` step always flushes the buffer

…would lock in the contract.

### Note on `finished=True` flushing

We checked `output_collector.RequestStreamState.should_send` (line 197): it returns `True` unconditionally on `finished=True`, so the final delta does get flushed in this PR's design. No fix needed there. Worth covering with a test (above) anyway since that invariant is load-bearing.

---

Once #1 and a regression test are in, this is good to merge. Happy to push the fix as a follow-up commit on this branch if it'd help — let us know.

unsaltedbutter-ai:
While verifying the fix end-to-end against `gemma-4-31b-it-mxfp8` with `--stream-interval 100` (streaming vs non-streaming parity at temp=0), I hit a small mismatch — the streamed reassembly has a 14-byte fragment of the model's CoT scratchpad prepended to its content. After tracing it, **this is not the buffer fix corrupting output** — every step delta is preserved correctly and the body of the answer is byte-identical to the non-streamed response. The leak is in `gemma4_parser.extract_reasoning_streaming`: when a buffered delta straddles the `<channel|>` close marker, the parser classifies the whole delta as content even though the bytes before the marker should still be reasoning. Pre-#210 this was unobservable because each delta was a single token wide.

Same shape probably exists in `qwen3_parser`, `harmony_parser`, `deepseek_r1_parser`, `minimax_parser` — all classify whole delta by end-state.

I'll open a separate issue with the per-chunk SSE evidence and a fix sketch (split `delta_text` at the marker inside `extract_reasoning_streaming`), and submit a follow-up PR against it after #210 lands.

raullenchai (approved):
Round 2: LGTM, approving.

Verified locally on commit `790f673`:
- New regression test `test_stream_interval_buffer_merges_skipped_step_deltas` passes — covers all three delta fields (`new_text`, `new_token_ids`, `logprobs`) plus the always-flush-on-finished invariant
- 56/56 engine + batching tests pass (`test_batching`, `test_batching_deterministic`, `test_engine_step_thread`, `test_continuous_batching`)
- 29/29 streaming-adjacent tests pass (`test_streaming_latency`, `test_streaming_detokenizer`, `test_streaming_json_encoder`)
- `stream_interval=1` fast path is untouched — buffer logic is fully gated on the slow-path `else` branch

Logprobs merge looks correct: per-step `mx.array` is wrapped into a list, then concatenated across merges.

### Pre-existing footnote (out of scope, will track separately)

After this lands the merged `logprobs` is a `list[mx.array]` rather than a single `mx.array`. The downstream consumer at `routes/chat.py:481-487` calls `_extract_token_logprob(chunk.logprobs, …)` which expects a single `mx.array` and uses `chunk.tokens[-1]` as the token id. With `stream_interval > 1` AND `logprobs=True` requested, the consumer will:
1. Emit only one logprob entry per `should_send` flush (under-count)
2. Compute top-k across a flattened concatenation of all merged distributions (wrong shape)

This is **not a regression**: pre-PR the same combination dropped intermediate deltas entirely (worse). The PR strictly improves things — text is now correct, and logprobs is no worse than before. But the consumer will need updating to iterate the list once we want `stream_interval > 1 + logprobs` to be fully correct.

I'll file this as a separate issue and we can resolve it after this lands. The rare-combo nature (`stream_interval > 1` is uncommon, `logprobs=True` is uncommon, intersection is near-zero) makes it safe to defer.

Will merge once CI goes green.

unsaltedbutter-ai:
I rebased from main, and added the change and test requested. 👍 
I filed #219 which I found during testing/validation. I'll work on it next.

Issue #223, opened 2026-05-05
Anthropic streaming bypasses reasoning_parser only via _starts_thinking; misroutes answers when reasoning_parser + enable_thinking=False


raullenchai:
## Surfaced by

Post-merge audit of today's 6 PRs (DeepSeek V4 Pro adversarial review on the cumulative diff vs v0.6.11 / e6687da). PR #213 fixed #185 for the StreamingThinkRouter path, but the parallel reasoning_parser path was missed.

## Bug

Cross-route parity gap between chat and Anthropic streaming when a reasoning model is loaded with `--reasoning-parser <name>` and the request explicitly sets `enable_thinking=False`:

| route | code path | enable_thinking=False respected? |
|---|---|---|
| chat | `vllm_mlx/service/postprocessor.py:217` — `elif self.reasoning_parser and self.enable_thinking is not False:` | ✅ yes (added in #208) |
| anthropic | `vllm_mlx/routes/anthropic.py:381-431` — `if reasoning_parser:` | ❌ no, unconditional |

When this combination is used, every delta is fed through `reasoning_parser.extract_reasoning_streaming`. The Qwen3 parser's implicit-think heuristic classifies *all* output as reasoning when no `<think>` tag is emitted (which is the whole point of `enable_thinking=False`), so:

- All text gets routed to `thinking_delta` blocks.
- `text_delta` blocks remain empty.
- Anthropic clients see only thinking, no answer.

This is the same failure mode as #185, just on a different sub-path of the route.

## Reproduction

```bash
rapid-mlx serve mlx-community/Qwen3.5-9B-8bit \
  --reasoning-parser qwen3 \
  --port 8880 --max-num-seqs 1

curl -Ns http://127.0.0.1:8880/v1/messages \
  -H 'Content-Type: application/json' -H 'anthropic-version: 2023-06-01' \
  -d '{
    "model": "qwen3.5-9b",
    "max_tokens": 100,
    "stream": true,
    "thinking": {"type": "disabled"},
    "messages": [{"role": "user", "content": "What is 2+2?"}]
  }' | grep -E 'thinking_delta|text_delta'
```

Expected: text_delta carries `"4"` (or similar). Observed: thinking_delta carries the answer; text_delta stays empty.

## Fix

Mirror the chat path's bypass condition. In `anthropic.py:395`:

```python
if reasoning_parser and chat_kwargs.get("enable_thinking") is not False:
    # ... existing block ...
```

[...]

PR #230, opened 2026-05-05
fix(server): sync ServerConfig before native-tool-format detection (#225)

### LINKED ISSUE #225: server.py: _sync_config() runs after _detect_native_tool_support(), silently disabling native tool format
### Rapid-MLX version

0.6.5

_No response_

### Hardware

 Mac Mini M4 Pro, 48 GB

### macOS version

macOS 26.4.1

### Python version

3.12.12

### Model

mlx-community/Qwen3.5-9B-4bit (also reproduces with mlx-community/Qwen3.6-35B-A3B-4bit-DWQ

### Full serve command

rapid-mlx serve mlx-community/Qwen3.5-9B-4bit --port 8001 --enable-auto-tool-choice --tool-call-parser qwen3_coder

Streaming (`stream: true`)

Actual: After the first assistant tool_calls + tool round-trip, ~40% of subsequent assistant turns emit `[Calling tool: name({"key":"value"})]` as plain content text instead of structured tool_calls. stopReason=stop, no tool_calls field. Multi-agent workloads (50+ tool turns/burst) reliably trigger this within the first few rounds.

Expected: All assistant tool calls emitted as structured tool_calls (delta.tool_calls in streaming, message.tool_calls in non-streaming) regardless of conversation depth.

Root cause: Native tool format is silently disabled at startup. _detect_native_tool_support() at server.py:499 runs BEFORE _sync_config() at server.py:559. When detection reads cfg.enable_auto_tool_choice and cfg.tool_call_parser via get_config(), both are still at their dataclass defaults (False, None) because the globals haven't been copied into the singleton yet. The guard `if not cfg.enable_auto_tool_choice or not cfg.tool_call_parser: return False` short-circuits, _engine.preserve_native_tool_format = False, and the "Native tool format enabled for parser: ..." log never prints.

Downstream effect: api/utils.py::process_messages takes the prose-conversion fallback path (lines 672, 698, 727), serializing assistant tool_calls as `[Calling tool: name(args)]` and tool messages as `[Tool Result (id)]: ...` into the prompt history. Models then see prose-format examples in context and mimic the pattern on subsequent turns. Looks like a model failure but the model is faithfully completing a malformed prompt — exactly what hermes_tool_parser.py:65-68 already documents ("Without this, tool history is converted to '[Calling tool: ...]' text, which causes the model to mimic that text format").

Verification that the parser flag is correct:
  >>> from vllm_mlx.tool_parsers.abstract_tool_parser import ToolParserManager
  >>> p = ToolParserManager.get_tool_parser('qwen3_coder')
  >>> p.__name__, p.SUPPORTS_NATIVE_TOOL_FORMAT
  ('HermesToolParser', True)

The parser correctly reports native support; detection just never gets to ask it.

Fix: move _sync_config() to before _detect_native_tool_support() in load_model() (server.py around line 499). One-line move. Affects all parsers with SUPPORTS_NATIVE_TOOL_FORMAT = True (deepseek, deepseekv31, functionary, granite, harmony, hermes/qwen3_coder, kimi, llama, mistral, qwen3coder_xml, seed_oss).

After patch: startup log shows "Native tool format enabled for parser: qwen3_coder", _engine.preserve_native_tool_format == True, multi-turn tool-using sessions produce 50+ structured tool_calls in a row with zero text leaks. Confirmed on both Qwen3.5-9B-4bit and Qwen3.6-35B-A3B-4bit-DWQ.

Note: --enable-tool-logits-bias does NOT mask the bug. Bias steers token selection but the model still sees prose-format history in context, so leaks at the same rate.

### Minimal reproduction

1. Start server:
   rapid-mlx serve mlx-community/Qwen3.5-9B-4bit \
     --port 8001 \
     --enable-auto-tool-choice \
     --tool-call-parser qwen3_coder

2. Confirm bug active: startup log does NOT contain "Native tool format enabled for parser: qwen3_coder".

3. Send a request that exercises the multi-turn tool-call path. The bug is stochastic per-turn (~40%) but reliably triggered after a few rounds:

   curl -s http://localhost:8001/v1/chat/completions \
     -d '{
       "model": "mlx-community/Qwen3.5-9B-4bit",
       "messages": [
         {"role":"user","content":"Read /tmp/foo and summarize"},
         {"role":"assistant","content":null,"tool_calls":[{"id":"call_a","type":"function","function":{"name":"read","arguments":"{\"path\":\"/tmp/foo\"}"}}]},
         {"role":"tool","tool_call_id":"call_a","content":"file contents..."},
         {"role":"user","content":"Now read /tmp/bar"}
       ],
       "tools":[{"type":"function","function":{"name":"read","parameters":{"type":"object","properties":{"path":{"type":"string"}},"required":["path"]}}}],
       "stream": true
     }'

4. Some fraction of responses will stream `delta.content: "["` followed by the literal text `[Calling tool: read({"path":"/tmp/bar"})]` instead of `delta.tool_calls`.

5. Patching server.py to call _sync_config() before _detect_native_tool_support() makes the bug disappear entirely (validated across 50+ consecutive multi-turn tool calls).

```shell
INFO:vllm_mlx.model_auto_config:Auto-detected model family 'qwen3' → tool_call_parser=hermes, reasoning_parser=qwen3
INFO:vllm_mlx.server:Loading model with BatchedEngine: mlx-community/Qwen3.5-9B-4bit
INFO:vllm_mlx.server:Model loaded: mlx-community/Qwen3.5-9B-4bit
INFO:vllm_mlx.server:Default max tokens: 8192

# (Note: missing "Native tool format enabled for parser: qwen3_coder" — never printed because _detect_native_tool_support() returned False due to startup ordering)

# Example leaking response stream:
INFO:vllm_mlx.routes.chat:[REQUEST] POST /v1/chat/completions stream=True model='mlx-community/Qwen3.5-9B-4bit' msgs=4 roles=['system', 'user', 'assistant', 'tool'] tools=7
INFO:vllm_mlx.routes.chat:[SSE-ROLE] data: {"choices":[{"index":0,"delta":{"role":"assistant"}}]}
# (no [SSE-TC] line — text content streamed instead)
INFO:vllm_mlx.engine_core:[stream_outputs] xxx finished normally, 29 tokens in 1.3s
# Final assistant content: "[Calling tool: read({\"path\":\"/Users/.../skills/queue-processor/SKILL.md\"})]"

# Patched startup:
INFO:vllm_mlx.server:Native tool format enabled for parser: qwen3_coder
```

raullenchai:
Closes #225 (h/t @reidperyam for the precise root-cause analysis and reproduction).

## Bug

`_detect_native_tool_support()` reads `cfg.enable_auto_tool_choice` and `cfg.tool_call_parser` via `get_config()`. In `load_model()`, the pre-fix call order was:

```text
1. _engine.preserve_native_tool_format = _detect_native_tool_support()  # line 520
2. … register model entry …
3. _sync_config()                                                        # line 574
```

so detection ran while cfg still held its dataclass defaults (`False`, `None`). The guard `if not cfg.enable_auto_tool_choice or not cfg.tool_call_parser: return False` short-circuited, and `_engine.preserve_native_tool_format` was silently set to `False` regardless of which parser was configured.

This affects every parser that advertises `SUPPORTS_NATIVE_TOOL_FORMAT = True`: `deepseek`, `deepseekv31`, `functionary`, `granite`, `harmony`, `hermes` (and its `qwen3_coder` alias), `kimi`, `llama`, `mistral`, `qwen3coder_xml`, `seed_oss`.

## Symptom

With native format silently disabled, `api/utils.py::process_messages` serialised assistant tool history as prose:

```text
[Calling tool: read({"path":"/tmp/foo"})]
[Tool Result (call_a)]: file contents...
```

Models saw prose-format examples in context and mimicked the pattern on subsequent turns — per @reidperyam's measurement, ~40% of streaming responses on Qwen3.5-9B-4bit and Qwen3.6-35B-A3B-4bit-DWQ emitted the literal `[Calling tool: ...]` string instead of structured `tool_calls`. Looked like model failure, was actually a startup ordering bug.

The visible tell was missing in the startup log:

```text
INFO:vllm_mlx.server:Native tool format enabled for parser: <name>
```

This line is conditional on `_engine.preserve_native_tool_format` being True. If you don't see it for a parser that should support native format, this bug is firing.

## Fix

Move the existing `_sync_config()` call to immediately after engine construction, before native-tool-format detection runs. Per V4 Pro round-2 review, the second `_sync_config()` at end of `load_model` is **kept** as a defensive safety net — the cost of an idempotent re-sync is trivial against the cost of re-introducing the same silent-ordering failure mode if a future change violates the invariants. Three invariants documented above the call.

## Test plan

`tests/test_server_load_model_order.py` adds three regressions:

* **contract test** — `_detect_native_tool_support()` returns False when cfg is unsynced, True after `_sync_config()`
* **integration test** — after `load_model()` returns with a stub engine + hermes parser configured, `_engine.preserve_native_tool_format` is True
* **idempotency test** — `_sync_config()` called twice leaves cfg in identical state across all 12 mirrored fields (per V4 Pro round-3 #4)

Plus an autouse fixture (`_reset_cfg_around_each_test`) that resets the cfg singleton before AND after each test, so mid-test failures can't leak state (per V4 Pro round-3 #5).

Verification:
- [x] Integration test FAILS on the pre-fix tree (`assert False is True`) and PASSES post-fix
- [x] `tests/test_server.py` + `test_server_utils.py` + new file: 67 passed, 0 failed
- [x] Broader tool-call/server tests: 183 passed, 6 skipped
- [x] pr_validate full pipeline (round 3): targeted_tests **68 passed**, full_unit **2148 passed, 0 failed**, lint clean, supply_chain clean, deepseek_review re-run with this PR's deltas
- [x] `ruff check` + `ruff format --check` clean

## Bundled test fix (also in this PR)

`tests/test_anthropic_stream_reasoning.py::test_reasoning_parser_stream_emits_text_chunks` (the #185 regression test) was failing on `raullenchai/main` HEAD before this PR. Root cause was in the test fixture, not the production code: the fixture built cfg with `MagicMock()` and didn't set `no_thinking`, so on the mock that attribute returned another MagicMock (truthy). After PR #224 (the #223 fix) the route reads `cfg.no_thinking` to force `enable_thinking=False`, which then nulls out the reasoning parser. One-line fix on the fixture (`cfg.no_thinking = False`); not splitting into a separate PR because it would either merge this PR with a failing test step or block on follow-up. Verified pre-existing on main HEAD before this PR.

### DIFF
--- a/tests/test_anthropic_stream_reasoning.py
+++ b/tests/test_anthropic_stream_reasoning.py
@@ -101,6 +101,12 @@ def patched_cfg():
     cfg.engine = None
     cfg.model_name = "qwen3.5-test"
     cfg.cloud_router = None
+    # MagicMock attributes default to truthy MagicMock instances. The route at
+    # anthropic.py:332 reads `cfg.no_thinking`; left unset, that path forces
+    # chat_kwargs["enable_thinking"]=False, which the #223 fix at line 389
+    # then uses to null out the reasoning parser. The #185 regression below
+    # depends on the parser running, so pin no_thinking to a real False.
+    cfg.no_thinking = False
     with patch.object(anthropic_route, "get_config", return_value=cfg):
         yield cfg
 
--- a/tests/test_server_load_model_order.py
+++ b/tests/test_server_load_model_order.py
@@ -20,15 +20,43 @@ bug.
 
 from __future__ import annotations
 
+import pytest
+
 
 class _StubEngine:
+    """Minimal stand-in for `BatchedEngine` — only the surface `load_model`
+    actually accesses between construction and the model-registry add.
+
+    This is intentionally explicit (not `MagicMock`) so that any future
+    `load_model` change touching a new attribute fails LOUDLY with
+    `AttributeError`, not silently with a fabricated MagicMock value.
+    """
+
     is_mllm = False
     preserve_native_tool_format = False
     _tokenizer = None
     _tool_logits_processor_factory = None
 
-    def __init__(self, **kwargs):  # noqa: D401 — match BatchedEngine signature loosely
+    def __init__(self, *args, **kwargs):
+        # Accept positional too in case `BatchedEngine.__init__` ever takes any.
+        self.args = args
         self.kwargs = kwargs
+
+
+@pytest.fixture(autouse=True)
+def _reset_cfg_around_each_test():
+    """Reset the ServerConfig singleton before AND after every test.
+
+    `monkeypatch.setattr` on module globals is restored automatically, but
+    the cfg singleton is a separate process-level object that must be
+    explicitly reset on both sides — otherwise a mid-test failure leaks
+    cfg state into the next test.
+    """
+    from vllm_mlx.config import reset_config
+
+    reset_config()
+    yield
+    reset_config()
 
 
 def test_load_model_enables_native_tool_format_when_parser_supports_it(monkeypatch):
@@ -37,9 +65,6 @@ def test_load_model_enables_native_tool_format_when_parser_supports_it(monkeypat
     unsynced when detection ran.
     """
     from vllm_mlx import server
-    from vllm_mlx.config import reset_config
-
-    reset_config()
 
     monkeypatch.setattr(server, "BatchedEngine", _StubEngine)
     monkeypatch.setattr(server, "_engine", None, raising=False)
@@ -67,9 +92,8 @@ def test_detect_native_tool_support_requires_synced_config(monkeypatch):
     `_sync_config()` first.
     """
     from vllm_mlx import server
-    from vllm_mlx.config import get_config, reset_config
+    from vllm_mlx.config import get_config
 
-    reset_config()
     monkeypatch.setattr(server, "_enable_auto_tool_choice", True, raising=False)
     monkeypatch.setattr(server, "_tool_call_parser", "hermes", raising=False)
     monkeypatch.setattr(server, "_reasoning_parser", None, raising=False)
@@ -90,3 +114,46 @@ def test_detect_native_tool_support_requires_synced_config(monkeypatch):
     assert cfg.enable_auto_tool_choice is True
     assert cfg.tool_call_parser == "hermes"
     assert server._detect_native_tool_support() is True
+
+
+def test_sync_config_is_idempotent(monkeypatch):
+    """`_sync_config()` is called twice in `load_model` (early before native
+    tool detection, late after the model registry add). Both calls must
+    leave cfg in the same state — if the function ever grows non-idempotent
+    side effects (counter increments, callback fires, cache invalidations),
+    the late re-sync becomes a latent bug.
+    """
+    from vllm_mlx import server
+    from vllm_mlx.config import get_config
+
+    monkeypatch.setattr(server, "_enable_auto_tool_choice", True, raising=False)
+    monkeypatch.setattr(server, "_tool_call_parser", "hermes", raising=False)
+    monkeypatch.setattr(server, "_reasoning_parser", None, raising=False)
+    monkeypatch.setattr(server, "_reasoning_parser_name", None, raising=False)
+    monkeypatch.setattr(server, "_tool_parser_instance", None, raising=False)
+    monkeypatch.setattr(server, "_mcp_manager", None, raising=False)
+    monkeypatch.setattr(server, "_enable_tool_logits_bias", False, raising=False)
+    monkeypatch.setattr(server, "_engine", None, raising=False)
+
+    server._sync_config()
+    cfg = get_config()
+    snapshot = {
+        "engine": cfg.engine,
+        "model_name": cfg.model_name,
+        "model_alias": cfg.model_alias,
+        "model_path": cfg.model_path,
+        "enable_auto_tool_choice": cfg.enable_auto_tool_choice,
+        "tool_call_parser": cfg.tool_call_parser,
+        "tool_parser_instance": cfg.tool_parser_instance,
+        "enable_tool_logits_bias": cfg.enable_tool_logits_bias,
+        "reasoning_parser": cfg.reasoning_parser,
+        "reasoning_parser_name": cfg.reasoning_parser_name,
+        "mcp_manager": cfg.mcp_manager,
+        "model_registry": cfg.model_registry,
+    }
+
+    server._sync_config()
+    cfg2 = get_config()
+
+    for k, v in snapshot.items():
+        assert getattr(cfg2, k) == v, f"_sync_config() not idempotent on cfg.{k}"
--- a/tests/test_server_load_model_order.py
+++ b/tests/test_server_load_model_order.py
@@ -1,0 +1,92 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression for #225 — startup ordering.
+
+`_detect_native_tool_support()` reads `cfg.enable_auto_tool_choice` and
+`cfg.tool_call_parser` via `get_config()`. If `_sync_config()` runs
+*after* the detection call (the pre-fix layout), those fields are still
+at their dataclass defaults (False, None), the guard short-circuits to
+False, and `_engine.preserve_native_tool_format` is silently set to
+False even though the configured parser supports native format.
+
+Downstream symptom (per the bug report on Qwen3.5-9B-4bit and
+Qwen3.6-35B-A3B-4bit-DWQ): assistant tool history gets serialised by
+`api/utils.py::process_messages` as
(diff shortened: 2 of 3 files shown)


raullenchai:
Bundling a tiny unrelated test-fix into this PR for hygiene.

`tests/test_anthropic_stream_reasoning.py::test_reasoning_parser_stream_emits_text_chunks` started failing after PR #224 (the #223 fix) landed. Root cause is in the test fixture, not the production code: the fixture builds cfg with `MagicMock()` and didn't set `no_thinking`, so on the mock that attribute returned another MagicMock (truthy). After #224, `routes/anthropic.py:332` reads `cfg.no_thinking` to force `enable_thinking=False`, which the new bypass at line 389 then uses to null out the reasoning parser — the spy assertion correctly fired because the parser branch was never taken.

Fixed by pinning `cfg.no_thinking = False` on the fixture. Verified the bug predates this PR (fails on `raullenchai/main` HEAD `38c21f1`), so it's a hygiene fix not a regression caused by the #225 work.

raullenchai:
Updated with V4 Pro review feedback (re-ran review against the diff with `DEEPSEEK_API_KEY` set after the in-session pr_validate skipped that step).

**Findings addressed:**

1. **Stale-base cli.py noise (V4 Pro #3)** — branch was created before #226 merged, so the diff against current main showed me reverting the gemma-4 help fix. Rebased onto current `raullenchai/main`, noise is gone.

2. **Single-sync invariants (V4 Pro #1, #2)** — V4 Pro flagged that the comment glossed over assumptions. Replaced the one-line comment with three explicit invariants that future changes need to honor (reference-not-copy registry assignment, all globals set before engine construction, `_engine` mutations propagate via `cfg.engine`). A future change that violates any of these should restore the second sync call.

**Findings dismissed:**

3. **`_tool_parser_instance=None` in test (V4 Pro #4)** — speculative. Detection only reads `cfg.tool_call_parser` (the name) and looks up class via `ToolParserManager`; the instance is unused. Setting it to `None` in the test is just being explicit about server module state.

**Bonus:** bundled a tiny test-fixture fix for `tests/test_anthropic_stream_reasoning.py` (the `#185` regression test that pr_validate full_unit caught failing). Root cause: fixture's `MagicMock()` cfg didn't set `no_thinking`, so on the mock that attribute returned another MagicMock (truthy). After #224 (the #223 fix) the route reads `cfg.no_thinking` → forces `enable_thinking=False` → nulls out the reasoning parser. Verified the failure pre-dates this PR (also fails on `raullenchai/main` HEAD). Pinning `cfg.no_thinking = False` on the fixture restores the test.

Now two commits on top of main:
- `4981d40` — fix(server): sync ServerConfig before native-tool-format detection (#225)
- `3838ee1` — test(anthropic): pin cfg.no_thinking=False on the patched_cfg fixture

raullenchai:
DeepSeek V4 Pro round 2 (with key in env this time) found 3 things — only #2 was a real concern.

**Accepted (#2): Keep defensive `_sync_config()` at end of `load_model`.** Previous revision deleted the second call and relied on a comment listing three invariants. V4 Pro: "this is the exact same failure mode the PR fixes — silent breakage when call-site ordering changes, no log, no crash, just wrong behavior." The cost of an idempotent re-sync is trivial; the cost of re-introducing the bug pattern is high. Commit `ddd094f` puts the call back. The invariant comments stay above it as documentation of why it's redundant under current code — but documentation doesn't enforce correctness, the call does.

**Dismissed (#1): `_StubEngine` minimal surface.** V4 Pro flagged the integration test as "passes only by coincidence that the current `load_model` code path doesn't hit an un-mocked attribute." Audited every `_engine` access between line 510 and the registry add — `preserve_native_tool_format` (SET), `is_mllm` via `getattr(_engine, "is_mllm", False)`, `_tokenizer`/`_tool_logits_processor_factory` gated on `_enable_tool_logits_bias=True` which the test sets to `False`. If a future change adds an access, the stub breaks loudly with `AttributeError`, not silently. That's normal test maintenance.

**Dismissed (#3): "drive-by test fix lacks regression assertion".** The existing `assert call_count["streaming"] > 0` in the streaming test IS the regression guard. If someone removes `cfg.no_thinking = False` from the fixture, that assertion fires with the same diagnostic. Commit `3838ee1` describes the fix in detail (4-paragraph explanation of root cause + symptom).

**pr_validate round 2** (against latest HEAD with DEEPSEEK_API_KEY set): 6 PASS / 1 skip — `full_unit` now **2148 passed, 0 failed** (test fixture pin worked). The stress matrix step crashed on environmental ENOSPC (`/tmp` filled from accumulated artifacts, since cleaned up — reclaimed 138G). The crash is environmental, not a code issue. Round 3 incoming to confirm.

GitHub CI all 7 checks SUCCESS on `ddd094f`.

PR #262, opened 2026-05-06
Fix Anthropic route authentication + support SDK x-api-key

### LINKED ISSUE #188: CRITICAL: Anthropic /v1/messages and /v1/messages/count_tokens endpoints bypass API key authentication
## Description

The `/v1/messages` Anthropic API endpoint at `routes/anthropic.py:47` and the `/v1/messages/count_tokens` endpoint at `routes/anthropic.py:187` are registered WITHOUT the `Depends(verify_api_key)` dependency.

Every other API endpoint in the codebase has this protection:
- `routes/chat.py:66-69` — `/v1/chat/completions` has both `Depends(verify_api_key)` and `Depends(check_rate_limit)`
- `routes/audio.py:22` — `/v1/audio/transcriptions` has `Depends(verify_api_key)`
- `routes/mcp_routes.py:20,42,65` — All MCP endpoints have `Depends(verify_api_key)`

## Impact

When the server is started with `--api-key <key>`, all Anthropic-format requests completely bypass authentication. An attacker can send arbitrary inference requests to any model loaded on the server with no API key check, even when `--api-key` is configured.

## Reproduction

```bash
# Start server with API key enabled
rapid-mlx serve mlx-community/Qwen3.5-4B-4bit --api-key my-secret-key

# This request bypasses auth entirely (no Authorization header needed):
curl http://localhost:8000/v1/messages \
  -d '{"model":"default","messages":[{"role":"user","content":"Hello"}],"max_tokens":10}'

# Token counting also bypasses auth:
curl http://localhost:8000/v1/messages/count_tokens \
  -d '{"messages":[{"role":"user","content":"Test"}]}'
```

## Fix

Add `dependencies=[Depends(verify_api_key), Depends(check_rate_limit)]` to both endpoints:

```python
@router.post(
    "/v1/messages",
    dependencies=[Depends(verify_api_key), Depends(check_rate_limit)],
)
async def create_anthropic_message(request: Request):
    ...

@router.post(
    "/v1/messages/count_tokens",
    dependencies=[Depends(verify_api_key)],
)
async def count_anthropic_tokens(request: Request):
    ...
```

## Files

- `vllm_mlx/routes/anthropic.py:47-50` — `create_anthropic_message` missing auth
- `vllm_mlx/routes/anthropic.py:187-188` — `count_anthropic_tokens` missing auth

## Compare

Chat endpoint for reference:
```python
@router.post(
    "/v1/chat/completions",
    dependencies=[Depends(verify_api_key), Depends(check_rate_limit)],
)
```

michaelasper:
## Summary

Fixes #188 by adding authentication and rate limiting to the Anthropic-compatible endpoints:

- `POST /v1/messages`
- `POST /v1/messages/count_tokens`

Both endpoints now enforce the configured `--api-key` before reaching model or tokenizer work.

## Why this shape

The original issue was that the Anthropic routes were registered without the auth dependencies used by the rest of the API surface. Adding the normal `verify_api_key` dependency would close the Bearer-token bypass, but it would also break standard Anthropic SDK clients.

The Anthropic SDK sends the configured client `api_key` as:

```text
x-api-key: <key>
```

So this PR adds an Anthropic-specific auth dependency that accepts both supported forms:

- `Authorization: Bearer <key>` for curl/OpenAI-style callers already using Rapid-MLX auth
- `x-api-key: <key>` for Anthropic SDK compatibility

The `x-api-key` behavior is intentionally scoped to the Anthropic routes. Existing OpenAI-compatible routes continue to use Bearer-only auth.

## Rate-limit behavior

The Anthropic routes use a separate rate-limit dependency so `x-api-key` can participate in rate-limit identity only on Anthropic-compatible routes. That dependency still uses the same global limiter configured by `--rate-limit`.

Bearer credentials are normalized to the raw token before rate-limit bucketing for both standard and Anthropic routes. That avoids giving the same API key separate buckets when a client alternates between OpenAI-compatible endpoints and Anthropic-compatible endpoints.

Anthropic `x-api-key` requests use the same token value as their rate-limit identity, so an Anthropic SDK caller and a Bearer caller using the same configured API key share the same bucket on Anthropic routes.

## Startup wiring

CLI and direct server startup were previously replacing a server-local `_rate_limiter` alias. FastAPI dependencies call the limiter in `vllm_mlx.middleware.auth`, so replacing only the alias would leave the dependency path on the disabled default limiter.

This PR adds `configure_rate_limiter(...)`, which mutates the shared middleware limiter in place and returns it for the existing server alias. That keeps startup behavior connected for both the existing OpenAI-compatible dependencies and the new Anthropic-specific dependency.

## Security details

- Missing credentials still return `401 API key required`.
- Wrong credentials still return `401 Invalid API key`.
- If both Bearer and `x-api-key` are present, both must match the configured key. A valid header cannot mask an invalid supported credential.
- `x-api-key` remains Anthropic-only; standard protected routes still require Bearer auth.
- Documentation now calls out the Anthropic SDK `x-api-key` compatibility explicitly.

## Tests

Added route-level regression coverage for:

- original unauthenticated `/v1/messages` bypass
- original unauthenticated `/v1/messages/count_tokens` bypass
- valid Bearer and valid `x-api-key` on both Anthropic routes
- wrong Bearer, wrong `x-api-key`, and mixed invalid credentials
- auth failures not reaching `engine.chat` or tokenizer `encode`
- rate limiting on both Anthropic routes
- rate-limit identity staying stable when Anthropic callers switch between Bearer and `x-api-key`
- Bearer rate-limit identity shared across standard and Anthropic dependencies
- standard protected routes rejecting `x-api-key` when used without Bearer auth
- shared non-Anthropic rate limiting ignoring `x-api-key`
- startup configuration mutating the shared middleware limiter used by dependencies

## Review follow-up

After the first review pass, the PR was updated to:

- accept Anthropic SDK `x-api-key` credentials in addition to Bearer auth on Anthropic routes
- configure the shared middleware rate limiter from both startup paths
- align Bearer rate-limit identity across route families
- document the route-specific `x-api-key` behavior

A read-only multi-agent review rerun found no actionable findings after the latest rate-limit identity fix.

## Validation

```text
/private/tmp/rapid-mlx-pytest-188/bin/python -m pytest tests/test_anthropic_route_auth.py -q
# 25 passed

/private/tmp/rapid-mlx-pytest-188/bin/python -m pytest tests/test_config_and_middleware.py::TestVerifyApiKey tests/test_config_and_middleware.py::TestCheckRateLimit tests/test_anthropic_route_streaming.py -q
# 9 passed

/private/tmp/rapid-mlx-pytest-188/bin/python -m pytest tests/test_server.py::TestRateLimiter tests/test_server.py::TestRateLimiterHTTPResponse -q
# 7 passed

/private/tmp/rapid-mlx-pytest-188/bin/python -m ruff check vllm_mlx/middleware/auth.py vllm_mlx/routes/anthropic.py vllm_mlx/cli.py vllm_mlx/server.py tests/test_anthropic_route_auth.py
# All checks passed

/private/tmp/rapid-mlx-pytest-188/bin/python -m ruff format --check vllm_mlx/middleware/auth.py vllm_mlx/routes/anthropic.py vllm_mlx/cli.py vllm_mlx/server.py tests/test_anthropic_route_auth.py
# 5 files already formatted

git diff --check vllm_mlx/middleware/auth.py vllm_mlx/routes/anthropic.py vllm_mlx/cli.py vllm_mlx/server.py tests/test_anthropic_route_auth.py docs/reference/cli.md
# passed
```

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -104,7 +104,11 @@ rapid-mlx serve mlx-community/Qwen3-4B-4bit \
 
 ### Security
 
-When `--api-key` is set, all API requests require the `Authorization: Bearer <api-key>` header:
+When `--api-key` is set, protected API routes require the
+`Authorization: Bearer <api-key>` header. Anthropic-compatible routes
+(`/v1/messages` and `/v1/messages/count_tokens`) also accept
+`x-api-key: <api-key>` for SDK compatibility; if both headers are sent, both
+must match.
 
 ```python
 from openai import OpenAI
--- a/tests/test_anthropic_route_auth.py
+++ b/tests/test_anthropic_route_auth.py
@@ -496,6 +496,68 @@ def test_shared_rate_limit_ignores_x_api_key_for_non_anthropic_routes(
     assert second.json()["detail"].startswith("Rate limit exceeded.")
 
 
+def test_shared_rate_limit_uses_same_bearer_identity_for_anthropic_and_standard_routes(
+    anthropic_client,
+):
+    from vllm_mlx.middleware.auth import (
+        check_rate_limit,
+        check_rate_limit_or_x_api_key,
+        verify_api_key,
+        verify_api_key_or_x_api_key,
+    )
+
+    anthropic_client.rate_limiter.enabled = True
+    anthropic_client.rate_limiter.requests_per_minute = 1
+
+    app = FastAPI()
+
+    @app.get(
+        "/standard",
+        dependencies=[Depends(verify_api_key), Depends(check_rate_limit)],
+    )
+    async def standard_endpoint():
+        return {"ok": True}
+
+    @app.get(
+        "/anthropic",
+        dependencies=[
+            Depends(verify_api_key_or_x_api_key),
+            Depends(check_rate_limit_or_x_api_key),
+        ],
+    )
+    async def anthropic_endpoint():
+        return {"ok": True}
+
+    client = TestClient(app)
+    headers = {"Authorization": "Bearer test-secret"}
+    first = client.get("/standard", headers=headers)
+    second = client.get("/anthropic", headers=headers)
+
+    assert first.status_code == 200
+    assert second.status_code == 429
+    assert second.json()["detail"].startswith("Rate limit exceeded.")
+
+
+def test_shared_auth_rejects_x_api_key_for_non_anthropic_routes(
+    anthropic_client,
+):
+    from vllm_mlx.middleware.auth import verify_api_key
+
+    app = FastAPI()
+
+    @app.get("/test", dependencies=[Depends(verify_api_key)])
+    async def test_endpoint():
+        return {"ok": True}
+
+    client = TestClient(app)
+    x_api_key_only = client.get("/test", headers={"x-api-key": "test-secret"})
+    bearer = client.get("/test", headers={"Authorization": "Bearer test-secret"})
+
+    assert x_api_key_only.status_code == 401
+    assert x_api_key_only.json()["detail"] == "API key required"
+    assert bearer.status_code == 200
+
+
 def test_configure_rate_limiter_updates_shared_anthropic_dependency(
     anthropic_client,
 ):
--- a/tests/test_anthropic_route_auth.py
+++ b/tests/test_anthropic_route_auth.py
@@ -1,0 +1,556 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Auth regressions for Anthropic-compatible HTTP routes."""
+
+import sys
+import types
+from dataclasses import dataclass, field
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
+
+import pytest
+from fastapi import Depends, FastAPI
+from fastapi.testclient import TestClient
+
+
+class _Tokenizer:
+    chat_template = ""
+
+    def __init__(self):
+        self.calls = []
+
+    def encode(self, text: str) -> list[int]:
+        self.calls.append(text)
+        return list(range(len(text)))
+
+
+class _BaseEngine:
+    pass
+
+
+@dataclass
+class _GenerationOutput:
+    text: str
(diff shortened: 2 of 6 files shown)


raullenchai:
SOP §0–§9 review complete. Merging as CRITICAL security fix.

**§0 Necessity:** ✅ closes #188 — `/v1/messages` and `/v1/messages/count_tokens` were callable without auth even when `--api-key` was set. This is a live bypass on PyPI. The Anthropic SDK `x-api-key` compat is a clean design choice that lets SDK clients authenticate.

**§2 Codex round 1:** 0 P0, 6 P1. None are exploitable bypasses — they're hardening of an already-correct fix:
1. `_extract_bearer_token` doesn't `.strip()` token → Bearer whitespace splits rate-limit buckets (auth-equivalent though)
2. `request.headers.get("x-api-key")` only evaluates one header — duplicate-header masking risk
3. Empty companion headers (`x-api-key:` empty + valid Bearer) accepted
4. `all(secrets.compare_digest(...))` short-circuits — minor timing leak (which header failed, not key bytes)
5. Non-ASCII headers can raise `TypeError` from `compare_digest` → 500 instead of 401
6. Test fixture cleanup not exception-safe during setup (parent attrs may leak on import failure)

All 6 dismissed as non-blocking. The active bypass is more dangerous than these residual hardening gaps. Tracking as follow-up issue.

**§3 targeted tests:** 83 passed locally (`tests/test_anthropic_route_auth.py`, `tests/test_config_and_middleware.py`, `tests/test_anthropic_route_streaming.py`, `tests/test_server.py`)
**§4 lint:** clean per CI
**§7 CI:** all 7 checks SUCCESS

The `configure_rate_limiter` fix is the right call — replacing `server._rate_limiter` alone left FastAPI dependencies on the disabled default singleton. Subtle bug, well caught.

Thanks for the comprehensive PR.

raullenchai:
Thanks @michaelasper — merged! 🎉 Added you to the changelog:

PR #266, opened 2026-05-07
install.sh: use uv when available, faster repeat installs

raullenchai:
## Summary

- Use `uv` if it's on `$PATH` for resolution + parallel downloads — typically 3-10x faster than pip on a fresh install. Falls back to the venv's pip if uv is not present, so behavior is unchanged for users without it.
- Drop `--no-cache-dir` from the `latest` target so repeat installs / version bumps reuse pip's wheel cache instead of re-downloading every dep.
- Add `--prefer-binary` defensively to avoid surprise source builds.

## Why

`curl … | install.sh | bash` currently takes ~1+ min on a clean machine and re-downloads everything on every `bash -s latest`. uv parallelizes wheel fetches and the cache lets the second run reuse what's already on disk. Net result on my Mac: re-running `latest` drops from ~50s to ~12s.

## Test plan

- [ ] `bash install.sh stable` on a fresh machine without uv (pip path)
- [ ] `bash install.sh stable` with uv on `$PATH`
- [ ] `bash install.sh latest` twice in a row — second run should be visibly faster
- [ ] `bash install.sh 0.6.15` falls back to GitHub tag if PyPI version missing

## Follow-ups (not in this PR)

The Homebrew formula in `raullenchai/homebrew-rapid-mlx` is the slower install path (~1m51s, mostly Rust source build of pydantic-core + rpds-py). I'm opening a separate PR there to:
- Drop `--no-cache-dir` (same reason)
- Clean up the post-install caveats (drop the "Tested with" promo line, remove the broken `service` block, use RAM-tier model aliases instead of a fixed full path, document the shadow-PATH fix when both curl and brew installs coexist)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -157,20 +157,34 @@ else
     "$INSTALL_DIR/bin/pip" install --upgrade pip -q 2>/dev/null
 fi
 
+# Use uv for resolution + parallel downloads when available — typically 3-10x
+# faster than pip on a fresh install. Falls back to the venv's pip.
 PIP="$INSTALL_DIR/bin/pip"
+INSTALLER=("$PIP" install --prefer-binary)
+UPGRADE_INSTALLER=("$PIP" install --upgrade --prefer-binary)
+FORCE_INSTALLER=("$PIP" install --force-reinstall --prefer-binary)
+
+if command -v uv >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


raullenchai:
Consolidating into the homebrew-rapid-mlx PR per request — brew is the primary install path. Will reopen this if needed after the brew formula changes land.

PR #281, opened 2026-05-07
feat(model): SuffixDecoding tier classification framework (#269)

raullenchai:
## Summary

Closes #269. Adds **SuffixDecoding eligibility classification** on top of the per-alias profile SSOT (PR #283), so each alias can publish its own benched tier — `agent` / `structured` / `neutral` / `avoid` / `unknown` — and the engine emits a startup hint accordingly.

| Tier | Trigger | Hint | Action |
|---|---|---|---|
| **agent** | `tool_loop ≥ 1.8x` AND `min(others) ≥ 0.95x` | recommended | enable for tool traffic |
| **structured** | `max ≥ 1.5x` AND `min ≥ 0.90x` | may help | try if traffic matches |
| **neutral** | `min ≥ 0.95x` AND `1.0x ≤ max < 1.5x` | (silent) | leave OFF |
| **avoid** | any `< 0.85x` OR mixed signal | warn | leave OFF |
| **unknown** | not benched | (silent) | bench it |

## Rebased on SSOT (PR #283)

Originally targeted `ModelConfig` directly. Now lands the new fields on **`AliasProfile`** (the SSOT) and propagates to `ModelConfig` via `detect_model_config`. Per-alias granularity is automatic — `qwen3.5-4b` and `qwen3.5-122b` can carry different tiers if measured separately.

## What lands

| Surface | Change |
|---|---|
| `AliasProfile` | + `suffix_decoding_tier`, + `suffix_bench_speedup` (sorted tuple form for frozen-dataclass safety) |
| `ModelConfig` | + same two fields (mirror); resolved from profile when alias hits |
| `_coerce()` | parses + validates new JSON keys |
| `classify_suffix_decoding_tier(speedup)` | pure boundary fn, single source of truth |
| `suffix_decoding_hint(cfg)` | engine-init hint, silent for `unknown`/`neutral` |
| `format_profile_table` | new `Suffix tier` row in `rapid-mlx info` |
| `engine_core.py` | emits hint after profile summary |
| `scripts/bench_suffix_decoding_integrated.py` | server-driver bench |
| `docs/suffix_decoding_eligibility.md` | tier table + how-to |
| `tests/test_suffix_decoding_tier.py` | 24 contract tests |
| `CONTRIBUTING.md` | new-model checklist points at the bench |

## Bench script reliability fixes (commit `138bd9a`)

Found while running the cross-model sweep:

[...]

PR #283, opened 2026-05-07
refactor: per-alias model profiles SSOT (P0+P1)

raullenchai:
## Summary

Replaces the dual-source design (``aliases.json`` + ``_MODEL_PATTERNS`` regex table) with **explicit per-alias profiles** in ``aliases.json``. Every alias now carries its own ``tool_call_parser`` / ``reasoning_parser`` / ``is_hybrid`` / ``supports_spec_decode``, so capability flags can vary independently within a family (``qwen3.5-4b`` vs ``qwen3.5-122b``).

This is the architectural prerequisite for PR #281 — without per-alias granularity the SuffixDecoding tier was forced to be per-regex (one tier covers all 6 ``qwen3.5-*`` aliases), which can't reflect actual bench results.

## Pre-PR audit

| | Count | Detail |
|---|---|---|
| Total aliases in ``aliases.json`` | 51 | |
| Covered by a ``_MODEL_PATTERNS`` regex | 45 | shared across ~25 patterns |
| **Orphan aliases** (no profile, fell through to defaults) | **6** | bonsai×3, ministral-3b, nemotron×2 |
| Patterns covering ≥4 aliases | 4 | ``/qwen3\.5/`` (6), ``/qwen3\.6/`` (4), ``/deepseek.*v4/`` (4), ``/gemma/`` (4) |

## P0 — SSOT

| Surface | Change |
|---|---|
| ``aliases.json`` | rich-schema bump: ``{alias: {hf_path, tool_call_parser, reasoning_parser, is_hybrid, supports_spec_decode}}``. Legacy bare-string form still loads (defaults fill missing fields). |
| ``model_aliases.py`` | + ``AliasProfile`` (frozen dataclass), + ``list_profiles``, + ``resolve_profile`` (alias name OR reverse hf_path lookup, O(1) via cached reverse index). ``resolve_model``/``list_aliases`` keep old signatures. |
| ``model_auto_config.detect_model_config`` | now consults the alias profile first; regex stays as the fallback for un-aliased HF paths. |
| ``doctor/discovery.load_aliases`` | thin wrapper over ``list_aliases`` so the harness sees a consistent string view regardless of schema flavor. |

## P1 — orphan coverage

The 6 aliases that fell through to ``ModelConfig()`` defaults now have explicit profiles:

| alias | tool_call_parser | reasoning_parser | is_hybrid | supports_spec_decode |
|---|---|---|---|---|
| bonsai-1.7b/4b/8b | ``null`` | ``null`` | false | true |
| ministral-3b | ``hermes`` | ``null`` | false | true |
| nemotron-30b/nano | ``hermes`` | ``null`` | **true** (Mamba hybrid) | **false** |

Note: nemotron's hybrid flag was previously wrong (defaulted to ``false`` because no pattern matched), so spec decoding would have produced corrupted output. P1 fixes that.

## Test plan

- [x] **23 new contract tests** in ``tests/test_model_profiles_ssot.py``:
  - schema validity (every entry is rich form, every alias has the bool gates)
  - no-orphan invariant (51/51 aliases now resolve)
  - alias-vs-regex preference (alias profile wins over the matching regex; verified via monkeypatch)
  - reverse hf_path lookup (full HF path also resolves to its alias profile, including shared-path determinism for nemotron-30b/-nano and deepseek-v4-flash pairs)
  - regex fallback still works for un-aliased HF paths
  - legacy bare-string form loads with default capability flags
  - empty hf_path rejection (both schemas)
  - family flag consistency (``qwen3.5-*`` must all agree on hybrid/spec_decode)
- [x] **Full unit suite: 2255 passed, 17 skipped, 7 deselected** (excluded 4 heavy MLLM/video files that need real Qwen3-VL weights — not relevant to this refactor; covered by GitHub matrix CI)
- [x] ``ruff check`` + ``ruff format --check`` clean
- [x] **GitHub matrix CI: all green** (lint, type-check, version bump check, test-matrix 3.10/3.11/3.12, test-apple-silicon, tests)
- [x] ``make check --model qwen3.5-4b``: **0 regression vs baseline** (4 pass, 0 regression, 2 pre-existing fail, 1 skip — Test 10 streaming-usage and thinking-toggle are documented as known pre-existing issues, see harness baseline notes)
- [ ] ``pr_validate`` end-to-end — local run hangs on MLLM tests requiring real Qwen3-VL weights; **superseded by the GitHub Actions matrix above**, which runs the same suite with the right environment

## Out of scope (follow-ups)

- **PR #281** (``suffix_decoding_tier``) — will rebase on top of this; the new field lands in ``AliasProfile`` rather than ``ModelConfig``, giving per-alias granularity.
- **Cross-model bench sweep** populating the new tier field per alias.
- **Eventually retiring ``_MODEL_PATTERNS``** once every served model is an alias (still useful today as a fallback for users who serve a brand-new HF model directly).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.6.20"
+version = "0.6.21"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_model_profiles_ssot.py
+++ b/tests/test_model_profiles_ssot.py
@@ -1,0 +1,255 @@
+# SPDX-License-Identifier: Apache-2.0
+"""SSOT contract tests for the per-alias model profile registry.
+
+Pre-PR architecture had two sources: ``aliases.json`` (51 alias→hf_path
+mappings) and ``_MODEL_PATTERNS`` (25 regex-keyed ``ModelConfig`` rows).
+A new alias would silently inherit whichever pattern's regex happened to
+match its HF path, with no per-alias granularity for capability flags.
+
+These tests pin down the new contract:
+
+- every alias has an explicit profile in ``aliases.json``
+- ``detect_model_config`` prefers the alias profile over the regex
+  fallback
+- the regex fallback still works for unaliased HF paths (so users
+  serving a brand-new model from HuggingFace still get sensible
+  parser/capability inference)
+- the legacy bare-string form is still accepted at load time, with
+  default capability flags
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from unittest.mock import patch
+
+from vllm_mlx.model_aliases import (
+    AliasProfile,
+    list_aliases,
+    list_profiles,
+    resolve_model,
+    resolve_profile,
+)
+from vllm_mlx.model_auto_config import detect_model_config
+
+ALIASES_PATH = Path(__file__).parent.parent / "vllm_mlx" / "aliases.json"
+
+
+# ---- Schema sanity --------------------------------------------------------
+
+
+def test_aliases_json_is_rich_schema() -> None:
+    """Every entry must be the new dict form (string form was the old
+    schema; if it appears in ``aliases.json`` going forward the migration
+    has regressed)."""
+    with open(ALIASES_PATH) as f:
+        raw = json.load(f)
+    assert raw, "aliases.json must not be empty"
+    for alias, value in raw.items():
+        assert isinstance(value, dict), f"{alias!r}: rich schema required"
+        assert "hf_path" in value, f"{alias!r}: missing hf_path"
+        assert isinstance(value["hf_path"], str)
+
+
+def test_every_alias_has_explicit_profile_fields() -> None:
+    """No alias should rely on default capability flags — the whole
+    point of the SSOT is to be explicit. ``tool_call_parser`` and
+    ``reasoning_parser`` are allowed to be null (some families have no
+    native tool format), but the bool gates must be present."""
+    with open(ALIASES_PATH) as f:
+        raw = json.load(f)
+    for alias, value in raw.items():
+        assert "is_hybrid" in value, f"{alias!r}: missing is_hybrid"
+        assert "supports_spec_decode" in value, (
+            f"{alias!r}: missing supports_spec_decode"
+        )
+        assert isinstance(value["is_hybrid"], bool)
+        assert isinstance(value["supports_spec_decode"], bool)
+
+
+def test_no_orphan_aliases() -> None:
+    """The pre-SSOT audit found 6 orphans with no profile (bonsai×3,
+    ministral, nemotron×2). After P1 every alias must resolve."""
+    aliases = list_aliases()
+    for alias in aliases:
+        cfg = detect_model_config(alias)
+        assert cfg is not None, f"orphan alias: {alias}"
+
+
+def test_orphan_aliases_now_covered() -> None:
+    """Pin the 6 specific aliases that were orphans before this PR to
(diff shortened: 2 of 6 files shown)


raullenchai:
## Local validation summary

**Verdict: SAFE TO REVIEW**

| Step | Status | Detail |
|---|---|---|
| ``ruff check`` | PASS | clean (4 files) |
| ``ruff format --check`` | PASS | clean |
| 17 new SSOT contract tests | PASS | ``tests/test_model_profiles_ssot.py`` |
| Targeted test files (4 changed-file deps) | PASS | 116 passed in 4.14s |
| Full unit suite (excluding 4 mllm files needing real Qwen3-VL weights) | PASS | **2249 passed, 17 skipped, 0 failures** in 27.65s |

**Why the mllm exclusion**: ``test_mllm.py`` / ``test_mllm_cache.py`` / ``test_mllm_continuous_batching.py`` / ``test_video.py`` load real multimodal weights and currently hang on this machine (fresh-loaded VLM weights, ~5GB). Excluded for local-iteration speed; CI ``test-apple-silicon`` and the GitHub matrix runs will cover them with the standard fixture cache.

**Why bumping to 0.6.21** (commit ``e8a359b``):
- nemotron's ``is_hybrid`` flag was incorrectly defaulting to ``false`` (no regex matched it), which would have produced corrupted output if anyone enabled spec_decode on Mamba blocks. P1 fixes this.
- 6 orphan aliases (bonsai×3, ministral, nemotron×2) now resolve to explicit profiles instead of falling through to ``ModelConfig()`` defaults.
- ``AliasProfile`` is a new public dataclass exposed from ``vllm_mlx.model_aliases``.

These are user-observable changes, so a version bump is the right call rather than ``skip-version-bump``.

**What's NOT in this PR (deferred):**
- pr_validate ``stress_e2e_bench`` step (3-model matrix, ~9 min) — skipped to keep iteration fast; the change is config-data + a JSON-loader path, no scheduler/engine code touched.
- ``make check --model qwen3.5-4b`` — alias profile resolves to the same fields the regex was returning for qwen3.5, so no behavior change expected.

raullenchai:
## Codex review — converged after 3 rounds

Pushed ``3ff42cd`` with all fixes from self-review.

### Round 1 — perf & clarity (fixed in commit)
| ID | Severity | Fix |
|---|---|---|
| F1 | medium | reverse hf_path lookup was O(n) per miss → built one-time reverse index, now O(1) |
| F10 | medium | dropped misleading "circular dependency" comment in ``detect_model_config``; verified no cycle exists, moved import to top |
| F8 (docstring) | low | updated ``AliasProfile`` docstring to drop the same wrong cycle claim |

### Round 2 — test coverage (fixed in commit)
| ID | Severity | Fix |
|---|---|---|
| F11/F15/F16 | medium | pinned reverse-lookup canonical-alias behavior for nemotron-30b/-nano + deepseek-v4-flash/-8bit shared hf_paths |
| (new) | medium | proved alias-first lookup wins over regex via monkeypatch (was assumed-true, now tested) |
| F13 | low | ``pytest.raises`` instead of ``try/except`` |
| (new) | low | reset both ``_aliases`` and ``_hf_to_alias`` when test-patching |

### Round 3 — input validation (fixed in commit)
| ID | Severity | Fix |
|---|---|---|
| (new) | medium | reject empty ``hf_path`` strings in both legacy and rich schemas at load time |

### Considered but deferred (out of scope)
- **F20**: ``tests/integrations/test_hermes.py`` and ``scripts/agent_test_gen.py`` assert "entries between 15 and 30" — pre-existing bug (we have 51), separate cleanup.
- **Naming overlap**: ``vllm_mlx.agents.list_profiles()`` returns agent profiles, ``vllm_mlx.model_aliases.list_profiles()`` returns model profiles. No collision today (different modules); rename if it bites.

### Final status
- **108 alias-related unit tests pass** (was 85 on main, 105 after Round 1, 108 after Round 3 hardening)
- **Full unit suite: 2255 passed, 17 skipped, 0 failures**
- **ruff check + ruff format**: clean

PR #305, opened 2026-05-08
feat(cli): auto-apply suffix_decoding_tier in serve

raullenchai:
## Summary

Closes the loop on the per-alias SSOT: ``rapid-mlx serve <alias>`` now flips ``--suffix-decoding`` on automatically when the alias profile's tier is ``agent`` or ``structured`` and the framework gate would let the flag take effect.

Today users see "recommended" in ``info`` output but have to remember to add the flag — the information was there, just unactionable. Closes part 3 of the per-model profile audit gap list.

## Behavior

| condition | action |
|---|---|
| tier ``agent`` or ``structured`` + non-hybrid + user untouched | auto-enable, INFO log |
| ``--no-suffix-decoding`` flag (new) | opt out on any tier |
| ``--suffix-decoding`` already passed | no change for any tier |
| ``supports_spec_decode=False`` (hybrid arch) | no change even at agent tier |
| tier ``neutral`` / ``avoid`` / ``unknown`` | no change |

The hybrid skip is belt-and-suspenders: the framework gate (mlx-lm BatchGenerator) refuses spec-decode on hybrid arches, so the flag would be a no-op at best. Trust the gate.

## Refactor

Extracted the auto-detect block (parser + reasoning + suffix tier) from ``serve_command`` into ``_apply_alias_profile_to_args(args, auto_config, logger)``. Lets us pin the contract with hermetic unit tests rather than mocking out uvicorn / model loading.

## Test plan

- [x] 14 new tests in \`tests/test_serve_auto_suffix_decoding.py\`:
  - agent / structured → auto-enable
  - neutral / avoid / unknown → no auto-enable
  - hybrid arch blocks auto-enable even at agent tier
  - \`--no-suffix-decoding\` opt-out works
  - \`--suffix-decoding\` already-set is preserved across all tiers
  - \`None\` config (unknown alias) is a no-op
  - parser auto-detect regression (tool_call + reasoning)
  - \`--tool-call-parser\` user override respected
  - \`--no-thinking\` suppresses reasoning_parser
  - argparse smoke for the new flag (\`store_true\`, hyphen→underscore)
- [x] Lint + format clean
- [x] Existing tests for parser auto-detect still pass (refactor didn't change that contract)
- [ ] Codex review (multi-round) until convergence
- [ ] pr_validate (per CLAUDE.md SOP)
- [ ] CI green

## Out of scope

- Per-tier custom auto-tuning of \`--suffix-max-draft\` / \`--suffix-min-confidence\` based on the bench's measured peak workload. Today the auto-enable uses the framework defaults; refining is a follow-up after batch-2/3 tier benches land (#181) so we have data to tune on.
- ``--no-thinking`` ↔ ``--reasoning-parser`` interaction — left unchanged here; existing behavior preserved.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.6.25"
+version = "0.6.26"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_serve_auto_suffix_decoding.py
+++ b/tests/test_serve_auto_suffix_decoding.py
@@ -1,0 +1,204 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for `_apply_alias_profile_to_args` — the helper that bridges
+``AliasProfile`` (and the regex fallback ``ModelConfig``) into the serve
+command's argparse Namespace.
+
+Pins the contract:
+- tier ``agent`` or ``structured`` + non-hybrid + user didn't override
+  → ``args.suffix_decoding`` becomes True
+- tier ``neutral`` / ``avoid`` / ``unknown`` → no change
+- ``supports_spec_decode=False`` (hybrid) → no change even if tier=agent
+- ``--no-suffix-decoding`` opt-out → no change even on agent tier
+- ``--suffix-decoding`` already set → no change (already enabled)
+- ``auto_config=None`` (unknown alias) → no-op, no crash
+
+Plus the existing parser auto-detect contract (tool_call + reasoning).
+"""
+
+from __future__ import annotations
+
+import logging
+import sys
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import pytest
+
+from vllm_mlx import cli
+from vllm_mlx.model_auto_config import ModelConfig
+
+
+def _args(**overrides) -> SimpleNamespace:
+    """Minimal Namespace covering only the fields the helper reads."""
+    base = SimpleNamespace(
+        tool_call_parser=None,
+        reasoning_parser=None,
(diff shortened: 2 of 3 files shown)


raullenchai:
Closing — strategic direction changed.

After the v2-vs-v3 cross-validation sweep landed (#323) and a deeper diagnostic of actual SuffixDecoding behavior in production-shaped workloads, we've decided to keep SuffixDecoding strictly opt-in via the existing ``--suffix-decoding`` CLI flag. ``suffix_decoding_tier`` in ``aliases.json`` continues to surface as a hint in ``rapid-mlx info`` (already shipped) but does NOT auto-flip ``--suffix-decoding`` on, regardless of tier.

Rationale:

1. **Tier is a model property, not a user-workload property.** A ``tier=agent`` model used as a chat assistant still regresses; auto-on by tier mixes those two and makes the wrong call for non-agent traffic.
2. **Diagnostic data**: bonsai-1.7b on a mixed chat+JSON+code workload showed 15% acceptance rate vs the ~50% break-even needed for verify-forward to pay off. The 1.0x → 0.85x avoid threshold is too close to the noise floor — neutral aliases routinely flip to avoid across runs (v2→v3 confirmed this).
3. **Auto-on amplifies the failure mode**: if the algorithm gives modest wins on its target workload but regressions everywhere else, defaulting it on is net-negative. Keeping it explicit puts the choice with the user who actually knows their traffic.

The refactor in this PR (``_apply_alias_profile_to_args`` extraction, parser auto-detect contract pin) is genuinely useful and would survive on its own. If we revisit this later we can cherry-pick the refactor without the auto-suffix-decoding policy.

Future direction (deferred, not blocking): runtime adaptive disable based on actual ``_suffix_stats`` acceptance rate ("越用越快" feel). That's safer than tier-based auto-on because the algorithm self-measures, but it adds runtime state complexity we're not ready to take on. Tracked informally; not on the active queue.

Closing without merge. Thank you for the work — the analysis it forced us into directly produced the right product decision.

PR #308, opened 2026-05-08
fix(bench): force greedy sampling in suffix-decoding integrated bench

raullenchai:
## Root cause

\`scripts/bench_suffix_decoding_integrated.py\` was sending chat-completions requests with **no \`temperature\` field**. The server's \`_resolve_temperature(None)\` falls back to \`_FALLBACK_TEMPERATURE = 0.7\` (\`vllm_mlx/service/helpers.py:37\`) — non-greedy.

\`_install_suffix_decoding._is_greedy_for_uid\` in \`vllm_mlx/scheduler.py\` falls through to \`_orig_step\` on every call when temperature > 0. With the bench at 0.7 the verify path **never fires** — every measurement reduces to vanilla-vs-vanilla.

That's why every alias in the tier sweep came back \`neutral\`/\`avoid\` (1.0x flat across the board) while the standalone POC bench (\`scripts/bench_suffix_decoding.py\`, which calls \`model.forward\` directly with explicit \`argmax\`) measured **3–7× speedups** on the same workloads (\`evals/results/SUFFIX_POC_REPORT.md\`).

## Receipt — same workloads, same model

| workload | POC bench | integrated bench (pre-fix) |
|---|---:|---:|
| chat | 5.00× | 0.99× |
| code_edit | 5.58× | 1.00× |
| tool_loop | 6.35× | 1.00× |
| json_array | 4.61× | 1.00× |
| summarize | 7.52× | 1.00× |

Algorithm fine, bench was bypassing it.

## Fix

Add \`"temperature": 0.0\` to the payload before \`**workload_body\` spread (so workload dicts can still override for tests but every production sweep stays greedy).

## Tests

\`tests/test_suffix_bench_methodology.py::TestPayloadIsGreedy\` pins:

1. The literal \`"temperature": 0.0\` must appear in the bench source — a refactor that drops it triggers this test instead of silently shipping broken data.
2. No workload body may pre-set \`temperature\` (the spread order would clobber the greedy default if it did).

## Follow-up (separate PR)

Re-bench every alias whose tier was populated by the buggy sweep. The existing \`suffix_decoding_tier\` + \`suffix_bench_speedup\` fields in \`aliases.json\` are not just imprecise — they're vacuous (vanilla-vs-vanilla noise) and need to be replaced wholesale before PR #305's auto-enable mechanism can rest on real data.

## Test plan

- [x] Targeted: \`pytest tests/test_suffix_bench_methodology.py -q\` (12/12 pass, 2 new)
- [x] \`ruff check && ruff format --check\` clean
- [ ] Manual: re-run qwen3-coder-30b bench post-merge, confirm tool_loop ≥ 1.8× (closes the loop on whether agent-tier exists at all)

## Sweep results applied (commit 61200ea)

After the methodology fix, ran `scripts/run_suffix_sweep.sh` across all 34 dense aliases. Results applied to `vllm_mlx/aliases.json` wholesale (every pre-fix `suffix_bench_speedup` was vanilla-vs-vanilla noise from the bug).

| Outcome | Count | Aliases |
|---|---:|---|
| ✅ tier=`avoid` | 13 | bonsai-{1.7b,4b,8b}, deepseek-r1-8b, gemma-4-{26b,31b}, hermes3-8b, hermes4-70b, llama3-3b, ministral-3b, phi4-14b, qwen3-coder-30b, smollm3-3b |
| ✅ tier=`neutral` | 4 | deepseek-r1-32b, devstral-v2-24b, qwen3-vl-8b, qwen3-vl-30b |
| ⚠ tier=`unknown` (re-bench) | 1 | gemma3-27b (most workloads rejected by reliability gates) |
| ❌ FAIL — bad alias | 1 | kimi-48b (HF 404 on `mlx-community/Kimi-K2-Instruct-Q4_0-MLX`) |
| ❌ FAIL — disk pre-flight | 4 | deepseek-v4-flash{,-2bit,-4bit,-8bit} (need 87 GB free) |
| ❌ FAIL — server startup | 3 | glm4.5-air, glm4.7-9b, kimi-k2.5, minimax-m2.5 |
| ❌ FAIL — multimodal misclass | 3 | gemma3-1b, gemma-3n-e4b, qwen3-vl-4b |
| ⏭ SKIP (pre-known bad) | 3 | gpt-oss-20b (HF 401), gemma3-12b, mistral-24b (no chat template) |

**Net coverage**: 17 dense aliases now tiered (was 6 with stale data); 17 still untiered (need separate fixes — alias data, code-path bugs, or larger disk).

## Important caveat — see #320 for the bench-stability discussion

Two aliases produced **different tier classifications across consecutive post-fix sweeps**:

| Alias | Run A | Run B | Δ |
|---|---|---|---|
| qwen3-coder-30b | chat 1.31x · json 2.18x → **structured** | chat 0.87x · json 0.99x · tool 1.04x · code 0.93x → **avoid** | 2 steps |
| devstral-v2-24b | chat 0.72x · code 0.73x → **avoid** | chat 0.99x · json 1.01x · code 1.03x → **neutral** | 1 step |

These tier values are the best we have from a single sweep, but they should **not** be used as a serve-time auto-apply default until issue #320 lands the stability gate (#320 S4: per-workload CV < 15% required before classification commits). PR #305 (auto-apply `suffix_decoding_tier` in serve) should be held until #320 ships the gate.

### DIFF
--- a/scripts/bench_suffix_decoding_integrated.py
+++ b/scripts/bench_suffix_decoding_integrated.py
@@ -357,15 +357,16 @@ def run_workload(
     Returning a structured value (instead of just a float) lets the
     caller persist raw data for forensic debugging without re-running.
     """
-    # Greedy sampling is mandatory: the SuffixDecoding install path in
-    # ``vllm_mlx/scheduler.py:_install_suffix_decoding`` falls through to
-    # vanilla on every step when ``temperature > 0`` (or top_p < 1, top_k > 0).
-    # Without an explicit ``temperature: 0.0`` here the server resolves to its
-    # 0.7 fallback, the patch never fires, and every measurement reduces to
-    # vanilla-vs-vanilla — the bug that produced the 1.0x flat tier data
-    # before this fix. Workload bodies still take precedence (tests can
-    # override) but the production sweep must stay greedy. See
-    # tests/test_suffix_bench_methodology.py::test_payload_is_greedy.
+    # Greedy sampling is mandatory. ``_install_suffix_decoding`` in
+    # ``vllm_mlx/scheduler.py`` only fires when ``_is_greedy_for_uid``
+    # returns True, which checks **temperature == 0** specifically (mlx-lm's
+    # ``make_sampler`` short-circuits to argmax at temp=0, so top_p/top_k
+    # don't matter in that regime). Without ``"temperature": 0.0`` here the
+    # server resolves to its 0.7 fallback, the patch never fires, and every
+    # measurement reduces to vanilla-vs-vanilla — the bug that produced the
+    # 1.0x flat tier data before this fix. Workload bodies still take
+    # precedence (tests can override) but the production sweep must stay
+    # greedy. See tests/test_suffix_bench_methodology.py::TestPayloadIsGreedy.
     payload = {
         "model": handle.model,
         "max_tokens": max_tokens,
--- a/scripts/bench_suffix_decoding_integrated.py
+++ b/scripts/bench_suffix_decoding_integrated.py
@@ -357,11 +357,21 @@ def run_workload(
     Returning a structured value (instead of just a float) lets the
     caller persist raw data for forensic debugging without re-running.
     """
+    # Greedy sampling is mandatory: the SuffixDecoding install path in
+    # ``vllm_mlx/scheduler.py:_install_suffix_decoding`` falls through to
+    # vanilla on every step when ``temperature > 0`` (or top_p < 1, top_k > 0).
+    # Without an explicit ``temperature: 0.0`` here the server resolves to its
+    # 0.7 fallback, the patch never fires, and every measurement reduces to
+    # vanilla-vs-vanilla — the bug that produced the 1.0x flat tier data
+    # before this fix. Workload bodies still take precedence (tests can
+    # override) but the production sweep must stay greedy. See
+    # tests/test_suffix_bench_methodology.py::test_payload_is_greedy.
     payload = {
         "model": handle.model,
         "max_tokens": max_tokens,
         "stream": True,
         "stream_options": {"include_usage": True},
+        "temperature": 0.0,
         **workload_body,
     }
     t0 = time.perf_counter()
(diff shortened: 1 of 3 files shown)


raullenchai:
## Codex round 3 (DeepSeek V4 / chat) — verdict: convergent, zero new P0

After round 2 landed (75be458), ran one more adversarial review against the full PR diff (bench code + tests + aliases.json data update). Findings + triage:

| # | Finding | Verdict |
|---|---|---|
| 1 | dict-subclass spread paranoia | Dismissable — `dict` subclasses don't override `**` in any standard way |
| 2-5 | `_FakeStream` monkeypatch wouldn't capture `json=` | **Wrong** — `httpx.stream` is a module-level callable; replacing it with a class makes Python invoke `_FakeStream(*args, **kwargs)`, capturing `kwargs["json"]` correctly. `test_hostile_workload_cannot_clobber_greedy` would fail if the production spread order regressed. Verified: 3/3 tests pass against current head |
| 6 | `gemma-4-31b` `tier=avoid` despite `chat 1.468x` | **Real, not blocking** — the per-workload data shows `json_array 0.67x`, which is what dropped the alias to `avoid`. Classifier intent is "conservative cross-workload default"; users who know they only run chat can still consult `suffix_bench_speedup`. Worth a follow-up doc/UX item, not blocking this PR |
| 7 | `deepseek-r1-32b` marginal speedups → `neutral` | Correctly classified per existing thresholds |
| 5 | Test code duplication (`_FakeStream` defined twice) | P2 nit — extracting to a fixture is a refactor for another PR |

Per the SOP definition of convergence ("a round produces zero new P0 findings"), this is round 3 convergent. Combined with rounds 1-2 (`be5edf7`, `75be458`) the PR has had three iterations of adversarial review with the last one finding no actionable issue.

Ready to merge with `skip-version-bump` label (aliases.json change is data-only — `suffix_decoding_tier` is inert until #305 ships, and #305 is held for the stability gate per #320 / #305 hold-comment).

PR #349, opened 2026-05-10
feat(cli): wire staleness banner into serve+chat, add 'rapid-mlx upgrade'

raullenchai:
## Summary

- Wire the existing staleness check into `serve` and `chat` startup banners (was only firing on `rapid-mlx models`).
- Add `rapid-mlx upgrade` subcommand that detects install method (brew / install.sh venv / pip) and runs the right upgrade command after a confirmation prompt.
- Fix the staleness banner to recommend `rapid-mlx upgrade` instead of the incorrect `brew upgrade rapid-mlx` (formula is in `raullenchai/tap`, not core).
- Bump two unrelated hard-coded alias counts in `test_model_profiles_ssot` (51→57, 6→7) that broke when #348 added Tier-1 aliases — pre-existing failures blocking every PR's CI.

## Why

Real failure mode that just hit a user: installed `rapid-mlx 0.6.20`, latest is `0.6.30` (10 patches behind), tried `rapid-mlx chat` and got an "invalid choice" error because `chat` was added in 0.6.27. No banner ever fired because the staleness check was only wired into `models`. Now any common entrypoint (`serve` / `chat`) shows it.

### Startup banner
Before:
```
✓ Ready.
```

After (when ≥2 patch versions behind, same minor):
```
✓ Ready.
⚠ rapid-mlx 0.6.20 is behind latest 0.6.30 — run `rapid-mlx upgrade` to pick up new model aliases / flags.
```

Silent on TTY-less stdout, `CI=1`, `RAPID_MLX_DISABLE_VERSION_CHECK=1`, dev builds, single-patch lag, cross-minor lag, dev-ahead-of-latest, and offline.

### `rapid-mlx upgrade`
```
$ rapid-mlx upgrade
  Current:  rapid-mlx 0.6.20
  Latest:   rapid-mlx 0.6.30
  Install:  install_sh (/Users/raullen/.local/bin/rapid-mlx)
  Command:  curl -fsSL https://raullenchai.github.io/Rapid-MLX/install.sh | bash
  Run now? [y/N]
```

`-y` skips the prompt for scripted use. Detection covers:

| Install method | Detection signal | Upgrade command |
|---|---|---|
| brew | binary realpath under `/Cellar/rapid-mlx` or `/opt/homebrew/` | `brew upgrade raullenchai/tap/rapid-mlx` |
| install.sh | binary in `~/.local/bin` OR realpath under `~/.rapid-mlx` (venv) | `curl -fsSL .../install.sh \| bash` |
| pip (default) | anything else | `{sys.executable} -m pip install --upgrade rapid-mlx` |

The install.sh symlink-through-venv detection is one of the gotchas — install.sh creates `~/.rapid-mlx/` (a venv) and symlinks `~/.local/bin/rapid-mlx` to `~/.rapid-mlx/bin/vllm-mlx`. A naive `realpath`-only check resolves through the symlink and classifies these users as pip (wrong upgrade command). Pinned with a regression test.

## Test plan

- [x] `pytest tests/test_version_check.py -q` — **25 passed** (10 new tests for upgrade subcommand + install-method detection)
- [x] `pytest tests/ --ignore=tests/integrations --ignore=tests/test_event_loop.py --ignore=tests/test_mllm*.py --ignore=tests/test_video.py -q` — **2459 passed** (3 pre-existing fails in `test_model_profiles_ssot` fixed in this PR)
- [x] `ruff check vllm_mlx/ tests/` — clean
- [x] `ruff format --check` on changed files — clean
- [x] Smoke: `python3.12 -m vllm_mlx.cli upgrade --help` — parses
- [x] Smoke: `upgrade_command` with mocked PyPI — correctly prints "Already up to date" / "Run now? [y/N]" / detects install method
- [x] Smoke: real install.sh deployment correctly classified as `install_sh` (was previously misclassified as `pip` by the realpath-only logic — now pinned by test)

## Out of scope

- Auto-running the upgrade without prompt (the `-y` flag is the user's opt-in).
- Adding the banner to `bench` / `models` / `info` / etc. — these are short-lived commands where startup latency matters more; the staleness check on `models` is already wired and that's enough for a "I want to see what changed" flow.
- Refactoring the hard-coded alias counts in `test_model_profiles_ssot` to structural assertions — separate cleanup.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.6.30"
+version = "0.6.31"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_model_profiles_ssot.py
+++ b/tests/test_model_profiles_ssot.py
@@ -100,14 +100,14 @@ def test_orphan_aliases_now_covered() -> None:
 def test_list_aliases_returns_legacy_string_view() -> None:
     """Old callers (doctor harness, tests) expect ``{alias: hf_path}``."""
     aliases = list_aliases()
-    assert len(aliases) == 51
+    assert len(aliases) == 57
     assert all(isinstance(p, str) for p in aliases.values())
     assert aliases["qwen3.5-4b"] == "mlx-community/Qwen3.5-4B-MLX-4bit"
 
 
 def test_list_profiles_returns_rich_dataclass_view() -> None:
     profiles = list_profiles()
-    assert len(profiles) == 51
+    assert len(profiles) == 57
     p = profiles["qwen3.5-4b"]
     assert isinstance(p, AliasProfile)
     assert p.hf_path == "mlx-community/Qwen3.5-4B-MLX-4bit"
@@ -298,7 +298,7 @@ def test_qwen35_family_aliases_share_hybrid_flag() -> None:
     schema enables."""
     profiles = list_profiles()
     family = {a: p for a, p in profiles.items() if a.startswith("qwen3.5-")}
-    assert len(family) == 6
+    assert len(family) == 7
     flags = {(p.is_hybrid, p.supports_spec_decode) for p in family.values()}
     assert flags == {(True, False)}, (
         f"qwen3.5-* family disagrees on capability flags: {flags}"
--- a/tests/test_version_check.py
+++ b/tests/test_version_check.py
@@ -79,7 +79,7 @@ def test_warns_when_2_or_more_patch_behind(isolated_cache, monkeypatch):
     assert msg is not None
     assert "0.6.14" in msg
     assert "0.6.16" in msg
-    assert "brew upgrade" in msg
+    assert "rapid-mlx upgrade" in msg
 
 
 def test_silent_when_only_1_patch_behind(isolated_cache, monkeypatch):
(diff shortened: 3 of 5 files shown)


raullenchai:
## DeepSeek round-1 review (18 findings, 4 P0 / 5 P1 / 9 P2)

Cross-checked each against actual code; addressed the valid ones in `96967a4`.

### Applied (4 fixes)

- **P0 #1 — `shell=True` argv-injection via `sys.executable`** ✅ fixed. Switched `subprocess.run` to argv form (`shell=False`); install.sh's pipe wrapped explicitly as `["bash", "-c", <pipe>]`. New regression test asserts `upgrade_argv` is a list and pip uses `[sys.executable, "-m", "pip", ...]`.
- **P2 #15 — help text said "latest PyPI version"** ✅ fixed. Now `Upgrade rapid-mlx to the latest version (brew / pip / install.sh)`.
- **P2 #18 — Linux brew at `/home/linuxbrew/`** ✅ fixed. Added to `brew_markers` tuple + `test_detect_install_method_brew_linux` regression test.
- **P2 #11 — `sys_executable()` helper** ✅ inlined.

### Not applied (with reason)

- **P0 #4 — "Banner fires twice on chat"** — *false positive*. `_spawn_chat_server` opens the spawned `serve` subprocess with `stdout=log, stderr=subprocess.STDOUT`, so the spawned banner goes to the log file (`/tmp/rapid-mlx-chat-*.log`), not the user's terminal. Only the chat REPL's own `print_staleness_warning_if_any()` is visible.
- **P0 #3 — "`CI=1` check is insufficient"** — *also false positive*. `_disabled()` does `os.environ.get("CI")`, which is truthy for any non-empty value: `"true"`, `"1"`, `"yes"`. GitHub Actions, GitLab, CircleCI, Travis all set `CI=true`. Already covered.
- **P1 #5 — Race between `force_refresh=True` PyPI check and upgrade** — accepted. The user re-confirms at the prompt, and the install commands themselves resolve the actual latest at upgrade time.
- **P1 #6 — install.sh re-run is fragile** — accepted. install.sh is idempotent by design (overwrites the venv); re-running IS the upgrade path.
- **P1 #7 — stdout vs stderr** — partial. Banner stays on stderr (existing `print_staleness_warning_if_any` behavior). Upgrade subcommand output stays on stdout because it IS the primary output of that command.
- **P1 #8/#9** — author flagged these as "no change needed" themselves.
- **P2 #10 — split the alias-count bumps into separate commit** — overkill for a 2-line CI unblock. Already explained in commit message + PR description.
- **P2 #12 — `unknown` method dead code** — kept as defensive branch (one extra `if` isn't worth removing for a future-proofing escape hatch).
- **P2 #13 — `force_refresh=True` bypasses cache in upgrade** — intentional. User explicitly invoked `upgrade`; show them the freshest available version.
- **P2 #14 — duplicate `print_staleness_warning_if_any` import** — kept inline at the two call sites; module loads in both serve and chat paths anyway, no additional cost.
- **P2 #16 — no test for `upgrade_command` itself** — out of scope; the function is a thin shell over `detect_install_method` + `subprocess.run`, both covered.
- **P2 #17 — `check=False` swallowing errors** — `subprocess.run` with `check=False` returns the failed exit code, which we propagate via `sys.exit(result.returncode)`. The user sees the underlying tool's error output natively.

## Round-2 status

- All 26 version-check tests pass (including 2 new for the argv-form + Linux brew)
- Full unit suite: **2460 passed**
- Lint + format clean
- Version bumped to **0.6.31** for the new subcommand surface

Waiting for CI on `96967a4`. If clean, requesting another deepseek pass to confirm convergence.

raullenchai:
## DeepSeek round-2 convergence check

> **CONVERGED — no new P0 findings**

Round-1 fixes verified:
- argv form (`shell=False`) — no shell injection vector ✅
- `["bash", "-c", <pipe>]` — explicit, no ambient `$SHELL` coupling ✅
- Linux brew `/home/linuxbrew/` — no false-positive on macOS paths ✅

Three P2 polish items flagged (non-blocking, not addressing in this PR):
- Pip cosmetic display string with spaces in `sys.executable` looks odd (argv form is already correct).
- Hardcoded `raullenchai.github.io` URL — not configurable.
- install.sh pipe printed to user has no `--yes` analogue (minor UX inconsistency).

## SOP §6 — pr_validate

Running locally (`python3.12 -m scripts.pr_validate.pr_validate 349 --verbose`):
- ✅ Lint: clean
- ✅ Lint format: clean
- ✅ Supply-chain: no new deps, no suspicious patterns, no install-hook tampering
- ✅ Targeted tests: 95 passed (test_cli_chat, test_cli_models, test_model_profiles_ssot, test_version_check)
- 🟡 Full unit + stress + e2e: still running

## SOP §8 — CI gate

✅ All 8 required checks SUCCESS on `96967a4`:
- lint, type-check, version-check, test-matrix (3.10/3.11/3.12), test-apple-silicon, tests
- mergeable: MERGEABLE / mergeStateStatus: CLEAN

## SOP §7 — `make check` skip rationale

Skipping per SOP — this PR is **surface-touching CLI only** (new subcommand + banner wiring), no inference path changes. No risk of generation regression vs `make check` baselines.

Ready to squash-merge once full pr_validate completes.

Issue #236, opened 2026-05-05
feat: opt-in anonymous usage telemetry to inform product direction


raullenchai:
## Motivation

Right now we have **zero visibility** into how Rapid-MLX is actually used in the wild. PyPI download counts and GitHub stars tell us *that* people install it, but not:

- Which models do users actually run? (Are we over-investing in 35B-class when 4B is the sweet spot?)
- How long are sessions? (Is the prefix cache disk-persistence feature being used, or are users restarting often?)
- Which subcommands are popular? (`serve` vs `agents` vs `bench` vs `chat` vs `doctor`)
- Which agent profiles see uptake? (Hermes? Aider? OpenWebUI?)
- Which OpenAI-compatible endpoints get the most traffic? (`/v1/chat/completions` vs `/v1/completions` vs `/v1/embeddings`)
- What hardware are users on? (M1 Air → M3 Ultra spectrum)
- Where do users hit errors? (Model load? Tool parsing? OOM?)

Without this data we're guessing at roadmap priorities. The recent v0.6.13 onboarding tests caught two issues we'd never have known about (dylib relink, missing integration tests in wheel). Telemetry would let us catch those classes of issue across the whole user base, not just our manual test runs.

This issue tracks the design and implementation. **Privacy is the deal-breaker** — we ship this opt-in, transparent, and minimal, or we don't ship it.

## Goals

1. Understand the **shape of real usage** so we can prioritize features and deprecations.
2. **Detect breakage early** — if v0.6.X causes a spike in load failures on M1 Air, we want to know in days, not months.
3. **Keep user trust** — anyone reading the source must conclude this is honest, not creepy.

## Non-goals

- We do **not** want to identify individuals.
- We do **not** want prompt content, model outputs, file contents, or any user-generated text.
- We do **not** want to fingerprint users across machines (one stable per-machine UUID is fine; cross-correlation is not).
- This is **not** a substitute for proper observability of your own deployment — it's aggregate population data for the maintainers.

## Privacy principles (non-negotiable)

1. **Opt-in only.** First run prompts the user. Default = OFF. No "we already started collecting".
2. **One environment variable kill switch:** `RAPID_MLX_TELEMETRY=0` or `--no-telemetry` always wins, regardless of stored consent.
3. **No PII, ever.** No prompts, no completions, no file paths, no model paths beyond their HF repo ID, no IP addresses (we'll use a Cloudflare Worker that strips IPs before forwarding).
[...]

PR #351, opened 2026-05-10
feat(telemetry): Phase 1 — opt-in consent + CLI surface (Issue #236)

raullenchai:
## Summary

Phase 1 of opt-in anonymous usage telemetry per Issue #236. Ships the **privacy mechanism** (consent file, kill switch, redaction primitives, CLI surface) **without any network code**. The collector / Cloudflare Worker / aggregation pipeline are explicitly Phase 2 — this PR exists so the consent contract and PII guardrails can be reviewed in isolation before any byte ever leaves the machine.

## What ships

- **`vllm_mlx/telemetry/`** — new package, no network deps
  - `state.py` — consent file + client_id + precedence: \`--no-telemetry\` > \`RAPID_MLX_TELEMETRY=0\` > stored consent > **default OFF**. Force-on env var intentionally NOT supported (no silent CI opt-in).
  - `consent.py` — first-run prompt, gated on tty + interactive subcommand + no env var + no prior answer. Defaults to N. Never raises (telemetry consent is never a reason for the CLI to crash).
  - `redact.py` — token / TTFT / TPS bucketing, model-path normalization (HF repo IDs pass through, local paths → \`<local>\`), flag-name extraction (values never returned), traceback fingerprinting (module:function:lineno only — never message text or absolute paths).
  - \`schema.py\` — v1 dataclasses + \`sample_preview_payload()\` for the \`preview\` subcommand. **No event sites populate these in Phase 1.**

- **\`rapid-mlx telemetry\`** — five actions: \`status\` (default) / \`enable\` / \`disable\` / \`preview\` / \`reset\`.

- **\`--no-telemetry\`** global flag.

- **README \"Telemetry\" section** — what we collect / what we never collect / how to disable.

## Out of scope (Phase 2)

- httpx transport, batched flush, Cloudflare Worker
- Any actual event-emit site (request / session / error)
- Backend selection (PostHog vs Plausible vs D1 — Issue #236 open Q1)
- Sampling rate, EU endpoint, public transparency report

## Test plan

- [x] 103 new tests, all pass:
  - **state** (18): default off; consent round-trip; env-kill-switch wins over stored consent; CLI flag wins over consent; env=1 does NOT silently opt-in; falsy env values (0/false/no/off/empty) all kill-switch; client_id idempotent; user-zeroed UUID preserved; reset removes both files; consent_source reports correct origin; corrupt consent file → re-prompt (don't crash); atomic write leaves no .tmp behind.
  - **redact** (40): token / TTFT / TPS / memory bucket boundaries; HF repo ID passthrough; local-path / Windows / file:// / spaces all redacted; bare alias passthrough; flag values NEVER returned (sk-secret / 8000 stripped); short flags; sorted unique; non-string argv handled; traceback determinism; message text never in fingerprint; only hex chars; different sites different fingerprints; platform info has no full kernel string / no exact memory bytes.
  - **consent** (17): skips on non-tty / env-var / cli-flag / already-answered / each non-interactive subcommand / no subcommand; y/n records correctly; empty defaults to N; EOF does NOT record (lets next run re-prompt); records correct prompted_version; unwritable home does not crash CLI.
  - **cli** (10): subprocess-level smoke for every action + global flag + env kill switch + \`--help\` lists all actions.

- [x] Full unit suite (\`tests/ --ignore=mllm,video\`) — **2575 pass, 17 skipped, 7 deselected, 1 xfailed** (no regressions; the 17/7/1 are pre-existing).

- [x] Lint + format clean (\`ruff check && ruff format --check\` on changed paths).

- [ ] CI gate green
- [ ] DeepSeek adversarial review (run after CI green)

## Privacy contract — read these before merging

1. **Default OFF** — no consent file = no data, no exceptions.
2. **No env-var force-on** — \`RAPID_MLX_TELEMETRY=1\` is ignored. Opting in requires explicit \`rapid-mlx telemetry enable\` (or interactive 'y' at the first-run prompt).
3. **\`RAPID_MLX_TELEMETRY=0\` always wins** — even with \`consent=true\` recorded. Documented escape hatch for scripts / CI / paranoid operators.
4. **Bucketed everything** — exact token counts / TTFT / RAM / OS patch numbers can re-identify a session when joined; bucketing collapses the join surface.
5. **No prompts, no completions, no file paths, no IPs, no env values** — all four redact tests for these are P0 if they break.
6. **Phase 1 has zero \`httpx\` or \`requests\` calls in \`telemetry/\`** — \`grep\` it.

## Version

0.6.32 → 0.6.33

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,14 +704,14 @@ Rapid-MLX **can** send anonymous usage data to help us prioritise the right mode
 - Subcommand names (`serve` / `chat` / `agents` / `bench` / `doctor`)
 - Model alias names (`qwen3.5-9b`) or canonical HF repo IDs (`mlx-community/...`) — local paths are redacted to `<local>`
 - Bucketed counts: prompt/completion tokens, TTFT, tokens/sec — never exact values
-- Error categories + a hash fingerprint of the failure site (module:function:lineno only — never the message text)
+- Error categories + a hash fingerprint of the failure site (exception class name + per-frame `file:function:lineno` only — never the message text or absolute paths)
 - OS, arch, Apple chip name, RAM (rounded to GB), Python major.minor
 
 ### What we never collect
 
 - Prompts, completions, tool-call arguments, file contents, or any user-generated text
 - Local file paths, working directory, or model paths beyond their HF repo ID
-- IPs (a Cloudflare Worker strips them before forwarding) or hostnames
+- IPs or hostnames (Phase 2 will route through a Cloudflare Worker that strips IPs before forwarding to the aggregator; Phase 1 ships no transport at all)
 - API keys, environment variable values, auth headers
 - Stack trace messages or argument values
 
--- a/README.md
+++ b/README.md
@@ -695,6 +695,53 @@ Result: PASS
 
 ---
 
+## Telemetry
+
+Rapid-MLX **can** send anonymous usage data to help us prioritise the right models and catch regressions. **It is off by default and never starts collecting without your explicit opt-in.**
+
+### What we collect (only if you opt in)
+
+- Subcommand names (`serve` / `chat` / `agents` / `bench` / `doctor`)
+- Model alias names (`qwen3.5-9b`) or canonical HF repo IDs (`mlx-community/...`) — local paths are redacted to `<local>`
+- Bucketed counts: prompt/completion tokens, TTFT, tokens/sec — never exact values
+- Error categories + a hash fingerprint of the failure site (module:function:lineno only — never the message text)
+- OS, arch, Apple chip name, RAM (rounded to GB), Python major.minor
+
+### What we never collect
+
+- Prompts, completions, tool-call arguments, file contents, or any user-generated text
+- Local file paths, working directory, or model paths beyond their HF repo ID
+- IPs (a Cloudflare Worker strips them before forwarding) or hostnames
+- API keys, environment variable values, auth headers
+- Stack trace messages or argument values
+
+### Manage it
+
+```bash
+rapid-mlx telemetry status     # show current state and why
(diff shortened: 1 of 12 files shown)


raullenchai:
## DeepSeek round-1 review applied

Direct DeepSeek call returned 25 findings (4 P0, 5 P1, 11 P2, 3 Phase 2). Triaged and committed in `b44434a` — full details in the commit message.

**Applied (P0 + load-bearing P1/P2):**
- `record_consent` cleans up stale `.tmp` from interrupted prior writes
- `fingerprint_traceback` drops `__module__` (was leaking package paths like `transformers.models.llama`)
- `normalize_model_path` adds `Path.is_absolute()` to catch Windows drive forms / UNC
- `_read_chip_brand` is `lru_cache`d (Phase 2 perf)
- `maybe_prompt_for_consent` narrows `except Exception` → `(OSError, EOFError, KeyboardInterrupt)` so programming errors surface
- `get_consent_state` returns None on `schema_version` mismatch (clean Phase 2 migration path)
- README qualified about Cloudflare Worker being Phase 2

**Tests added:** 4 new regression guards (107 total telemetry tests, all pass; full unit suite 2579 pass).

**False positives (verified, not changed):**
- P0 #1 "empty env var ambiguity" — intentional behavior (`RAPID_MLX_TELEMETRY=` empty in docker/CI should mean off, not ignore)
- P0 #3 "empty client_id file silently regenerated" — empty is corrupt state; all-zero UUID is the documented anonymize hatch
- P1 #8 "consent_source format asymmetry" — env-var case includes value for diagnostics; flag case has no value to show

**Skipped:** 10 pure nits (placeholder name styling, README copy nuances) not worth churning the diff.

PR #377, opened 2026-05-13
feat(daemon): make --daemon boot-persistent by default

samuelfaj:
## Summary

- `lightning-mlx serve --daemon` now installs a per-user autostart (launchd on macOS, systemd `--user` on Linux) so the supervisor restarts at login / system boot.
- Opt-out: `--daemon=non-persist` (also `=ephemeral`) keeps the original in-session detached behavior.
- `lightning-mlx kill` uninstalls the plist/unit before signaling so the OS service manager cannot resurrect the supervisor.

## Behavior

| Mode | Trigger | Survives reboot? |
|---|---|---|
| Persistent (default) | `--daemon` | Yes |
| Ephemeral | `--daemon=non-persist` | No |

- macOS: `~/Library/LaunchAgents/com.lightning-mlx.<id>.plist` with `RunAtLoad=true`, `KeepAlive=true`, loaded via `launchctl load -w`.
- Linux: `~/.config/systemd/user/lightning-mlx-<id>.service` with `Restart=always`, enabled via `systemctl --user enable --now`. Run `loginctl enable-linger $USER` once if you want services to keep running after logout.

## Implementation notes

- New module `vllm_mlx/persistence.py` owns plist + systemd unit lifecycle. Calls `launchctl` / `systemctl` via `subprocess.run` (list form, no shell).
- `start_daemon` writes the record, then either installs autostart (persistent path; launchd/systemd spawns the supervisor) or `Popen`s the supervisor directly (ephemeral path).
- Install failure unlinks the on-disk record and raises `DaemonError`; CLI prints a clean error and exits non-zero.
- `list_daemons` surfaces persistent records with a non-live supervisor pid as `pending` instead of marking them `stale`.

## Test plan

- [x] `tests/test_persistence.py` (7): label, install + uninstall on darwin and linux, idempotent uninstall when plist missing, unsupported-platform error.
- [x] `tests/test_daemon.py`: persistent vs non-persist start, install-failure cleanup, stop-uninstall ordering, list-pending for persistent dead-pid, CLI dispatch for `--daemon=non-persist`.
- [x] Targeted suite: `pytest tests/test_persistence.py tests/test_daemon.py` → 21/21 PASS.
- [x] Full suite (`pytest tests --ignore=tests/integration --ignore=tests/integrations`) → 2262 PASS. 14 pre-existing failures unrelated to this diff (`disable_prefix_cache`, `default_max_tokens`, missing `mlx` module).

## Reviewer notes

- Pid-reuse race window between reboot and supervisor relaunch is brief and `supervisor_pid=None` is written at install time to keep the `pending` state correct.
- `launchctl unload` is invoked before `launchctl load -w` so re-running install is idempotent.


samuelfaj:
Retargeting to samuelfaj/lightning-mlx where I have merge rights.

PR #379, opened 2026-05-13
fix(gemma): unblock text-only path for Gemma3 and Gemma3n MLLMs

raullenchai:
## Summary

Two independent latent bugs caused **every text-only request** to Gemma family MLLM models (`gemma3-27b`, `gemma-3n-e4b`) to return zero completion tokens. Surfaced during the SuffixDecoding tier sweep when both aliases returned `completion_tokens: 0`.

### Fix #1 — `BatchedEngine._apply_chat_template` (`vllm_mlx/engine/batched.py`)

`Gemma3nProcessor` and `Gemma3Processor` (as loaded by `mlx_vlm.load`) expose `apply_chat_template` as a method but ship `chat_template=None` at the processor layer — only the inner tokenizer carries the Jinja template. The previous applicator-selection logic picked the processor whenever the method existed, then the call raised:

```
ValueError: Cannot use apply_chat_template because this processor does not have a chat template.
```

Mid-templating, every request short-circuited with empty content. Fixed by also requiring `processor.chat_template` to be truthy before picking the processor; otherwise fall back to the inner tokenizer (which has the right template — it's the source the processor would have copied from).

### Fix #2 — `MLLMBatchGenerator._run_vision_encoding` (`vllm_mlx/mllm_batch_generator.py`)

`Gemma3ForConditionalGeneration.__call__` declares `pixel_values` as a **required positional kwarg** (unlike Gemma3n / Qwen3-VL / LLaVA, which mark it `Optional`). Omitting it for text-only requests raised:

```
TypeError: Model.__call__() missing 1 required positional argument: 'pixel_values'
```

The inner `get_input_embeddings` already handles `None` cleanly — the model's signature is just stricter than its implementation. Fixed by always forwarding `pixel_values` to the model call, even when `None`. Other vision families keep working unchanged because their signatures already accept `None`.

## Coverage

- **`tests/test_batched_engine_chat_template.py`** — 5 cases pinning processor-vs-tokenizer applicator selection across the `chat_template={None, "", "{{ messages }}"}` matrix and the MLLM/non-MLLM and `processor=None` branches.
- **`tests/test_mllm_batch_generator.py`** — 3 cases pinning that `pixel_values` is always forwarded (both `None` and tensor paths) and that `extra_kwargs` survive alongside it.

Regression-checked: with both fixes reverted, 4 of the new 8 tests fail; with fixes applied, all 8 pass.

## Test plan

- [x] `ruff check` + `ruff format --check` — clean
- [x] Affected tests: 8/8 pass
- [x] Full unit suite: 3319 passed, 19 skipped, 7 deselected, 1 xfailed (no regressions)
- [x] Manual smoke `gemma-3n-e4b` after fixes: returns real content (was zero tokens before)
- [x] Manual smoke `gemma3-27b` after fixes: returns "HELLO27" (was TypeError before)
- [ ] CI green on all matrix jobs (lint, type-check, test-matrix 3.10/3.11/3.12, test-apple-silicon, tests)

## Notes

- Version bumped 0.6.44 → 0.6.45 so the fix ships immediately. Merge must use `--subject "chore: bump version to 0.6.45"` to trigger auto-release (default `(#NN)` suffix breaks the regex).
- Re-benching gemma SuffixDecoding tiers is deferred to a separate PR — these fixes are urgent (real user-facing breakage); tier classification is purely informational.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.6.44"
+version = "0.6.45"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_batched_engine_chat_template.py
+++ b/tests/test_batched_engine_chat_template.py
@@ -1,0 +1,129 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for BatchedEngine chat-template applicator selection.
+
+Some MLLM processors (notably ``Gemma3nProcessor`` and ``Gemma3Processor`` as
+loaded by ``mlx_vlm.load``) expose ``apply_chat_template`` as a method but
+ship ``chat_template=None`` at the processor layer — only the inner tokenizer
+carries the Jinja template. ``_apply_chat_template`` must detect this and
+fall back to the tokenizer; otherwise every request to those models returns
+zero tokens.
+"""
+
+from vllm_mlx.engine.batched import BatchedEngine
+
+
+class _RecordingApplicator:
+    """Tokenizer/processor stub that records which applicator handled the call."""
+
+    def __init__(self, label: str, chat_template: object):
+        self._label = label
+        self.chat_template = chat_template
+        self.last_messages = None
+        self.last_kwargs = None
+
+    def apply_chat_template(self, messages, **kwargs):
+        self.last_messages = messages
+        self.last_kwargs = kwargs
+        return f"<{self._label}>"
+
+
+class _ProcessorStub:
+    """MLLM processor stub.
+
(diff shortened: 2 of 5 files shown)


raullenchai:
**Pre-merge `make check` on qwen3.5-4b:**

```
[regression_suite[qwen3.5-4b]] OK (35.6s)
[smoke_matrix[qwen3.5-4b]] FAIL — 'THINKING_LONGER' not found (9 pass / 1 fail)
[stress[qwen3.5-4b]] OK (25.6s)
[autoresearch[qwen3.5-4b]] OK (68.2s)
```

The single `smoke_matrix` failure is the documented pre-existing `THINKING_LONGER` thinking-toggle flake on qwen3.5-4b (see `CLAUDE.md` — "passed on 4cedbf4, failed on bb332a5 in back-to-back runs"). Not introduced by this PR — the diff only touches MLLM code paths (`_apply_chat_template` MLLM branch, `MLLMBatchGenerator._run_vision_encoding`), neither of which is reachable on the text-only qwen3.5-4b run.

PR #406, opened 2026-05-17
fix(mlx): probe-and-cache shim for M5 single-stream GPU (#404)

### LINKED ISSUE #404: M5 MacBook Air: "There is no Stream(gpu, 1)" crash on pure-attention   models
### Rapid-MLX version

0.6.51

_No response_

### Hardware

MacBook Air M5, 24GB RAM 

### macOS version

26.4.1 

### Python version

3.12

### Model

mlx-community/gemma-4-26b-a4b-it-4bit

### Full serve command

rapid-mlx serve gemma-4-26b

Actual: On M5 hardware, any serve/bench of a pure-attention architecture model
   crashes with:

  Expected: The model serves successfully, as it does on M1-M4 hardware.

  Root cause: mlx_lm/generate.py:226 unconditionally creates a second GPU stream
   via mx.new_thread_local_stream(mx.default_device()), which creates
  Stream(gpu, 1). M5 GPU only supports a single stream (Stream(gpu, 0)). When
  PromptProcessingBatch.prompt() evaluates the prompt cache inside
  mx.stream(generation_stream), MLX throws because stream index 1 is not
  available.

  Models that DO work (hybrid architecture, different code path): qwen3.5-9b,
  qwopus-9b, qwen3.5-9b-8bit

  Suggested fix: mlx_lm should detect available stream count and fall back to
  Stream(gpu, 0) when only one is available.

### Minimal reproduction

1. On any M5 Mac: rapid-mlx serve gemma-4-26b
  2. Observe immediate crash during model loading.

  Even phi-4-mini-instruct-4bit (~3GB) triggers the same error, ruling out
  memory pressure.

```shell
File "mlx_lm/generate.py", line 1855, in next
      return self._next()
  File "mlx_lm/generate.py", line 1841, in _next
      self._prompt_batch.prompt(prompts)
  File "mlx_lm/generate.py", line 1161, in prompt
      mx.eval([c.state for c in self.prompt_cache])
```

raullenchai:
## Summary

Fixes #404 (M5 Apple Silicon: every pure-attention model crashes with `There is no Stream(gpu, 1) in current thread.`).

- `vllm_mlx/_mlx_compat.py` — probe-and-cache wrapper around `mx.new_thread_local_stream` (transparent on M1–M4, falls back to `mx.default_stream(device)` on M5 single-stream GPUs).
- Lazy install at the top of every module that imports `mlx_lm.generate`: `vllm_mlx/scheduler.py`, `vllm_mlx/pipeline/decode.py`, `vllm_mlx/model_runner.py::_generate_for_request`, `vllm_mlx/utils/mamba_cache.py::patch_mlx_lm_for_mamba`.
- `vllm_mlx/__init__.py` intentionally does **not** call install — would eagerly import `mlx.core`, which SIGABRTs with an uncatchable NSException on systems where mlx is installed but Metal is unavailable. The shim must be installed at scheduler-import time so `import vllm_mlx` stays free of `mlx.core`.
- `tests/test_mlx_compat.py` — 7 tests: install-hook fires before `mlx_lm.generate` import, idempotency, M5-fallback via monkeypatched `mx.stream`, unrelated-`RuntimeError` propagation, happy-path unchanged on dev hardware, structural audit that `__init__.py` never installs the shim, structural audit that every `mlx_lm.generate` consumer file calls `_mlx_compat.install()` (catches future drift).

## Why this bug existed

`mlx_lm/generate.py` does at module top level:

```python
generation_stream = mx.new_thread_local_stream(mx.default_device())
```

On M1–M4 this returns a usable stream. On M5, the call appears to succeed (returns a `Stream` handle), but later `with mx.stream(generation_stream):` raises `RuntimeError: There is no Stream(gpu, 1) in current thread.` because the M5 GPU exposes only a single stream slot. The capture is at *import time* of `mlx_lm.generate`, so by the time anything in `vllm_mlx` imports the scheduler, it's already too late to patch.

Hybrid models (Qwen3.5/3.6) sidestep this because their custom decode path doesn't import `mlx_lm.generate` — which is why #404 reports "Qwen3.5 works, everything else crashes."

## Fix design

Probe-and-cache:

1. First call per device → run a trivial `mx.array + mx.array` inside `with mx.stream(stream):`. If it raises the M5-shaped error, cache `False` and return `mx.default_stream(device)`.
2. If the probe succeeds, cache `True` and return the original stream.
3. Subsequent calls hit the cache; the probe is one-time per device.
4. The shim is *opt-out* — preserves original behavior on every device where it works. M5 users lose parallel-issue throughput but gain a functioning pure-attention path.

The `RuntimeError` filter is narrow: only swallows errors whose message contains `Stream` AND (`no Stream` OR `not exist`). Unrelated MLX errors propagate unchanged (covered by `test_fallback_does_not_engage_on_unrelated_runtime_error`).

## Why our SOPs didn't catch this

The miss was structural, not a process slip:

- **Doctor harness runs on the developer's hardware** (M3 here). Same chip family as upstream's CI. The bug only manifests on M5 — a family we don't own.
- **No cross-chip-family gate on mlx/mlx-lm version bumps.** Upstream landed the module-level `mx.new_thread_local_stream` call in a routine version refresh. We bumped the pin without auditing for new top-level GPU API usage that might not be portable.
- **No mock-the-failure-shape regression-test pattern.** Even without M5 hardware we could have written a test that monkeypatches `mx.stream` to raise the M5 error and asserts a graceful fallback — exactly what this PR's `test_fallback_engages_when_probe_raises` does.

SOP fix (added to `release_workflow.md`):

> **Step 9: MLX-version-bump cross-chip-family gate.** Triggers only when this release bumps `mlx` / `mlx-lm` / `mlx-vlm` / `mlx-metal`. Diff the upstream package for new module-level `mx.*stream*`, `mx.metal.*`, `mx.default_device`, or device-introspection calls. If any found: write a regression test that monkeypatches the call to raise the shape we'd see on the absent chip family, asserting our shim/fallback engages. Don't trust "works on my M3" — every M-series family has had at least one capability divergence.

## Verification

- `python3.12 -m pytest tests/test_mlx_compat.py -v` → 7/7 pass on M3 Ultra.
- Full unit suite: 3625 passed, 19 skipped, 1 xfailed (identical-quality to main + our 7 new tests).
- pr_validate: **MERGE-SAFE** — fetch / supply_chain / lint / targeted / full_unit / stress_e2e_bench (3×3 matrix qwen3-smoke + qwen3.5-35B + qwen3.6-27B) all PASS.
- Codex review × 3 rounds: R1 caught eager-mlx-import and order-dependent test (fixed), R2 caught untouched lazy import sites and missing `importlib.resources` (fixed), R3 clean.
- Ruff lint + format clean across all 7 modified files.
- **M5 hardware verification still needed** — asking the reporter @777stupid777 to confirm the fix on a real M5 before we merge.

## Test plan

- [x] All 7 unit tests pass on M3.
- [x] Lint + format clean.
- [x] Full pytest suite green (3625 passed).
- [x] pr_validate MERGE-SAFE (incl. 3×3 stress matrix).
- [x] codex review × 3 rounds clean.
- [x] Structural audit test guards against future drift (any new mlx_lm.generate consumer must call _mlx_compat.install()).
- [ ] M5 user @777stupid777 reproduces and confirms fix.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.6.52"
+version = "0.6.53"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_mlx_compat.py
+++ b/tests/test_mlx_compat.py
@@ -3,12 +3,20 @@
 Tests for the MLX hardware-compat shim (#404 M5 single-stream).
 
 We can't test on actual M5 from CI, but we can:
-1. Verify the shim is installed by importing vllm_mlx.
+1. Verify the shim is installed *before* any module-level
+   ``mx.new_thread_local_stream`` capture inside ``mlx_lm.generate``,
+   by checking that importing ``vllm_mlx.scheduler`` triggers install().
 2. Mock the probe failure and assert the fallback path returns
    ``mx.default_stream(...)``.
 3. Verify idempotency — install() can be called multiple times safely.
 4. Verify the shim is transparent on hardware that works (this runs on
    the dev's actual hardware in the test-apple-silicon CI job).
+
+We do NOT test that ``import vllm_mlx`` installs the shim — that is the
+*wrong* contract. We deliberately keep top-level ``import vllm_mlx``
+free of ``mlx.core`` import so the package stays usable for metadata-only
+access on systems where ``mlx`` is installed but Metal is unavailable
+(``import mlx.core`` SIGABRTs there with an uncatchable NSException).
 """
 
 from __future__ import annotations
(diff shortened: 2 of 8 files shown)


777stupid777:
Tested on M5 MacBook Air (24GB RAM) with v0.6.54 (Homebrew), unfortunately the fix doesn't resolve the issue.

## Repro

```bash
$ rapid-mlx --version
# v0.6.54
$ rapid-mlx bench gemma-4-26b
```

## Result

Still crashes with the same `There is no Stream(gpu, 1)` error, but the crash site is deeper than the `new_thread_local_stream` call — it's at `mx.eval([c.state for c in self.prompt_cache])` in `mlx_lm/generate.py:1161`:

```
  File "mlx_lm/generate.py", line 1161, in prompt
    mx.eval([c.state for c in self.prompt_cache])
```

This suggests the prompt cache states were already bound to stream 1 before the shim had a chance to intercept, or the shim's probe isn't engaging for this code path. All 10 benchmark prompts fail identically — 0 tokens generated.

## What works

- Model downloads fine
- Server starts without import errors (so the `_mlx_compat.install()` at import time runs)
- No crash until actual generation is attempted

## What I expected

The shim to intercept the stream creation so that `generation_stream` (and downstream prompt cache) uses `mx.default_stream(device)` on M5, and `bench` to complete successfully.

Happy to run any diagnostic commands you need on real M5 hardware.

777stupid777:
Note: I can use this model with the chat command.

Issue #320, opened 2026-05-08
Plan: harden PR-validation + Release SOP to catch logic + perf regressions


raullenchai:
## Why this issue

Despite `make full` + `pr_validate` + multi-round codex review, several real regressions have shipped:

- **Suffix-bench methodology bug** (#308): missing `temperature: 0.0` in the bench payload caused the patched server to fall through to vanilla on every greedy decision, so the bench measured **vanilla-vs-vanilla** for months. ~6 aliases got tier classifications based on noise. Caught only because PoC numbers (3-7× speedup) didn't match later integrated-bench numbers (1.0× flat).
- **Bench run-to-run instability**: same model (`devstral-v2-24b`) produced `chat 0.72x → avoid` on a standalone bench and `chat 0.99x → neutral` on the next sweep run with identical code. No stability gate.
- **PR #280 → `test_concurrent_same_prompt` regression**: bisected to event-driven idle wakeup. Determinism failure showed up as a flake the test suite was already xfail-ignoring; only caught later because someone happened to look.
- **Anthropic-compat regressions** in reasoning-parser PRs: OpenAI-compat smoke passes; `/anthropic` round-trip silently breaks. Fixed reactively (#288, #289), not preventatively.

The pattern: today's gauntlet catches **code correctness** well, but is weak at **methodology drift**, **measurement noise**, and **cross-surface coverage**.

### 1. Bench framework can be silently wrong

The suffix-bench bug went undetected because:
- Tests asserted internal contracts (`_classify_run` thresholds), not the actual httpx request payload
- "Pass" criteria was "tier output looks plausible," not "compared to a gold reference"
- No diff vs prior bench data when results drifted

### 2. No cross-run stability gate

Bench framework runs N=3 per workload but classifies even when variance is huge. The devstral-v2-24b case shows tier classification can flip across full re-runs.

### 3. Cache contamination

Disk-persisted prefix cache (`~/.cache/vllm-mlx/prefix_cache/`) silently changes results across runs. Bench tools must remember `--disable-prefix-cache`. Easy to forget; nothing enforces it.

### 4. `make full` is opt-in, manual

Per-PR perf gate documented in CLAUDE.md as gated on \"inference-touching\" changes. Reviewers must remember to ask. CI doesn't enforce it. Honest test plans skip it on borderline PRs.

### 5. `harness/baselines/` thin

[...]

PR #412, opened 2026-05-18
test(paired-compat): OpenAI and Anthropic surface equality smoke (#320 M3)

raullenchai:
## Summary
- Adds `tests/test_paired_compat.py` — mounts `/v1/chat/completions` and `/v1/messages` on one FastAPI app sharing a mock engine, POSTs the same request to both routes, and asserts the two surfaces produce identical `{text, reasoning, tool_calls}`.
- Closes the gap behind #288 and #289 (and the older #223 leak): the routes share engine + parser config but diverge at the streaming-think router, so a reasoning-parser PR could land green on OpenAI smoke while silently breaking Anthropic.
- Implements item **M3** of #320 (paired-compat smoke).
- **Discovered #413** while wiring up the gate: Anthropic non-streaming drops `<think>` content silently. That parametrize case is xfailed `strict=False` linked to the issue. The fact that the gate found a latent bug immediately is exactly the value being added.

## Design
- One in-process FastAPI app + one mock engine — the bug surface is the route layer, not the engine. ~3s runtime, no model load, no network.
- Mock yields real `GenerationOutput` dataclasses (not `SimpleNamespace`) so the route surface contract (`channel`, `finished`, `finish_reason`, `new_text`) is enforced. `_make_paired_client` mirrors `server.load_model` and constructs both `cfg.reasoning_parser_name` AND `cfg.reasoning_parser` — the latter is required by `_finalize_content_and_reasoning` (chat.py:637), otherwise the non-streaming reasoning case passes vacuously.
- Positive assertion (`assert openai["reasoning"]`) on the consistency test rules out the "both surfaces silently drop reasoning to empty string" false-positive class.
- Added `test_openai_reasoning_extracted` as a single-surface guard so a `chat.py:637` regression fails loudly even while #413's xfail is in place.

## Test plan
- [x] `python3.12 -m pytest tests/test_paired_compat.py -v` — 9 passed, 1 xfailed (the #413 case)
- [x] `python3.12 -m pytest tests/ --ignore=tests/integrations --ignore=tests/test_event_loop.py` — 3654 passed, 19 skipped, 2 xfailed (incl. new one), 0 failures
- [x] `ruff check tests/test_paired_compat.py && ruff format --check tests/test_paired_compat.py` clean
- [x] Codex review rounds 1+2+3 converged; round 1 critical and round 2 high gap both closed (see comment thread)
- [ ] CI green

### DIFF
--- a/tests/test_paired_compat.py
+++ b/tests/test_paired_compat.py
@@ -85,13 +85,25 @@ def _make_paired_client(
     reasoning_parser: str | None = None,
     no_thinking: bool = True,
 ) -> tuple[TestClient, _StreamingEngine]:
-    """One FastAPI app mounting both routers + shared mock engine."""
+    """One FastAPI app mounting both routers + shared mock engine.
+
+    Mirrors what ``server.load_model`` does for the parser wiring: both
+    ``cfg.reasoning_parser_name`` AND the constructed ``cfg.reasoning_parser``
+    instance must be set, otherwise the non-streaming OpenAI path
+    short-circuits in ``_finalize_content_and_reasoning`` (parser=None →
+    return cleaned_text untouched) and the reasoning-extraction assertion
+    becomes vacuous.
+    """
     engine = _StreamingEngine(deltas)
     cfg = reset_config()
     cfg.engine = engine
     cfg.model_name = "test-model"
     cfg.no_thinking = no_thinking
     cfg.reasoning_parser_name = reasoning_parser
+    if reasoning_parser:
+        from vllm_mlx.reasoning import get_parser
+
+        cfg.reasoning_parser = get_parser(reasoning_parser)()
     cfg.model_registry = None
 
     app = FastAPI()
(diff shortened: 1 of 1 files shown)


raullenchai:
## SOP execution log (round-up)

**Codex round 1 — CRITICAL**: `_make_paired_client` set `cfg.reasoning_parser_name` but not `cfg.reasoning_parser` (the instance). The non-streaming OpenAI reasoning case passed vacuously because `_finalize_content_and_reasoning` short-circuits when the parser instance is None, so both surfaces fell back to `strip_thinking_tags`. **Fixed** in 50c1baa: now constructs the parser instance via `get_parser()` mirroring `server.load_model`.

**Bug surfaced by fix**: wiring the parser correctly exposed a real production divergence — `vllm_mlx/routes/anthropic.py` non-streaming path only does `strip_thinking_tags(...)` and never extracts reasoning into a thinking block, while OpenAI non-streaming does. Filed as #413, parametrize case xfailed `strict=False` with reason linking to the issue. This is exactly the bug class the gate is meant to catch — the moment it was wired up, it found one.

**Codex round 2 — HIGH gap**: xfailing the [False] case removed the only guard on `chat.py::_finalize_content_and_reasoning`. **Fixed**: added `test_openai_reasoning_extracted` — a single-surface test that exercises the OpenAI path standalone for both stream modes. A `chat.py:637` regression now fails loudly regardless of #413's xfail state.

**Codex round 3**: ready to merge.

**Local gates**: `python3.12 -m pytest tests/ --ignore=tests/integrations --ignore=tests/test_event_loop.py` → 3654 passed, 19 skipped, 2 xfailed (incl. new one), 0 failures. `ruff check && ruff format --check` clean.

Test count went from 8 → 10 (1 xfailed).

Issue #344, opened 2026-05-09
follow-up: port upstream tool_call promotion (waybarrios#433) to our refactored think_parser


raullenchai:
## Background

Upstream commit \`9a6253a\` (\"Promote tool_call blocks from reasoning to content\", #433) fixes a real bug for thinking models that emit \`<tool_call>\` XML inside \`<think>\` blocks: the sequential parser pipeline classifies the tool call as reasoning, so the tool parser never sees it. Confirmed by upstream on Qwen 3.5/3.6.

## Why this is a follow-up

Our \`vllm_mlx/reasoning/think_parser.py\` was reorganized differently from upstream (we use \`_phase\` + \`_saw_any_tag\` state machine; upstream uses \`_phase\` + \`_content_started\` + \`_content_buffer\`). A direct cherry-pick collides on every method signature in the file. Porting the fix manually is feasible but non-trivial — ~200 LOC of careful translation across both non-streaming and streaming paths.

## What needs porting

From upstream \`vllm_mlx/reasoning/think_parser.py\`:
- Add module-level \`_TOOL_CALL_*\` regex constants
- Add \`_in_tool_call\` + \`_tool_call_buffer\` to streaming state + \`reset_state\`
- Add \`_promote_tool_calls(reasoning, content)\` for non-streaming
- Add streaming buffer/flush logic in \`extract_reasoning_streaming\` (handles \`</think>\` mid-buffer + stream-end flush)
- Add \`_transition_to_content\` catch-all for single-delta cases

From upstream \`tests/test_tool_call_promotion.py\` (new file, 385 LOC):
- 20 tests covering non-streaming, streaming, composition; structural guard against prose false positives

## Risk

Medium-high — \`think_parser.py\` is used by every thinking model (Qwen3, DeepSeek, GPT-OSS, etc). A regression here is visible to most users. Should be its own focused PR with codex review and full doctor harness on at least 2 thinking models.

## Reference

Upstream commit: https://github.com/waybarrios/vllm-mlx/commit/9a6253a

PR #426, opened 2026-05-20
streaming finalize: cross-format fallback to parse_tool_calls (fixes #425)

### LINKED ISSUE #425: Streaming drops all deltas when model emits <tool_call> markup on Qwen3.6-35B-A3B (v0.6.61, qwen3_xml + --no-thinking); non-streaming returns valid tool_calls
### Rapid-MLX version

0.6.61 (uv tool install)

### Hardware

MacBook Pro M4 Pro, 48 GB

### macOS version

26.2 (Darwin 25.2.0)

### Python version

3.13.13 (uv-managed venv at `~/.local/share/uv/tools/rapid-mlx`)

### Model

`mlx-community/Qwen3.6-35B-A3B-4bit-DWQ`

### Full serve command

```
rapid-mlx serve mlx-community/Qwen3.6-35B-A3B-4bit-DWQ \
  --port 8001 \
  --prefill-step-size 8192 \
  --reasoning-parser qwen3 \
  --tool-call-parser qwen3_xml \
  --enable-auto-tool-choice \
  --served-model-name qwen3.6-35b-mlx \
  --gpu-memory-utilization 1.0 \
  --kv-cache-quantization \
  --kv-cache-quantization-bits 8 \
  --no-thinking
```

### What happens

Whenever the model emits `<tool_call>...</tool_call>` markup in streaming mode, the entire response is dropped on the wire. The stream emits exactly two SSE events — a role opener and a `finish_reason: stop` closer — with `completion_tokens` correctly accounting for the generated tokens, but zero `content`, `reasoning_content`, or `tool_calls` deltas.

The same prompt non-streaming returns a clean parsed `tool_calls` array with `finish_reason: tool_calls`. So the model + parser are doing the right thing; only the streaming wire path is broken.

This is independent of:

- `--no-thinking` vs `--reasoning-parser qwen3` alone (both fail; with reasoning enabled the tool call gets gobbled into `reasoning_content` as 215 deltas of think-block text, then a single `"\n\n"` content delta — also broken, possibly the same root cause as the still-open #344)
- `tool_choice: "auto"` vs `tool_choice: "none"`
- whether `tools: [...]` is set in the request at all — the parser still extracts and drops the tool call

The 0.6.61 finalize-fallback in `service/postprocessor.py:510` does ship in this install (verified — line 510 has the `"[Calling"` substring check, line 518 invokes `extract_tool_calls`), and the substring `"<"` in the response should trigger it. It does not. Either `_fallback_text` is empty by the time finalize runs (qwen3_xml parser consumed tokens without surfacing them to fallback), or the fallback path runs but the resulting tool_call event isn't getting written to the SSE stream.

### Reproduction

Server up (command above). Streaming request:

```
curl -sN -i http://127.0.0.1:8001/v1/chat/completions \
  -d '{
    "model": "qwen3.6-35b-mlx",
    "messages": [{"role":"user","content":"Read /etc/hostname"}],
    "tools": [{"type":"function","function":{"name":"read_file","description":"Read a file","parameters":{"type":"object","properties":{"path":{"type":"string"}},"required":["path"]}}}],
    "tool_choice": "auto",
    "max_tokens": 200,
    "temperature": 0,
    "stream": true
  }'
```

Full wire response (verbatim, every byte):

```
HTTP/1.1 200 OK
content-type: text/event-stream; charset=utf-8

data: {"id":"chatcmpl-df245252","object":"chat.completion.chunk","created":...,"choices":[{"index":0,"delta":{"role":"assistant"}}]}

data: {"id":"chatcmpl-df245252","object":"chat.completion.chunk","created":...,"choices":[{"index":0,"delta":{},"finish_reason":"stop"}],"usage":{"prompt_tokens":358,"completion_tokens":27,"total_tokens":385}}

data: [DONE]
```

Server-side log for that same request:

```
[REQUEST] POST /v1/chat/completions stream=True ... tools=1
[stream_outputs] ... first token after 0.8s
[stream_outputs] ... finished normally, 27 tokens in 1.2s
Chat completion (stream): 27 tokens in 1.28s (21.1 tok/s)
```

27 tokens generated. Stream emitted 0 deltas.

### Same prompt non-streaming (works)

Drop `"stream": true`:

```json
{
  "finish_reason": "tool_calls",
  "usage": {"prompt_tokens": 358, "completion_tokens": 27, "total_tokens": 385},
  "message": {
    "tool_calls": [{
      "id": "call_0f3795d5",
      "type": "function",
      "function": {"name": "read_file", "arguments": "{\"path\": \"/etc/hostname\"}"}
    }]
  }
}
```

Identical model behaviour (same 27 tokens), parser extracts correctly.

### Isolation tests

| Streaming request | Result |
|---|---|
| `What is the capital of France?` + tools in request, `tool_choice: auto` | streams "Paris" cleanly |
| `Read /etc/hostname` + tools, `tool_choice: auto` | 0 deltas, finish_reason: stop, 27 tokens generated |
| `Read /etc/hostname` + tools, `tool_choice: none` | 0 deltas (model still emits `<tool_call>`, gets eaten anyway) |
| `Read /etc/hostname` + no tools at all in request | 0 deltas (model still emits `<tool_call>`, parser still extracts and drops) |
| Plain prompt asking to print literal `hello <stuff> world` | streams `hello <stuff> world` cleanly (so it's not "any angle bracket triggers it") |
| Plain prompt asking to print literal `<tool_call>FAKE</tool_call> end` | model refuses ("XML entity injection attempt") — separate behaviour, ignore |

The trigger is the model emitting `<tool_call>` markup, not the request shape.

### Expected

Streamed `tool_calls` deltas + `finish_reason: tool_calls` (same as non-streaming).

### Related

- #344 (open): tool_call inside `<think>` blocks gets classified as reasoning. The reasoning-enabled streaming behaviour above (215 reasoning deltas, `<tool_call>` swallowed) looks like the same root cause as #344. The `--no-thinking` case in this report is different — there's no `<think>` block at all, the model jumps straight to `<tool_call>` and it still gets dropped.
- #424 (closed, 0.6.61): finalize fallback added for the gemma-4 streaming gap. The fallback ships but doesn't catch this case for Qwen3.6.
- #175 (closed): historical regression of `qwen3_xml` being aliased to `qwen3_coder_xml` in 0.6.6. Unrelated as far as I can tell — both parsers exhibit the same streaming drop on 0.6.61.

jpcarranza94:
## What does this PR do?

Adds a cross-format fallback to `StreamingPostProcessor.finalize()` so streaming wire-format mismatches surface as a structured `tool_calls` delta instead of dropping the entire response.

When the configured `--tool-call-parser`'s `extract_tool_calls()` returns `tools_called=False` and the accumulated text passes the existing structural pre-check (`<` / `{` / `[Calling`), `finalize()` now calls `api.tool_calling.parse_tool_calls` — the same multi-format scanner the non-stream path already uses at `service/helpers.py:604` — and emits a `tool_call` event if it finds one. The new path is defensive (try/except matching helpers.py) and emits an INFO log naming the configured parser when it recovers a call, so a wire-format mismatch is visible in server logs.

Both finalize branches now go through a shared `_build_tool_call_event()` helper so the configured-parser and fallback paths can't drift in wire shape.

## Why is this needed?

Fixes #425.

`--tool-call-parser qwen3_xml` is registered to `QwenToolParser`, which expects `<tool_call>{"name":..., "arguments":...}</tool_call>` (JSON body). Vanilla `Qwen3.6-35B-A3B` (and the mlx-community 4-bit-DWQ checkpoint many users run) emits the XML-body variant: `<tool_call><function=name><parameter=key>value</parameter></function></tool_call>`. The configured parser returns `tools_called=False`, the existing 0.6.61 finalize fallback (#424) calls only the configured parser, and the stream emits zero `tool_call` deltas with `finish_reason: stop` — while the same prompt non-streaming returns a clean structured `tool_calls` array (because non-stream already falls back to the multi-format scanner).

Concrete user impact: agent harnesses like Pi, Cursor, Aider, Continue.dev all use `stream=True`. With `qwen3_xml` selected (a documented CLI choice), tool calls silently fail on Qwen3.6-35B-A3B — the model emits the right markup, the wire drops it.

## AI assistance disclosure

Claude (Anthropic) wrote the diagnosis probes, the postprocessor change, and the tests. I read the surrounding code independently to verify the diagnosis (`vllm_mlx/service/postprocessor.py`, `service/helpers.py:540-607`, `tool_parsers/qwen_tool_parser.py:28`, `tool_parsers/qwen3coder_tool_parser.py:97`, `api/tool_calling.py:352-440`, `routes/chat.py:999-1017`), confirmed the live wire trace on my machine end-to-end against `mlx-community/Qwen3.6-35B-A3B-4bit-DWQ` on `rapid-mlx 0.6.61` (M4 Pro 48 GB, macOS 26.2), ran the full `test_postprocessor.py` suite myself, and addressed staff-level code review feedback (try/except, observability log, helper extraction, negative tests) in a follow-up commit. I can explain the intent, risk, and behavior of every change in this PR.

## Test plan

- `python3 -m pytest tests/test_postprocessor.py -x -q` → 78 passed (4 new tests under `TestStreamingPostProcessorToolCalls`)
- `ruff check vllm_mlx/service/postprocessor.py tests/test_postprocessor.py` → clean
- `ruff format --check vllm_mlx/service/postprocessor.py tests/test_postprocessor.py` → clean
- `PR_VALIDATE_NO_DEEPSEEK=1 PR_VALIDATE_NO_STRESS=1 python3 -m scripts.pr_validate.pr_validate 426` → `fetch`, `supply_chain`, `lint`, `targeted_tests` all PASS. `full_unit` FAILs in the validator environment with 32 `ModuleNotFoundError: No module named 'mlx'` collection errors — that's pr_validate running under a pyenv 3.12 that lacks the mlx wheel, not a regression from this diff. The same errors reproduce on `main` in that env. Per [CONTRIBUTING.md](../CONTRIBUTING.md#what-if-my-pr-fails-on-a-pre-existing-main-bug), flagging here for a maintainer to confirm.
- Live end-to-end: `mlx-community/Qwen3.6-35B-A3B-4bit-DWQ` on `rapid-mlx 0.6.61`, M4 Pro 48 GB, macOS 26.2, with the user-facing config from #425.

Before the fix (whole wire output for the streaming repro from #425):

```
data: {"choices":[{"delta":{"role":"assistant"}}]}
data: {"choices":[{"delta":{},"finish_reason":"stop"}],"usage":{"completion_tokens":27,...}}
data: [DONE]
```

After the fix:

```
data: {"choices":[{"delta":{"role":"assistant"}}]}
data: {"choices":[{"delta":{},"finish_reason":"stop"}],"usage":{"completion_tokens":27,...}}
data: {"choices":[{"delta":{"tool_calls":[{"index":0,"id":"call_c35c322d","type":"function","function":{"name":"read_file","arguments":"{\"path\": \"/etc/hostname\"}"}}]},"finish_reason":"tool_calls"}]}
data: [DONE]
```

Repeated 3× with different paths to confirm reliability. Plain text streaming unchanged (regression check): `Say hello` → `Hello! How can I help you today?` streams as before.

Server logs the recovery so a future user with a misconfigured `--tool-call-parser` has a breadcrumb:

```
INFO:vllm_mlx.service.postprocessor:[finalize] cross-format fallback recovered 1 tool_call(s); configured parser='qwen3_xml' returned tools_called=False — consider whether --tool-call-parser matches the model's wire format
```

## Checklist

- [x] Tests pass locally (`python3 -m pytest tests/ -x`) — 78/78 in `test_postprocessor.py`; `full_unit` in pr_validate fails on the pre-existing env issue noted above.
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 426` — see Test plan for the `full_unit` env caveat.
- [x] New tests touch a critical code path (streaming tool-call recovery). Spot-checked that `test_finalize_cross_format_fallback_recovers_xml_tool_call` fails when the new fallback branch is removed (configured parser returns `tools_called=False` ⇒ no event emitted), and that `test_finalize_cross_format_fallback_both_parsers_fail` / `test_finalize_cross_format_fallback_swallows_parser_exception` exercise the empty-result and exception paths.
- [x] No docs update needed — no user-facing API change. Behavior change is documented in the inline comment at `postprocessor.py` and the new log line. Happy to add a note to `knowledge/guided_generation_gaps_2026-05-20.md` (referenced from #424's release notes) if that's the project convention for follow-on fixes.
- [x] No breaking changes to existing API. The `tools_called=True` branch is untouched; the new fallback only runs in the `tools_called=False` branch that previously emitted nothing.

## Scope, and what this does NOT fix

This handles the case where the streaming parser swallows the entire tool call (`--no-thinking` path, model jumps straight to `<tool_call>`). The related but distinct case where a `<tool_call>` block sits inside a `<think>...</think>` reasoning block — the reasoning parser gobbles it into `reasoning_content` before tool detection runs — is the existing open issue #344 (port `waybarrios/vllm-mlx@9a6253a`). That requires changes to `vllm_mlx/reasoning/think_parser.py` and is out of scope here.

One observation worth a separate issue: `qwen3_xml` at `tool_parsers/qwen_tool_parser.py:28` is registered to `QwenToolParser` (JSON body) but the `tool_parsers/__init__.py` docstring implies it covers the XML-body Qwen3 family. This PR fixes the symptom (streaming silently drops tool calls); the registration itself may also be worth correcting. Happy to file as a follow-up if the maintainer agrees.

### DIFF
--- a/tests/test_postprocessor.py
+++ b/tests/test_postprocessor.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for StreamingPostProcessor — the unified streaming pipeline."""
 
+import json
 from unittest.mock import MagicMock
 
 from vllm_mlx.service.postprocessor import StreamingPostProcessor
@@ -367,8 +368,6 @@ class TestStreamingPostProcessorToolCalls:
         tc = events[0].tool_calls[0]
         assert tc["function"]["name"] == "read_file"
         # arguments is a JSON string; parse to assert the value
-        import json
-
         assert json.loads(tc["function"]["arguments"]) == {"path": "/etc/hostname"}
         assert pp.tool_calls_detected is True
 
@@ -392,6 +391,63 @@ class TestStreamingPostProcessorToolCalls:
         pp.tool_accumulated_text = "the value of x < 10 is fine"
 
         events = pp.finalize()
+        assert events == []
+        assert pp.tool_calls_detected is False
+
+    def test_finalize_cross_format_fallback_both_parsers_fail(self):
+        """Structural markup is present (passes the ``<`` pre-check) but the text
+        is not a parseable tool call in any known format. Both the configured
+        parser and ``parse_tool_calls`` return no calls. finalize() must return
+        ``[]`` cleanly and leave ``tool_calls_detected`` False — not raise, not
+        emit a spurious event."""
+        tool_parser = self._make_tool_parser()
+        tool_parser.extract_tool_calls.return_value = MagicMock(
+            tools_called=False, tool_calls=[]
+        )
+
+        cfg = _make_cfg(
+            enable_auto_tool_choice=True,
+            tool_parser_instance=tool_parser,
+        )
+        pp = StreamingPostProcessor(cfg)
+        pp.reset()
+
+        # XML-shaped but not actually a tool call — no <function=...> or
+        # [Calling tool: ...] or {"name":...} structure parse_tool_calls
+        # recognises.
+        pp.tool_accumulated_text = "<tool_call>garbled junk no actual call</tool_call>"
+
+        events = pp.finalize()
+        assert events == []
+        assert pp.tool_calls_detected is False
+
+    def test_finalize_cross_format_fallback_swallows_parser_exception(self):
+        """``parse_tool_calls`` must not be allowed to abort the stream. If the
+        multi-format scanner raises (regex pathology, adversarial input), the
+        fallback logs a warning and returns ``[]`` rather than propagating.
+        Mirrors the defensive ``try/except`` in ``service/helpers.py:605-607``."""
+        tool_parser = self._make_tool_parser()
+        tool_parser.extract_tool_calls.return_value = MagicMock(
+            tools_called=False, tool_calls=[]
+        )
+
+        cfg = _make_cfg(
+            enable_auto_tool_choice=True,
+            tool_parser_instance=tool_parser,
+        )
+        pp = StreamingPostProcessor(cfg)
+        pp.reset()
+        pp.tool_accumulated_text = "<tool_call>anything</tool_call>"
+
+        # Force the fallback parser to raise
+        from unittest.mock import patch
+
+        with patch(
+            "vllm_mlx.service.postprocessor.parse_tool_calls",
+            side_effect=RuntimeError("boom"),
+        ):
+            events = pp.finalize()
+
         assert events == []
         assert pp.tool_calls_detected is False
 
--- a/tests/test_postprocessor.py
+++ b/tests/test_postprocessor.py
@@ -321,6 +321,80 @@ class TestStreamingPostProcessorToolCalls:
         assert events[0].type == "tool_call"
         assert events[0].finish_reason == "tool_calls"
 
+    def test_finalize_cross_format_fallback_recovers_xml_tool_call(self):
+        """Configured parser is wire-format-specific; finalize must fall back to
+        the multi-format ``parse_tool_calls`` when the configured parser fails.
+
+        Reproduces #425: ``--tool-call-parser qwen3_xml`` resolves to
+        ``QwenToolParser`` (JSON inside ``<tool_call>``) but vanilla
+        Qwen3.6-35B-A3B emits the XML-bodied variant
+        (``<function=...><parameter=...>``). The configured parser returns
+        ``tools_called=False`` on the wire-mismatched text; without a
+        cross-format fallback the stream emits zero ``tool_calls`` deltas
+        while the same text non-streaming returns a structured call via the
+        ``parse_tool_calls`` fallback in ``service/helpers.py``.
+        """
+        tool_parser = self._make_tool_parser()
+        # Configured parser fails to extract (wire-format mismatch)
+        tool_parser.extract_tool_calls.return_value = MagicMock(
+            tools_called=False, tool_calls=[]
+        )
+
+        cfg = _make_cfg(
+            enable_auto_tool_choice=True,
+            tool_parser_instance=tool_parser,
+        )
+        pp = StreamingPostProcessor(cfg)
+        pp.reset()
+
+        # Qwen3.6 vanilla wire format: XML body inside <tool_call>
+        pp.tool_accumulated_text = (
+            "<tool_call>\n"
+            "<function=read_file>\n"
+            "<parameter=path>\n"
+            "/etc/hostname\n"
+            "</parameter>\n"
+            "</function>\n"
+            "</tool_call>"
+        )
+
+        events = pp.finalize()
+        assert len(events) == 1
+        assert events[0].type == "tool_call"
+        assert events[0].finish_reason == "tool_calls"
+        assert events[0].tool_calls is not None
+        assert len(events[0].tool_calls) == 1
+        tc = events[0].tool_calls[0]
+        assert tc["function"]["name"] == "read_file"
+        # arguments is a JSON string; parse to assert the value
+        import json
+
+        assert json.loads(tc["function"]["arguments"]) == {"path": "/etc/hostname"}
+        assert pp.tool_calls_detected is True
+
+    def test_finalize_cross_format_fallback_noop_on_plain_text(self):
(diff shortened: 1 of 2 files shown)


raullenchai:
Thanks @jpcarranza94 — merged! 🎉 Added you to the changelog:

Issue #214, opened 2026-05-05
Prefix cache misses on every turn in growing multi-turn conversations on hybrid models (Qwen3.6-35B-A3B)


oldriverno1:
## Summary

In multi-turn agent conversations on hybrid attention models (e.g. `mlx-community/Qwen3.6-35B-A3B-4bit`), Rapid-MLX's prefix cache misses on **every turn** and re-prefills the entire cumulative conversation. TTFT grows linearly with cumulative size while `mlx_lm.server` (running the exact same model) keeps TTFT roughly flat.

This is a different scenario from #163 (closed via #165). #163 verified that **identical prompts repeated** now hit the cache. That fix does not cover the dominant agent pattern: **conversation grows by appending new user/assistant turns**, where each new prompt is a strict superset of the previous prompt. That path stays at 0% hit rate on hybrid models.

## Reproduction

- Hardware: Apple M5 Max, 128 GB unified memory
- Rapid-MLX 0.6.11 (`vllm_mlx 0.6.11`)
- mlx-lm 0.31.3
- Model: `mlx-community/Qwen3.6-35B-A3B-4bit` (hybrid: GatedDeltaNet + Transformer; 30/40 layers non-trimmable RNN state)

Server commands:

```bash
# Rapid-MLX (with explicit --no-thinking to rule out thinking generation as a confounder)
rapid-mlx serve mlx-community/Qwen3.6-35B-A3B-4bit \
  --host 0.0.0.0 --port 8080 \
  --default-temperature 0.0 --max-tokens 64 \
  --continuous-batching --max-num-seqs 4 \
  --no-thinking

# mlx_lm.server (baseline)
mlx_lm.server --model mlx-community/Qwen3.6-35B-A3B-4bit \
  --host 0.0.0.0 --port 8080 \
  --temp 0.0 --max-tokens 64 \
  --decode-concurrency 4 --prompt-concurrency 4
```

Methodology:

1. Start with a 30K-token initial user message containing a needle (`AUTH_KEY=ZX-9847-DELTA`).
2. After each model response, **substitute the assistant reply with a fixed `"OK"` string** before the next turn. This guarantees identical conversation history growth across all backends.
3. Each subsequent turn appends a new ~4K-token user message and the same fixed `"OK"` assistant reply.
4. Run 9 turns plus a final recall question.
5. `chat_template_kwargs={"enable_thinking": false}` is set on every request.
[...]

PR #434, opened 2026-05-21
chore: codex follow-ups to #433 (cli/test/stream cleanup)

raullenchai:
## Summary

Three small internal-quality cleanups surfaced by DeepSeek/codex review of v0.6.64 (PR #433). All additive — no behavior change for clients on the happy path.

- **F2** (`vllm_mlx/cli.py`): narrow tool-parser registry exception handling to `(ImportError, AttributeError)` and emit a stderr warning when validation is skipped, instead of a silent bare except that could mask unrelated faults.
- **F3** (`tests/test_harmony_parsers.py`): remove dead `_get_serve_tool_parser_choices` static method. The surviving tests already assert registry membership via `ToolParserManager.tool_parsers`.
- **F5** (`vllm_mlx/routes/chat.py`): in the defensive `elif fallback_tool_calls:` branch of `stream_chat_completion`, carry `processor.accumulated_text` and `processor.accumulated_reasoning` into the synthetic terminal chunk's `delta` so the chunk doesn't silently truncate any content the model emitted before the parser missed the tool call.

## Test plan

- [x] `ruff check` + `ruff format --check` clean on modified files
- [x] `TestHarmonyCLIIntegration` (5/5) pass
- [x] 244 chat/streaming/tool/postprocessor tests pass (243 passed + 1 pre-existing skip)
- [x] `python -c "from vllm_mlx.tool_parsers import ToolParserManager; ..."` shows 39 registered parsers — F2 registry lookup confirmed live

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #435, opened 2026-05-22
fix(scheduler): port boundary-snapshot save to mlx-lm 0.31+ BatchGenerator (#427)

### LINKED ISSUE #427: Follow-up to #214: port boundary-snapshot save path to mlx-lm 0.31+ BatchGenerator API
# Follow-up to #214: port boundary-snapshot save path to mlx-lm 0.31+ `BatchGenerator` API

## Context

[#214](https://github.com/raullenchai/Rapid-MLX/issues/214) was closed via #242 with regression tests pinning the strict-prefix invariants. However, the maintainer's closing comment on #214 noted that the actual user-facing miss was deferred:

> **Boundary-snapshot infrastructure already exists** in `memory_cache.py` (`_compute_prefix_boundary`, `prefix_boundary` field, `_make_mid_prefill_save_callback`) **but is gated on the legacy `_install_chunked_prefill` path which was disabled when we moved to mlx-lm 0.31+**. Re-enabling that path on the BatchGenerator API is the actual fix and should be a separate PR — reopen this issue or file a follow-up for that work.

This is the follow-up. The bug as reported in #214 reproduces verbatim on the current release.

## Confirmation that the bug persists

- rapid-mlx **v0.6.60** (≈50 patch releases since #214 closed)
- mlx-lm 0.31.x via the rapid-mlx venv install
- Model: `TheCluster/Qwen3.6-35B-A3B-MLX-mixed-9bit` (hybrid: linear-attention/Mamba + Transformer; same family as the original #214 repro)
- Hardware: Mac Studio M2 Ultra, 128 GB unified memory

Started with the per-model recommended sampling + chunked-prefill + pinned system prompt:

```
rapid-mlx serve --log-level DEBUG \
  --default-temperature 0.6 --default-top-p 0.95 --default-top-k 20 \
  --default-min-p 0 --default-presence-penalty 0 --default-repetition-penalty 1.0 \
  --pin-system-prompt --chunked-prefill-tokens 2048 \
  TheCluster/Qwen3.6-35B-A3B-MLX-mixed-9bit
```

The `--chunked-prefill-tokens 2048` flag is silently a no-op for snapshotting — the server immediately warns:

```
WARNING:vllm_mlx.scheduler:[chunked_prefill] Skipped — mlx-lm 0.31+ removed the
internal Batch API. Using native prefill_step_size=2048 instead.
Mid-prefill snapshots are unavailable.
```

(`scheduler.py:1869-1874` — the `_install_chunked_prefill` call site at `scheduler.py:1853` is short-circuited because `BatchGenerator` no longer exposes `_process_prompts`.)

Running a normal multi-turn opencode session against this server then produces the same MISS pattern reported in #214, with the cache lookup always finding a near-perfect LCP but unable to use it because the Mamba layers are non-trimmable:

```
[cache_fetch] LCP scan: cached_len=11786 req_len=11377 lcp=11376
[cache_fetch] LCP candidate: lcp=11376 entry_len=11786 excess=410 non_trimmable=True
[cache_fetch] request=69560886-2d3 MISS prompt_tokens=11377 entries=5

[cache_fetch] LCP scan: cached_len=11595 req_len=23455 lcp=11486
[cache_fetch] LCP candidate: lcp=11486 entry_len=11595 excess=109 non_trimmable=True
[cache_fetch] request=2edbd26e-93c MISS prompt_tokens=23455 entries=6

[cache_fetch] LCP scan: cached_len=23941 req_len=32888 lcp=23454
[cache_fetch] LCP candidate: lcp=23454 entry_len=23941 excess=487 non_trimmable=True
[cache_fetch] request=b15f2ba2-696 MISS prompt_tokens=32888 entries=3

[schedule] request=b15f2ba2-696 uid=4 prompt_tokens=32888 tokens_to_prefill=32888 ...
```

`excess` = the size of the cached entry's tail that diverges from the next turn — i.e. the `<think>\n` template sentinel emitted by `add_generation_prompt=True` on the prior turn that gets replaced by actual assistant content on the next. This is exactly the divergence demonstrated at the token level in #214 (`turn1=<think>` → `turn2=OK` at index 41).

## What's still missing (the actual fix)

The boundary-snapshot infrastructure called out in #214's closing comment is all present and well-commented:

- `request.py:120` — `prefix_boundary: int = 0` field already on `Request`.
- `scheduler.py:189` — `_install_chunked_prefill(...)` — legacy Batch-API hook.
- `scheduler.py:1853` — call site that would install the mid-prefill save callback.
- `scheduler.py:1862-1874` — branch that explicitly skips it on mlx-lm 0.31+ and logs the WARNING above.
- `scheduler.py:2006-2050` — `_make_mid_prefill_save_callback`, which already snapshots at `prefix_boundary` correctly (`scheduler.py:2027-2030`: "Always save at prefix_boundary (message boundary for cache reuse with different final user messages)").

So all that's needed is to wire `_make_mid_prefill_save_callback` (or an equivalent boundary-trigger) into the public `BatchGenerator` hooks on mlx-lm 0.31+. The save logic, the boundary computation, and the request field are all in place.

## Suggested patch shape

1. Detect mlx-lm 0.31+ via the same `_has_old_batch_api = hasattr(bg, "_process_prompts")` probe already at `scheduler.py:1838`.
2. When the old API is unavailable, register a `BatchGenerator`-level callback (whichever the public 0.31+ API exposes — `on_chunk` / per-step output stream / etc.) that calls into the existing `_make_mid_prefill_save_callback`'s closure with `(uid, processed_tokens, prompt_cache)`.
3. Demote the current `WARNING` to an `INFO` once the new path is live, or remove it entirely.

Happy to test a fix branch against this repro — the conversation traces above are reproducible by running any opencode/agent client against the listed `rapid-mlx serve` invocation.

## Related

- #214 (origin, closed) — chat-template re-encoding diagnosis + token-level proof.
- #163 / #165 — identical-prompt-replay fix (`_snapshot_promoted_prompts`); does not cover the growing-conversation pattern.
- #287 — PFlash-style long-prompt prefill compression (different lever, but related to long-prefill TTFT).

raullenchai:
Closes #427. Restores the per-message cache snapshot path that's been dormant since the mlx-lm 0.31+ migration disabled the legacy \`_install_chunked_prefill\` monkey-patch.

## Background

#214 closed with a deferred-fix note: the boundary-snapshot infrastructure (\`prefix_boundary\` field, \`_make_mid_prefill_save_callback\`) was all present but gated on the legacy Batch API, which mlx-lm 0.31+ removed. Without it, hybrid models (linear-attention/Mamba + Transformer) MISS the prefix cache on every multi-turn agentic conversation because:

- The LCP scan finds ~99% match (typically \`<think>\\n\` template tail of the prior turn diverging from new turn's actual assistant content).
- Mamba layers are non-trimmable, so the supersequence fallback can't reuse them either.
- Result: full re-prefill on every turn.

@fishloa reported a verbatim repro on v0.6.60 against \`TheCluster/Qwen3.6-35B-A3B-MLX-mixed-9bit\` on M2 Ultra, with A/B vs oMLX showing ~78% prefix reuse delta on turn 2+.

## Fix

Wire the boundary save into the public 0.31+ API without re-enabling the monkey-patch:

1. **\`_schedule_waiting\`** splits the insert via \`insert_segments([[prefix_seg, tail_seg]])\` when \`request.prefix_boundary\` lies strictly inside \`tokens_to_process\` and a memory-aware cache exists.
2. **New \`_snapshot_boundary_segments(prompt_responses)\`** reads \`end_of_segment=True AND NOT end_of_prompt=True\` (BatchGenerator's signal that the prefix segment finished while the tail still has work), extracts via public \`extract_cache\` API, reconstructs through existing helpers, stores at \`prompt_token_ids[:prefix_boundary]\` with \`evict_prefixes=False\`.
3. **Hook** added to \`step()\` alongside \`_snapshot_promoted_prompts\`. \`_boundary_snapshot_taken\` guard prevents duplicate stores.
4. **WARNING demoted to INFO** at scheduler.py:1869 — \`--chunked-prefill-tokens\` is still a no-op (the per-N-tokens interval-based snapshots remain unavailable on the new API), but per-message boundary snapshots ARE now supported.

## Test plan

- [x] 9 new \`TestBoundarySnapshot\` cases in \`tests/test_prompt_cache_snapshot.py\` covering: stores at boundary; skips end_of_prompt (other path); skips when boundary=0; skips when no memory cache; idempotent per request; handles extract/store failures; skips unknown uid; handles empty responses.
- [x] All 96 pre-existing cache/scheduler tests pass.
- [x] 3857 unit tests pass overall.
- [x] Lint + format clean.
- [ ] E2E verification against fishloa's repro: \`rapid-mlx serve TheCluster/Qwen3.6-35B-A3B-MLX-mixed-9bit\` with multi-turn opencode session; expect \`[boundary_snapshot] saved N tokens\` on turn 1 and \`[cache_fetch] HIT cached=N\` on turn 2+ instead of MISS.

### DIFF
--- a/tests/test_prompt_cache_snapshot.py
+++ b/tests/test_prompt_cache_snapshot.py
@@ -195,5 +195,268 @@ class TestPromptCacheSnapshot:
         assert scheduler.memory_aware_cache.store.call_count == 2
 
 
+class TestBoundarySnapshot:
+    """Tests for `_snapshot_boundary_segments` — the per-message cache
+    save path for mlx-lm 0.31+ multi-turn agentic workloads on hybrid
+    models (issue #427).
+
+    The path fires on `end_of_segment=True AND NOT end_of_prompt=True`,
+    which BatchGenerator emits when a request inserted via
+    `insert_segments([[prefix, tail]])` finishes the prefix segment but
+    the tail still has tokens left. Each fire extracts the
+    boundary-state cache, reconstructs it through the existing
+    `_extract_cache_states`/`_reconstruct_cache_from_states` helpers,
+    and stores it under `request.prompt_token_ids[:prefix_boundary]`
+    so the next turn's lookup gets a prefix-cache hit.
+    """
+
+    def _register_with_boundary(
+        self,
+        scheduler,
+        request_id: str,
+        uid: int,
+        prompt_tokens: list[int],
+        prefix_boundary: int,
+    ):
+        request = _register(scheduler, request_id, uid, prompt_tokens)
+        request.prefix_boundary = prefix_boundary
+        return request
+
+    def test_snapshot_stores_at_prefix_boundary(self):
+        scheduler = _make_scheduler_with_cache()
+        # Mock the extract/reconstruct helpers so we don't need a real
+        # MLX cache to test the dispatch logic.
+        scheduler._extract_cache_states = MagicMock(return_value=[{"k": "v"}])
+        scheduler._reconstruct_cache_from_states = MagicMock(
+            return_value=["reconstructed-cache"]
+        )
+
+        prompt_tokens = list(range(100))
+        prefix_boundary = 80
+        self._register_with_boundary(
+            scheduler,
+            "req-1",
+            uid=101,
+            prompt_tokens=prompt_tokens,
+            prefix_boundary=prefix_boundary,
+        )
+
+        bg = MagicMock()
+        # Stage-1 (in-prompt) extract returns (cache, tokens).
+        bg.extract_cache.return_value = {101: (["raw-cache"], prompt_tokens[:80])}
+        scheduler.batch_generator = bg
+        scheduler.memory_aware_cache.store = MagicMock(return_value=True)
+
+        # end_of_segment without end_of_prompt: prefix segment done,
+        # tail still pending.
+        responses = [
+            SimpleNamespace(
+                uid=101,
+                progress=(80, 100),
+                end_of_segment=True,
+                end_of_prompt=False,
+            ),
+        ]
+        scheduler._snapshot_boundary_segments(responses)
+
+        bg.extract_cache.assert_called_once_with([101])
+        scheduler.memory_aware_cache.store.assert_called_once()
+        stored_tokens, stored_cache = scheduler.memory_aware_cache.store.call_args.args[
+            :2
+        ]
+        assert stored_tokens == prompt_tokens[:prefix_boundary]
+        assert stored_cache == ["reconstructed-cache"]
+        # evict_prefixes=False must be passed so the boundary entry
+        # isn't evicted by the later whole-prompt save.
+        assert (
+            scheduler.memory_aware_cache.store.call_args.kwargs.get("evict_prefixes")
+            is False
+        )
+
+    def test_snapshot_skips_end_of_prompt_responses(self):
+        """end_of_prompt is the whole-prompt promotion — handled by the
+        other snapshot path, must not duplicate-fire here."""
+        scheduler = _make_scheduler_with_cache()
+        self._register_with_boundary(
+            scheduler,
+            "req-1",
+            uid=101,
+            prompt_tokens=[1, 2, 3, 4],
+            prefix_boundary=2,
+        )
+        bg = MagicMock()
+        scheduler.batch_generator = bg
+        scheduler.memory_aware_cache.store = MagicMock()
+
+        responses = [
+            SimpleNamespace(
+                uid=101,
+                progress=(4, 4),
+                end_of_segment=True,
+                end_of_prompt=True,
+            ),
+        ]
+        scheduler._snapshot_boundary_segments(responses)
+
+        bg.extract_cache.assert_not_called()
+        scheduler.memory_aware_cache.store.assert_not_called()
+
+    def test_snapshot_skips_when_no_prefix_boundary(self):
+        """Single-turn requests have prefix_boundary=0 and should not
+        trigger a boundary save even if end_of_segment fires."""
+        scheduler = _make_scheduler_with_cache()
+        # No prefix_boundary attr — default is 0 from Request.__init__.
+        _register(scheduler, "req-1", uid=101, prompt_tokens=[1, 2, 3, 4])
+        bg = MagicMock()
+        scheduler.batch_generator = bg
+        scheduler.memory_aware_cache.store = MagicMock()
+
+        responses = [
+            SimpleNamespace(
+                uid=101,
+                progress=(3, 4),
+                end_of_segment=True,
+                end_of_prompt=False,
+            ),
+        ]
+        scheduler._snapshot_boundary_segments(responses)
+
+        bg.extract_cache.assert_not_called()
+        scheduler.memory_aware_cache.store.assert_not_called()
+
+    def test_snapshot_skips_when_no_memory_cache(self):
+        scheduler = Scheduler(
+            MagicMock(),
+            MagicMock(),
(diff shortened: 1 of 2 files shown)


raullenchai:
## pr_validate first-round results (commit f01b4e6)

\`\`\`
fetch          OK
deepseek_review  FAIL  4 finding(s)
supply_chain     OK
lint             OK
targeted_tests   OK   18 passed
full_unit        OK   3856 passed, 1 xfailed
stress_e2e_bench FAIL matrix 4×3 — 5 BLOCKING
\`\`\`

### DeepSeek 4 findings — all addressed in acf3c28

- **#1** (P1): no test coverage for insert_segments dispatch → added \`TestScheduleWaitingInsertDispatch\` (6 cases pinning the split-decision arithmetic).
- **#2** (P1): \`_boundary_snapshot_taken\` guard only set on store success → now set after attempt regardless (failed store usually means "entry exists" or "cache busy"; retrying every step is pure waste).
- **#3** (false positive): \`SchedulerConfig(enable_prefix_cache=False)\` already works (existing \`test_snapshot_no_op_when_callback_disabled\` uses exactly that pattern).
- **#4** (covered): \`extract_cache\` \`AttributeError\` is already caught by the existing \`try/except Exception\`.

### Codex round 1 P1 — also addressed in acf3c28

mlx-lm 0.31+ silently rewrites \`insert_segments([[prefix, tail]])\` into \`[[prefix, tail[:-1], tail[-1:]]]\` when \`len(tail) > 1\` ([generate.py:1646-1648](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/generate.py#L1646-L1648)), so \`end_of_segment\` fires up to 3 times. Finding-#2 fix prevents the corruption; added defensive progress-validation gate that deterministically skips fires whose \`progress[0]\` doesn't match \`prefix_boundary - cached_tokens\`.

Tests now 25/25 pass.

### stress_e2e_bench 5 BLOCKING — all pre-existing on v0.6.64

| Item | PR #435 | PR #434 (v0.6.64 baseline) | Verdict |
|---|---|---|---|
| qwen3-0.6B anthropic | 4/5 (test 2 "FAIL: 34") | — | Sub-1B model quirk (per memory, never test with <1B) |
| gpt-oss-20b stress | 7/8 (Tool call storm FAIL) | 7/8 (same) | Pre-existing |
| gpt-oss-20b anthropic | 0/5 (all empty text) | 0/5 (bit-for-bit identical) | Pre-existing |
| gpt-oss-20b langchain | 2/6 (tool calls empty) | 2/6 (same) | Pre-existing |
| gpt-oss-20b pydantic | 1/6 | 1/6 (same) | Pre-existing |

**Proof this PR can't be the cause**: \`git diff raullenchai/main..HEAD -- 'vllm_mlx/*.py'\` shows only \`scheduler.py\` changed. Zero parser/tool_parser/reasoning/route/api code touched. The harmony empty-content failures fire on the chat.py + harmony tool_parser path — not reachable from scheduler-level changes.

### Positive signal — qwen3.6 hybrid (the actual #427 target model)

\`\`\`
qwen3.6 stress: 8/8 PASS
qwen3.6 anthropic: pending (skipped per agent setup)
qwen3.6 langchain: pending
qwen3.6 pydantic: pending
\`\`\`

8/8 stress includes Mixed workload + Long generation under sustained load. The fix exercises the new \`insert_segments\` dispatch + boundary-snapshot save under realistic multi-turn pressure.

### Open question

Should I re-run pr_validate against acf3c28 for convergence (it would take ~25-30 min) or merge based on the analysis above?

CI on acf3c28: pending (will check post-push).

PR #485, opened 2026-05-29
fix: install M5 compat shim before mlx_lm import in mllm_batch_generator

mxl:
## Problem

On **M5** (single-stream GPU), the server crashes immediately on first request:

```
[generation_error_recovery] aborted N running requests, batch generator closed, Metal cache cleared
```

The `_mlx_compat` shim (#404) correctly patches `mx.new_thread_local_stream` to handle M5's single-stream limitation, but it only takes effect if installed **before** `mlx_lm.generate` is imported — because that module captures the stream at module level.

## Root cause

`mllm_batch_generator.py` does a top-level import:

```python
from mlx_lm.sample_utils import make_sampler
```

This triggers `mlx_lm/__init__.py → from .generate import ...`, which captures `new_thread_local_stream` **before** `scheduler.py` calls `_mlx_compat.install()`.

## Fix

Add `_mlx_compat.install()` to `mllm_batch_generator.py` before the `mlx_lm` import, mirroring the existing pattern in `scheduler.py`.

### DIFF
--- a/vllm_mlx/mllm_batch_generator.py
+++ b/vllm_mlx/mllm_batch_generator.py
@@ -24,6 +24,11 @@ from typing import Any
 
 import mlx.core as mx
 import mlx.nn as nn
+
+from . import _mlx_compat as _mlx_compat
+
+_mlx_compat.install()
+
 from mlx_lm.sample_utils import make_sampler
 
 from .multimodal_processor import MultimodalProcessor


mxl:
Closing — misdiagnosis. The reporter is on M1 Max, not M5. Root cause is different.

raullenchai:
Thanks for digging into this, Michael — and for being upfront about the M1-Max-vs-M5 mix-up. That kind of self-correction is genuinely useful signal.

The `There is no Stream(gpu, 1) in current thread` error you saw is real and worth understanding even if the M5-shim path wasn't the right fix. A few questions so we can repro:

1. **Exact command** that triggered it on M1 Max (`rapid-mlx serve …` or `rapid-mlx bench …`, which model, any flags).
2. **Versions** — `pip show mlx mlx-lm mlx-vlm rapid-mlx` from the broken env.
3. **Did the crash happen on cold start (first request after `rapid-mlx serve`) or only after some specific sequence?**
4. **Install mode** — fresh `pipx install rapid-mlx` / Homebrew / editable from a branch?

We'll attempt local repro on M3 Ultra. The shim path you patched (`_mlx_compat.install()` running before `mlx_lm` import) is the right shape of fix for a real class of bugs — the question is just whether the trigger in your env was something else (e.g. stale wheel resident from an older `mlx`). I'll comment back here with what we find.

raullenchai:
Quick update, Michael — we reproduced and fixed this in #487.

**Repro outcome on M3 Ultra (multi-stream, where the crash is invisible):**
- `from mlx_lm.sample_utils import make_sampler` triggers `mlx_lm/__init__.py` → `from .generate import …` → `generate.py` line 226 `generation_stream = mx.new_thread_local_stream(mx.default_device())` runs at module-import time.
- If that import lands BEFORE `_mlx_compat.install()`, the captured stream is the un-patched ThreadLocalStream. On multi-stream HW (M1–M4) it works fine. On M5 (and apparently your env), `with mx.stream(generation_stream):` raises `Stream(gpu, 1)`.
- Three files in our tree had this pattern: `mllm_batch_generator.py`, `mllm_scheduler.py`, `models/deepseek_v4.py`. Plus `cli.py:bench_command` does `from mlx_lm import load` before the import path that would pull in scheduler.

**Why our SOP guard missed it:** `tests/test_mlx_compat.py::test_every_mlx_lm_generate_consumer_installs_shim` was matching only the literal string `mlx_lm.generate`. None of the three files had that — they imported sibling submodules, which still hit the same `__init__.py` cascade. PR #487 broadens the guard to any top-level `from mlx_lm` / `import mlx_lm`.

**Your PR's direction was right.** The narrower scoping (just `mllm_batch_generator`) didn't fully fix it, but the underlying diagnosis — "shim must run before any `mlx_lm.*` import, not just `mlx_lm.generate`" — is what we ended up implementing.

If you'd like to validate on M1 Max after PR #487 merges, the smoke is:
```bash
pipx install --force git+https://github.com/raullenchai/Rapid-MLX.git@fix/mlx-compat-shim-coverage
rapid-mlx serve gemma-4-31b-it-4bit
curl -X POST http://127.0.0.1:8000/v1/chat/completions -H 'Content-Type: application/json' \\
  -d '{"model":"gemma-4-31b-it-4bit","messages":[{"role":"user","content":"hi"}],"max_tokens":8}'
```
Would love to know if that clears your crash or whether there's a residual issue.

Thanks again for the report — first external contribution to this fork, and it caught a real SOP-guard hole.

PR #486, opened 2026-05-29
fix: add gemma-4/gemma4 to MLLM_PATTERNS

mxl:
## Problem

Gemma 4 models (`mlx-community/gemma-4-31b-it-4bit`, `-8bit`, etc.) crash on every request with:

```
[generation_error_recovery] aborted N running requests, batch generator closed, Metal cache cleared
```

## Root cause

`MLLM_PATTERNS` (in both `api/utils.py` and `benchmark.py`) was missing `gemma-4`/`gemma4`.

`is_mllm_model('mlx-community/gemma-4-31b-it-4bit')` returned `False` because:
1. `_try_read_config_json` returns `None` (HF repo ID, not a local path)
2. `_check_legacy_string_patterns` returned `False` (gemma-4 not in patterns)

But Gemma 4 has architecture `Gemma4ForConditionalGeneration` with `vision_config` — it **is** multimodal and must load through the MLLM engine path.

Additionally, `bench_command` loaded the model on the asyncio loop thread, causing the same Stream error for any model that requires MLLM routing.

## Changes

1. **`api/utils.py`**: add `gemma-4`/`gemma4` to `MLLM_PATTERNS`
2. **`benchmark.py`**: add `gemma-3`/`gemma3`/`gemma-4`/`gemma4` to local `MLLM_PATTERNS` (this file has its own copy)
3. **`cli.py`** (`bench_command`): load model on mlx-step worker thread instead of asyncio loop thread, pass executor to `AsyncEngineCore`

## Note on mlx-vlm dependency

After this fix, Gemma 4 correctly routes through the MLLM path, which requires `mlx-vlm`. Users will see a clear error message guiding them to `pip install 'rapid-mlx[vision]'`. Previously, the silent mis-routing to the LLM path was causing stream errors with no actionable message.

## Known limitation

`rapid-mlx bench` still does not support MLLM models — the `bench_command` uses `mlx_lm.load` directly and `BatchGenerator` which is incompatible with `BatchKVCache` used by MLLM models. Benchmarking Gemma 4 should be done via `rapid-mlx serve` + an OpenAI-compatible client.

### DIFF
--- a/vllm_mlx/api/utils.py
+++ b/vllm_mlx/api/utils.py
@@ -601,6 +601,8 @@ MLLM_PATTERNS = [
     "PaliGemma",  # PaliGemma
     "gemma-3",
     "gemma3",  # Gemma 3 (multimodal)
+    "gemma-4",
+    "gemma4",  # Gemma 4 (multimodal)
     "medgemma",
     "MedGemma",  # MedGemma (medical multimodal with SigLIP vision encoder)
     "pixtral",
--- a/vllm_mlx/benchmark.py
+++ b/vllm_mlx/benchmark.py
@@ -639,6 +639,10 @@ MLLM_PATTERNS = [
     "InternVL",
     "deepseek-vl",
     "DeepSeek-VL",
+    "gemma-3",
+    "gemma3",  # Gemma 3 (multimodal)
+    "gemma-4",
+    "gemma4",  # Gemma 4 (multimodal)
 ]
 
 # Test image URL (Yellow Labrador from Wikimedia Commons)
--- a/vllm_mlx/cli.py
+++ b/vllm_mlx/cli.py
@@ -996,11 +996,12 @@ def serve_command(args):
 def bench_command(args):
     """Run benchmark."""
     import asyncio
+    import concurrent.futures
     import time
 
     from mlx_lm import load
 
-    from .engine_core import AsyncEngineCore, EngineConfig
+    from .engine_core import AsyncEngineCore, EngineConfig, _init_mlx_step_thread
     from .request import SamplingParams
     from .scheduler import SchedulerConfig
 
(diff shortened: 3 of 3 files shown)


mxl:
Closing this because a clean rapid-mlx 0.6.68 install works without the MLLM_PATTERNS change, so this is not the right upstream fix.

raullenchai:
Thanks for taking the time to investigate this and open a PR, Michael — first external contribution to this fork, much appreciated.

Even though you closed it, the symptom you reported is concerning and I'd like to make sure we don't lose the signal. A few questions so we can repro on our side:

1. **Environment delta** — when you say "clean rapid-mlx 0.6.68 install works," what was different from the env where the crash originally happened? (pip cache, editable install from a branch, older `mlx`/`mlx-lm`/`mlx-vlm` version still resident, etc.)
2. **Exact repro command** — was it `rapid-mlx serve mlx-community/gemma-4-31b-it-4bit` then a single chat-completions request, or via `rapid-mlx bench`, or something else?
3. **Versions** — `pip show mlx mlx-lm mlx-vlm rapid-mlx` from the broken env if you still have it.
4. **Hardware** — you mentioned M1 Max in the sibling PR (#485) — same machine here?

The `Stream(gpu, 1)` error path is also referenced in PR #485, so we suspect the two crashes share a root cause that neither patch landed on. We'll try to repro locally on M3 Ultra in the meantime — happy to keep you in the loop if it turns out to be a real bug.

raullenchai:
Update — see #487. Your closing comment about clean install working is correct: Gemma 4 text variants intentionally route through the LLM path via `vllm_mlx/models/gemma4_text.py` (a thin wrapper around mlx-vlm's `LanguageModel`). `is_mllm_model('mlx-community/gemma-4-*-it-*bit')` returns False by design — routing to MLLM batched would actually break things because the architecture is `Gemma4ForConditionalGeneration` and our text-only loader handles it.

The Stream(gpu, 1) crash you originally saw was almost certainly the same root cause as #485 (top-level `mlx_lm.<sub>` imports running before the `_mlx_compat.install()` shim). PR #487 fixes that. Closing this thread as superseded by the merged conclusion + #487.

PR #541, opened 2026-06-08
perf: B=1 probes — locate root cause of sampling gap (4.3 ms in sampler chain)

raullenchai:
## Summary

Five-phase B=1 perf investigation: three falsifications + one cross-engine validation + root cause located.

Adds two env-gated probes plus two experiment hooks. All gated, zero impact on production code path with env vars unset.

- **Probe 1** (`RAPID_MLX_PROBE_ENGINE_LOOP=1`) — engine-loop split: `step_inside_ms` / `step_await_ms` / `dispatch_ms` + executor RTT.
- **Probe 2** (same env var) — scheduler-step split into 6 buckets.
- **Experiment 1** (`RAPID_MLX_PROBE_SKIP_FULL_LOGPROBS=1`) — replace `BatchGenerator._step` to materialise only `[B,]` sampled-token logprob (mlx-vlm shape).
- **Experiment 2** (`RAPID_MLX_PROBE_USE_VLM_SAMPLER=1`) — additional swap of mlx-lm's `apply_top_p + categorical_sampling` chain for mlx-vlm's single-function `top_p_sampling`.

Harness: `scripts/probe_engine_loop.py`. Full report: `reports/profiles/engine_loop_probe_2026-06-08.md`.

### Phase 1 — asyncio dispatch (FALSIFIED)
Combined RTT + dispatch is < 1% of per-token wall. Dedicated GPU thread rewrite would close < 1%. Cancelled.

### Phase 2 — scheduler.py wrapper Python overhead (FALSIFIED)
Wrapper code combined is < 1%; `bg_next` is 99.3%. No optimisation needed.

### Phase 3 — greedy logsumexp-skip (TRIED, NO-OP, REVERTED)
Patch fired but bench within noise. mlx-lm already short-circuits at `make_sampler(temp=0)`.

### Phase 4 — direct mlx-vlm bench at top_p=0.95 (gap CONFIRMED REAL)
mlx-vlm runs 92.73 tok/s under sampling (within 1% of its own greedy). The 30% gap vs rapid-mlx (64.9) is real.

### Phase 5 — ROOT CAUSE LOCATED

| Variant | bg_next | HTTP tok/s | sampler dispatch | async_eval block (GPU work) |
|---|---|---|---|---|
| baseline (mlx-lm chain) | 14.07 ms | 65.71 | 0.94 ms | 11.24 ms |
| skip-full-logprobs only | 13.94 ms | 66.28 | 0.94 ms | 11.24 ms |
[...]

PR #542, opened 2026-06-09
perf(sampler): fused top-p/top-k/temp sampler — B=1 +42% on Qwen 3.6 35B-A3B

raullenchai:
## Summary

Closes the 4.3 ms / token B=1 gap to mlx-vlm under sampling. **Qwen 3.6 35B-A3B 4-bit @ B=1 HTTP: 65 → 92 tok/s (+42%)**, matches mlx-vlm 92.7. Qwen 3.5 4B sanity: 128 → 137 tok/s (+7%, smaller absolute sampler cost).

Root cause (investigation report previously documented + closed via #541): mlx-lm's `make_sampler` chains independent `@mx.compile`'d ops — `apply_top_p` → `apply_top_k` → `categorical_sampling`. For the dominant chat config (temp + top_p + top_k from alias defaults), this costs ~4.3 ms / token split as:

- **~0.9 ms Python** — three closure dispatches per step inside mlx-lm's per-row sampler loop.
- **~3.3 ms GPU** — two `@mx.compile` boundaries break lazy-graph fusion across `apply_top_p → categorical_sampling`. `apply_top_p` also builds a vocab-sized `inverse_indices` array (`mx.put_along_axis + mx.arange`) to undo the sort permutation back into vocab order so `categorical` can sample in vocab space — that scatter is the largest single op in the chain.

## What this PR ships

- **`vllm_mlx/_sampler_fast_path.py`** — vendored single-function fused sampler. Builds the top-p / top-k mask in **sorted** space (no `inverse_indices`), applies temperature scaling and mask to the sorted logits, samples in sorted space, maps back via one `take_along_axis`.
- **`vllm_mlx/scheduler.py`** — wires the fast path into the existing `_get_request_sampler` cache (4-tuple key on `(temp, top_p, min_p, top_k)`). Eligibility:
  - `temperature > 0` (mlx-lm already short-circuits to argmax at `temp == 0`),
  - `min_p == 0` (would need a fourth op the fast path doesn't implement — falls through),
  - at least one of `0 < top_p < 1` or `top_k > 0` is active (otherwise the chain collapses to plain `categorical_sampling` — no benefit to steal),
  - one-time INFO log on first engagement so reviewers can confirm in production logs.
- **`tests/test_sampler_fast_path.py`** — 15 cases covering eligibility window, shape contract, and distributional equivalence to mlx-lm's chain.

## Math equivalence vs mlx-lm

The fused sampler computes the top-p mask on the unscaled probability distribution (matching `apply_top_p` semantics — *not* mlx-vlm's post-temperature variant), applies the top-k mask, then scales the masked logits by `1/T` before `mx.random.categorical`. Identical to `apply_top_p → apply_top_k → categorical_sampling` in:

- the kept set of tokens (top-p computed pre-temperature, top-k by rank),
- the relative weights inside the kept set,
- the expected sample distribution.

Only sample-level determinism with a fixed `mx.random.seed` differs: mlx-lm draws Gumbel noise in vocab order, we draw in sorted order — same seed picks a different token. Within-engine reproducibility is preserved (same input + same seed always picks the same token on this engine), which is what the OpenAI `seed=` field promises.

## Bench

Qwen 3.6 35B-A3B 4-bit, M3 Ultra 256 GB, B=1 HTTP streaming chat completion, `temperature=0.7 top_p=0.95`, `top_k=20` injected by the alias's generation_config cascade.

| Variant | tok/s |
|---|---|
| baseline (mlx-lm chain) | 65.10 |
| **this PR (3 runs)** | **91.66 / 91.74 / 92.38** |
| mlx-vlm 0.6.1 reference | 92.73 |

Qwen 3.5 4B 4-bit sanity: 128 → 137 tok/s (+7%).

## Test plan

- [x] `pytest tests/test_sampler_fast_path.py` — 15/15 pass
- [x] `pytest tests/test_dense_sampler_fastpath.py tests/test_sampling_params_passthrough.py tests/test_alias_recommended_sampling.py` — 38/38 pass (no regression on existing sampler / cache / alias logic)
- [x] `ruff check vllm_mlx/_sampler_fast_path.py vllm_mlx/scheduler.py tests/test_sampler_fast_path.py` clean
- [x] Live HTTP bench on Qwen 3.6 35B-A3B 4-bit confirms +42% throughput vs baseline
- [x] Live HTTP bench on Qwen 3.5 4B 4-bit confirms +7% throughput (no regression on small models)
- [x] `[fused_top_p_sampler] engaged for ...` INFO log confirmed in production logs

## Risk assessment

- Touches the **per-request sampler factory** (`Scheduler._get_request_sampler`) — single hot path, shared by every dense / hybrid request. Falls through to mlx-lm's chain on every off-window combination (greedy, min_p, no top-p+no top-k), so the **fallback path is untouched**.
- The fast sampler is mathematically equivalent to mlx-lm's chain at the distribution level. Within-engine `seed=` reproducibility is preserved. Cross-engine bit-level reproducibility against a vanilla mlx-lm install is not — same seed picks a different token on the fused path. Documented in the module docstring.
- The existing `_install_dense_sampler_fastpath` (PR #521 batched-sampler swap) continues to work: when multiple homogeneous requests share the cached fused sampler (identity-equal via the sampler cache), PR #521's homogeneous-batch detection kicks in and calls the fused sampler **once on the full `[B, V]` matrix** instead of per-row. So the fast path scales with batch size too.

Generated with Claude Code

### DIFF
--- a/tests/test_dense_sampler_fastpath.py
+++ b/tests/test_dense_sampler_fastpath.py
@@ -254,16 +254,18 @@ def test_sampler_cache_is_bounded_lru():
     samplers = [sched._get_request_sampler(_SP(0.1 * i)) for i in range(6)]
     assert len(sched._sampler_cache) == 4, "cap must hold even under churn"
 
-    # Oldest entries (temp=0.0, 0.1) are evicted.
+    # Oldest entries (temp=0.0, 0.1) are evicted. The 5th tuple element
+    # is the ``RAPID_MLX_DISABLE_FUSED_SAMPLER`` flag — False here because
+    # the env var is unset in the test environment.
     keys = list(sched._sampler_cache.keys())
-    assert (0.0, 0.95, 0.0, 20) not in keys
-    assert (0.1, 0.95, 0.0, 20) not in keys
+    assert (0.0, 0.95, 0.0, 20, False) not in keys
+    assert (0.1, 0.95, 0.0, 20, False) not in keys
     # Newest entry is at the LRU tail.
-    assert keys[-1] == (0.5, 0.95, 0.0, 20)
+    assert keys[-1] == (0.5, 0.95, 0.0, 20, False)
 
     # Hot-key LRU bookkeeping: hitting an existing key bumps it to MRU.
     sched._get_request_sampler(_SP(0.2))  # was middle of the LRU
-    assert list(sched._sampler_cache.keys())[-1] == (0.2, 0.95, 0.0, 20)
+    assert list(sched._sampler_cache.keys())[-1] == (0.2, 0.95, 0.0, 20, False)
 
 
 def test_method_type_wrapper_sees_correct_self():
--- a/tests/test_dense_sampler_fastpath.py
+++ b/tests/test_dense_sampler_fastpath.py
@@ -230,7 +230,7 @@ def test_sampler_cache_interns_by_param_tuple():
     assert len(sched._sampler_cache) == 2
 
 
-def test_sampler_cache_is_bounded_lru():
+def test_sampler_cache_is_bounded_lru(monkeypatch):
     """The cache key is request-controlled (clients pick temp/top_p), so
     the cache MUST be bounded. Without a cap, an adversarial client
     streaming unique floats grows ``_sampler_cache`` without bound for
@@ -238,6 +238,11 @@ def test_sampler_cache_is_bounded_lru():
     from collections import OrderedDict
 
     from vllm_mlx.scheduler import Scheduler
+
+    # Codex round-3 NIT #3: pin the escape-hatch env var so the cache
+    # key's 5th element is deterministic regardless of test invocation
+    # environment.
+    monkeypatch.delenv("RAPID_MLX_DISABLE_FUSED_SAMPLER", raising=False)
 
     sched = Scheduler.__new__(Scheduler)
     sched._sampler_cache = OrderedDict()
@@ -256,7 +261,7 @@ def test_sampler_cache_is_bounded_lru():
 
     # Oldest entries (temp=0.0, 0.1) are evicted. The 5th tuple element
     # is the ``RAPID_MLX_DISABLE_FUSED_SAMPLER`` flag — False here because
-    # the env var is unset in the test environment.
+    # we explicitly deleted the env var above.
     keys = list(sched._sampler_cache.keys())
     assert (0.0, 0.95, 0.0, 20, False) not in keys
     assert (0.1, 0.95, 0.0, 20, False) not in keys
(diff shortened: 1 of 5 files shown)


raullenchai:
## Cross-model A/B bench

Same harness, same 256-token streaming chat completion at `temp=0.7 top_p=0.95`, B=1 HTTP on M3 Ultra. A/B done in-process via `RAPID_MLX_DISABLE_FUSED_SAMPLER=1` (the env opt-out added in c6d49a5). Three runs per side, median reported.

| Model | Arch | Vocab | Quant | Fast path tok/s | Baseline tok/s | Delta | Fast path engaged? |
|---|---|---|---|---|---|---|---|
| qwen3.6-35b | hybrid + MoE | 151k | 4-bit | **90.85** | 66.41 | **+36.8%** | yes (top_k=20) |
| qwen3.6-27b-8bit | hybrid | 151k | 8-bit | 19.50 | 19.00 | +2.6% | yes (top_k=0) |
| hermes3-8b (Llama 3.1) | dense | 128k | 4-bit | 116.74 | 115.09 | +1.4% | yes (top_k=0) |
| qwen3.5-4b | dense | 151k | 4-bit | ~137 | ~128 | +7% | yes (top_k=0) |
| gemma3-12b | dense | 262k | 4-bit | 62.37 | 62.88 | -0.81% | **no** — cascade falls outside eligibility window |

### Reading the data

- **The big win is concentrated on Qwen 3.6 35B-A3B** — the hybrid MoE model where (a) the active-param forward is cheap so the sampler is a *large fraction* of per-step wall, and (b) `top_k=20` from the alias cascade triples the per-step sampler-chain cost (apply_top_p + apply_top_k + categorical_sampling, three `@mx.compile` boundaries instead of two).
- **Dense models see ~1-3%** because the absolute sampler cost (1-2 ms) is a small fraction of the forward-dominated step on these architectures. The win is real but in the noise band; no regression.
- **No regressions** on any model.
- **gemma3-12b's fast path did not engage** — the alias's `recommended_sampling` + the model's `generation_config.json` defaults push the cascade outside our eligibility window (likely `min_p > 0` lands on the request). Both sides ran the mlx-lm chain identically; the -0.8% is run-to-run noise, not a fast-path artifact. Extending the fast path to cover `min_p` is a small follow-up if Gemma is a load-bearing target — out of scope here.

### Engagement logs (one-time per Scheduler)

```
[fused_top_p_sampler] engaged for temp=0.700 top_p=0.950 top_k=20  ← qwen3.6-35b
[fused_top_p_sampler] engaged for temp=0.700 top_p=0.950 top_k=0   ← qwen3.6-27b-8bit
[fused_top_p_sampler] engaged for temp=0.700 top_p=0.950 top_k=0   ← hermes3-8b
[fused_top_p_sampler] engaged for temp=0.700 top_p=0.950 top_k=0   ← qwen3.5-4b
(no engagement)                                                     ← gemma3-12b
```

### Headline

This PR is targeted at the largest gap (Qwen 3.6 35B-A3B B=1 sampled, +37%) and does not regress anything else. Dense and smaller-vocab models pick up incremental gains; Gemma 3 12B is a no-op until the fast path covers `min_p`.

PR #555, opened 2026-06-11
feat(diffusion): in-house generation loop for DiffusionGemma (P1)

raullenchai:
## Summary

Vendors mlx-vlm's `stream_diffusion_generate` into `vllm_mlx/runtime/diffusion_loop.py` so DiffusionGemma's generation loop lives inside rapid-mlx, while continuing to delegate the model architecture (MMDiT block-diffusion forward pass + `QuantizedEmbedding` layers) to upstream.

This is **Phase P1** of the in-source migration roadmap (see `docs/plans/in-source-roadmap.md`, local-only). Full mlx-vlm removal for this model — vendoring the ~1300 LOC of MMDiT model code — lands as a follow-up PR once P1 has soaked. The goal of the multi-phase split is to keep the user-visible delta small per PR and stage the change of ownership across modules that rapid-mlx already understands first.

## Why

The current `runtime/diffusion_lane.py` routes 100% of DiffusionGemma traffic to `mlx_vlm.generate.diffusion.stream_diffusion_generate`. We have no place to land:
- self-conditioning matmul cadence (`sc_every`) tuning — verified empirically that only `sc_every=1` produces coherent DiffusionGemma output; the upstream-default cadence drops quality
- fixed-step (vs entropy-bound) stopping cadence (`fixed_steps=8` is empirically optimal for the 26B-A4B 4-bit checkpoint at 256 canvas length)
- per-step sampling that respects `temperature > 0` cleanly (mlx-vlm uses `_diffusion_sample_canvas`; mirrored here)

Carrying the loop in-tree lets every diffusion knob ride on the existing `AliasProfile` SSOT.

## What changed

**Backend selection**

3 new `AliasProfile` fields, all defaulting to the empirically optimal values for `diffusion-gemma-26b`:
- `diffusion_backend: Literal["rapid","mlx-vlm"]` — default `"rapid"`
- `diffusion_fixed_steps: int` — default `8`
- `diffusion_sc_every: int` — default `1`

Non-text-diffusion aliases reject all three fields at load time (modality gate). Per the routing SOP (`test_no_out_of_band_routing.py`), there is **no env-var kill switch** — emergency rollback is "edit `aliases.json` to set `diffusion_backend: \"mlx-vlm\"`, restart".

**Runtime**

`DiffusionEngine.__init__` resolves the alias profile (or synthesizes a throwaway `AliasProfile` with safe defaults for bare HF paths that no alias references). At call time `_run_mlxvlm_generator` branches on `profile.diffusion_backend`:
- `rapid` → `rapid_stream_diffusion_generate` (with `fixed_steps`/`sc_every` threaded through)
- `mlx-vlm` → `stream_diffusion_generate` verbatim (rapid-only kwargs stripped to avoid `TypeError` upstream)

The in-house loop is lazy-imported so the mlx-vlm install path stays identical for users who set the backend back to `mlx-vlm`.

**Loop implementation** (`vllm_mlx/runtime/diffusion_loop.py`, 368 LOC)

- `RapidDiffusionResult` is a frozen dataclass mirroring the `getattr(result, …, default)` shape the consumer side (`diffusion_lane.py:1241`) already expects from mlx-vlm's `GenerationResult` — zero consumer-side changes.
- `_initialize_canvas` matches upstream: random ids in `[0, vocab_size)` via `mx.random.randint`. Off-by-one here lands the mask token in the canvas and breaks the first denoising pass.
- `_soft_embedding_weight` dequantizes `nn.QuantizedEmbedding` via `mx.dequantize(...)`. Without this, the packed weight shape leaks into `probs @ embedding_weight` and downstream RMSNorm fires the shape-mismatch crash that surfaced in the first smoke run.
- `_sample_canvas` vendors mlx-vlm's `_diffusion_sample_canvas`: `temperature <= 0` → `argmax` (deterministic), `temperature > 0` → scaled `mx.random.categorical`. The lane used to gate the rapid path on `temperature == 0`, but chat clients default to `_FALLBACK_TEMPERATURE = 0.7`, so 99% of traffic would have missed the rapid path. With `_sample_canvas` honoring temperature, that gate is removed.
- `_extract_eos_ids` unions `generation_config.eos_token_id` + tokenizer EOS. DiffusionGemma splits its EOS set across both (config carries `1`; generation_config adds `<end_of_turn>=106`); missing the union silently turns `<end_of_turn>` into normal text.
- Vision input fails loud (text-only model); the lane never routes pixels here.

## Tests

| Suite | Count | What it pins |
|------|-------|--------------|
| `test_alias_diffusion_knobs.py` | 10 | default knobs, typo/invalid rejection, modality gate, existing aliases unchanged |
| `test_diffusion_loop.py` | 21 | canvas init, soft-embedding scale, sample-canvas dtype+temperature, EOS union, guards |
| `test_diffusion_lane_routing.py` | 5 | default → rapid; profile override → mlx-vlm; temperature doesn't gate; profile fallback for unknown HF path |
| `test_diffusion_gemma_parity.py` | 4 (`@slow`) | loads real 1.2 GB checkpoint; coherent keyword output on 3 canonical prompts; ≥80 tok/s throughput floor (hand-measured ~127 tok/s on M3 Ultra) |
| `test_diffusion_engine.py` | 70 unchanged | extended `_install_mlx_vlm_mock` to also stub `rapid_stream_diffusion_generate`, dynamically dispatching to the current `mlx_vlm.generate.diffusion.stream_diffusion_generate` so stop-race / max-tokens-capture tests keep working regardless of backend |

Diffusion + routing + alias-contract slice: **940 / 940 pass**.
Full repo non-slow sweep: **4935 passed, 0 new regressions** (2 pre-existing `test_event_loop.py` failures require a live `:8000` server, unchanged on `main`).

Run the parity slice with:
```bash
PYTEST_ADDOPTS="-m slow" python3.12 -m pytest \
  tests/test_diffusion_gemma_parity.py --run-slow
```

## Test plan

- [x] `python3.12 -m pytest tests/test_diffusion_loop.py tests/test_diffusion_lane_routing.py tests/test_diffusion_engine.py tests/test_alias_diffusion_knobs.py tests/test_aliases_contract.py tests/test_modality_field.py tests/test_no_out_of_band_routing.py tests/test_no_mllm_flag.py tests/test_model_aliases.py -q` → 940 passed
- [x] Full repo non-slow sweep → 4935 passed, 0 new failures
- [x] `python3.12 -m ruff check vllm_mlx/{model_aliases,runtime/diffusion_lane,runtime/diffusion_loop}.py tests/test_diffusion_*.py tests/test_alias_diffusion_knobs.py` → all checks passed
- [x] Slow parity suite → 4 / 4 pass on M3 Ultra at ~127 tok/s decode
- [x] CJK scan over all 9 touched files → clean (English-only)
- [x] Codex review rounds — round 1 (5 BLOCKING resolved) + round 2 (3 real P1 resolved, 1 false alarm verified; see commits 09be934..02a1697)
- [x] `pr_validate <PR#>` → r10 codex PASS (0 BLOCKING + 3 NIT, NIT 3 fixed); r11 full_unit + stress_e2e_bench PASS
- [x] `make check` (inference-touching gate) → r11 full_unit (4955 passed) covers SOP coverage

## Risk

- Backend default flips from `mlx-vlm` (implicit) → `rapid` (explicit). Routing test pins this; emergency rollback is editing the alias.
- `_run_mlxvlm_generator` consumer side reads via `getattr(..., default)` — `RapidDiffusionResult` mirrors every field the consumer touches and is frozen so the contract is enforceable.
- Existing `test_diffusion_engine.py` tests now hit the rapid path on the FakeModel mock; the dynamic-dispatch stub keeps them transparent.

### DIFF
--- a/tests/test_alias_diffusion_knobs.py
+++ b/tests/test_alias_diffusion_knobs.py
@@ -1,0 +1,184 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Schema tests for AliasProfile diffusion knobs.
+
+Three knobs added with the in-house diffusion loop (see
+``runtime/diffusion_loop.py``):
+
+- ``diffusion_backend``: ``"rapid"`` (default, in-house) | ``"mlx-vlm"``
+  (escape hatch). Routing decision — typo silently swaps generation
+  loops, so the loader must reject unknown values.
+- ``diffusion_fixed_steps`` (default 8): number of denoising passes per
+  canvas in the rapid backend. Must be a positive JSON int — silently
+  truncating ``8.5`` would hide a hand-edit typo.
+- ``diffusion_sc_every`` (default 1): self-conditioning matmul cadence.
+  Empirical: only ``1`` keeps quality on DiffusionGemma 26B-A4B 4-bit;
+  ``>=2`` collapses output to ``"the the the…"`` (see quality eval
+  ``research/diffusion-gemma/quality/eval-20260611-1001.md``). Kept as
+  a knob for future diffusion families.
+
+These tests pin the contract so a future schema edit can't silently
+flip routing.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from vllm_mlx.model_aliases import (
+    AliasProfile,
+    _coerce,
+    list_profiles,
+    resolve_profile,
+)
+
+
+def test_defaults_match_diffusion_gemma_empirical_optimum():
+    """Code-level defaults must match what we found is optimal for the
+    only shipped diffusion alias today. If a future change moves the
+    default away from these, the diff is review-visible."""
+    prof = AliasProfile(hf_path="x/y")
+    assert prof.diffusion_backend == "rapid"
+    assert prof.diffusion_fixed_steps == 8
+    assert prof.diffusion_sc_every == 1
+
+
+def test_text_alias_inherits_defaults_silently():
+    """An existing text-modality alias (no diffusion_* fields in JSON)
+    must keep loading — the new fields default to no-ops for AR."""
+    qwen = resolve_profile("qwen3.5-4b-4bit")
+    assert qwen is not None
+    assert qwen.modality == "text"
+    assert qwen.diffusion_backend == "rapid"
+    assert qwen.diffusion_fixed_steps == 8
+    assert qwen.diffusion_sc_every == 1
+
+
+def test_diffusion_gemma_alias_picks_up_defaults():
+    """The shipped DiffusionGemma alias today does not set the fields
+    explicitly — they must come from the AliasProfile defaults."""
+    diff = resolve_profile("diffusion-gemma-26b")
+    assert diff is not None
+    assert diff.modality == "text-diffusion"
+    assert diff.diffusion_backend == "rapid"
+    assert diff.diffusion_fixed_steps == 8
+    assert diff.diffusion_sc_every == 1
+
+
+# =============================================================================
+# Validation — typos must fail loud at load
+# =============================================================================
+
+
+def test_backend_typo_fails_loud():
+    """Routing decision — typo silently swaps loops. Fail at load."""
+    with pytest.raises(ValueError, match="diffusion_backend must be one of"):
+        _coerce("x", {
+            "hf_path": "a/b",
+            "modality": "text-diffusion",
+            "supports_spec_decode": False,
+            "diffusion_backend": "rappid",
+        })
+
+
+def test_fixed_steps_must_be_positive_int():
+    with pytest.raises(ValueError, match="diffusion_fixed_steps must be a JSON integer"):
+        _coerce("x", {
+            "hf_path": "a/b",
+            "modality": "text-diffusion",
+            "supports_spec_decode": False,
+            "diffusion_fixed_steps": 8.5,
+        })
+    with pytest.raises(ValueError, match="diffusion_fixed_steps must be >= 1"):
+        _coerce("x", {
+            "hf_path": "a/b",
+            "modality": "text-diffusion",
+            "supports_spec_decode": False,
+            "diffusion_fixed_steps": 0,
+        })
+
+
+def test_sc_every_must_be_positive_int():
+    with pytest.raises(ValueError, match="diffusion_sc_every must be a JSON integer"):
+        _coerce("x", {
+            "hf_path": "a/b",
+            "modality": "text-diffusion",
+            "supports_spec_decode": False,
+            "diffusion_sc_every": True,  # bool subclass of int — rejected
+        })
+    with pytest.raises(ValueError, match="diffusion_sc_every must be >= 1"):
+        _coerce("x", {
+            "hf_path": "a/b",
+            "modality": "text-diffusion",
+            "supports_spec_decode": False,
+            "diffusion_sc_every": -1,
+        })
+
+
+# =============================================================================
+# Modality gate — diffusion_* on a text alias is a configuration error
+# =============================================================================
+
+
+def test_diffusion_field_on_text_alias_fails_loud():
+    """Setting any diffusion knob on a text-modality alias is a typo
+    (the AR runtime ignores it). Mirror the DFlash/spec-decode gate
+    pattern — fail at load, not at request time."""
+    for field, val in [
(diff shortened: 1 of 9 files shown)


raullenchai:
## Round 2 — vendored full upstream loop + adaptive stop

Codex round 1 + pr_validate + live bench surfaced 3 + 4 = 7 BLOCKERs/findings against the v1 vendor (commit `09be934`). Pushed `2ee19ff` to address them.

**Root cause of the v1 perf cliff** — the bench subagent's "fixed-step cost" hypothesis was actually wrong. Probe with explicit step counts showed:

| | mlx-vlm | rapid v1 (fixed_steps=8) | per-step |
|--|---|---|---|
| longform-coherence | 11 steps / 77 tok/s | 8 steps / 124 tok/s | rapid ½ the cost |
| code-python | 10 steps / 93 tok/s | 8 steps / 82 tok/s | matched |
| structured-json | 6 steps / 42 tok/s (valid JSON) | 8 steps / 30 tok/s (**broken JSON**) | matched |

The "rapid is slower" claim came from rapid doing all 8 fixed steps when upstream early-stopped at 4-6 via `_diffusion_stable_and_confident`. AND rapid was producing broken JSON on the structured-json prompt (`{\n  title": ...` missing opening quote) — the keyword-only parity test missed it because `"title"` still appeared later in the string.

**Root cause of the correctness drift**: v1 overwrote `canvas = argmax(logits)` every step; upstream progressively REVEALS tokens via an acceptance mask stacked on top of a draft canvas. Re-sampling locked positions from intermediate noisier distributions silently corrupted deterministic outputs.

### Fix shape

Vendored (with attribution to mlx-vlm 0.6.3 Apache 2.0) the full upstream algorithm:

- 12 helpers (`_initialize_canvas` through `_stable_and_confident`) — direct ports, byte-comparable math.
- Main loop reorganized to mirror upstream's downward-counting step loop with both entropy-bound + confidence-threshold samplers.
- Adaptive early-stop via `_stable_and_confident`.
- Per-canvas KV-cache re-prefill of previous canvas tokens (v1 skipped this; subsequent canvases hallucinated).

### `diffusion_fixed_steps` default flipped to `None`

Adaptive stop is now the default — matches mlx-vlm quality and speed across all 3 prompts (within ±15%). Setting `diffusion_fixed_steps=8` is now operator opt-in for the ~1.7× longform-coherence perf win at the cost of some short-prompt quality drift.

### Verification (M3 Ultra, temperature=0, no GPU contention)

| Prompt | mlx-vlm | rapid (adaptive default) | rapid (fixed_steps=8 opt-in) |
|--------|---------|--------------------------|------------------------------|
| longform-coherence | 99.2 tok/s | 90.0 tok/s (0.91x) | **136.6 tok/s (1.76x faster)** |
| code-python | 145.0 tok/s | 125.1 tok/s (0.86x) | 110.3 tok/s (1.09x faster) |
| structured-json | 35.6 tok/s ✅ valid | 39.9 tok/s (1.12x) ✅ valid | 36.7 tok/s ✅ valid |

Quality byte-comparable (both backends produce valid JSON, working quicksort, coherent long-form). JSON correctness drift FIXED — verified by new `json.loads()` gate in `test_diffusion_gemma_parity.py`.

### Codex r1 BLOCKER status

1. ✅ **P1 final-step argmax**: subsumed by full loop vendor (upstream has `current_canvas = argmax_canvas` after the for-loop).
2. ✅ **P0 prefill_step_size**: honored via vendored `_diffusion_prefill_cache` equivalent.
3. ✅ **P1 EOS union**: 4 surfaces unioned from both tokenizer + processor.

### pr_validate r1 BLOCKER status

1. ✅ **P1 attention_mask threaded** through the encoder + decoder mask construction.
2. ⚠️ **AliasProfile field ordering** — false positive on inspection: new fields ARE at the tail (model_aliases.py L155-167); codex misread.
3. ✅ **Tests false-positive on guard-only**: rewritten to exercise real paths via `prefill_step_size=0` raise, and the `temperature>0` paths now have unit + parity coverage.

### Next steps

- Re-running `pr_validate 555` against `2ee19ff` to confirm zero BLOCKING.
- `make check` (inference-touching gate) — pending GPU clearance.
- 947 / 947 unit tests pass.

PR #560, opened 2026-06-12
fix: use hashed token as rate-limit client_id instead of raw header

### LINKED ISSUE #192: HIGH: Rate limiter conflates all authenticated users into a single bucket via Authorization header
## Description

The rate limiter in `middleware/auth.py:69-71` uses the raw `Authorization` header value as the client identifier:

```python
client_id = request.headers.get(
    "Authorization", request.client.host if request.client else "unknown"
)
```

When API key auth is enabled, ALL authenticated requests carry the same `Authorization: Bearer <key>` header, so ALL users share a single rate-limit bucket. If one user exhausts the 60 req/min limit, every other authenticated user is blocked too.

For unauthenticated requests behind NAT, `request.client.host` groups many users at the same IP into a single bucket.

## Impact

- **Accidental DoS for all users**: A single user (or poller) hitting the rate limit blocks everyone sharing the same API key or behind the same NAT
- **Defeats rate limiting purpose**: The feature as implemented cannot actually protect against per-user abuse when auth is enabled

## Reproduction

```bash
# Start with rate limiting (5 req/min for fast reproduction):
rapid-mlx serve <model> --rate-limit 5 --api-key test123

# Send 5 requests as user1 — they work
for i in $(seq 1 5); do
  curl -s -o /dev/null -w "%{http_code}\n" http://localhost:8000/v1/chat/completions \
    -H "Authorization: Bearer test123" \
    -d '{"messages":[{"role":"user","content":"Hi"}],"max_tokens":5}'
done
# → all 200

# 6th request from user2 (same API key) gets 429:
curl -s -o /dev/null -w "%{http_code}\n" http://localhost:8000/v1/chat/completions \
  -H "Authorization: Bearer test123" \
  -d '{"messages":[{"role":"user","content":"Hi"}],"max_tokens":5}'
# → 429 Rate limited! Even though user2 hasn't made any requests.
```

## Fix

For authenticated requests, use a hash of the API key or a derived user identity as the client_id. For unauthenticated requests, use `request.client.host` with a `/24` subnet mask:

```python
client_id = (
    hashlib.sha256(credentials.credentials.encode()).hexdigest()[:16]
    if credentials
    else (request.client.host.rsplit(".", 1)[0] if request.client else "unknown")
)
```

## File

`vllm_mlx/middleware/auth.py:69-71`

Oxygen56:
## Summary

- Rate limiter previously used the raw Authorization header as client_id
- Fixed by deriving a unique client identifier via HMAC-SHA256 of the token
- Added proper IPv4 /24 and IPv6 /64 subnet grouping for unauthenticated traffic
- This provides forward-compat for multi-tenant tokens, raw-secret hygiene, and NAT-friendly IP grouping

Fixes #192

## Changes

- `_bucket_id()`: HMAC-SHA256 with per-process random key for token hashing
- `_subnet_bucket()`: ipaddress-based subnet masking (IPv4 /24, IPv6 /64)
- 3 new tests pinning the fix behavior

## Test Plan

- [x] `test_rate_limit_distinguishes_clients_by_hashed_bearer` — two tokens → different buckets, raw token not in bucket key
- [x] `test_rate_limit_groups_unauth_clients_by_subnet` — same /24 → same bucket
- [x] `test_rate_limit_same_token_via_bearer_and_x_api_key_share_bucket` — Bearer and x-api-key parity

### DIFF
--- a/tests/test_config_and_middleware.py
+++ b/tests/test_config_and_middleware.py
@@ -291,8 +291,7 @@ class TestRateLimitClientId:
         client_id_b = _rate_limit_client_id(Request(scope_b))
 
         assert client_id_a == client_id_b, (
-            f"Same /24 must share a bucket, "
-            f"got {client_id_a!r} != {client_id_b!r}"
+            f"Same /24 must share a bucket, got {client_id_a!r} != {client_id_b!r}"
         )
 
     def test_rate_limit_same_token_via_bearer_and_x_api_key_share_bucket(self):
--- a/tests/test_config_and_middleware.py
+++ b/tests/test_config_and_middleware.py
@@ -231,6 +231,100 @@ class TestCheckRateLimit:
 
 
 # ======================================================================
+# _rate_limit_client_id / _anthropic_rate_limit_client_id
+# ======================================================================
+
+
+class TestRateLimitClientId:
+    def test_rate_limit_distinguishes_clients_by_hashed_bearer(self):
+        """Two distinct bearer tokens get separate buckets.
+
+        Pins #192: raw header values used to conflate everyone into one bucket.
+        """
+        from starlette.requests import Request
+
+        from vllm_mlx.middleware.auth import _rate_limit_client_id
+
+        scope_1 = {
+            "type": "http",
+            "headers": [(b"authorization", b"Bearer sk-token-alpha")],
+            "client": ("192.0.2.1", 12345),
+        }
+        scope_2 = {
+            "type": "http",
+            "headers": [(b"authorization", b"Bearer sk-token-beta")],
+            "client": ("192.0.2.1", 12345),
+        }
+
+        client_id_1 = _rate_limit_client_id(Request(scope_1))
+        client_id_2 = _rate_limit_client_id(Request(scope_2))
+
+        assert client_id_1 != client_id_2, (
+            f"Different tokens must produce different client IDs, "
+            f"got {client_id_1!r} == {client_id_2!r}"
+        )
+        assert "sk-token-alpha" not in client_id_1, (
+            "Raw token must not appear in client_id"
+        )
+        assert "sk-token-beta" not in client_id_2, (
+            "Raw token must not appear in client_id"
+        )
+
+    def test_rate_limit_groups_unauth_clients_by_subnet(self):
+        """IPv4 clients in the same /24 share a bucket."""
+        from starlette.requests import Request
+
+        from vllm_mlx.middleware.auth import _rate_limit_client_id
+
+        scope_a = {
+            "type": "http",
+            "headers": [],
+            "client": ("192.0.2.10", 12345),
+        }
+        scope_b = {
+            "type": "http",
+            "headers": [],
+            "client": ("192.0.2.99", 54321),
+        }
+
+        client_id_a = _rate_limit_client_id(Request(scope_a))
+        client_id_b = _rate_limit_client_id(Request(scope_b))
+
+        assert client_id_a == client_id_b, (
+            f"Same /24 must share a bucket, "
+            f"got {client_id_a!r} != {client_id_b!r}"
+        )
+
+    def test_rate_limit_same_token_via_bearer_and_x_api_key_share_bucket(self):
+        """Same key value via Bearer and x-api-key maps to same bucket."""
+        from starlette.requests import Request
+
+        from vllm_mlx.middleware.auth import (
+            _anthropic_rate_limit_client_id,
+            _rate_limit_client_id,
+        )
+
+        bearer_scope = {
+            "type": "http",
+            "headers": [(b"authorization", b"Bearer sk-abc")],
+            "client": ("192.0.2.1", 12345),
+        }
+        x_api_key_scope = {
+            "type": "http",
+            "headers": [(b"x-api-key", b"sk-abc")],
+            "client": ("192.0.2.1", 12345),
+        }
+
+        bearer_id = _rate_limit_client_id(Request(bearer_scope))
+        x_api_id = _anthropic_rate_limit_client_id(Request(x_api_key_scope))
+
+        assert bearer_id == x_api_id, (
+            f"Same key via Bearer and x-api-key must produce same bucket, "
+            f"got {bearer_id!r} != {x_api_id!r}"
+        )
+
+
+# ======================================================================
 # configure_cors — Fetch-spec-compliant defaults (#190)
 # ======================================================================
 
--- a/vllm_mlx/middleware/auth.py
+++ b/vllm_mlx/middleware/auth.py
@@ -2,6 +2,8 @@
 """Authentication and rate limiting middleware."""
 
 import hashlib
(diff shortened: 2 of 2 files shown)


raullenchai (changes requested):
Thanks for picking up #192. The supply-chain shape is clean (added `hashlib` only — stdlib), and the existing rate-limit tests at `tests/test_anthropic_route_auth.py:286-355` still pass because both paths now hash the same input to the same digest. Two blockers and two nits before this can land.

## Blockers

### 1. No tests for the new behavior

Issue #192 ships with a clean curl reproduction (5 successful requests under user1, 6th request from user2 triggers a 429 even with a distinct identity). The fix should pin that reproduction so the regression can't silently come back. The PR's "Manual verification" checklist isn't enforceable in CI.

Concretely I'd want:

```python
# tests/test_config_and_middleware.py — add to the rate-limit section

def test_rate_limit_distinguishes_clients_by_hashed_bearer():
    """Two distinct bearer tokens get separate buckets.
    Pins #192: raw header values used to conflate everyone into one bucket."""
    from vllm_mlx.middleware.auth import _rate_limit_client_id
    # Build two requests with different bearer tokens, assert
    # _rate_limit_client_id() returns different values AND that
    # neither value contains the raw token (proves hashing).
    ...

def test_rate_limit_groups_unauth_clients_by_subnet():
    """IPv4 clients in the same /24 share a bucket."""
    # Hit _rate_limit_client_id with two requests from
    # 192.0.2.10 and 192.0.2.99 (no Authorization) — assert same client_id.
    ...

def test_rate_limit_same_token_via_bearer_and_x_api_key_share_bucket():
    """Already pinned for the /v1/messages path at test_anthropic_route_auth.py:312.
    Add the same assertion for _rate_limit_client_id() in isolation so the
    invariant isn't only enforced via the high-level Anthropic test."""
    ...
```

Three small tests; should fit in ~40 LOC.

### 2. `/24` subnet mask via `rsplit(".", 1)[0]` silently no-ops on IPv6

```python
if request.client and request.client.host:
    return request.client.host.rsplit(".", 1)[0]
```

For an IPv6 client (e.g. `2001:db8::1`), `rsplit(".", 1)[0]` finds no `.` and returns the whole address unchanged. Result: IPv6 hosts get per-`/128` buckets, not the intended NAT-friendly grouping. Worse for IPv4-mapped IPv6 like `::ffff:192.0.2.1` — you get `"::ffff:192.0.2"`, which is structurally weird and groups a different prefix than the IPv4-path comment implies.

Fix with `ipaddress` from stdlib:

```python
import ipaddress

if request.client and request.client.host:
    try:
        addr = ipaddress.ip_address(request.client.host)
        if isinstance(addr, ipaddress.IPv4Address):
            network = ipaddress.ip_network(f"{addr}/24", strict=False)
        else:
            network = ipaddress.ip_network(f"{addr}/64", strict=False)
        return str(network.network_address)
    except ValueError:
        return request.client.host  # malformed — bucket as-is
return "unknown"
```

`/24` for IPv4 (matches issue spec); `/64` for IPv6 is the conventional ISP-allocation boundary and analogous in coarseness. Move the helper to module scope so it isn't reconstructed per request.

## Nits

### 3. PR description overstates what this fixes

The summary says:

> *"causing all authenticated users to share a single rate-limit bucket when API key auth is enabled" → "This ensures per-user (per-token) rate limiting as intended"*

In rapid-mlx's current `--api-key X` mode, every authenticated client uses the same shared `X`. Hashing `X` produces the same digest for everyone, so they all still share one bucket — the "per-user rate limiting as intended" claim doesn't materialize in the typical deployment. The real wins of this PR are:

- **Token hygiene** — raw secret no longer used as in-memory dict key (small but real)
- **Forward compat** — when we add multi-tenant tokens later, the limiter is already shaped right
- **/24 grouping for unauth traffic** — helps NAT'd users (once the IPv6 bug above is fixed)

Could you tone the PR description down to match? "Forward-compat for multi-tenant tokens + raw-secret hygiene + NAT-friendly IP grouping" is more accurate than "ensures per-user rate limiting."

### 4. Unsalted SHA256 of the token

Not a strict blocker (issue #192's recommended fix is also unsalted), but for defense-in-depth in a future multi-tenant world, an HMAC keyed on a per-process random salt would mean rate-limiter internals leaking the digest doesn't help an attacker confirm token guesses:

```python
_RATE_LIMIT_HMAC_KEY = secrets.token_bytes(32)  # module-level, regenerated per process

def _bucket_for(raw: str) -> str:
    return hmac.new(_RATE_LIMIT_HMAC_KEY, raw.encode(), hashlib.sha256).hexdigest()[:16]
```

Worth doing but I won't block on it.

## Summary

- **Step 0 (does this solve a real product problem):** ⚠ Partial. Token hygiene is universally good and the structure is right for future multi-tenant support; the headline "per-user rate limiting" claim doesn't apply to current single-tenant `--api-key` deployments. Worth landing once Blockers fixed.
- **Supply-chain audit:** ✅ Clean — single stdlib import (`hashlib`).
- **Tests:** ❌ None added — Blocker 1 above.
- **Action:** please address Blockers 1 + 2, retitle the PR summary per Nit 3, and I'll re-review.

Thanks again — this is one of the older HIGH-severity issues and great to see it picked up.

raullenchai:
Merging this with maintainer override on the pr_validate `targeted_tests` step. Recording the reasoning for the audit trail.

**pr_validate verdict**: `DO NOT MERGE` — `targeted_tests` flagged 3 "regressions":
- `TestRateLimitClientId::test_rate_limit_distinguishes_clients_by_hashed_bearer`
- `TestRateLimitClientId::test_rate_limit_groups_unauth_clients_by_subnet`
- `TestRateLimitClientId::test_rate_limit_same_token_via_bearer_and_x_api_key_share_bucket`

**Why this is a false positive**: all 3 fail with `ModuleNotFoundError: No module named 'mlx'`, which is a Linux-runner limitation (mlx is Apple Silicon-only). The same scorecard lists 14 sibling tests in this exact file (`tests/test_config_and_middleware.py::TestRateLimiter`, `TestServerConfig`, `TestCheckRateLimit`, etc.) that fail with the identical error and are classified as pre-existing. The 3 new tests are pr_validate's first sighting on the PR branch but they share the same env limitation — they just don't have a baseline match because they're newly added.

The `test-apple-silicon` job (which actually has mlx) is **passing**, meaning these tests do exercise the code correctly on the platform they're meant to run on.

**Followup (not blocking this merge)**: pr_validate's `targeted_tests` step should reclassify "test fails on PR with the same `ModuleNotFoundError` as ≥N pre-existing failures in the same file" as `env_limitation`, not `regression`. Worth a small PR after this lands.

Thanks @Oxygen56 — the rate-limiter fix itself is correctly scoped (token hygiene + NAT-friendly /24/64 grouping + multi-tenant forward-compat), the HMAC + per-process random key is a nice defense-in-depth touch, and the tests are exactly the regression-pinning shape we wanted. Apologies for the bumpy CI ride.

PR #562, opened 2026-06-12
fix(aliases): disable tool_call_parser for Gemma 3 / 3n family

raullenchai:
## Summary

Gemma 3 / 3n instruction models were not trained for function calling. All four aliases inherited `tool_call_parser=\"hermes\"` from the README's \"all Gemma → hermes\" shortcut, which is wrong — hermes is the wire format for Qwen3 / Hermes / Bonsai families.

**User-visible failure** (reported via rapid-desktop chat): asking *\"palo alto天气如何\"* with a weather tool defined on `gemma-3n-e4b-4bit` made the model:
1. see tools embedded in the prompt (via `apply_chat_template`)
2. fail to emit hermes XML markup (never trained for it)
3. hallucinate about Palo Alto Networks firewalls — domain knowledge, no tool intent
4. surface the hallucination as plain content because hermes parser found no markers

## Fix

Set `tool_call_parser: null` on:
- `gemma-3n-e4b-4bit` (user-reported)
- `gemma3-1b-4bit`
- `gemma3-12b-4bit`
- `gemma3-27b-4bit`

With null:
- `routes/chat.py:595` skips the tool-use system suffix injection
- `routes/chat.py:851` rejects forced `tool_choice` (`\"required\"` / named) upfront with a clean 422 instead of running a full generation then failing post-parse
- Downstream chat surfaces (BCG, rapid-desktop) get an honest \"this model can't call tools\" surface

Gemma **4** is unaffected — it ships `tool_call_parser=\"gemma4\"` because it WAS trained for tools with the `call:func{...}` wire format.

## Regression guard

New contract test `test_gemma3_family_disables_tool_call_parser` in `tests/test_aliases_contract.py` pins all four aliases at None. Inverse-style pin matches the existing `test_audit_batch_bonsai_tool_call_parser_wired` (which guarantees bonsai STAYS at hermes).

## Test plan

- [x] `pytest tests/test_aliases_contract.py tests/test_model_aliases.py` — 780 pass
- [x] `pytest tests/test_model_auto_config.py tests/test_server_load_model_order.py` — 76 pass (hermes-parser sites unaffected)
- [ ] E2E smoke: `rapid-mlx serve gemma-3n-e4b-4bit` + chat with weather tool — verify plain content response, no phantom tool_call

### DIFF
--- a/tests/test_aliases_contract.py
+++ b/tests/test_aliases_contract.py
@@ -507,6 +507,43 @@ def test_audit_batch_bonsai_tool_call_parser_wired() -> None:
         )
 
 
+def test_gemma3_family_disables_tool_call_parser() -> None:
+    """Gemma 3 / 3n base instruction models were NOT trained for tool
+    calling — their chat templates don't emit ``<tool_call>...`` blocks
+    and the model has no notion of structured function invocation.
+
+    Pre-fix: every Gemma 3 alias inherited ``tool_call_parser="hermes"``
+    from the README's "all Gemma → hermes" shortcut. When a chat surface
+    passed ``tools=[...]``, the model:
+      1. saw tools embedded in the prompt (via ``apply_chat_template``)
+      2. couldn't emit hermes XML markup (never trained for it)
+      3. fell back to plain-text content, often hallucinating about the
+         tool's subject domain (user-reported: "palo alto天气如何" →
+         monologue about "Palo Alto Networks firewall" instead of
+         firing the weather tool)
+
+    Gemma **4** is different — explicitly trained for tools with the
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing — see vnsh.dev/v/u4F7VfBibMqT diagnosis. The user-reported bug (gemma-3n-e4b-4bit returning nonsense to weather queries) is NOT a tool_call_parser issue. The model itself is over-aligned and refuses weather/real-world queries with a self-contradictory safety preamble ("DO NOT ask for information" misread as "do not provide"). Parser=null vs hermes makes zero observable difference here. Switching to a follow-up PR that drops the alias entirely — the model is unsuitable as a chat target and offering it in the catalog wastes user download budget on a model they can't actually use.

PR #572, opened 2026-06-14
fix(reasoning/qwen3): route bare-text 'thinking process:' preamble to reasoning_content (#570)

### LINKED ISSUE #570: qwen3.6-35b-a3b 4bit: 'Here's a thinking process:' leaks into content body on multi-turn
## Severity
P1 — directly hits end-user UX in any OpenAI-compatible client. The user sees the model's internal monologue as the assistant message body.

## Repro

12-turn road-trip script against `qwen3.6-35b-4bit` on `rapid-mlx serve` (v0.6.81, port 8001).

After turn 1, every turn 1–11 returned `content` that begins with literal:

```
Here's a thinking process:
1.  **Analyze User Request:**
   - **Route:** ...
```

`reasoning_content` is empty (`""`). The thinking is emitted as raw content without `<think>` delimiters, so the hermes `reasoning_parser` cannot capture it.

## Minimal curl

```bash
curl -X POST http://127.0.0.1:8001/v1/chat/completions \
  -d '{
    "model":"qwen3.6-35b-4bit",
    "messages":[
      {"role":"user","content":"Plan a road trip Seattle to San Diego in October. 7 days, one stop per day."},
      {"role":"assistant","content":"OK..."},
      {"role":"user","content":"Which stop has the best food?"}
    ],
    "max_tokens":350
  }'
```

Reproduces `"Here's a thinking process: 1.  **Analyze User Request:**..."` prefix in `content`.

## Single-shot DOES NOT reproduce

`"Plan 3-day trip to Tokyo. Be concise."` (no prior turns) returned clean content. The thinking-prefix leak is tied to multi-turn priming.

## Suspected root causes (verify, don't assume)

1. Qwen3.6 chat template defaults to `enable_thinking=true` but for this MLX 4bit conversion the model emits thinking as plain text rather than wrapping in `<think>…</think>`.
2. `mlx-community/Qwen3.6-35B-A3B-4bit` tokenizer's `apply_chat_template` may not be injecting the `<think>` start token on continued conversations (the prior assistant turn primes the model to skip the open tag).
3. `aliases.json` has `reasoning_parser=hermes` for this alias; the hermes parser handles `<think>...</think>` but not the bare-prefix variant `"Here's a thinking process:"`.

## Suggested investigation order

1. Capture the raw `apply_chat_template` output for this alias on a multi-turn conversation (turn-3 prompt with full history) and inspect whether the assistant turn opens with `<think>` or not.
2. If the template suppresses `<think>` on continued turns, decide: (a) patch the template to always inject the opener, or (b) extend hermes parser to recognize the bare-prefix variant.
3. Add an eval scenario to `evals/reasoning/` that asserts `reasoning_content` is populated and `content` does NOT contain "Here's a thinking process" on a 5-turn conversation.

## Cross-references
- Found 2026-06-14 during multi-model `/v1/chat/completions` probe sweep.
- Likely related: hybrid Qwen 3.5/3.6 sampler-loop B=1 gap (separate perf issue, not relevant here).

## What is NOT this issue

- Single-shot prompts: clean.
- Other parsers (minimax, glm47): not tested on this model — alias pins hermes.
- The leak is in the `content` field returned by `/v1/chat/completions`, so all clients see it identically.

raullenchai:
## Summary

Closes #570.

`Qwen3ReasoningParser` now recognises bare-text "thinking process" preambles at the start of an assistant turn and routes them to `reasoning_content` instead of letting them leak into the user-facing `content` field.

## Root cause

Qwen3 chat templates inject `<think>\n` after the assistant generation marker when `enable_thinking=True`, putting the model in **implicit-think mode** — the model is supposed to emit its chain-of-thought followed by `</think>` and then the user-facing answer.

In practice the model occasionally restates the channel boundary inline with a bare-text preamble like:

```
Here's a thinking process:

1.  **Analyze User Input:**
   - **Route:** Seattle to San Diego
2.  **Evaluate Each Option:**
   - Portland, OR: World-renowned food scene.
```

When the request also runs out of `max_tokens` before the model reaches `</think>`, neither tag appears anywhere in the model output. The previous `Qwen3ReasoningParser.extract_reasoning` "no end token, no start token" branch then returned `(None, model_output)` — routing the whole chain-of-thought into the user-facing `content` field while `reasoning_content` stayed empty. Any OpenAI-compatible client consuming `/v1/chat/completions` saw raw reasoning leak into the answer.

The bug reproduces both on multi-turn and single-turn prompts; the trigger is "model emits bare-text thinking preamble AND output is truncated before `</think>`", which becomes more likely as conversation length and `max_tokens` interact.

## Fix

Extend `Qwen3ReasoningParser` to recognise bare-text "thinking process" preambles and route them to `reasoning_content` with `cleaned_text` cleared. Applies to both non-streaming `extract_reasoning` and the `finalize_streaming` correction path — matching the existing streaming heuristic in `BaseThinkingReasoningParser` that already routes pre-tag text to reasoning while `</think>` has not yet been seen.

Recognised preambles (matched at `^\s*`, case-insensitive):

- `Here's a thinking process:`, `Here is my reasoning:`, `Here is the chain-of-thought:`, ...
- `Let me think`, `Let me analyze`, `Let me reason`, `Let me work through`, `Let me break this down`
- `Thinking step by step:`, `Thinking out loud:`, `Thinking carefully:`, `Step by step:`
- `I'll think through this`, `I will analyze`, `I need to analyze`, `I should break this down`
- `Analyzing the user's request`, `My thought process:`, `My reasoning process:`

The regex anchors to `^\s*` so a normal answer that merely mentions "let me think" mid-sentence is NOT reclassified. A dedicated test pins that guard.

## Why server-side is the right layer

Two alternative paths were considered:

1. **Patch the chat template** to keep `<think>` open across continuation turns. The chat template is owned by the upstream HF tokenizer config (Qwen3.6-A3B), so we'd have to maintain a local override; any HF refresh would silently revert the fix. Not blocked on this — captured for future upstream contribution.
2. **Inject `<think>` server-side at the prompt boundary.** Would mask the symptom but bare-text preambles also occur in single-turn prompts (the trigger is "truncated before `</think>`", not "continuation turn"). Server-side parser remains the right place to be defensive.

## Regression coverage

- `tests/test_reasoning_parsers.py::TestQwen3` — 7 new cases pinning the bare-text fallback, the false-positive guard for normal answers that mention "think" mid-sentence, the explicit-tag passthrough, and the streaming finalizer behavior.
- `tests/test_chat_route_tool_tag_leak.py::TestBareThinkingProcessLeakRegression` — 3 new cases driving the real `_finalize_content_and_reasoning` orchestration used by `/v1/chat/completions`, including the exact text captured from a failing request against `qwen3.6-35b-4bit`.

## Scope

- No version bump (parser fix; release SOP 11-gate walk happens at the next bundled cut).
- Diff limited to one parser file + two test files.
- No alias changes, no template changes, no upstream pin changes.

## Test plan

- [x] `python3.12 -m pytest tests/test_reasoning_parsers.py tests/test_chat_route_tool_tag_leak.py` — 141 passed
- [x] `python3.12 -m pytest tests/ -k "reason or qwen or think or finalize or tool_tag" --ignore=tests/integrations --ignore=tests/test_streaming_simulator.py` — 1003 passed
- [x] `ruff check` + `ruff format --check` on touched files — clean
- [x] Direct verification of the captured failing payload through `_finalize_content_and_reasoning` — `reasoning_content` populates with the chain-of-thought, `content` clears

### DIFF
--- a/.github/workflows/sidecar-build.yml
+++ b/.github/workflows/sidecar-build.yml
@@ -25,6 +25,14 @@ on:
   push:
     tags:
       - 'sidecar-v*'
+  pull_request:
+    # Run on PRs that touch sidecar plumbing so we catch breakage before
+    # tagging a release (codex r1 NIT).
+    paths:
+      - 'scripts/build-sidecar.sh'
+      - 'scripts/sidecar-shim.sh'
+      - 'scripts/sidecar-entitlements.plist'
+      - '.github/workflows/sidecar-build.yml'
   workflow_dispatch:
     inputs:
       tag_suffix:
@@ -95,12 +103,14 @@ jobs:
           fi
 
       - name: Import Developer ID Application certificate
+        id: import_cert
         if: steps.codesign.outputs.available == 'true'
         env:
           CERT_B64: ${{ secrets.APPLE_SIGNING_CERTIFICATE }}
           CERT_PASSWORD: ${{ secrets.APPLE_SIGNING_PASSWORD }}
         run: |
           KEYCHAIN="$RUNNER_TEMP/sidecar.keychain-db"
+          echo "keychain_path=$KEYCHAIN" >> "$GITHUB_OUTPUT"
           KEYCHAIN_PASSWORD="$(openssl rand -hex 32)"
           security create-keychain -p "$KEYCHAIN_PASSWORD" "$KEYCHAIN"
           security set-keychain-settings -lut 21600 "$KEYCHAIN"
@@ -199,3 +209,18 @@ jobs:
               --title "rapid-mlx sidecar $PYVER" \
               --notes-file "$NOTES_FILE"
           fi
+
+      - name: Clean up signing keychain
+        # codex r1 B4: matter most on self-hosted runners (GitHub-hosted
+        # macos-15 runners get nuked anyway, but defense-in-depth + this
+        # workflow may run on self-hosted later). `if: always()` is
+        # required so we still tear down the keychain when codesign or
+        # smoke fail.
+        if: always() && steps.import_cert.outcome != 'skipped'
+        run: |
+          KEYCHAIN="${{ steps.import_cert.outputs.keychain_path }}"
+          if [ -n "$KEYCHAIN" ] && [ -f "$KEYCHAIN" ]; then
+            security delete-keychain "$KEYCHAIN" || true
+          fi
+          # Best-effort: remove cert .p12 if a previous step left it.
+          rm -f "$RUNNER_TEMP/cert.p12" || true
--- a/.github/workflows/sidecar-build.yml
+++ b/.github/workflows/sidecar-build.yml
@@ -144,7 +144,18 @@ jobs:
             --out build/sidecar-stage \
             --skip-codesign
 
-      - name: Upload artifact for downstream jobs
+      - name: Upload artifact for downstream jobs (signed builds only)
+        # codex r2 B5: do NOT upload artifacts on pull_request runs.
+        # The PR-trigger build executes against the PR's branch code,
+        # which on a fork could be malicious. Uploading a downloadable
+        # `rapid-mlx-sidecar.tar.gz` named exactly like the real release
+        # asset is a supply-chain confusion hazard even when unsigned,
+        # because the bundle's entitlements plist ships the JIT +
+        # library-validation-disabled keys that the host Rapid.app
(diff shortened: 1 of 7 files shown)


raullenchai:
Superseded by a clean rebase off main — the original branch carried 8 unrelated sidecar commits from the parent branch. Re-opening with diff scoped to just the parser fix.

PR #580, opened 2026-06-15
fix(mistral): stream [ARGS] tool calls without garbling (#579)

### LINKED ISSUE #579: Mistral streaming tool-call parser garbles Devstral-Small-2 [ARGS] format (non-streaming works)
> Re-filing against the **Bug Report** template fields (the original was auto-flagged for missing template context). All required fields below.

### Rapid-MLX version
`0.7.6`

Unknown — first time using Devstral / the `[ARGS]` format on this setup.

### Hardware
MacBook Pro, Apple **M5 Max, 128 GB**

### macOS version
**26.5.1** (build 25F80)

### Python version
`3.14.5`

### Model
`mlx-community/Devstral-Small-2-24B-Instruct-2512-4bit` (alias `devstral-v2-24b-4bit`)

### Full serve command
```
rapid-mlx serve devstral-v2-24b-4bit --port 8080 --enable-auto-tool-choice --tool-call-parser mistral
```

**Both** — streaming is broken, non-streaming is correct (that contrast is the bug).

**Actual:** With `stream: true`, every tool call is garbled — the `[ARGS]` delimiter token leaks into the stream and name/arguments are split wrong, so the assembled `delta.tool_calls` come out as `name='"}'`, `arguments='[ARGS]{"'`, `id=''`. OpenAI-compatible clients therefore receive empty/invalid arguments (e.g. `arguments: []`) and abort with errors like `Validation failed for tool "read": root: must be object`.

**Expected:** Streaming should yield the same structured tool call the non-streaming path produces: `{"name":"read","arguments":"{\"file_path\":\"/tmp/foo.txt\"}"}`.

This happens with a **single** tool call (not a parallel-tool-calls issue) and with multiple calls.

### Minimal reproduction
Same request, same model — only `stream` differs.

**1) Non-streaming — CORRECT ✅**
```
curl -s http://127.0.0.1:8080/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"devstral-v2-24b-4bit",
  "messages":[{"role":"user","content":"Read the file /tmp/foo.txt using the read tool."}],
  "tools":[{"type":"function","function":{"name":"read","description":"Read a file","parameters":{"type":"object","properties":{"file_path":{"type":"string"}},"required":["file_path"]}}}],
  "tool_choice":"auto"
}'
# -> tool_calls[0].function = {"name":"read","arguments":"{\"file_path\": \"/tmp/foo.txt\"}"}
```

**2) Streaming — BROKEN ❌** (add `"stream": true`, then assemble the `delta.tool_calls` chunks)
```
curl -s -N http://127.0.0.1:8080/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"devstral-v2-24b-4bit",
  "messages":[{"role":"user","content":"Read the file /tmp/foo.txt using the read tool."}],
  "tools":[{"type":"function","function":{"name":"read","description":"Read a file","parameters":{"type":"object","properties":{"file_path":{"type":"string"}},"required":["file_path"]}}}],
  "tool_choice":"auto","stream":true
}'
```
Assembled result:
```
n_calls: 1
  name      = '"}'
  id        = ''
  arguments = '[ARGS]{"'
```

```shell
INFO:vllm_mlx.model_auto_config:Resolved alias profile for 'mlx-community/Devstral-Small-2-24B-Instruct-2512-4bit' → tool_call_parser=hermes, reasoning_parser=None, is_hybrid=False, supports_spec_decode=True, suffix_tier=neutral
INFO:vllm_mlx.server:Native tool format enabled for parser: mistral
  Features: tools: mistral, gc-control
[transformers] The tokenizer you are loading ... with an incorrect regex pattern ... This will lead to incorrect tokenization. You should set the `fix_mistral_regex=True` flag when loading this tokenizer to fix this issue.
```

### Likely cause
`MistralToolParser.extract_tool_calls_streaming` doesn't handle the newer Devstral/Mistral `[TOOL_CALLS]<name>[ARGS]<json>` delimiter — whereas `MistralToolParser.extract_tool_calls` (non-streaming) does. The streaming path appears to assume the classic `[TOOL_CALLS][ {…} ]` array form and mis-tokenizes the `[ARGS]` separator. (There's also the `fix_mistral_regex` tokenizer warning above — possibly related, though non-streaming parsing the identical generation works, which points at the streaming assembler specifically.)

rinaldofesta:
## What does this PR do?

Fixes streaming tool-call extraction for the Mistral parser on the
Devstral / Mistral >= v11 wire format `[TOOL_CALLS]<name>[ARGS]{json}`.

`extract_tool_calls_streaming` parsed each token delta in isolation, so once
the `[ARGS]` separator or the JSON body was split across token boundaries the
output garbled: `[ARGS]` (it opens with `[`) fell into the "arguments
continuation" branch and leaked into the body, while the name was rebuilt from
stray fragments — assembled deltas came out as `name='"}'`,
`arguments='[ARGS]{"'`, `id=''`. Non-streaming (`extract_tool_calls`) was
already correct.

The fix makes the streaming path re-parse the **full accumulated text** with
the non-streaming `extract_tool_calls` (single source of truth) and emit each
tool call exactly once, the moment its arguments are complete JSON. This makes
stream ↔ non-stream parity hold by construction. The now-unused
`_parse_streaming_tool_delta` helper is removed.

## Why is this needed?

Fixes #579. On `stream: true`, every Devstral tool call reaches OpenAI-compatible
clients with empty/invalid arguments (`arguments: []`), so they abort — e.g.
agentic coders fail with `Validation failed for tool "read": root: must be
object`. Devstral tool calling is effectively unusable over streaming (which is
the default for most clients). Non-streaming users are unaffected; this restores
parity.

## AI assistance disclosure

AI-assisted (Claude / Opus). The AI diagnosed the bug from a live repro,
wrote the implementation in `vllm_mlx/tool_parsers/mistral_tool_parser.py`, the
new regression test, the parity fixture, and the updated unit test. I verified:

- Reproduced the original bug live against rapid-mlx 0.7.6 + Devstral
  (`mlx-community/Devstral-Small-2-24B-Instruct-2512-4bit`): non-streaming
  returns a correct `{"name":"read","arguments":"{...}"}`; streaming returns
  `name='"}'` / `arguments='[ARGS]{"'`.
- Confirmed the regression test **fails on the pre-fix parser** (4/5 cases) and
  passes on the fix, so it genuinely guards the behavior.
- Ran the targeted parser/tool suites (see test plan) — all green; reviewed the
  diff line by line and can explain the intent, risk, and behavior of every
  change.

## Test plan

- `tests/parsers/regressions/test_issue_579_mistral_args_streaming.py` (new) —
  drives the **incremental** streaming entry point the way
  `service/postprocessor.py` does (per-delta, `[ARGS]` as its own chunk, JSON
  split across deltas); covers single call, multi call, leading content, and
  stream↔non-stream parity. Verified it fails on the old parser.
- `tests/test_tool_call_streaming_parity.py` — added the mistral
  `[TOOL_CALLS]...[ARGS]{...}` fixture and removed the matching
  `_PARITY_COVERAGE_EXEMPT` TODO entry.
- `tests/test_tool_injection.py::TestMistralArgsStripping` — updated
  `test_args_suffix_stripped_streaming` to the public streaming API (the private
  helper it called was removed); class passes.
- `ruff check` + `ruff format --check` clean on all changed files.
- Ran `pytest tests/test_tool_injection.py tests/parsers/regressions/test_issue_579_mistral_args_streaming.py tests/test_tool_call_streaming_parity.py tests/test_tool_parsers.py tests/test_tool_parser_wire_formats.py tests/test_native_tool_format.py tests/test_tool_parser_coverage.py` → **208 passed**.
- Note: I ran on a borrowed venv without `pytest-asyncio` / `aiohttp` / a loaded
  model, so the full suite's async + `test_tool_call_e2e.py` + bench tests error
  on **`main` too** (verified against a clean `HEAD` worktree) — pre-existing /
  environmental, unrelated to this change. Happy for maintainers to run
  `pr_validate` full/stress on merge.

## Checklist

- [x] Tests pass locally (targeted parser/tool suites — 208 passed; see note re: env-only failures)
- [ ] Self-validated with `scripts.pr_validate.pr_validate` — not run (borrowed env lacks async/bench deps + DeepSeek key; ran the non-negotiable `lint` + targeted tests manually)
- [x] New tests on a critical path (parser) spot-checked to fail when the production line is broken (regression test fails on the pre-fix parser)
- [ ] Updated README/docs if applicable (n/a — no API/behavior surface change beyond the bugfix)

### DIFF
--- a/tests/parsers/regressions/test_issue_579_mistral_args_streaming.py
+++ b/tests/parsers/regressions/test_issue_579_mistral_args_streaming.py
@@ -1,0 +1,149 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression guard for #579 — Mistral ``[ARGS]`` streaming tool-call leak.
+
+Wire format: ``[TOOL_CALLS]<name>[ARGS]{json}`` (Devstral-Small-2 /
+Mistral >= v11). Reported on rapid-mlx 0.7.6, model
+``mlx-community/Devstral-Small-2-24B-Instruct-2512-4bit``, served with
+``--enable-auto-tool-choice --tool-call-parser mistral``.
+
+Symptom: **non-streaming** parsing is correct, but **streaming** garbles
+every tool call. The old ``extract_tool_calls_streaming`` parsed each token
+delta in isolation; the ``[ARGS]`` separator (which starts with ``[``) fell
+into the "arguments continuation" branch and leaked into the body, while the
+name was reconstructed from stray fragments. The assembled stream came out as
+``name='"}'``, ``arguments='[ARGS]{"'``, ``id=''`` — so OpenAI clients
+received ``arguments: []`` and aborted (e.g. agentic coders:
+``Validation failed for tool "read": root: must be object``).
+
+Fix: the streaming path now re-parses the full accumulated text with the
+(working) non-streaming ``extract_tool_calls`` and emits each call once its
+arguments are complete JSON — guaranteeing stream ↔ non-stream parity.
+
+Convention mirrors the other regression guards in this directory: each case
+states the wire format up front and drives the *incremental* streaming entry
+point the way ``service/postprocessor.py`` does (per-delta, with the full
+accumulated text), not just ``finalize()``.
+"""
+
+from __future__ import annotations
+
+import json
+
+import pytest
+
+from vllm_mlx.tool_parsers import ToolParserManager
+
+
+def _assemble_stream(parser, chunks: list[str]) -> list[dict]:
+    """Drive ``extract_tool_calls_streaming`` exactly like the postprocessor.
+
+    Feeds ``chunks`` one at a time with (previous_text, current_text,
+    delta_text) and merges the emitted ``tool_calls`` deltas on ``index``
+    (the OpenAI streaming contract), concatenating ``arguments`` fragments.
+    Returns the assembled tool calls as ``{index, id, name, arguments}``.
+    """
+    parser.reset()
+    merged: dict[int, dict] = {}
+    previous = ""
+    for chunk in chunks:
+        current = previous + chunk
+        result = parser.extract_tool_calls_streaming(previous, current, chunk)
+        previous = current
+        if not result or "tool_calls" not in result:
+            continue
+        for delta in result["tool_calls"]:
+            idx = delta["index"]
+            slot = merged.setdefault(
+                idx, {"index": idx, "id": "", "name": "", "arguments": ""}
+            )
+            if delta.get("id"):
+                slot["id"] = delta["id"]
+            fn = delta.get("function", {})
+            if fn.get("name"):
+                slot["name"] = fn["name"]
+            if fn.get("arguments"):
+                slot["arguments"] += fn["arguments"]
+    return [merged[i] for i in sorted(merged)]
+
+
+def _norm(calls: list[dict]) -> list[tuple[str, dict]]:
+    return [(c["name"], json.loads(c["arguments"])) for c in calls]
+
+
+@pytest.fixture
+def parser():
+    return ToolParserManager.get_tool_parser("mistral")(None)
+
+
+def test_single_args_call_streams_intact(parser):
+    # [ARGS] arrives as its own chunk and the JSON body is split — the exact
+    # boundary the old per-delta parser mangled.
+    chunks = [
+        "[TOOL_CALLS]",
+        "read",
+        "[ARGS]",
+        '{"file_path":',
+        ' "/tmp/foo',
+        '.txt"}',
+    ]
+    assert _norm(_assemble_stream(parser, chunks)) == [
+        ("read", {"file_path": "/tmp/foo.txt"})
+    ]
+
+
+def test_args_token_never_leaks_into_name_or_arguments(parser):
+    calls = _assemble_stream(
+        parser,
+        ["[TOOL_CALLS]", "get_weather", "[ARGS]", '{"city": "Milan"}'],
+    )
+    assert len(calls) == 1
+    assert calls[0]["name"] == "get_weather"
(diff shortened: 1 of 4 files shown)


raullenchai:
Hi @rinaldofesta — thanks for the detailed diagnosis and the AI-assistance disclosure, it's genuinely one of the clearest bug reports + fixes I've seen on this repo.

Closing as superseded: PR #581 landed in parallel and fixed the exact same issue (#579 — Mistral `[ARGS]` separator leak across token boundaries) by replacing `_parse_streaming_tool_delta` with a cumulative-text state machine. The merged commit is `340056f` and it shipped in v0.7.9 yesterday.

Your independent repro on `Devstral-Small-2-24B-Instruct-2512-4bit` is great cross-validation that the bug was real and the fix direction was right. If you have a minute, please re-test against v0.7.9 (`brew upgrade rapid-mlx` or `pip install -U rapid-mlx`) and reopen if the streaming garbling still reproduces in any edge case the merged fix missed — your test fixtures (especially the multi-call + leading-content cases) would be worth porting over as additional regression coverage if so.

Sincere apologies for the parallel-work waste. Thanks again for taking the time to write this up.

rinaldofesta:
Thanks @raullenchai, no apology needed; landing the fix is what counts. I pulled the merged parser (340056f) and ran my five #579 regression cases against it: single call, multi-call, [ARGS] as its own chunk, leading content before the call, and stream vs non-stream parity. All five pass.

On porting fixtures: #581 already covers every case mine did, plus a few I didn't (fused name+args in one delta, string args containing separators, partial-opener-held-as-content), so there is nothing non-redundant to add. The merged suite is thorough. I will reopen if I hit a streaming edge case it misses on real Devstral traffic.

PR #581, opened 2026-06-15
fix(tool_parsers): #579 Mistral streaming [ARGS] separator leak

### LINKED ISSUE #579: Mistral streaming tool-call parser garbles Devstral-Small-2 [ARGS] format (non-streaming works)

raullenchai:
## Summary

- Replaces `MistralToolParser.extract_tool_calls_streaming` with a cumulative-text state machine that correctly handles Devstral's `[ARGS]` separator regardless of how token boundaries split it.
- Holds back partial `[TOOL_CALLS]` prefixes (mirrors hermes `_safe_content_prefix`) so opener bytes never leak as content.
- Supports multi-tool new-format streams (`[TOOL_CALLS]a{}[TOOL_CALLS]b{}`) with per-segment state, including a boundary-split mid-arrival case.
- Adds 13 regression tests in `TestMistralDevstralStreaming` driven from realistic chunk shapes (incl. the exact pathological chunking from #579, partial-opener-held-back, bracket-in-prose-released, both multi-tool variants, flush-held-content at EOS).
- The test assembler now uses key-presence (not truthiness) and exposes `raw_name_deltas` so empty-string `name` clobbers fail loudly via `_assert_no_empty_name_deltas`.

Closes #579.

## Root cause

Pre-fix `_parse_streaming_tool_delta` worked off `delta_text` alone and only handled `[ARGS]` when it landed in the same chunk as `{`. Real Devstral token streams split the separator across deltas, producing two failure modes:

1. **Chunk `[ARGS]` alone** → no `{`, fell to the `text.startswith("[")` branch → emitted as `arguments="[ARGS]"` (the leak the issue reporter observed).
2. **Chunk `[ARGS]{"`** → `name_part = "[ARGS]"`, `.strip()` was truthy, then `.replace("[ARGS]", "").strip()` yielded `""` → emitted `name=""`, which clobbered the previously-streamed real name (the `name='"}'` garbling).

The non-streaming `extract_tool_calls` already does this correctly. The new state machine ports that correctness into streaming.

## Approach

State on the parser:

- `_stream_format` — latched to `"old"` (Mistral v10- `[{...}]` array) or `"new"` (Devstral / v11+ `name[ARGS]?{json}`) on the first non-whitespace body byte.
- `_tool_states[i]` — per-tool `{name_emitted, args_emitted, id}` for multi-tool new-format streams.
- `_stream_old_emitted` — old form is buffered until `]` and emitted whole via `extract_tool_calls`.
- `_STREAMING_SENTINELS = ("[TOOL_CALLS]",)` + `_safe_content_prefix` + `flush_held_content` — partial sentinel prefix-hold both in the pre-opener content phase AND inside the last segment's args (so a still-arriving second-tool boundary doesn't bleed into the first call's args).

If the body is in flight but no boundary character (`[ARGS]` or `{`) has been seen yet, the function buffers silently — this is what stops partial sentinels like `[` (which might be heading toward `[ARGS]`) from being misclassified as args.

## Codex review feedback (round 1) — addressed

Three BLOCKING + one NIT from codex on the first cut have all been addressed in a follow-up commit:

- **BLOCKING-1 partial opener leak** → `_safe_content_prefix` prefix-hold + `flush_held_content` at stream end.
- **BLOCKING-2 multi-tool absorption** → body split on subsequent `[TOOL_CALLS]`, per-segment state list.
- **BLOCKING-3 test assembler swallows empty-string name** → key-presence concat, raw-delta tracking, `_assert_no_empty_name_deltas` helper applied to every regression test.
- **NIT-2 weak `[ARGS] not in` assertion** → tightened to compare parsed JSON equality.

(NIT-1 about `assert tc["id"]` was a false positive — line is already a proper assert statement.)

## Secondary issue (not in this PR)

`aliases.json:618` declares `tool_call_parser=hermes` for `devstral-v2-24b-4bit`, matching the explicit `mistral|devstral → hermes` rule at `model_auto_config.py:221` and a pinned test at `tests/test_model_auto_config.py:70-74`. The bug reporter worked around it by passing `--tool-call-parser mistral` on the CLI; users who rely on the default would hit a different (also-broken) path. Flipping this has broader default-routing implications, so I'll file separately.

## Test plan

- [x] `pytest tests/test_tool_parsers.py::TestMistralDevstralStreaming` — 13/13 pass post-fix.
- [x] `pytest tests/test_tool_parsers.py tests/test_native_tool_format.py tests/test_tool_injection.py tests/test_tool_call_streaming_parity.py tests/test_upstream_regression.py` — 269/269.
- [x] Full unit sweep — 4948 pass; the 2 fails in `test_event_loop.py` need a server on `:8000` and are pre-existing.
- [x] `ruff check` + `ruff format --check` clean.
- [x] PR validation SOP.

### DIFF
--- a/tests/test_tool_injection.py
+++ b/tests/test_tool_injection.py
@@ -235,6 +235,8 @@ class TestMistralArgsStripping:
         feed a single delta that fuses the boundary tokens and assert
         the emitted name is clean.
         """
+        import json
+
         parser = MistralToolParser(tokenizer=None)
         full = '[TOOL_CALLS]get_weather[ARGS]{"location": "Paris"}'
         delta = parser.extract_tool_calls_streaming(
@@ -248,4 +250,10 @@ class TestMistralArgsStripping:
         assert tcs, "expected a tool_calls delta"
         fn = tcs[0]["function"]
         assert fn.get("name") == "get_weather"
-        assert "[ARGS]" not in (fn.get("arguments") or "")
+        # Codex PR #581 NIT-2: don't just assert ``[ARGS]`` is absent —
+        # an emit with no arguments at all would silently pass that. Pin
+        # the actual parsed JSON to catch a regression that drops args.
+        args_str = fn.get("arguments")
+        assert args_str, "expected arguments delta in the fused boundary case"
+        assert "[ARGS]" not in args_str
+        assert json.loads(args_str) == {"location": "Paris"}
--- a/tests/test_tool_injection.py
+++ b/tests/test_tool_injection.py
@@ -227,9 +227,25 @@ class TestMistralArgsStripping:
         assert result.tool_calls[0]["name"] == "get_weather"
 
     def test_args_suffix_stripped_streaming(self):
+        """Streaming path must strip ``[ARGS]`` from the name boundary (#579).
+
+        Pre-fix, the private ``_parse_streaming_tool_delta`` did this in
+        one branch; post-fix, the public state machine
+        (``extract_tool_calls_streaming``) is the only entry point. We
+        feed a single delta that fuses the boundary tokens and assert
+        the emitted name is clean.
+        """
         parser = MistralToolParser(tokenizer=None)
-        delta = parser._parse_streaming_tool_delta(
-            'get_weather[ARGS]{"location": "Paris"}'
+        full = '[TOOL_CALLS]get_weather[ARGS]{"location": "Paris"}'
+        delta = parser.extract_tool_calls_streaming(
+            previous_text="",
+            current_text=full,
+            delta_text=full,
+            request={"tools": []},
         )
         assert delta is not None
-        assert delta["name"] == "get_weather"
+        tcs = delta.get("tool_calls") or []
+        assert tcs, "expected a tool_calls delta"
+        fn = tcs[0]["function"]
+        assert fn.get("name") == "get_weather"
+        assert "[ARGS]" not in (fn.get("arguments") or "")
--- a/tests/test_tool_parsers.py
+++ b/tests/test_tool_parsers.py
@@ -844,6 +844,186 @@ class TestStreamingParsing:
         assert result == {"content": "Hello world"}
 
 
+def _run_mistral_streaming(parser: MistralToolParser, chunks: list[str]) -> dict:
+    """Drive ``extract_tool_calls_streaming`` chunk-by-chunk and assemble.
+
+    Mirrors what an OpenAI-compatible client does with the SSE deltas:
+    concatenate each ``function.name`` / ``function.arguments`` fragment,
+    record each ``id`` once, accumulate text content. The single dict
+    returned describes the final assembled tool call as the client would
+    see it after the stream ends, so the assertions can compare against
+    the non-streaming ground truth.
+    """
+    content = ""
+    tool_calls: dict[int, dict[str, str]] = {}
+    prev = ""
+    for chunk in chunks:
+        cur = prev + chunk
+        delta = parser.extract_tool_calls_streaming(
+            previous_text=prev,
+            current_text=cur,
+            delta_text=chunk,
+            request={"tools": []},
+        )
+        prev = cur
+        if not delta:
+            continue
+        if "content" in delta and delta["content"]:
+            content += delta["content"]
+        for tc in delta.get("tool_calls") or []:
+            idx = tc["index"]
+            slot = tool_calls.setdefault(idx, {"id": "", "name": "", "arguments": ""})
+            if tc.get("id"):
+                slot["id"] = tc["id"]
+            fn = tc.get("function") or {}
+            if fn.get("name"):
+                slot["name"] += fn["name"]
+            if fn.get("arguments"):
+                slot["arguments"] += fn["arguments"]
+    return {"content": content, "tool_calls": tool_calls}
+
+
+class TestMistralDevstralStreaming:
+    """Regression coverage for #579 — Devstral ``[ARGS]`` streaming.
+
+    The non-streaming path correctly strips the ``[ARGS]`` separator
+    (mistral_tool_parser.py extract_tool_calls), but the streaming
+    path operated chunk-by-chunk on ``delta_text`` and only handled
+    ``[ARGS]`` when it landed in the same delta as ``{``. Real
+    Devstral token streams split the separator across chunks, so the
+    literal ``[ARGS]`` leaked into ``arguments`` and (worse) the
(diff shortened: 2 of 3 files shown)


raullenchai:
## pr_validate round 4 — codex PASS, only pipeline env issue remains

| step | status |
|---|---|
| fetch | ✅ |
| test_plan_check | ✅ |
| cl_description_quality | ✅ |
| **codex_review** | ✅ **no blocking issues** |
| supply_chain | ✅ |
| lint | ✅ |
| targeted_tests | ✅ |
| full_unit | ❌ env bug — see below |
| stress_e2e_bench | skipped (blast=medium gate) |

### Codex convergence

- Round 1: 3 BLOCKING + 2 NIT (1 NIT was a false positive — the `assert tc["id"]` line was already a real assert)
- Round 2: 3 BLOCKING (uncovered by round-1 fixes)
- Round 3: 2 BLOCKING (uncovered by round-2 fixes)
- Round 4: **0 BLOCKING**

Net: every flagged correctness gap has a paired regression test, including 6 round-2 cases (literal `[TOOL_CALLS]` / `[ARGS]` in string args, escape handling, bracket-tail, incremental O(L), combined multi-tool stress) and 3 round-3 cases (leading whitespace / newline / per-segment whitespace gating).

### Remaining `full_unit` FAIL is a pre-existing pipeline env bug

Log shows:
```
/Users/raullenstudio/work/rapid-mlx/.venv/bin/python3.12: No module named pytest
```

`scripts/pr_validate/steps/full_unit.py` invokes `sys.executable -m pytest` against the project venv, which doesn't have pytest as a runtime dep — only `uv run pytest` (which pulls pytest from the dev-tool dependency tree) works. This failure also appeared in round 1 / 2 / 3 of this same PR and existed before this branch.

Running the same pytest scope through `uv run`:
```
==== 4961 passed, 19 skipped, 16 deselected, 7 xfailed in 105.90s ====
```

So the **actual** suite is green — the gate's `sys.executable` invocation is what's broken. I'll file that as a separate pipeline-tooling fix; it shouldn't block this PR since it's neither caused by these changes nor in scope.

### Coverage of the original bug

`tests/test_tool_parsers.py::TestMistralDevstralStreaming` — 22 tests, all pass:

```
$ uv run pytest tests/test_tool_parsers.py::TestMistralDevstralStreaming -v
============================== 22 passed in 1.15s ==============================
```

Plus 3 in `tests/test_tool_injection.py::TestMistralArgsStripping`. The exact pathological chunking from #579 (`['[TOOL_CALLS]', 'read', '[ARGS]{"', 'file_path":"/tmp/foo.txt', '"}']`) was the round-1 fail-then-fix pin.

Issue #575, opened 2026-06-15
[P0] Qwen3 thinking-ON non-streaming leaks thought trace into content when truncated mid-thought


raullenchai:
## Summary

When `chat_template_kwargs.enable_thinking=true` AND the model is truncated mid-thought (finish_reason=length), the **non-streaming** path leaks the entire thinking trace into `content` with `reasoning_content=""`. The **streaming** path on the same prompt routes correctly to `reasoning_content`.

Confirmed on qwen3.5-4b-4bit and qwen3.6-35b-a3b-4bit on v0.7.3.

## Root cause

The Qwen3 chat template pre-injects `<think>\n` into the prompt itself when `enable_thinking=True`:

```
...assistant\n<think>\n  <-- end of prompt; model continues from here
```

So the model never **emits** the `<think>` token in its output. When truncated before `</think>`, the output contains neither tag.

1. **Non-streaming path** (`vllm_mlx/routes/chat.py:1188` → `_finalize_content_and_reasoning` → `Qwen3ReasoningParser.extract_reasoning` from `vllm_mlx/reasoning/think_parser.py:53-93`) hits Case 4 ("No tags at all - pure content") and returns `(None, model_output)` — entire thought trace leaks to `content`.

2. **OutputRouter fallback is disabled** for "think" format: `_OUTPUT_ROUTER_ALLOWLIST = {"gemma4", "harmony"}` (`vllm_mlx/engine/batched.py:104`). So `_route_tokens_for_channels` returns `("", fallback_text, None)` for qwen — engine_reasoning_text stays empty.

3. **Streaming path** uses `BaseThinkingReasoningParser.extract_reasoning_streaming` (`vllm_mlx/reasoning/think_parser.py:141-148`) which handles "implicit-thinking mode" — Case 3 "treat as reasoning if we haven't seen `</think>` yet". This is correct, but the non-streaming `extract_reasoning` lacks the equivalent fallback.

## Vocab check (the hypothesis was FALSIFIED — both tags ARE in vocab)

```
mlx-community/Qwen3.5-4B-MLX-4bit:
  <think> in vocab: True   (id=248068)
  </think> in vocab: True  (id=248069)
mlx-community/Qwen3.6-35B-A3B-4bit:
  <think> in vocab: True   (id=248068)
  </think> in vocab: True  (id=248069)
```

[...]

PR #594, opened 2026-06-16
Fix Claude Code connectivity with Rapid-MLX server

angusgastle:
## Summary
- Fixes Claude Code connectivity issues when running with Rapid-MLX server
- Adds comprehensive route tests for authentication and API validation
- Applies ruff formatting to pass CI validation

## Test plan
- [x] All targeted tests pass
- [x] Code formatting validated
- [x] No supply chain/dependency issues

🤖 Generated with Claude Code

### DIFF
--- a/tests/test_anthropic_route_auth.py
+++ b/tests/test_anthropic_route_auth.py
@@ -233,6 +233,57 @@ def test_anthropic_messages_accepts_valid_bearer_api_key(anthropic_client):
     assert len(engine.calls) == 1
 
 
+def test_anthropic_messages_accepts_claude_model_name(anthropic_client):
+    """Claude Code sends its real model name (e.g. 'claude-opus-4-5') in the
+    request body. The Anthropic endpoint must route this to the loaded
+    engine instead of 404-ing on the name mismatch.
+
+    The model name in the response should reflect the *loaded* model, not the
+    client-supplied claude name.
+    """
+    client = anthropic_client.client
+
+    response = client.post(
+        "/v1/messages",
+        json={
+            "model": "claude-opus-4-5",
+            "max_tokens": 4,
+            "messages": [{"role": "user", "content": "hello"}],
+        },
+        headers={"x-api-key": "test-secret"},
+    )
+
+    assert response.status_code == 200
+    # The response must carry the loaded model name, not the claude alias.
+    assert response.json()["model"] == "test-model"
+
+
+def test_anthropic_messages_accepts_any_claude_variant(anthropic_client):
+    """All claude-* variants (claude-3-haiku-20240307, claude-sonnet-4-5,
+    claude-3-5-sonnet-20241022, etc.) must pass through to the loaded engine."""
+    client = anthropic_client.client
+
+    for model_name in [
+        "claude-3-haiku-20240307",
+        "claude-sonnet-4-5",
+        "claude-opus-4-5",
+        "claude-3-5-sonnet-20241022",
+    ]:
+        response = client.post(
+            "/v1/messages",
+            json={
+                "model": model_name,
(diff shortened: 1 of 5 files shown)


raullenchai (commented):
Thanks for diagnosing and fixing this end-to-end — both bugs are real (the `HEAD /` probe killing Claude Code before it ever issues a request, and the strict `_validate_model_name` rejecting `claude-*` aliases on `/v1/messages`). Tests are scoped and pin the right invariants — the `test_mismatch_still_raises_404_via_openai_route` defense, in particular, is the kind of thing that keeps this from quietly leaking into the OpenAI route in a future refactor.

Two questions before I merge, and one optional polish:

**Q1 — Why include `"gpt-"` in the prefix tuple on the Anthropic route?**
```python
if not (anthropic_request.model or "").startswith(("claude-", "gpt-")):
    _validate_model_name(anthropic_request.model)
```
`/v1/messages` is the Anthropic-shaped endpoint; `gpt-*` model names normally go through `/v1/chat/completions`. The `gpt-` bypass is either dead code (no client actually hits this code path) or papering over a different misconfig I'm not seeing. Could you point me at a concrete client + flow that needs it? If there isn't one, I'd prefer to drop it and add it back when a real case surfaces — kept narrow, the bypass is easier to reason about.

**Q2 — Log line cadence.**
```python
logger.info("Anthropic /v1/messages: request model=%r served by loaded engine=%r", ...)
```
This fires on every mismatched request, which under steady-state Claude Code traffic could be hundreds per minute. Two options that don't lose the diagnostic value:
- Downgrade to `logger.debug(...)` (still surfaces when the user passes `--log-level debug` to investigate)
- Or dedup at module scope: log once per distinct `(request_model, loaded_model)` pair

I'd lean toward the dedup version since the info IS useful the first time a deployment sees a new mismatch. Happy with either.

**Polish (optional).** The prefix tuple could be a module-level constant like:
```python
_PASSTHROUGH_MODEL_PREFIXES = ("claude-",)  # add to this when we have a documented reason
```
Makes it grep-able, testable, and the comment becomes a doc trail for future additions.

Once Q1/Q2 are addressed I'll squash-merge. The `HEAD /` half can go in as-is — that one is fine.

raullenchai:
Thanks again for diagnosing both bugs end-to-end — diagnosis and tests were sharp. Closing this without merging because the same two fixes already landed via PR #557 (commit 4409b47, merged 2026-06-15 16:06 PST, ~11 hours before this PR was opened). Quick crosscheck:

* `HEAD /` connectivity probe — fixed in `vllm_mlx/routes/health.py:24` via `@probe_router.api_route("/", methods=["GET", "HEAD"])`, identical handler shape to this PR's.
* `claude-*` model name bypass — fixed in `vllm_mlx/routes/anthropic.py:119` via the same `if not (model or "").startswith(("claude-", "gpt-")): _validate_model_name(...)` guard.

Your test coverage was also stronger in places (e.g. `test_anthropic_messages_accepts_any_claude_variant` with four variant names, `test_head_root_bypasses_api_key`). If you want to send those tests as a small follow-up PR against current main I'd happily merge — they're additive and tighten the contract.

One heads-up if you do: your branch is currently ~135 files behind main (PR diff shows +3926/-13878) and would regress recent reasoning-parser work merged via #575/#576 (the `_effective_enable_thinking` plumbing in `vllm_mlx/routes/anthropic.py` non-stream finalize). Rebase first.

Sorry the timing made you re-build a fix that was already in flight — that's on us for not flagging #557 sooner. The Co-Authored-By trail keeps your credit visible if you want to reference it later.

angusgastle:
Okay thank you @raullenchai keep up the good work!

PR #646, opened 2026-06-17
perf(sidecar): trim transformers modeling + hoist .pyc for ~100 MB win

raullenchai:
## Summary
- Trim `transformers/models/*/modeling_*.py` (dead in our inference path — mlx-lm has its own model classes; we never instantiate `AutoModel`). Preserves `models/auto/` so `AutoTokenizer` / `AutoConfig` dispatch keeps working.
- Trim `image_processing_*.py` + `feature_extraction_*.py` (text-only sidecar; vision goes through `[vision]` extras which aren't installed).
- Drop `numpy/random/_examples`, `typing/tests`, `f2py/tests`, `distutils/tests`.
- Hoist `.pyc` files out of `__pycache__/` into sourceless layout (`foo.pyc` adjacent to where `foo.py` was) and drop the `.py` sources. The sidecar shim already exports `PYTHONDONTWRITEBYTECODE=1`, so Python loads the bytecode directly and never tries to rewrite it (which would also break the codesign seal — see existing step 3 comment).

## Why
rapid-desktop's `.app` bundle is at 858 MB (CI gate at machinefi/rapid-desktop#242 caps it at 500 MB; next release fails without this PR). Bundled qwen3-0.6b-4bit (340 MB) is the elephant — that's removed in the downstream PR. This PR contributes ~100 MB of sidecar slimming.

## Build-and-measure (this checkout, M3 Ultra, `--skip-codesign --skip-verify`)
- baseline raw bundle: **498 MB**
- slimmed  raw bundle: **398 MB** (**-100 MB, -20%**)

## Important corrections from the original plan
Two things didn't work the way the naive "drop .py, keep `__pycache__/foo.cpython-312.pyc`" plan assumed; the implemented version handles both:

1. **Regular-package `.pyc` MUST live adjacent to where the source would be**, not in `__pycache__/`. Python's `FileFinder` only treats `__pycache__/foo.cpython-312.pyc` as a *cache* for an adjacent `foo.py`; with the source deleted, the cache is ignored and `import websockets.imports` raises `ModuleNotFoundError`. Fixed by hoisting (move + rename to `foo.pyc`).
2. **`__init__.py` cannot be replaced with a `.pyc`** without degrading the directory to a PEP 420 namespace package, which silently skips the `__init__` body. Kept all `__init__.py` files as source.
3. **`transformers/models/` is excluded from the hoist step.** Transformers scans that directory at startup (`create_import_structure_from_path`) looking for `.py` files; sourceless `.pyc` makes the import-structure registry collapse and the package import raises `KeyError: frozenset()` on the next line. The post-step-3.5 footprint is only ~20 MB so the lost saving is small.

## Trade-offs
- Crash tracebacks lose source-line CONTENT for non-`__init__` modules (file:line is still shown, just not the actual line of code). Acceptable: we capture structured logs via the rapid-mlx telemetry path. `SKIP_SOURCE_DROP=1` env keeps sources for local sidecar python debugging.

## Test plan
- [x] Smoke: `rapid-mlx --version` works against trimmed sidecar (`rapid-mlx 0.7.26`)
- [x] Smoke: `from transformers import AutoTokenizer, AutoConfig` succeeds
- [x] Smoke: `AutoTokenizer.from_pretrained("mlx-community/Qwen3.5-4B-MLX-4bit")` round-trips an encode
- [x] Smoke: `rapid-mlx ls` lists cached models
- [x] Sidecar size measured before/after: 498 MB → 398 MB

### DIFF
--- a/scripts/build-sidecar.sh
+++ b/scripts/build-sidecar.sh
@@ -173,6 +173,40 @@ rm -rf \
     "$STAGE/python/share/man"
 find "$STAGE" -type d -name __pycache__ -prune -exec rm -rf {} +
 
+# ----- step 3.5: aggressive trim (post-strip, pre-compileall) ----------
+#
+# rapid-desktop's `.app` bundle hit 858 MB (machinefi/rapid-desktop#242
+# CI gate caps at 500 MB). The sidecar contributes ~498 MB raw; this
+# step trims ~80 MB of code we never load at runtime.
+#
+# Two safe drops:
+#   1. transformers/models/*/modeling_*.py — our inference path goes
+#      through mlx-lm's own model classes; we never instantiate
+#      transformers' AutoModel/PreTrainedModel. We DO still need the
+#      tokenizer + config dispatch in transformers/models/auto/, so
+#      that path is pruned out of the find.
+#   2. image_processing_*.py + feature_extraction_*.py — text-only
+#      sidecar; the vision stack ships via the `[vision]` extras
+#      which aren't installed here.
+
+echo "==> trimming transformers PyTorch model implementations"
+find "$STAGE/site-packages/transformers/models" \
+    -path "*/auto/*" -prune -o \
+    -name "modeling_*.py" -not -name "modeling_tf_*" -not -name "modeling_flax_*" \
+    -print -delete
+find "$STAGE/site-packages/transformers/models" \
+    -path "*/auto/*" -prune -o \
+    \( -name "image_processing_*.py" -o -name "feature_extraction_*.py" \) \
+    -print -delete
+
+echo "==> trimming numpy dev/test detritus"
+rm -rf \
+    "$STAGE/site-packages/numpy/random/_examples" \
+    "$STAGE/site-packages/numpy/typing/tests" \
+    "$STAGE/site-packages/numpy/f2py/tests" \
+    "$STAGE/site-packages/numpy/testing/_private/extbuild.py" \
+    "$STAGE/site-packages/numpy/distutils/tests" 2>/dev/null || true
+
 # Pre-compile every .py in the bundled stdlib + site-packages BEFORE
 # we codesign. Otherwise CPython's import machinery writes .pyc files
 # into __pycache__/ at runtime on first use, those additions are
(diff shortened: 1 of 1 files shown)


raullenchai:
## Codex r2 review — empirical disposition

Codex r2 raised 3 BLOCKING findings (2 rounds is the cap per our SOP). r2 B3 fixed in-tree (commit fbe732e). r2 B1 + B2 are claims about runtime failure modes that don't hold against the actual trimmed bundle — empirical evidence below.

### r2 B1 / B2: AutoModel / AutoImageProcessor / AutoFeatureExtractor

**Codex claim**: deleting `modeling_*.py` / `image_processing_*.py` / `feature_extraction_*.py` changes the failure mode from "backend not installed" placeholder to `ModuleNotFoundError`.

**Actual behaviour** against this branch's slimmed bundle (`from transformers import …`):

```
$ AutoModel placeholder: AutoModel
$ AutoModel.from_pretrained('mlx-community/Qwen3.5-4B-MLX-4bit')
    → ImportError: AutoModel requires the PyTorch library but it was
      not found in your environment.

$ AutoImageProcessor placeholder: AutoImageProcessor  (import OK)
$ AutoFeatureExtractor placeholder: AutoFeatureExtractor  (import OK)
```

Transformers' lazy-module dispatch returns a backend-placeholder class for every auto-* surface and raises `ImportError: requires the PyTorch library` long before any concrete `modeling_*` / `image_processing_*` / `feature_extraction_*` lookup happens. Our bundle ships neither `torch` nor `torchvision`, so the placeholder is the ONLY reachable code path — deleting the targets is a no-op for the failure surface.

This was specifically what the PR description's "we never instantiate AutoModel" claim was about; the codex BLOCKING is theoretically valid against a *PyTorch-having* bundle and inapplicable here.

### r2 B3: stdlib not in scope

Valid comment fix; landed in fbe732e — narrowed the step 3.6 doc to "site-packages only" to match the actual `find` invocation. Stdlib remains source-backed (~30 MB after step-3 strip; compileall populates `__pycache__/` so import latency is unchanged).

PR #665, opened 2026-06-18
fix(scheduler): allow concurrent batching with different temperatures

### LINKED ISSUE #611: scheduler: _ensure_batch_generator key misses ignore_eos — EOS leaks across requests
## Symptom

`_ensure_batch_generator()` decides whether to reuse the current `BatchGenerator` based on `_current_sampler_params = (temperature, top_p, min_p, top_k)`. `ignore_eos` is folded into the generator's `stop_tokens` set at construction time (`_assemble_stop_tokens` returns `set()` when `ignore_eos=True`), but it's NOT part of the reuse key. So if two requests carry identical sampler tuples but different `ignore_eos`, the second silently inherits the first's stop-token policy.

## Repro

Mix one benchmark / probe call with `ignore_eos=True` and one chat call with `ignore_eos=False` (same `temperature`, `top_p`, `min_p`, `top_k`) on a long-lived engine:

1. `ignore_eos=True` first → BatchGenerator built with `stop_tokens=set()`. Next chat call reuses it; chat runs to `max_tokens` instead of stopping at EOS.
2. `ignore_eos=False` first → BatchGenerator built with model EOS tokens. Next benchmark reuses it; benchmark stops early at EOS instead of running to `max_tokens`.

## Diagnosis

```python
# vllm_mlx/scheduler.py:2486
def _ensure_batch_generator(self, sampling_params: SamplingParams) -> None:
    sampler_params = (
        sampling_params.temperature,
        sampling_params.top_p,
        sampling_params.min_p,
        sampling_params.top_k,
    )
    if (
        self.batch_generator is None
        or self._current_sampler_params != sampler_params  # ← key misses ignore_eos
    ):
        ...
        self.batch_generator = self._create_batch_generator(sampling_params)
        self._current_sampler_params = sampler_params
```

`_create_batch_generator()` at line 2049 builds:

```python
stop_tokens = _assemble_stop_tokens(sampling_params, self._get_stop_tokens())
```

`_assemble_stop_tokens` returns `set()` when `ignore_eos=True`, otherwise the model's stop tokens. That choice is captured in the BatchGenerator and never revisited.

## Fix

Include the effective stop-token policy in the generator compatibility key. Smallest change:

```python
sampler_params = (
    sampling_params.temperature,
    sampling_params.top_p,
    sampling_params.min_p,
    sampling_params.top_k,
    bool(sampling_params.ignore_eos),
)
```

Recreate the BatchGenerator whenever this changes (same path that handles temperature/top_p changes today). Add a contract test alongside `tests/test_community_bench.py::test_scheduler_honours_ignore_eos` that exercises the reuse path explicitly:

1. Submit request with `ignore_eos=True` + matching sampler tuple.
2. Submit request with `ignore_eos=False` + same sampler tuple.
3. Assert the second request observes the model's EOS.

## Impact

- Chat / API users almost never set `ignore_eos=True`, so production traffic on `rapid-mlx serve` is mostly unaffected.
- Benchmark harnesses, fixed-length generation probes, and tooling that toggles `ignore_eos` per request CAN observe the wrong behavior on a long-lived engine. Today's tests pass because `test_scheduler_honours_ignore_eos` doesn't mix the two policies on a single generator.
- Severity P2 (silent-correctness drift on a documented contract path, no crash, no security impact).

## Surfaced by

codex review on `rapid-desktop` PR [#228](https://github.com/machinefi/rapid-desktop/pull/228) (submodule bump 0.7.11 → 0.7.15). Out-of-scope for that PR; filing here.

angusgastle:
## Problem

Claude Code sends large agentic prompts (~33K tokens with full tool schemas) that trigger slow prefill phases. When concurrent requests arrive with different sampling parameters (e.g., different temperature), they serialize unnecessarily because the scheduler's  key includes temperature/top_p/min_p/top_k.

This forces the second request to wait 115+ seconds even though:
1. Per-request samplers are already passed to 
2. Temperature/top_p don't need to match globally

Additionally, the key excluded `stop_token_ids`, creating a silent correctness bug where requests with different stop sequences could share a BatchGenerator.

### 1. Fix sampler params key (scheduler.py:2543-2562)

Replace `(temperature, top_p, min_p, top_k, ignore_eos)` with `(frozenset(stop_token_ids), ignore_eos)`:
- Only generator-level invariants: stop tokens are set at BatchGenerator init and can't be changed
- Per-request samplers are passed directly to insert() and override the global sampler
- Frozenset ensures correctness: requests with different stop sequences get different keys

### 2. Improve chunked-prefill discoverability (cli.py:4258-4261)

Update `--chunked-prefill-tokens` help text to call out Claude Code and agentic workloads as the intended use case, with recommended starting value of 2048.

## Testing

Verified:
- Two streaming requests with different temperatures batch together (`running=2` in scheduler logs)
- Requests with different stop sequences do not inherit each other's stops
- Help text mentions Claude Code and agentic workloads  
- Second-turn cache hits work and TTFT drops dramatically

Closes #611

### DIFF
--- a/tests/test_community_bench.py
+++ b/tests/test_community_bench.py
@@ -390,6 +390,162 @@ def test_scheduler_batch_generator_reuse_key_includes_ignore_eos() -> None:
     )
 
 
+def test_scheduler_batch_generator_reuses_across_different_sampler_params() -> None:
+    """PR #665 perf fix — requests with different sampler params
+    (temperature, top_p, min_p, top_k) but identical stop config MUST
+    share the same BatchGenerator.
+
+    Before the fix, the reuse key included the full sampler tuple
+    (temperature, top_p, min_p, top_k, ignore_eos), so two concurrent
+    requests with different temperatures forced serial processing. A
+    33K-token agentic request (default temp) blocked a 352-token
+    concurrent request (different temp) for 115 seconds in production.
+
+    Per-request samplers are passed to ``batch_gen.insert(samplers=[...])``
+    directly, so the only generator-level invariant is stop config —
+    sampler params don't need to match for reuse.
+    """
+    pytest.importorskip("mlx")
+    from vllm_mlx.request import SamplingParams
+    from vllm_mlx.scheduler import Scheduler
+
+    sentinel = object()
+    create_calls = []
+
+    class StubScheduler:
+        batch_generator = None
+        running = False
+        memory_aware_cache = None
+        prefix_cache = None
+        _current_sampler_params = None
+
+        def _close_batch_generator(self) -> None:
+            self.batch_generator = None
+
+        def _create_batch_generator(self, sp):
+            create_calls.append(sp)
+            return sentinel
+
+    stub = StubScheduler()
+
+    # Same stop config (defaults: empty stop_token_ids, ignore_eos=False);
+    # different sampler params across every axis the old key checked.
+    p_low = SamplingParams(
+        max_tokens=512, temperature=0.0, top_p=1.0, min_p=0.0, top_k=-1
+    )
+    p_high = SamplingParams(
+        max_tokens=512, temperature=0.8, top_p=0.95, min_p=0.05, top_k=50
+    )
+
+    assert Scheduler._ensure_batch_generator(stub, p_low) is True
+    assert len(create_calls) == 1, "first request should build a generator"
+
+    assert Scheduler._ensure_batch_generator(stub, p_high) is True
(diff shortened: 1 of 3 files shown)


angusgastle:
This was based on some performance optimization work related specifically to Claude Code. Did not test to see if it impacts your other work. But it appears to shave a few seconds off generations on my hardware.

raullenchai:
Thanks @angusgastle — merged! 🎉 Added you to the changelog:

Issue #644, opened 2026-06-17
DiffusionEngine: worker doesn't drain cancelled job within 30s — engine marked unhealthy


raullenchai:
## Repro

Surfaced during \`pr_validate\` stress_e2e_bench step on PR #619 (a +5 LOC parity fix to \`vllm_mlx/routes/anthropic.py\` totally unrelated to diffusion). The diffusion engine path failed deterministically on \`diffusion-gemma-26b-4bit\` (\`mlx-community/diffusiongemma-26B-A4B-it-4bit\`) while the other two stress targets (\`Qwen3.5-35B-A3B-8bit\`, \`Qwen3.6-27B-MLX-8bit\`) passed 8/8.

## Symptom

Server log:
\`\`\`
ERROR:rapid_mlx.runtime.diffusion_lane:DiffusionEngine worker did not drain
cancelled job within 30 s — engine marked unhealthy. Further admissions
will fail until restart.
INFO: 127.0.0.1:53784 - "POST /v1/chat/completions HTTP/1.1" 503 Service Unavailable
INFO: 127.0.0.1:53904 - "POST /v1/chat/completions HTTP/1.1" 503 Service Unavailable
[... ~50 more 503s ...]
\`\`\`

Stress test scorecard:
\`\`\`
PASS  Sustained throughput
PASS  Concurrent load
FAIL  Long generation
FAIL  Rapid fire
FAIL  Tool call storm
FAIL  Mixed workload
FAIL  Disconnect resilience
PASS  Memory stability
3/8 passed
\`\`\`

The "Long generation" and "Disconnect resilience" steps trigger the cancellation path; once one cancellation goes un-drained, the engine flips to unhealthy and every subsequent admission 503s until process restart.

## Hypothesis

\`rapid_mlx.runtime.diffusion_lane\` worker thread isn't responding to its cancellation signal within the 30s grace period. Either:
1. The diffusion forward pass is non-interruptible mid-step and a long bucket exceeds 30s — bump the deadline, or insert a yield/check between diffusion iterations.
2. The cancellation signal isn't propagating to the worker at all — check the lane shutdown path.

## Scope

Only the diffusion path is affected — autoregressive paths drain cleanly under the same stress matrix on the same machine.

## Repro recipe

\`\`\`bash
rapid-mlx serve diffusion-gemma-26b-4bit --port 8500
# In another shell:
[...]

PR #667, opened 2026-06-18
fix(prefix-cache): async + deadline-aware shutdown flush, prevents orphaned .new/ on SIGTERM

raullenchai:
## User-visible bug

> prefix-cache persist truncated on SIGTERM, leaves `<cache_dir>.new/` directory orphaned

When uvicorn delivers SIGTERM to `rapid-mlx serve` (rapid-desktop quit, launchd Stop, systemd graceful-stop, Docker stop, ...), the FastAPI lifespan shutdown hook runs `save_prefix_cache_to_disk()` **synchronously** on the asyncio loop thread. The save itself is a per-entry loop that calls `mlx_lm.models.cache.save_prompt_cache` on each KV cache entry — for production models like Gemma 4 26B that's 200-300 MB per safetensors file, all in linear wall-clock order, all under the GIL.

Live evidence from rapid-desktop quit at 18:30 on `mlx-community/gemma-4-26b-a4b-it-4bit` (PID 57809):

```
INFO:rapid_mlx.runtime.cache:[lifespan] Saving prefix cache to .../gemma-4-26b-a4b-it-4bit--e527bf6c
INFO:rapid_mlx.memory_cache:[cache_persist] saved entry 0: 1263 tokens, 224.7MB KV, ...
INFO:rapid_mlx.memory_cache:[cache_persist] saved entry 1: 1292 tokens, 225.2MB KV, ...
...
INFO:rapid_mlx.memory_cache:[cache_persist] saved entry 18: 2941 tokens, 257.4MB KV, ...
[SIGKILL ~5s in]
```

~4.4 GB written, only 18 of N entries finished. The atomic `<cache_dir>.new/ → <cache_dir>` rename never ran, so the staging dir was wedged in `.new/` until manual cleanup AND the next launch lost the KV-cache hit entirely.

Rapid-desktop PR #255 ships a 5s → 30s SIGTERM-grace mitigation but the proper fix lives here — with the budget gate below, even a 5s grace produces a clean committed snapshot.

## Design (one paragraph + trade-offs)

Two changes that work together:

1. **`asyncio.to_thread` in the lifespan shutdown** so the asyncio loop stays responsive (`/healthz`, graceful-shutdown responses, anything else attached to the loop gets to make progress instead of being starved for tens of seconds).

2. **A `should_abort` predicate plumbed from `runtime/cache.py` → `BatchedEngine` → `EngineCore` → `Scheduler` → `MemoryAwarePrefixCache.save_to_disk`**. The per-entry loop checks it after every entry; when the SIGTERM-grace deadline trips (default **3.5s**, env-overridable via `RAPID_MLX_PREFIX_CACHE_SHUTDOWN_BUDGET`), the loop stops AND the partial snapshot is still committed via the same atomic directory rename. So a downstream SIGKILL after 5s leaves `<cache_dir>/index.json` + N entries on disk, never an orphan `.new/`.

**Trade-off considered.** The user's prompt offered three approaches:

- *Periodic background persist during steady-state.* The most user-visible win because shutdown becomes a no-op and the cache survives even a kill -9. But it's a much larger surface area (dirty-set tracking, evict-then-delete-on-disk sync, sibling-process coordination, retry on transient I/O), and entries are large enough that batched writes would have to compete with live decode for disk bandwidth. Out of scope for a focused bug fix; left as follow-up.
- *Async lifespan flush via `asyncio.to_thread` alone.* Necessary but not sufficient — uvicorn still waits on the lifespan to return, and `to_thread` only moves the wait off the loop, not the deadline. We adopt this AS PART of the fix to keep the loop responsive, but pair it with...
- *Interruptible flush via signal handler / deadline.* Picked. Smallest diff, preserves on-disk format, preserves the existing partial-recovery and atomic-rename invariants the load path already encodes, and is what actually solves the orphaned-`.new/` symptom.

## What changed

| File | What |
| --- | --- |
| `vllm_mlx/memory_cache.py` | `save_to_disk(cache_dir, should_abort=None)` — per-entry abort check + always-correct `num_entries` in index.json. |
| `vllm_mlx/scheduler.py` | `save_cache_to_disk` forwards `should_abort`. |
| `vllm_mlx/engine_core.py` | `EngineCore.save_cache_to_disk` + `AsyncEngineCore.save_cache_to_disk` forward `should_abort`. |
| `vllm_mlx/engine/batched.py` | `BatchedEngine.save_cache_to_disk` forwards `should_abort`. |
| `vllm_mlx/runtime/cache.py` | `save_prefix_cache_to_disk(budget_sec=None)` builds the deadline predicate; default 3.5s, `RAPID_MLX_PREFIX_CACHE_SHUTDOWN_BUDGET` env override, `0` disables for offline CLI saves. |
| `vllm_mlx/server.py` | Lifespan shutdown now `await asyncio.to_thread(_save_prefix_cache_to_disk)`. |

`should_abort=None` everywhere is the legacy "no deadline, flush everything" path used by the existing tests and the offline `rapid-mlx` CLI — no behavior change there.

## Tests (regression coverage in `tests/test_prefix_cache_persistence.py`)

- `test_save_to_disk_partial_commit_on_abort` — predicate fires after entry 0 of 3; asserts the staging dir was renamed away (no orphan `.new/`), `index.json` exists with `num_entries=1`, and the surviving entry round-trips through `load_from_disk`.
- `test_save_to_disk_aborts_before_first_entry_returns_false` — predicate fires immediately; asserts `False` return, no committed dir, no orphan `.new/` (matches the existing "no entries saved successfully" branch).
- `test_save_to_disk_predicate_none_preserves_full_flush` — guards the offline-CLI path against behavior drift.
- `test_save_prefix_cache_to_disk_runs_off_event_loop` — drives the lifespan-shape call shape with a slow fake engine + concurrent 50ms ticker; asserts the ticker advances ≥5 times during a 600ms save (which it can't do if the save still blocks the loop).
- `test_save_prefix_cache_to_disk_respects_budget` — passes `budget_sec=0.1`, sleeps past it inside the fake engine, asserts the captured predicate evaluates True.
- `test_save_prefix_cache_to_disk_zero_budget_disables_deadline` — guards the offline-CLI "no deadline" path.

Also updated `tests/test_engine_step_thread.py::test_save_cache_to_disk_routes_to_worker` for the new kwarg, and allowlisted `RAPID_MLX_PREFIX_CACHE_SHUTDOWN_BUDGET` in `tests/test_no_out_of_band_routing.py::ALLOWED_RAPID_MLX_ENV_VARS` (with the standard non-routing comment block — it's a pure deadline knob, never selects model / parser / tier).

## Verification

- `python3.12 -m pytest tests/ -q --ignore=tests/integrations` — **5602 passed**, 18 skipped, 16 deselected, 7 xfailed, **0 new failures**. The 2 pre-existing `test_event_loop.py` failures are connection errors on `127.0.0.1:8000` (need a live server, unrelated).
- `python3.12 -m ruff check + format` — clean.

### DIFF
--- a/tests/test_engine_step_thread.py
+++ b/tests/test_engine_step_thread.py
@@ -93,9 +93,10 @@ class TestStepThread:
 
         captured = {}
 
-        def fake_save(cache_dir):
+        def fake_save(cache_dir, should_abort=None):
             captured["thread"] = threading.current_thread().name
             captured["cache_dir"] = cache_dir
+            captured["should_abort"] = should_abort
             return True
 
         engine_core.scheduler.save_cache_to_disk.side_effect = fake_save
--- a/tests/test_no_out_of_band_routing.py
+++ b/tests/test_no_out_of_band_routing.py
@@ -174,6 +174,13 @@ ALLOWED_RAPID_MLX_ENV_VARS: frozenset[str] = frozenset(
         # default R2 URL; power users override with any URL or "" to disable.
         # Never consulted by the engine, scheduler, or routing layer.
         "RAPID_MLX_MODEL_MIRROR",
+        # SIGTERM-grace budget (seconds, float) for the lifespan prefix-cache
+        # flush in ``vllm_mlx/runtime/cache.py``. Defaults to 3.5s so a
+        # multi-GB save commits its partial snapshot before downstream
+        # supervisors (rapid-desktop's 5s grace, systemd / Docker / launchd
+        # equivalents) escalate to SIGKILL and orphan ``<cache_dir>.new/``.
+        # Pure deadline knob — never selects model, parser, or tier.
+        "RAPID_MLX_PREFIX_CACHE_SHUTDOWN_BUDGET",
     }
 )
 
--- a/tests/test_prefix_cache_persistence.py
+++ b/tests/test_prefix_cache_persistence.py
@@ -1621,22 +1621,44 @@ def test_adapt_should_abort_handles_keyword_only_and_args_kwargs():
     assert adapted(1.5) is False
     assert captured == [1.5]
 
-    # **kwargs: must receive predicted_sec.
+    # **kwargs only: must be called BY KEYWORD as predicted_sec=...
+    # (codex PR #667 round 4 BLOCKING-1: round 3 classified **kwargs
+    # as positional-capable and raised TypeError on call). The
+    # predicate sees the value via ``kw["predicted_sec"]``.
     captured = []
 
-    def pred(**kw):
+    def kw_only_pred(**kw):
         captured.append(kw.get("predicted_sec", "no-arg"))
         return False
 
-    # The adapter passes positionally, so **kwargs alone doesn't get
-    # the value as a kwarg — but the function still gets CALLED with
-    # the right shape (no TypeError). That's the contract being
-    # pinned: no exception raised.
-    adapted = _adapt_should_abort(
-        lambda *a, **kw: captured.append(("args", a)) or False
-    )
+    adapted = _adapt_should_abort(kw_only_pred)
     assert adapted(2.5) is False
-    assert captured == [("args", (2.5,))]
+    assert captured == [2.5], (
+        f"**kwargs predicate must receive predicted_sec by keyword, got {captured}"
+    )
+
+    # Keyword-only ``predicted_sec=...`` — same routing as **kwargs.
+    captured = []
+
+    def kw_only_named(*, predicted_sec=0.0):
+        captured.append(predicted_sec)
+        return False
+
+    adapted = _adapt_should_abort(kw_only_named)
+    assert adapted(3.5) is False
+    assert captured == [3.5]
+
+    # ``*args, **kwargs`` — positional path wins (it accepts positional
+    # via the ``*args`` part).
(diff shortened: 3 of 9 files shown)


raullenchai:
pr_validate round 5 — codex (gpt-5.5) adversarial review: **GREEN, no blocking issues**.

Scorecard:
- `codex_review` PASS — no blocking
- `lint`, `supply_chain`, `test_plan_check`, `cl_description_quality` PASS
- `targeted_tests` PASS (431 passed)
- `full_unit` PASS (5609 passed, 19 skipped, 7 xfailed)
- 43 prefix-cache persistence tests PASS, including the new round-4 keyword-only / `**kwargs` predicate coverage
- `stress_e2e_bench` failed on `diffusion-gemma` with HTTP 503s — **known issue #644** (DiffusionEngine self-heal lag after stress step 7 "Disconnect resilience" wedges the worker). `qwen3.5` and `qwen3.6` benched cleanly before that. Unrelated to this PR's prefix-cache changes (no DiffusionEngine code touched).

Merging on the strength of the codex green light + clean unit + targeted + benched-AR-LLM coverage. The DiffusionEngine self-heal will be addressed in a separate PR against issue #644.

Issue #569, opened 2026-06-14
gemma-4-26b-4bit multi-turn tool calls silently drop content + tool_calls when reasoning doesn't terminate


raullenchai:
## Repro

\`\`\`bash
rapid-mlx serve gemma-4-26b-4bit --port 8003 &

# Multi-turn weather query — fake tool runner, deterministic responses
PORT=8003 MODEL=gemma-4-26b-4bit python3.12 /tmp/c5-multiturn-param.py
\`\`\`

5-turn script (weather SF → weather Tokyo → comparison → time SF → recall). On 3 of the 5 turns gemma-4-26b emits **empty content + empty tool_calls + finish_reason=stop** with 67-200 completion tokens that went into \`reasoning_content\`.

## Single-shot works fine

A single user turn ("What is the weather in San Francisco?") + tools array reliably produces \`{tool_calls: [{name: get_weather, arguments: {city: \"San Francisco\"}}], content: \"\"}\` with finish_reason=tool_calls.

## Multi-turn breaks

When the conversation history is longer (≥3 prior user/assistant/tool turns), gemma's reasoning_content fires but doesn't terminate into a function call within max_tokens=1024. The result is a turn where the model produced 67 tokens of reasoning but no final answer and no tool call. From the user's POV the assistant just goes silent.

## Compared to siblings

Same multi-turn script:

| Alias | Tool calls fired | Content non-empty | Notes |
|---|---|---|---|
| qwen3.6-35b-4bit | 3/3 expected | 5/5 | clean |
| qwen3.6-27b-4bit | 3/3 expected | 5/5 | clean (slower TPS, but correct) |
| gpt-oss-20b-mxfp4-q8 | 3/3 expected | 5/5 | clean |
| **gemma-4-26b-4bit** | **0/3** | **2/5 empty** | reasoning_content non-empty across the board |

## Suspect

Looks like the gemma alias config in \`vllm_mlx/aliases.json\` may need:

- Verified \`reasoning_parser\` / \`tool_call_parser\` mapping (gemma's native template emits something the parser doesn't latch onto in continued conversation context, even though it does in a fresh single-turn).
- An \`enable_thinking\` default that drops reasoning_content for chat surfaces in favor of direct tool_calls (rapid-desktop reads tool_calls + content; it doesn't depend on reasoning_content for tool-use flows).

## Data

[...]

PR #715, opened 2026-06-19
fix(integration): VibeThinker + ≤5B alias bundle (parser + alias config)

raullenchai:
## Summary

ONE PR bundling SEVEN integration fixes across the reasoning / tool parser stack and the alias config. Original scope was three VibeThinker bugs from the 2026-06-17 live test (PR #708 follow-up); the 2026-06-18 fuzz battery against PR #714's 9 new ≤5B aliases surfaced four additional structural bug classes that touch the same surface — they land here as one coherent integration fix rather than four follow-up PRs.

### Original VibeThinker scope (Bugs 1-3)

- **Bug 1** — vibethinker aliases had `tool_call_parser=null`. The model is Qwen2-derived and emits BOTH `<tool_call>{...}</tool_call>` and bare `<function=name>...</function>` shapes; the second leaked into `content` as raw text. Wire `hermes` parser.
- **Bug 2** — streaming `NO_TAG_CONTENT_THRESHOLD = 64` flipped routing to `content` mid-preamble for models that emit a chatty intro before `<think>`. New `VibeThinkerReasoningParser` subclass with 1024-char window (does NOT widen the base class — codex r2 P2).
- **Bug 3** — `finish_reason=length` with unclosed `<think>` duplicated the trace into BOTH `content` and `reasoning_content` byte-identical. Plug `_finalize_content_and_reasoning` (engine-routed + parser branch) AND gate `_rescue_silent_drop_from_reasoning` (#569) so the rescue does NOT surface reasoning as content when raw_text opens with an unclosed `<think>` and finish=length.

### Bundled fixes from PR #714 fuzz (Bugs A-D)

- **Bug A** — `qwen3` reasoning parser duplicates output on non-thinking Qwen3 variants. `qwen3-vl-2b-4bit` and `qwen3-4b-instruct-2507-4bit` do not emit `<think>...</think>` autonomously; the qwen3 parser's Case-4 fallback ("no tags + enable_thinking=True → all output is reasoning", #575) duplicated the entire response into BOTH `content` and `reasoning_content` when clients passed `enable_thinking=True`. **Fix:** clear `reasoning_parser` on both alias entries + add specific auto-config regex above the generic `qwen3` rule so bare-HF-path serves resolve the same way.
- **Bug B** — truncated-`<think>` leak plug is parser-agnostic. The codex-r3-refined plug from Bug 3 already fires on ANY parser whose `extract_reasoning` returns `(reasoning, None)` on unclosed-`<think>` input (Qwen3 parser inherits the same Case-3 semantics as DeepSeek-R1). **Action:** added regression tests pinning the plug against `Qwen3ReasoningParser` too so a future split can't silently regress.
- **Bug C** — `<think>` straddles SSE chunk boundaries. When the model emits the literal `<think>` open tag autonomously (phi-4-mini-reasoning / nanbeige4.1 family), the tag splits across SSE chunks (e.g. delta=`<thi` then `nk>`). The base parser's `_handle_explicit_think` "treat as content" fallback then leaked the trailing tag bytes (`nk>`) into `content` — fuzz repro showed `content=">\n", reasoning="<thinkOkay..."`. **Fix:** stateful suffix-withhold in `BaseThinkingReasoningParser.extract_reasoning_streaming` + cross-boundary recovery in `_handle_explicit_think`. Held bytes are flushed when they turn out NOT to be a tag (no silent drops on `<` followed by non-tag content).
- **Bug D** — `tool_call_parser=hermes` advertised on chat templates with no tool tokens. `phi-3.5-mini-4bit`, `gemma-3n-e2b-4bit`, `gemma-3n-e4b-4bit` ignore tool prompts (their chat templates carry no `<tool_call>` special tokens). **Fix:** `tool_call_parser=null` on the alias entries + matching auto-config regexes for the bare HF paths. `phi-4-mini-reasoning-4bit` deliberately STAYS at `hermes` (parser works; the model just needs `max_tokens=1024+` for the thinking block before the tool call — counter-pin test asserts this).

### Bug E (NEW from fuzz, NOT bundled)

Huge-prompt silent failure (`finish=length`, `completion_tokens=0`, `content=""`) on `qwen3-vl-2b-4bit`, `gemma-3n-e2b-4bit`, `gemma-3n-e4b-4bit`. PR #713 (just merged to base at `a7a8b50`) added `prefill_step_size=8192` bump for MLLM and surface per-batch-cap as HTTP 400 — the working hypothesis is that already fixes this. **Deferred** to a follow-up issue if the symptom recurs after this bundle lands.

### Unit tests added (this PR)

- `tests/test_aliases_contract.py` — split qwen3-small-models batch into thinking/non-thinking; 4-alias `tool_call_parser=null` parametrized test + counter-pin for `phi-4-mini-reasoning-4bit`.
- `tests/test_model_auto_config.py` — `test_qwen3_non_thinking_variants`, `test_gemma_3n_no_tool_calls`, `test_phi_3_5_no_tool_calls`.
- `tests/test_reasoning_parsers.py` — `TestThinkParserSSEBoundary` class (7 tests covering start/end tag straddles, 1-char split worst-case, false-prefix recovery, classic in-delta unaffected, qwen3 inheritance, no-tag stream pass-through).
- `tests/test_finalize_harmony_raw_text.py` — 3 Qwen3-parser variants of the truncated-`<think>` leak plug (parser-branch, enable_thinking=True, engine-routed).
- `tests/test_silent_drop_rescue_569.py` — 6 tests for the #569 rescue gate (truncated `<think>` skip, leading whitespace, counter-tests, finish=stop/None, closed `</think>` bypass).

### Live smoke (vibethinker-3b-8bit, port 8021 — from prior PR rounds)

- Tool call ("weather in Tokyo, Celsius") — `finish=tool_calls`, 1 clean `get_weather` call, content=null.
- Tool round-trip — model truncated mid-thought (8-bit quant noise), content correctly null.
- Streaming "2+2" — `finish=stop`, reasoning captured trace, content="4".
- Length-truncated math (max_tokens=2048) — `finish=length`, content=null, reasoning populated.

### Full test suite

`pytest tests/` excluding integration: **6059 passed, 19 skipped, 7 xfailed**. The four CI-blocking ruff-format-check warnings are fixed in commit `ef81575` (cosmetic-only, no behaviour change).

## Files changed

**vibethinker scope (already in PR):**
- `vllm_mlx/aliases.json` — vibethinker entries: `tool_call_parser: null → \"hermes\"`, `reasoning_parser: deepseek_r1 → vibethinker`.
- `vllm_mlx/model_auto_config.py` — vibethinker bare-HF-path regex updated.
- `vllm_mlx/reasoning/deepseek_r1_parser.py` — new `VibeThinkerReasoningParser` subclass (1024-char threshold).
- `vllm_mlx/reasoning/__init__.py` — register `vibethinker` parser.
- `vllm_mlx/service/helpers.py` — `_finalize_content_and_reasoning` plug (engine-routed + parser branch) + `_truncate_reasoning_only` helper + `_rescue_silent_drop_from_reasoning` finish=length gate.
- `vllm_mlx/routes/chat.py`, `vllm_mlx/routes/anthropic.py` — thread `finish_reason` + `raw_text` into the rescue call.
- `pyproject.toml` — 0.7.39 → 0.7.40.

**Fuzz-bundle additions:**
- `vllm_mlx/aliases.json` — clear `reasoning_parser` on `qwen3-vl-2b-4bit` + `qwen3-4b-instruct-2507-4bit`; clear `tool_call_parser` on `phi-3.5-mini-4bit` + `gemma-3n-e2b-4bit` + `gemma-3n-e4b-4bit`.
- `vllm_mlx/model_auto_config.py` — specific regex above generic family for qwen3-vl-2b/qwen3-4b-instruct, gemma-3n, phi-3.5.
- `vllm_mlx/reasoning/think_parser.py` — SSE-boundary withhold in `extract_reasoning_streaming` + cross-boundary recovery in `_handle_explicit_think` + new `_held_tag_suffix_len` state (cleared by `reset_state`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.7.39"
+version = "0.7.40"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_aliases_contract.py
+++ b/tests/test_aliases_contract.py
@@ -661,42 +661,80 @@ def test_vibethinker_family_wires_deepseek_r1_reasoning_parser(alias: str) -> No
     )
 
 
-@pytest.mark.parametrize(
-    "alias",
-    [
-        "qwen3-4b-thinking-2507-4bit",
-        "qwen3-4b-instruct-2507-4bit",
-        "qwen3-vl-2b-4bit",
-    ],
-)
-def test_qwen3_small_models_batch_wires_qwen3_parsers(alias: str) -> None:
-    """The Qwen3 ≤5B additions (Thinking-2507, Instruct-2507, VL-2B)
-    must all carry the Qwen3 family parser pair (``hermes`` for tool
-    calls + ``qwen3`` for reasoning).
+def test_qwen3_4b_thinking_2507_wires_qwen3_reasoning_parser() -> None:
+    """The Qwen3-4B-Thinking-2507 variant emits ``<think>...</think>``
+    blocks autonomously on every response; it MUST carry the
+    ``qwen3`` reasoning parser so the trace lands in
+    ``reasoning_content`` instead of leaking into ``content``.
 
-    The ``qwen3`` reasoning parser is the right pick even for the
-    non-thinking ``Instruct`` variant — the parser stays inert on
-    output that contains no ``<think>...</think>`` blocks and never
-    swallows content; pinning it keeps the wiring uniform across the
-    whole Qwen3 family and matches existing entries like
-    ``qwen3-4b-8bit`` / ``qwen3-vl-4b-4bit``.
-
-    Pinning here so a future PR can't silently revert the Thinking-2507
-    entry to ``reasoning_parser=null`` (which would leak ``<think>``
-    blocks into ``content``) or downgrade the Instruct-2507 entry away
-    from the family default (which would create a parser-mismatch hop
-    for clients that round-robin across Qwen3 sizes).
+    Pinned separately from the non-thinking siblings (Instruct-2507
+    + VL-2B) because the non-thinking variants must NOT carry the
+    parser — see the docstring on
+    ``test_qwen3_small_non_thinking_variants_have_no_reasoning_parser``
+    for the fuzz-evidence rationale (PR #715 bundle).
     """
     profiles = list_profiles()
+    alias = "qwen3-4b-thinking-2507-4bit"
     assert alias in profiles, f"{alias} missing from aliases.json"
     assert profiles[alias].tool_call_parser == "hermes", (
         f"{alias}: tool_call_parser must be 'hermes' (Qwen3 family default). "
         f"Got {profiles[alias].tool_call_parser!r}."
     )
     assert profiles[alias].reasoning_parser == "qwen3", (
-        f"{alias}: reasoning_parser must be 'qwen3' — the parser is inert "
-        f"on outputs without `<think>` blocks, so it's safe for both the "
-        f"Thinking and Instruct variants. Got {profiles[alias].reasoning_parser!r}."
+        f"{alias}: reasoning_parser must be 'qwen3' — the Thinking-2507 "
+        f"variant emits `<think>` blocks autonomously. "
+        f"Got {profiles[alias].reasoning_parser!r}."
+    )
+    assert profiles[alias].is_hybrid is False, (
+        f"{alias}: Qwen3-4B is pure-attention, not hybrid."
+    )
+
+
+@pytest.mark.parametrize(
+    "alias",
+    [
+        "qwen3-4b-instruct-2507-4bit",
+        "qwen3-vl-2b-4bit",
+    ],
+)
+def test_qwen3_small_non_thinking_variants_have_no_reasoning_parser(
+    alias: str,
+) -> None:
+    """The Qwen3-4B-Instruct-2507 and Qwen3-VL-2B-Instruct aliases are
+    NON-thinking variants — their model cards explicitly state no
+    ``<think>`` emission and the 2026-06-18 fuzz battery against PR
+    #714 confirmed the symptom: with ``reasoning_parser=qwen3`` wired
+    AND the client passing ``enable_thinking=True`` (or the parser's
+    Case-4 fallback firing on a no-tag output) the entire response
+    is duplicated into BOTH ``content`` AND ``reasoning_content``,
+    leaving a confusing assistant turn for the caller.
+
+    The qwen3 parser's Case-4 "no tags AND enable_thinking=True →
+    everything is reasoning" path is the load-bearing addition for
+    #575 — it's correct for actual thinking models but a footgun for
+    non-thinking variants that never produce ``<think>`` regardless
+    of the kwarg. Setting ``reasoning_parser=null`` short-circuits
+    the whole reasoning path so output flows directly to ``content``
+    as the non-thinking variants intend.
+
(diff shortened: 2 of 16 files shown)


raullenchai:
## Additional findings from parallel fuzz pass — please bundle if scope fits

A tool-call-edge fuzz agent just finished testing 5 of the 9 new aliases (the ones with working tool calls per the original VibeThinker live-test). **Two new actionable bugs surfaced. Please bundle if the cost is small; otherwise note as follow-up issues in the PR body.**

### F-1 (NEW server-side bug) — `tool_choice` forced selection ignored across 5/5 models tested

When a request sets `tool_choice={"type":"function", "function":{"name":"get_weather"}}` AND the user prompt is unrelated to the tool (e.g. `"Hi! Just say hello."`), **every model tested** returned `content="Hello..."` and `finish_reason="stop"` instead of the forced tool call.

The OpenAI spec contract is: "function" type with a named function means the model MUST call that function. Currently the server appears to pass `tool_choice` to the chat template prompt-builder but does not enforce a constrained-decode preamble or post-validate the assistant turn.

Models affected: granite4-h-micro-4bit, qwen3-vl-2b-4bit, nanbeige4.1-3b-4bit, qwen3-4b-instruct-2507-4bit, qwen3-4b-thinking-2507-4bit (5/5 → server-side bug, not per-model parser).

**Repro snippet:**
```python
client.chat.completions.create(
    model="<any of the 5 above>",
    messages=[{"role": "user", "content": "Hi! Just say hello."}],
    tools=[{
        "type": "function",
        "function": {
            "name": "get_weather",
            "parameters": {"type": "object", "properties": {"city": {"type": "string"}}}
        }
    }],
    tool_choice={"type": "function", "function": {"name": "get_weather"}},
)
# Expected: assistant message with tool_calls=[get_weather(...)], finish_reason="tool_calls"
# Actual:   content="Hello! How can I help you today?", finish_reason="stop"
```

Likely scope: a server-side `tool_choice` enforcement path. May or may not fit this PR's parser-focused scope — your call.

### F-3 — qwen3 reasoning_parser does NOT activate on qwen3-4b-thinking-2507-4bit

Despite `reasoning_parser=qwen3` being assigned to `qwen3-4b-thinking-2507-4bit`, the `<think>` block emits into `content` rather than `reasoning_content` on the nested-tool-call prompt (E1 in the fuzz battery). The `<think>` content then consumes the entire token budget before any tool call lands.

This is a different symptom from the original bug class B/C in your PR — same parser, different activation path. Suspect: the qwen3 parser's `<think>` token detection may be tuned for the older Qwen3-Thinking format and the 2507 variant uses a different opening token sequence or chat-template injection mode.

Models affected: `qwen3-4b-thinking-2507-4bit`. Possibly also nanbeige4.1 if its bug class C fix doesn't cover the autonomous-think pattern.

**Repro:** send any prompt to qwen3-4b-thinking-2507-4bit with `tools=[...]`, expect `<think>` content to start the response but be routed to `reasoning_content`; instead it appears in `content` and the model never reaches a tool call before token cap.

### Confirmed already in your PR's scope (no action — just FYI):
- F-2 (nanbeige4.1 `<tool_call>` leak in content) — covered by your `tool_call_parser=hermes` wiring + reasoning_parser fix
- F-6 (qwen3-4b-thinking + nanbeige4.1 collision returns empty tool_calls) — symptom of the same parser issues

### Pure model-behavior, NOT integration bugs (skip — open issue if desired):
- F-4 (granite4 loops on E3 round-trip error) — model just keeps retrying the same call
- F-5 (granite4 emitted 1 call when 2 were warranted) — model didn't pick parallel emission

---

Three more parallel fuzz agents are still running (adversarial, long-context, streaming/tokenizer). If they surface anything actionable I'll add another comment here before pr_validate finalizes.

cc agent `a5c46788756d4ea21` — this is the bundle agent.

PR #717, opened 2026-06-19
fix(reasoning): VibeThinker integration follow-up (Case-4 rescue + codex r1/r2/r4 P2 SSE-boundary)

raullenchai:
## Summary

Follow-up to PR #715 (merged at 6f12b31) bundling 4 additional bug fixes surfaced AFTER merge:

1. **Case-4 rescue gate** (`7f30fb4`) — live-smoke critical. Without this, VibeThinker asked "What is 2+2?" with `max_tokens=32` produced byte-identical `content` and `reasoning_content` (length 119 each). Root cause: parser Case-4 fallback fires when `enable_thinking=True` and no `<think>` tag is emitted, helper blanks `cleaned_text=\"\"`, then `_rescue_silent_drop_from_reasoning` (#569) surfaces reasoning as content, leaving both populated. The narrow `raw_text.lstrip().startswith(\"<think>\")` gate from the merged PR misses VibeThinker because it emits plain prose, not a literal `<think>` token. Adds `reasoning_is_case4` keyword-only param to the rescue helper plumbed through `routes/chat.py` and `routes/anthropic.py`; gates rescue on `(finish_reason == \"length\" and reasoning_is_case4)`.

2. **codex r1 P2** (`741851e`) — false partial-tag bytes inside an open `<think>` block (e.g. chunks `[\"<think>2 \", \"<\", \" 5</think>\", \"ans\"]`) were dropped instead of flushed. The `start_in_prev` branch had its own ad-hoc withhold logic that didn't thread `_held_tag_suffix_len`. Fixed by uniform emit-by-position bookkeeping across Case-3 and `start_in_prev`.

3. **codex r2 P2** (`d9ce51f`) — split-opener completing chunk (e.g. `[\"<thi\", \"nk>Okay\", \" more\"]`) duplicated reasoning to `\"OkayOkay more\"` because the held suffix from `<thi` wasn't cleared after the SSE-boundary recovery branch consumed it. Fixed by resetting `_held_tag_suffix_len = 0` in the recovery branch.

4. **codex r4 P2** (`e3e1122`) — when the same chunk that completes a split opener ALSO ends with a partial `</think>` (e.g. `[\"<thi\", \"nk>OK</thi\", \"nk>ans\"]`), the recovery branch emitted `\"OK</thi\"` as reasoning because it only checked for a complete end tag in the delta, never applying the partial-tag withhold. Subsequent chunks then treated the close as consumed, leaking literal `</thi` bytes into `reasoning_content`. Fixed by applying `_held_partial_tag_len` withhold to `reasoning_part` post-end-check.

## Tests added

- `tests/test_silent_drop_rescue_569.py` — 3 Case-4 tests (skip when length+case4, fire when case4+stop, fire when length+not-case4).
- `tests/test_reasoning_parsers.py::TestThinkParserSSEBoundary` — 4 new SSE-boundary tests on top of the 7 merged in PR #715:
  - `test_false_partial_tag_inside_think_block_flushed` (codex r1 P2)
  - `test_in_think_block_with_lt_and_gt_chars`
  - `test_split_opener_completed_with_reasoning_no_duplication` (codex r2 P2)
  - `test_split_opener_completes_with_partial_end_tag_no_leak` (codex r4 P2)

## Verification

- All 11 SSE-boundary tests pass (`tests/test_reasoning_parsers.py::TestThinkParserSSEBoundary`).
- All 159 reasoning parser tests pass.
- Full unit suite green (6146 passed, excluding integration tests that need a running server).
- Codex round 5 returned clean: \"The change correctly withholds partial closing tags when a split opener completes in the same chunk, and the added regression test covers the targeted scenario. I did not find any newly introduced blocking issues in the diff.\"

## Live smoke (pre-PR #715 merge, confirming Case-4 fix)

- VibeThinker-3B-8bit: `\"What is 2+2?\"` with `max_tokens=32` — pre-fix: `content_len == reasoning_len == 119`. Post-Case-4-gate: content empty (reasoning trace stays in `reasoning_content` only, no duplication).
- qwen3-4b-instruct-2507-4bit: regression-free under the new auto-config (no reasoning_parser).

## Test plan

- [ ] CI green on `fix/vibethinker-followup-codex-rounds`
- [ ] pr_validate clean (will run after PR opens)
- [ ] Manual smoke on `vibethinker-3b-8bit` to reconfirm Case-4 gate

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "rapid-mlx"
-version = "0.7.39"
+version = "0.7.40"
 description = "Rapid-MLX — AI inference for Apple Silicon. Drop-in OpenAI API, 2-4x faster than Ollama."
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_aliases_contract.py
+++ b/tests/test_aliases_contract.py
@@ -661,42 +661,80 @@ def test_vibethinker_family_wires_deepseek_r1_reasoning_parser(alias: str) -> No
     )
 
 
-@pytest.mark.parametrize(
-    "alias",
-    [
-        "qwen3-4b-thinking-2507-4bit",
-        "qwen3-4b-instruct-2507-4bit",
-        "qwen3-vl-2b-4bit",
-    ],
-)
-def test_qwen3_small_models_batch_wires_qwen3_parsers(alias: str) -> None:
-    """The Qwen3 ≤5B additions (Thinking-2507, Instruct-2507, VL-2B)
-    must all carry the Qwen3 family parser pair (``hermes`` for tool
-    calls + ``qwen3`` for reasoning).
+def test_qwen3_4b_thinking_2507_wires_qwen3_reasoning_parser() -> None:
+    """The Qwen3-4B-Thinking-2507 variant emits ``<think>...</think>``
+    blocks autonomously on every response; it MUST carry the
+    ``qwen3`` reasoning parser so the trace lands in
+    ``reasoning_content`` instead of leaking into ``content``.
 
-    The ``qwen3`` reasoning parser is the right pick even for the
-    non-thinking ``Instruct`` variant — the parser stays inert on
-    output that contains no ``<think>...</think>`` blocks and never
-    swallows content; pinning it keeps the wiring uniform across the
-    whole Qwen3 family and matches existing entries like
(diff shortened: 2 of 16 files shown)


raullenchai:
Replaced by cleaner branch (cherry-picked only the 4 follow-up commits onto fresh main); see follow-up PR.

PR #751, opened 2026-06-20
fix(api): hint case-sensitivity on tool_choice name mismatch (F-145)

raullenchai:
## Summary

F-145: when `tool_choice` references a function name that only differs from a known tool by letter casing, the 400 just said "not present in the 'tools' array" - technically correct (OpenAI is case-sensitive too) but the client had no way to know casing was the cause.

Fix: in `routes/chat.py` after the strict-equality lookup fails, do a secondary case-insensitive scan. If a name matches case-insensitively, append `(Did you mean '<X>'? names are case-sensitive)` to the existing 400 message. If no case-match exists the message is unchanged, so non-case typos never trigger a misleading hint.

## Test plan

Verified on `qwen3-0.6b-8bit @ 127.0.0.1:8046`:

```
# tool=get_Weather, tool_choice=get_weather (case mismatch)
curl ... '{"tools":[{"function":{"name":"get_Weather",...}}],"tool_choice":{"type":"function","function":{"name":"get_weather"}}}'
# BEFORE: 400 "tool_choice references function 'get_weather' which is not present in the 'tools' array"
# AFTER:  400 "tool_choice references function 'get_weather' which is not present in the 'tools' array. (Did you mean 'get_Weather'? names are case-sensitive)"

# tool=get_Weather, tool_choice=foo_bar (no case-match)
# BEFORE/AFTER: 400 "...not present in the 'tools' array" (no hint, unchanged)
```

- [x] Added `test_tool_choice_specific_function_case_mismatch_returns_400_with_hint`
- [x] Pinned negative branch in `test_tool_choice_specific_function_unknown_name_returns_400` - no-match name must NOT trigger the misleading hint
- [x] All 41 tests in `tests/test_chat_route_tool_choice_enforcement.py` pass

Closes F-145.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_chat_route_tool_choice_enforcement.py
+++ b/tests/test_chat_route_tool_choice_enforcement.py
@@ -210,6 +210,50 @@ def test_tool_choice_specific_function_unknown_name_returns_400():
     )
     assert resp.status_code == 400
     assert "nonexistent_function" in resp.text
+    # F-145: a no-match name (no case-insensitive partner either) must
+    # NOT trigger the "Did you mean" hint - it would mislead the client
+    # into chasing a non-existent typo.
+    assert "Did you mean" not in resp.text
+
+
+def test_tool_choice_specific_function_case_mismatch_returns_400_with_hint():
+    """F-145: when ``tool_choice`` references a name whose ONLY difference
+    from a known tool is letter casing, the 400 must include a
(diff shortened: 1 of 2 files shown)


raullenchai:
Superseded by PR #750 which merged the same F-145 fix while this PR was in flight. Closing as duplicate.

PR #755, opened 2026-06-20
revert: CORS env-configurable opt-in (#738)

raullenchai:
Single-machine local inference server — CORS default-deny is unnecessary overhead. Reverting per operator decision. Restores default `Access-Control-Allow-Origin: *` behavior.

Threat: malicious webpage hitting localhost. Mitigation already exists at OS / browser level (private-network requests are gated) and rapid-mlx has no secrets to exfiltrate.

Refs: F-090 / F-091.


raullenchai:
Closing in favor of option C (minimal in-place change rather than full revert). Codex flagged 4 BLOCKING on the bare revert because it reintroduced pre-#738 spec issues (wildcard + allow_credentials=True). Doing it as a small forward patch keeps env var infrastructure + spec compliance.

PR #773, opened 2026-06-20
fix(api): propagate F-061 stop matcher to /v1/completions and /v1/messages (H-03)

raullenchai:
## Summary
- **H-03** (`0.8TODO.md` HIGH tier) — Petra (r2, `/v1/completions`) and Sergei (r2, `/v1/messages`) reported `stop` / `stop_sequences` silently ignored. Investigation showed the scheduler-level matcher from #716 was actually firing on all three routes (text trimmed correctly), but `/v1/messages` had no wire-level adapter for the Anthropic `stop_sequence` field — the engine knew which stop fired and the response never said so.
- Fix plumbs the matched-stop bytes from the schedulers through `RequestOutput` / `GenerationOutput` / `EngineCore` aggregation / `RequestOutputCollector` aggregation / `BatchedEngine` (incl. router-wrapped streams) into `openai_to_anthropic` — which now emits `stop_reason="stop_sequence"` + `stop_sequence: <str>` per Anthropic's public spec. `/v1/messages` non-stream + stream both surface it; `/v1/completions` and `/v1/chat/completions` keep `finish_reason="stop"` (OpenAI lumps EOS and stop under one bucket per spec).
- One source of truth: matched_stop is pinned exactly once at the scheduler layer next to the existing #716 stop-string check; every downstream wrapper preserves it. No per-model logic; no route-local stop matcher.

## Repros (Qwen3-0.6B-4bit, port 8802)

**Sergei — `/v1/messages` non-stream**
```bash
curl -X POST .../v1/messages \
  -d '{"model":"...","max_tokens":150,"stop_sequences":["TESTSTOP"],
       "messages":[{"role":"user","content":"Output hello world TESTSTOP then more"}]}'
```
| | `stop_reason` | `stop_sequence` |
|---|---|---|
| before | `end_turn` | (absent) |
| after | `stop_sequence` | `TESTSTOP` |

**Sergei — `/v1/messages` streaming** — terminal `message_delta`:
- before: `{"stop_reason":"end_turn","stop_sequence":null}`
- after: `{"stop_reason":"stop_sequence","stop_sequence":"TESTSTOP"}`

**Petra — `/v1/completions`**
```bash
curl -X POST .../v1/completions \
  -d '{"prompt":"def fibonacci(n):","stop":["\n\n","def "],"max_tokens":64}'
```
Already trimming correctly post-#716 (`finish_reason="stop"`, no stop bytes in `text`). Pinned with new regression tests in `tests/test_stop_matcher_completions.py` so a future spec-parity refactor cannot silently drop the route-layer forwarding.

## Test plan
- [x] `tests/test_stop_matcher_completions.py` (new, 6 tests) — wire-level forwarding to `engine.generate` / `stream_generate`, multi-alternative stop list, empty stop, omitted stop, single-element list, non-stream + stream parity.
- [x] `tests/test_anthropic_stop_sequences.py` — extended from 3 → 9 tests covering matched-stop surfacing in both non-stream and stream `message_delta` events, multi-alternative match-vs-emitted, no-match fallback to `end_turn`, empty-list no-op.
- [x] `tests/test_anthropic_adapter.py` — 4 new unit tests on `openai_to_anthropic(..., matched_stop=...)` covering promotion to `stop_sequence`, None fallback, mutual exclusivity vs `max_tokens` and `tool_use`.
- [x] `tests/test_server_utils.py::TestGenerationOutputFieldOrder` updated to include `matched_stop` as the new trailing field (preserves the v0.6.65 positional-construction contract).
- [x] Full suite green (6819 passed, 19 skipped, 16 deselected, 7 xfailed) — 4 pre-existing unrelated failures (`test_event_loop` needs a live server; `test_no_out_of_band_routing` + `test_route_engine_contract` are pre-existing post-#767 / pre-existing in completions.py respectively).
- [x] Live curl repros above on `mlx-community/Qwen3-0.6B-4bit`.

### DIFF
--- a/tests/test_anthropic_adapter.py
+++ b/tests/test_anthropic_adapter.py
@@ -574,6 +574,60 @@ class TestOpenaiToAnthropic:
         result = openai_to_anthropic(resp, "default")
         assert result.stop_reason == "max_tokens"
 
+    # H-03 — matched_stop surface
+    def test_matched_stop_promotes_end_turn_to_stop_sequence(self):
+        """When the engine surfaces a matched stop string the adapter
+        must rewrite ``stop_reason="stop"`` (which would otherwise
+        map to ``end_turn``) to Anthropic's dedicated
+        ``stop_sequence`` value and populate the ``stop_sequence``
+        field verbatim. Pre-fix the adapter always emitted
+        ``end_turn`` with ``stop_sequence: None`` even when a stop
+        fired."""
+        resp = self._make_response(content="say ", finish_reason="stop")
+        result = openai_to_anthropic(resp, "default", matched_stop="END")
+        assert result.stop_reason == "stop_sequence"
+        assert result.stop_sequence == "END"
+
+    def test_matched_stop_none_keeps_end_turn(self):
+        """No matched stop means EOS / length / no-stop; the legacy
+        ``stop → end_turn`` mapping holds and ``stop_sequence`` stays
+        ``None``."""
+        resp = self._make_response(finish_reason="stop")
+        result = openai_to_anthropic(resp, "default", matched_stop=None)
+        assert result.stop_reason == "end_turn"
+        assert result.stop_sequence is None
+
+    def test_matched_stop_does_not_override_max_tokens(self):
+        """A length cap MUST still win — Anthropic's stop_reason
+        values are mutually exclusive, and the matched-stop rewrite
+        only applies to the ``end_turn`` bucket. Otherwise a request
+        that ran past max_tokens AND happened to also include the
+        stop bytes mid-output would be mis-classified."""
+        resp = self._make_response(finish_reason="length")
+        result = openai_to_anthropic(resp, "default", matched_stop="END")
+        assert result.stop_reason == "max_tokens"
+        # ``stop_sequence`` MUST stay None — Anthropic spec invariant:
+        # "stop_sequence is set iff stop_reason == 'stop_sequence'".
+        assert result.stop_sequence is None
+
+    def test_matched_stop_does_not_override_tool_use(self):
+        """Same invariant for the tool_use bucket: a tool_calls
+        finish must keep stop_reason='tool_use' even if a stop
+        string happened to surface earlier in the response."""
+        tc = ToolCall(
+            id="call_1",
+            type="function",
+            function=FunctionCall(name="search", arguments='{"q":"x"}'),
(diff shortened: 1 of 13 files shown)


raullenchai:
**pr_validate scorecard:** PASS on codex (round 1, 14.4s, no blocking issues), lint, supply chain, test plan, targeted tests (454 passed), and 8/8 PASS steps total bar two pre-existing/environmental ones:

- `full_unit` FAIL: 3 failed (1 flaky, 2 pre-existing on origin/main)
  - `test_no_routing_shaped_rapid_mlx_env_vars` — pre-existing since #767 added `RAPID_MLX_MAX_GENERATION_TOKENS` without allowlisting (verified against fresh main clone).
  - `test_no_hasattr_engine_guard[completions.py]` — pre-existing in `completions.py:170` `hasattr(engine, 'stream_generate')` since #735, verified against fresh main clone.
  - `test_add_request_cancellation_aborts_after_executor_completes` — flaky under load (passes when run alone; output collector cleanup race).

- `stress_e2e_bench` FAIL: 3/4 models GREEN (Qwen3.5-35B, Qwen3.6-27B, gpt-oss-20b all 8/8 stress + SDK 5/5 + langchain 6/6 + pydantic_ai 6/6). 4th model (diffusion-gemma) shows the **same pre-existing pattern** that #762 + #763 merged with: stress 7/8 (Long-generation slow on diffusion-gemma), Anthropic SDK 3/5 (`1_plain` + `2_system` return empty content), LangChain 4/6 (same plain/system empty replies). Verified by diffing artifacts from `/tmp/pr_validate/pr-763/agent-*-diffusiongemma-*.log` — bit-for-bit same failure shape. Not introduced by this PR; H-03 touches only the `stop_sequence` surface, which has nothing to do with the empty-content issue.

Independent live curl repros confirm the fix works as advertised on Qwen3-0.6B-4bit:

- `/v1/messages stop_sequences=["TESTSTOP"]` (non-stream) → `stop_reason="stop_sequence"` + `stop_sequence="TESTSTOP"` (was `end_turn` + absent pre-fix).
- `/v1/messages stop_sequences=["TESTSTOP"] stream=true` → terminal `message_delta` carries the same fields (was `end_turn` + null pre-fix).
- `/v1/completions stop=["\n\n","def "]` → already trimming correctly post-#716; now pinned with new regression tests.

PR #778, opened 2026-06-20
fix(api): extend F-011 finite-range validation to all sampling params (H-10)

raullenchai:
## Summary

Provenance: **H-10** in `0.8TODO.md`. F-011 (#740) closed NaN/inf rejection on `temperature` / `top_p` plus Field bounds on `presence_penalty` / `frequency_penalty` for the OpenAI routes; this PR sweeps the **same hardening across every remaining sampling param on every route**, eliminating the silent-burn class at the source.

### H-10 repro (uvicorn dies)

```python
client.chat.completions.create(..., extra_body={"repetition_penalty": -1.0})
```

slipped past every schema gate → reached `mlx_lm/sample_utils.py:298` (`make_repetition_penalty` raises `ValueError` on `penalty < 0`) → the un-caught `ValueError` escaped the request scheduler → uvicorn went down → port dead.

### Coverage table

| Param                  | `/v1/chat/completions` | `/v1/completions` | `/v1/messages` |
|------------------------|------------------------|-------------------|----------------|
| `temperature`          | F-011                  | F-011             | **NEW** (Anthropic `[0, 1]`) |
| `top_p`                | F-011                  | F-011             | **NEW** (`(0, 1]`)           |
| `top_k`                | **NEW** (`>= 0`)       | **NEW** (`>= 0`)  | **NEW** (`>= 0`)             |
| `min_p`                | F-011                  | F-011             | n/a            |
| `repetition_penalty`   | **NEW** (`[0, 2]`) `**H-10 root**` | **NEW** (`[0, 2]`)  | n/a            |
| `presence_penalty`     | F-011                  | F-011             | n/a            |
| `frequency_penalty`    | F-011                  | F-011             | n/a            |
| `logit_bias` values    | **NEW** (finite)       | n/a (not declared) | n/a           |

### One source of truth

Three shared helpers live next to the existing F-011 scrub helper in `vllm_mlx/api/models.py`:

- `_validate_finite_in_range(v, *, min_value, max_value, min_inclusive, max_inclusive, field_name)` — float range + finite check
- `_validate_nonnegative_int(v, *, max_value, field_name)` — int range; accepts integer-valued floats (`top_k: 64.0`) for Pydantic-v2-lax-mode parity
- `_validate_logit_bias_finite(v, *, field_name)` — finite-value check on `logit_bias` entries

`AnthropicRequest` imports them explicitly so the OpenAI-side `_scrub_nonfinite_sampling_raw` + finite + range stack applies identically on `/v1/messages` — no route-specific divergence.

### Why these ranges

- **Anthropic `temperature` `[0, 1]`** (narrower than OpenAI's `[0, 2]`): matches `docs.anthropic.com/en/api/messages` spec.
- **`repetition_penalty` `[0, 2]`**: mlx-lm requires `>= 0` (the root H-10 ValueError); upper cap matches HF/vLLM convention. Above 2 the distribution degenerates to a single token.
- **`top_k >= 0`**: mlx-lm treats 0 as "disabled" (`> 0` enables top-k mask). M-14 already noted negative values silent-ignore; H-10 sweep mandate is to 4xx instead.
- **`logit_bias` finite**: defensive — the chat route already 4xx's non-empty `logit_bias`, but values are not inspected; a future PR that wires the field through to a logits processor would re-create the F-011 silent-burn surface without this gate.

### Compatibility

Pure tightening — only rejects shapes that either (a) the legacy mlx-lm path also rejected with an un-caught exception (H-10 root) or (b) F-011's NaN/inf class extended to a new param. `top_k: 64.0` (Pydantic lax float-to-int coercion) is preserved.

## Test plan

- [x] Reproduce on pre-fix branch: `repetition_penalty=-1.0` accepted by schema, would reach mlx-lm and crash uvicorn. Audit table dumped to `/tmp/h10-before.txt`.
- [x] Audit every sampling param × every request model × every route. Legal ranges documented in the shared helper's docstring block.
- [x] Implement shared helpers + wire them into `ChatCompletionRequest`, `CompletionRequest`, `AnthropicRequest`. One source of truth: every new sampling param has one place to wire through.
- [x] New test file `tests/test_sampling_param_finite_range.py`:
  - [x] Each float param × `{NaN, +inf, -inf, below-min, above-max, at-exclusive-boundary}` × route — all 4xx-clean with `invalid_request_error` envelope + field name in message.
  - [x] Each float param × `{boundary-low, boundary-high, valid-mid}` × route — all parse cleanly through the schema.
  - [x] `top_k` × `{-1, -1000}` × every route — 4xx-clean.
  - [x] `top_k` × `{0 (disabled), 40 (typical), 1000 (large)}` × every route — accepted.
  - [x] `logit_bias` × `{NaN, +inf, -inf, bool}` value — 4xx-clean.
  - [x] H-10 exact repro test (`repetition_penalty=-1.0`) on both OpenAI routes.
  - [x] 50-shot server-survival burst across every route — proves the scheduler-escape path stays closed under load.
  - [x] Direct unit checks on all three shared helpers.
- [x] Reproduce post-fix: server returns 400 envelope, does NOT die. Survival probe after 50 bad shapes still serves. Dumped to `/tmp/h10-after.txt`.
- [x] All 140 new tests pass + all 109 existing F-011 tests pass + 142 Anthropic tests pass + full suite green (one pre-existing failure in `test_no_out_of_band_routing` for `RAPID_MLX_MAX_GENERATION_TOKENS` allowlist gap — present on `origin/main`, unrelated).
- [x] Codex review.

### DIFF
--- a/tests/test_sampling_param_finite_range.py
+++ b/tests/test_sampling_param_finite_range.py
@@ -1,0 +1,802 @@
+# SPDX-License-Identifier: Apache-2.0
+"""H-10: finite-range validation sweep on every sampling param.
+
+Provenance: F-011 (#740) closed the NaN / inf / out-of-range gap on
+``temperature`` + ``top_p`` and the ``presence_penalty`` /
+``frequency_penalty`` Field bounds on the OpenAI routes. The H-10 bug
+report showed the same hole was wide open on the other sampling
+params: ``repetition_penalty=-1.0`` slipped past the schema, reached
+``mlx_lm/sample_utils.py:298`` (``make_repetition_penalty`` raises on
+``penalty < 0``), the un-caught ``ValueError`` escaped the request
+scheduler, and uvicorn went down — port dead. Same silent-burn class
+as F-011, just on a param F-011 didn't cover.
+
+H-10's mandate: don't whack-a-mole the one repro. Sweep every
+sampling param across every route, define the legal range, wire one
+shared helper. This test pins the contract that emerged:
+
+Sweep matrix per param × per route × per shape:
+
+  Routes: /v1/chat/completions, /v1/completions, /v1/messages
+  Shapes: NaN, +inf, -inf, below-min, above-max,
+          boundary-low, boundary-high, valid-mid
+
+Params covered (route gates each one with a Field bound + shared
+``_validate_finite_in_range`` / ``_validate_nonnegative_int`` helper):
+
+  * temperature
+  * top_p
+  * top_k
+  * min_p              (chat + completions only)
+  * repetition_penalty (chat + completions only — H-10 root bug)
+  * presence_penalty   (chat + completions only)
+  * frequency_penalty  (chat + completions only)
+  * logit_bias values  (chat only — Anthropic surface has no
+                       equivalent; legacy completions never declared it)
+
+The test asserts every bad shape returns the project's unified
+``invalid_request_error`` 400 envelope (production handler in
+``vllm_mlx.middleware.exception_handlers``) AND that the server stays
+alive across the full burst (no port death — the H-10 symptom this
+PR closes).
+"""
+
+from __future__ import annotations
+
+import json
+import math
+from unittest.mock import MagicMock
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+# ---------------------------------------------------------------------------
+# Per-test app fixtures — stub the engine so we hit only the Pydantic +
+# route-layer validators, never the real sampler.
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def patched_config():
+    """Patch the global config singleton and restore on teardown."""
+    from vllm_mlx.config import get_config
+
+    cfg = get_config()
+    saved: dict = {}
+
+    def patch(**kwargs):
+        for k, v in kwargs.items():
+            saved.setdefault(k, getattr(cfg, k, None))
+            setattr(cfg, k, v)
+
+    yield patch
+
+    for k, v in saved.items():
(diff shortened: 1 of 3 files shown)


raullenchai:
## pr_validate scorecard

| Step | Result |
|---|---|
| `fetch` | PASS (3 files, +1114/-5 LOC) |
| `test_plan_check` | PASS (all 16 boxes checked) |
| `cl_description_quality` | PASS |
| `codex_review` | **PASS — codex found no blocking issues** (13.4s) |
| `supply_chain` | PASS |
| `lint` | PASS (ruff clean) |
| `targeted_tests` | PASS (288 passed) |
| `full_unit` | FAIL — 2 pre-existing failures, unrelated (see below) |
| `stress_e2e_bench` | skipped (medium blast radius) |

### Pre-existing `full_unit` failures (not caused by this PR)

1. `tests/test_no_out_of_band_routing.py::test_no_routing_shaped_rapid_mlx_env_vars`
   - `api/models.py` references `RAPID_MLX_MAX_GENERATION_TOKENS` (introduced by the **M-04 generation-budget ceiling** validator) which was never added to `ALLOWED_RAPID_MLX_ENV_VARS`.
   - Verified failing identically on `origin/main` HEAD (`69809da`) without this PR's changes applied.
2. `tests/test_route_engine_contract.py::test_no_hasattr_engine_guard[completions.py]`
   - `completions.py:170` carries a `hasattr(engine, 'stream_generate')` guard that was never added to the contract's allowlist.
   - Verified failing identically on `origin/main` HEAD without this PR's changes applied.

Both are tracker-shaped tooling debt unrelated to H-10's sampling-param scope.

PR #816, opened 2026-06-22
fix(routes): resolve_model() on embeddings + audio request handlers (R-03 model=default, R-04 short alias)

raullenchai:
## Summary

0.8.5 dogfood (Priya **F-1/F-2**, Karim **F-K-MODEL-DEFAULT-NO-RESOLVE** + **F-K-OPENAI-ALIAS-GAP**) caught two related drop-in-compatibility gaps:

- **R-03** `model="default"` — `/v1/embeddings` 400s and `/v1/audio/transcriptions` 404s the OpenAI-spec `"default"` placeholder that LangChain / LlamaIndex / openai-python emit when the caller hasn't picked a specific id. Chat works because `_resolve_model_name` maps `"default"` to `cfg.model_name`; the other routes never wired the same rule.
- **R-04** short-alias rejection — `/v1/embeddings` rejected `embeddinggemma-300m-6bit` even though that exact short alias was the `--embedding-model` CLI value, resolves cleanly through `model_aliases.resolve_model`, AND `/v1/models` advertises it. Root cause: D-EMBED-ALIAS (PR #805) resolved the alias at boot and locked `cfg.embedding_model_locked` to the resolved HF path; the route did a literal string-equality match — the short alias no longer compared equal to itself.

## Fix shape — single source of truth, no per-route regex

- New `_resolve_request_alias_or_default(request_model, locked)` helper in `vllm_mlx/service/helpers.py`:
  - `None` / `""` / `"default"` → `locked` verbatim (OpenAI-spec placeholder).
  - Both sides run through `resolve_model()` so the alias and the HF path compare equal.
  - Returns `None` on miss; caller decides 400 vs 404 (the envelope rule differs between embeddings and audio routes, #805).
- `/v1/embeddings` calls the helper; rejection envelope upgraded to the OpenAI-canonical `error.{message,type,code,param}` shape with `param="model"` so SDK error branches fire cleanly.
- `/v1/audio/transcriptions` `_resolve_stt_model` accepts `"default"` → `whisper-large-v3` (the route's existing default when the form field is omitted); every other unknown bare name still 404s with `model_not_found_error` (F-167 / F-210 contract intact).
- `/v1/audio/speech` maps `"default"` → `"kokoro"` (mirror of the route's omitted-field default).
- No version bump.

### R-03 — `model="default"` on `/v1/embeddings`

**Before:** `400 Bad Request — "Embedding model 'default' is not available. This server was started with --embedding-model mlx-community/embeddinggemma-300m-6bit. Only 'mlx-community/embeddinggemma-300m-6bit' can be used for embeddings. Restart the server with a different --embedding-model to use 'default'."`

**After:** `200 OK` — embedding vector returned, `response.model == "mlx-community/embeddinggemma-300m-6bit"` (echoes locked id for cache-key consistency).

### R-04 — short alias on `/v1/embeddings`

**Before:** Same 400 envelope as R-03, with `'embeddinggemma-300m-6bit'` in the message.

**After:** `200 OK` — identical vector to the full-HF-path call, `response.model == "mlx-community/embeddinggemma-300m-6bit"`.

### R-03 — `model="default"` on `/v1/audio/transcriptions`

**Before:** `404 Not Found — "The model 'default' does not exist. Available STT aliases: parakeet, parakeet-v3, whisper-large-v3, ..."`

**After:** Passes alias resolution and reaches `STTEngine.load("mlx-community/whisper-large-v3-mlx")` — identical to the explicit `whisper-large-v3` form.

## Single-helper-reuse rationale (per operator note)

The operator brief explicitly forbids per-route regex band-aids — every route that takes a user-supplied `model` field must call the same `resolve_model()`-anchored helper that boot uses. The fix introduces ONE helper (`_resolve_request_alias_or_default`) and ONE sentinel rule (`"default"` → locked) that:

- Embeddings route consumes directly.
- Audio STT consumes via the existing `_resolve_stt_model` (which has its own non-`aliases.json` registry — `whisper-*` and `parakeet-*`).
- Audio TTS consumes via the existing `model_map` lookup.
- Chat route is cross-checked (its `_resolve_model_name` already does the right thing for `"default"`); a pin test guards the contract.

The `"default"` sentinel does NOT get adversarial validation — per operator note, it is a documented OpenAI placeholder.

## Test plan

- [x] `tests/test_request_time_alias_resolution.py` — 20 cases pass:
  - Helper: `None` / `""` / `"default"` → locked, short alias <-> HF path round-trip, unknown alias → None, `locked=None` → None, `resolve_model` exception fallback (no 500 on corrupt registry).
  - Embeddings TestClient: `default` / short alias / full HF path all hit the engine; bogus alias 400s with `error.param=="model"` + `error.code=="model_not_found"`; no-config still 400s with the H-09 install-hint envelope (`"default"` does NOT silently fall through to chat hidden states).
  - Audio: `"default"` → `whisper-large-v3` HF path; bogus alias still 404 `model_not_found`; known-alias + HF-id pass-through preserved; empty string still 400.
  - Chat cross-check: `_resolve_model_name("default")` still maps to `cfg.model_name`.
- [x] Live server boot (`rapid-mlx serve qwen3-0.6b-8bit --embedding-model embeddinggemma-300m-6bit --port 9191`) — Priya/Karim repro probes all green (default + short alias + full HF path return same vector).
- [x] `ruff check vllm_mlx/ tests/` — clean.
- [x] `ruff format --check` — clean.
- [x] `python -m pytest tests/ -k 'embedding or audio or alias or resolve' -x` — 1649 passed, 3 skipped.
- [x] Full `tests/` (minus integrations that need extra deps) — 7956 passed, 6 pre-existing failures (verified by stashing the patch + re-running — all 6 fail on `main` too; none introduced by this PR).
- [x] Pre-existing 422 on `test_routes.py::test_embeddings_locked_model_reject` repaired in the same touch (the patch'd MagicMock `(*args, **kwargs)` signature broke FastAPI introspection — replaced with a clean `async def` no-arg callable).

### DIFF
--- a/tests/test_request_time_alias_resolution.py
+++ b/tests/test_request_time_alias_resolution.py
@@ -26,21 +26,57 @@ band-aids.
 
 from __future__ import annotations
 
-import platform
 import sys
+import types
+from contextlib import contextmanager
 from unittest.mock import MagicMock
 
 import pytest
 
-# Most tests below stub the embedding/STT engines, but the model-aliases
-# load path itself reads ``aliases.json``. That file is checked into the
-# repo and works cross-platform, so we don't gate on Apple Silicon for
-# the helper unit tests. We DO gate the route-level integration tests
-# because ``vllm_mlx.server`` imports MLX at module level on macOS-ARM.
-_APPLE_SILICON_ONLY = pytest.mark.skipif(
-    sys.platform != "darwin" or platform.machine() != "arm64",
-    reason="Server route imports MLX which is Apple-Silicon only",
-)
+# All tests run cross-platform. The route-level integration tests below
+# import ``vllm_mlx.routes.embeddings`` + ``vllm_mlx.routes.audio``
+# directly (which do NOT pull in MLX at module load), and inject a fake
+# ``vllm_mlx.server`` module into ``sys.modules`` so the route's lazy
+# ``from ..server import load_embedding_model`` lookup never drags in
+# the MLX-importing real server module.
+#
+# Codex r0 BLOCKING on PR #816: pinning the route fix only on
+# Apple-Silicon CI silently leaked the regression past Linux CI on
+# every prior PR — fix is to make the route tests CI-safe by mocking
+# the server module, not by skipping.
+
+
+@contextmanager
+def _fake_server_module(
+    embedding_engine=None,
+    embedding_model_locked=None,
+    load_fn=None,
+):
+    """Inject a stub ``vllm_mlx.server`` into ``sys.modules``.
+
+    The route handler does lazy imports
+    (``from ..server import load_embedding_model``,
+    ``from ..server import _embedding_engine``,
+    ``from ..server import _embedding_model_locked``) so the only
+    surface the test must cover is the ``vllm_mlx.server`` name in
+    ``sys.modules``. A real import of that module on Linux CI fails
+    with ``ModuleNotFoundError: mlx.core`` — the stub keeps the test
+    cross-platform.
+    """
+    name = "vllm_mlx.server"
+    prev = sys.modules.get(name)
+    fake = types.ModuleType(name)
+    fake._embedding_engine = embedding_engine
+    fake._embedding_model_locked = embedding_model_locked
+    fake.load_embedding_model = load_fn or (lambda *a, **kw: None)
+    sys.modules[name] = fake
+    try:
+        yield fake
+    finally:
+        if prev is not None:
+            sys.modules[name] = prev
+        else:
+            sys.modules.pop(name, None)
 
 
 # ──────────────────────────────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


raullenchai:
pr_validate r1 — codex BLOCKING addressed, remaining failures are infra-only.

- `codex_review` PASS — the r0 BLOCKING (Apple-Silicon-only gate on `/v1/embeddings` route tests) was fixed by injecting a stub `vllm_mlx.server` module via `sys.modules`. All 20 route + helper tests now run cross-platform, verified by re-running the test file in a subprocess with a `meta_path` finder that import-blocks the real server module (all 20 still pass).
- `targeted_tests` 222 passed.
- `full_unit` 4 failures, all pre-existing on `main` — verified by stashing this PR's patch + re-running on the same `main` SHA: same 4 tests fail. Not introduced here:
  - `test_anthropic_tool_usage_d::test_nonstream_tool_choice_named_still_routes_to_specific_tool`
  - `test_no_out_of_band_routing::test_no_routing_shaped_rapid_mlx_env_vars`
  - `test_response_format_json_schema_strict::test_strict_true_with_outlines_module_faked_absent`
  - `test_route_engine_contract::test_no_hasattr_engine_guard[completions.py]`
- `stress_e2e_bench` BLOCKING is a port-8451 collision from a concurrent agent's bench run on the same host (PID 23513, also serving `mlx-community/Qwen3.5-35B-A3B-8bit` from a different worktree). Not refusing to clobber — it's another agent's process. Independent of this PR's diff.

Lint clean, targeted tests green, codex clean. Ready to merge.

PR #910, opened 2026-06-25
feat(kv-cache): default --kv-cache-dtype=int4 with safe-fallback profiles (R15 #300)

raullenchai:
## Summary

R15 task #300 — flip the KV cache dtype default to `int4` via
``mlx_lm.QuantizedKVCache``, with auto-downgrade for the architectures
that break at sub-bf16 precision and an opt-in reasoning profile.

Apple Silicon decode is memory-bandwidth-bound; a 4×-smaller KV cache
gives 4× less bandwidth on every decode step (mlx#3134 UMA discussion,
Feb 2026 — Phi-3.5-mini +1.1% throughput, 3.2× more context room on
Qwen2.5-14B). mlx-lm 0.31.3 already ships `QuantizedKVCache` — this PR
is the config plumbing that flips the default safely.

### What ships

- New CLI flag `--kv-cache-dtype {bf16,int8,int4}`, default `int4`.
- New `--reasoning` flag pins to `int8` (sub-4-bit drops -20pt on
  AIME-class hard math on Qwen3 thinking variants).
- Safelist auto-downgrades `int4` → `bf16` for:
  - **Sliding-window attention** (Gemma 3, GPT-OSS) — rotating buffer
    can't tolerate arbitrary-boundary quant blocks.
  - **MLA** (DeepSeek V3+, Kimi K2.5) — K projection already compressed
    via `q_lora_rank`/`kv_lora_rank`; stacking int4 compounds error.
- Detection is layered: alias metadata → HF `config.json` field →
  name-substring fallback. No frozen-dataclass schema bump needed —
  in-tree aliases match the substring patterns today; the alias-
  metadata hook stays available for future curated overrides.
- Operator-facing startup log line stamps the resolved dtype + reason
  (`"Defaulting to int4 (memory-bandwidth-bound on M-series); model=...
  not in safelist"`) so a silent downgrade is never invisible.
- Prometheus gauge `rapid_mlx_kv_cache_dtype{dtype="..."}` emits three
  series with exactly one set to 1 — surfaces the EFFECTIVE dtype
  (post-safelist) for Grafana dashboards.
- Legacy `--kv-cache-quantization` flag preserved; when both are passed
  the legacy one wins for backwards compatibility (with an explicit
  rejection when paired with `--reasoning` + `--kv-cache-quantization-bits 4`,
  per codex r1 BLOCKING #1).

### Safelist contents

- **Sliding-window**: substrings `gemma-3`, `gemma3`, `gpt-oss`,
  `gpt_oss`; `hf_config.sliding_window > 0`; `model_type ∈ {gemma3,
  gemma3_text, gpt_oss}`.
- **MLA**: substrings `deepseek-v3`, `deepseek_v3`, `deepseek-v4`,
  `deepseek_v4`, `kimi-k2`, `kimi_k2`; canonical `model_type ∈
  {deepseek_v3, deepseek_v4}`; rank-pair `q_lora_rank > 0 AND
  kv_lora_rank > 0` ONLY when paired with a family-signal (name or
  metadata hint) — per codex r1 BLOCKING #3.

### Tests

- 26 unit tests for the resolver (`tests/test_kv_cache_dtype.py`) —
  matrix of dtype × safelist × reasoning × alias-metadata override,
  plus the BLOCKING #3 false-positive class.
- 6 metrics tests for the gauge (`tests/test_kv_cache_dtype_metrics.py`)
  including the loaded-bf16 + stale-stash regression case.
- 5 CLI tests via subprocess (`tests/test_kv_cache_dtype_cli.py`),
  including the BLOCKING #1 rejection-path e2e test.
- Full sweep: **10177 passed / 29 skipped / 7 xfailed** locally; no
  regressions.

### Quick smoke (Qwen3-0.6B-4bit, 128 decode tokens × 3 prompts)

| dtype | tok/s |
|-------|-------|
| bf16  | 333.89 |
| int4  | 324.08 |

The small-model / short-decode regime is bandwidth-cheap, so no
measurable win at this size. This was a sanity check that the wiring
didn't break decode quality, not a perf assertion — the documented
+1.1% / 3.2× context wins land on bigger models with longer contexts
where KV bandwidth dominates (see mlx#3134).

### codex r1 review (round 1)

Round 1 raised 3 BLOCKING + 2 NIT findings; all addressed in commit
b2cf669e:

- **BLOCKING #1** — `--reasoning` was silently ignored under the
  legacy `--kv-cache-quantization` branch. Now rejects the
  conflicting bits=4 combo with an actionable error; bits=8 +
  `--reasoning` (equivalent profiles) advertises the reasoning pin
  in the reason string.
- **BLOCKING #2** — Prometheus gauge could let a stale pre-load
  stash override a live engine bf16 state. Now distinguishes "engine
  reports a value" from "no engine value available".
- **BLOCKING #3** — `_is_mla` rank-pair heuristic was too loose.
  Now requires a family signal (alias metadata / canonical
  `model_type` / name pattern) in addition to the rank pair.
- **NIT #1** — CLI test loosely asserted `"int4" in text`; tightened
  to exact `--kv-cache-dtype {bf16,int8,int4}` choices line.
- **NIT #2** — Added missing loaded-bf16 + stale-stash regression
  test for the gauge.

## Test plan

- [x] Unit tests cover the safelist matrix (Gemma 3, GPT-OSS, DeepSeek
      V3/V4, Kimi K2, plain Qwen, rank-pair false positives).
- [x] CLI smoke verifies `--kv-cache-dtype` + `--reasoning` parse cleanly
      and the BLOCKING #1 conflict rejection fires.
- [x] Live smoke: server starts, `/metrics` shows the gauge, chat
      completion returns a valid response.
- [x] Quick decode tok/s comparison done (Qwen3-0.6B-4bit, smoke
      regime). Larger-model bench left to a follow-up issue — the
      perf contract on the bandwidth-cheap smoke model wasn't
      load-bearing for this PR.

### DIFF
--- a/tests/test_kv_cache_dtype.py
+++ b/tests/test_kv_cache_dtype.py
@@ -237,10 +237,14 @@ def test_mla_rank_pair_without_family_signal_does_not_trigger():
     assert decision.downgraded is False
 
 
-def test_mla_rank_pair_with_name_hit_triggers():
-    """codex r1 BLOCKING #3 follow-up: the rank pair MUST still trigger
-    when paired with a name-pattern signal (defends against a future
-    DeepSeek release that ships an unknown ``model_type`` value).
+def test_mla_unknown_deepseek_v5_name_fails_closed():
+    """codex r2 NIT (was BLOCKING #3 follow-up): when a future
+    DeepSeek release ships a name/``model_type`` we don't recognise
+    yet (e.g. ``deepseek_v5``), the safelist must fail CLOSED —
+    leave ``int4`` in place rather than guess at MLA based on rank
+    pairs alone. The previous test name suggested the rank pair
+    *triggered* the downgrade, which is the opposite of what the
+    body asserts; renamed to make a future failure self-explain.
     """
     decision = resolve_kv_cache_dtype(
         "int4",
--- a/tests/test_kv_cache_dtype.py
+++ b/tests/test_kv_cache_dtype.py
@@ -1,0 +1,295 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Unit tests for :mod:`vllm_mlx.kv_cache_dtype` (R15 task #300).
+
+The resolver is pure (no I/O, no model loading) — these tests pin the
+matrix of {dtype × safelist hit × reasoning override} that the CLI
+banner and Prometheus gauge depend on.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from vllm_mlx.kv_cache_dtype import (
+    DEFAULT_KV_CACHE_DTYPE,
+    KV_CACHE_DTYPES,
+    REASONING_KV_CACHE_DTYPE,
+    dtype_to_quantization_bits,
+    resolve_kv_cache_dtype,
+)
+
+
+# ---------------------------------------------------------------------------
+# Basic invariants
+# ---------------------------------------------------------------------------
+
+
+def test_default_is_int4():
+    """Locks the R15 #300 contract — int4 is the new default."""
+    assert DEFAULT_KV_CACHE_DTYPE == "int4"
+    assert "int4" in KV_CACHE_DTYPES
+
+
+def test_reasoning_default_is_int8():
+    """Reasoning profile pins to int8 (sub-4-bit drops -20pt on AIME)."""
+    assert REASONING_KV_CACHE_DTYPE == "int8"
+
+
+def test_dtype_to_quantization_bits_matrix():
+    """Each dtype maps onto a single, well-defined SchedulerConfig pair."""
+    assert dtype_to_quantization_bits("bf16") == (False, 8)
+    assert dtype_to_quantization_bits("int8") == (True, 8)
+    assert dtype_to_quantization_bits("int4") == (True, 4)
+
+
+def test_dtype_to_quantization_bits_rejects_unknown():
+    with pytest.raises(ValueError):
+        dtype_to_quantization_bits("fp4")
+
+
+def test_resolve_rejects_unknown_dtype():
+    with pytest.raises(ValueError):
+        resolve_kv_cache_dtype("fp4")
+
+
+# ---------------------------------------------------------------------------
+# Happy path — plain Qwen-style transformer, no safelist hit
+# ---------------------------------------------------------------------------
+
+
+def test_default_qwen_keeps_int4():
+    """A vanilla Qwen3 config should NOT be downgraded."""
+    decision = resolve_kv_cache_dtype(
+        "int4",
+        model_name="qwen3.5-9b-4bit",
+        hf_path="mlx-community/Qwen3.5-9B-MLX-4bit",
+        hf_config={"model_type": "qwen3", "hidden_size": 4096},
+    )
+    assert decision.dtype == "int4"
+    assert decision.downgraded is False
+    assert "qwen3.5-9b-4bit" in decision.reason
+
+
+def test_bf16_request_is_passed_through_even_on_safelist():
+    """An explicit bf16 request must never be silently upgraded."""
+    decision = resolve_kv_cache_dtype(
+        "bf16",
+        model_name="gemma-3-27b-4bit",
+        hf_config={"model_type": "gemma3", "sliding_window": 4096},
+    )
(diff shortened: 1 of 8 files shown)


raullenchai:
## R15 #300 — Qwen3.5-9B bench (decode tok/s + TTFT + RSS)

Follow-up to the test-plan item "Full bench on a larger model (Qwen3.5-9B+) to confirm the decode tok/s gain".

### Setup

- **Model**: `mlx-community/Qwen3.5-9B-4bit` (alias `qwen3.5-9b-4bit`, 4-bit weights).
- **Hardware**: M3 (this dev box).
- **Prompt**: 292 prompt tokens (256-token padded context + count-to-200 instruction).
- **Decode**: `max_tokens=400` (fixed-length runs) and `max_tokens=1000` (natural-stop runs at ~692 tokens).
- **Trials**: 1 warmup pass (excluded) + 5 measured trials at 400-tok; 1 warmup + 3 trials at 1000-tok.
- **Metric**: server-reported `usage.completion_tokens / (wall_elapsed - TTFT)` for decode TPS.
- **TTFT**: wall time from request send to first SSE `delta.content` event.
- **RSS**: `psutil` peak RSS of the `rapid-mlx serve` PID across the run.
- **Reinstall hygiene**: re-installed `--no-deps --force-reinstall .` from this PR branch
  immediately before each `rapid-mlx serve` boot (sibling agents are doing parallel
  installs into the same venv; this avoids picking up a stale wheel).
- **Server invocation**:
  ```
  rapid-mlx serve qwen3.5-9b-4bit --host 127.0.0.1 --port 8000 \
    --kv-cache-dtype {bf16,int4}
  ```
- **/metrics gauge** verified `rapid_mlx_kv_cache_dtype{dtype="<x>"} 1` matched the requested flag on each boot.

### Results — 400-tok decode (5-trial median)

| Metric              | bf16 (old default) | int4 (new default) | Δ          |
|---------------------|-------------------:|-------------------:|-----------:|
| decode TPS (median) |             113.65 |             113.61 |  −0.04 %   |
| decode TPS (mean)   |             113.67 |             113.60 |  −0.07 %   |
| TTFT (median, ms)   |              119.8 |              119.4 |  −0.4 ms   |
| peak RSS (MB)       |             5391.6 |             5388.2 |  −3.4 MB   |

### Results — 1000-tok decode (natural-stop at ~692 tokens, 3-trial median)

| Metric              | bf16   | int4   | Δ        |
|---------------------|-------:|-------:|---------:|
| decode TPS (median) | 113.47 | 113.22 | −0.22 %  |
| TTFT (median, ms)   |  118.7 |  119.4 | +0.7 ms  |
| peak RSS (MB)       | 5386.4 | 5388.6 | +2.2 MB  |

### Interpretation

- **Decode tok/s**: tied at this regime — both KV4 and bf16 land at ~113.5 tok/s
  on Qwen3.5-9B-4bit. KV4 is **not** a regression. The mlx#3134 +1.1 % throughput
  win was reported on Phi-3.5-mini at longer contexts; a 9B with <1k decoded
  tokens and a 292-tok prompt is too small a regime for bandwidth-bound effects
  to dominate (weight bandwidth dwarfs KV bandwidth here).
- **TTFT**: tied (within ~1 ms run-to-run noise).
- **RSS**: tied (within ~3 MB — KV4's cache-size win is below noise floor at
  this seq length; the 4× cache savings show up at multi-k contexts).
- **The win is "free swap, no perf regression at this size, headroom unlocked for
  long-context workloads"** — which matches the PR's framing. The +1.1 % / 3.2×
  ceiling stays a long-context claim, surfaceable in a later sustained-context
  release bench. This is sufficient to flip the default safely.

### Caveats

- Earlier 3-trial runs showed bf16 outliers (one 47 tok/s trial out of 3 in a
  thermal/OS-load excursion). Increasing to 5 trials + 2-sec cooldowns
  eliminated those, so the median above is the trustworthy number.
- Warmup pass is critical — graph JIT + Metal compile costs the first call
  ~50–70 ms TTFT and ~50 tok/s on the first 400-tok decode.

### Raw artifacts

JSON dumps written to `/tmp/r15_bench/result_{bf16,int4}.json` on the dev box;
serve logs at `/tmp/r15_bench/serve_*.log`.

PR #954, opened 2026-06-27
fix(spec_decode/mtp): port quantize + load_weights from qwen3_next_mtp; bench 1.0-1.15x on 9B

raullenchai:
## Summary

The 0.9 MTP injection pipeline was structurally incomplete — it built a fresh MTP module from scratch but never quantized it or loaded the `mtp.*` weights. The working sibling `vllm_mlx/patches/qwen3_next_mtp.py:108-200` has both steps; this PR ports them into `vllm_mlx/spec_decode/mtp/qwen3_5_inject.py`.

Three compounding defects shipped together (each independently blocked the bench):

1. **VLM-wrapper resolution wrong** — mlx-lm 0.31.3 always loads Qwen3.5 as `Model(language_model=TextModel)`. The original `_resolve_inner_text_model` (qwen3_5_inject.py:47-68) short-circuited to "VLM not supported" and returned None before any MTP work fired.
2. **`mtp_num_hidden_layers` lookup wrong** — the field lives under `text_config` in config.json but `BaseModelArgs.from_dict` silently drops unknown fields, so `args.mtp_num_hidden_layers` was always 0 and inject bailed at the gate at line 109.
3. **No quantize, no sidecar load** — even if (1)+(2) passed, the MTP head was random-init; accept rate would be ~0%.

Compounding factor: mlx-lm 0.31.3 `qwen3_5.py:313` unconditionally strips `mtp.*` in `sanitize()` during the main `mlx_lm.load`. Loading from a separate sidecar blob (`mlx-community/Qwen3.5-9B-MTP-4bit`, 131 MB, 31 keys, no `mtp.` prefix) AFTER the main load completes is the correct workaround — no upstream patch required.

Plus a separately-needed runtime patch for lossless: `cache_patch.patch_gated_delta_net_for_mtp` wraps `GatedDeltaNet.__call__` with a chunk-split version that splits `gated_delta_update` at the `n_confirmed=1` boundary and writes `(conv_state, ssm_state)` snapshot to `cache.rollback_state`. The existing `_rollback_draft` path in `generator.py` consumes the snapshot. Without it, lossless byte-equal diverged within ~10 tokens at temp=0 (linear-attention SSM state drifted on the first reject).

## Defect file:line references

- Broken: `vllm_mlx/spec_decode/mtp/qwen3_5_inject.py:118-198` (pre-fix)
- Working sibling template: `vllm_mlx/patches/qwen3_next_mtp.py:108-200`
- Sanitize strip: `.venv/lib/python3.12/site-packages/mlx_lm/models/qwen3_5.py:313`
- Test gap: `tests/test_mtp_spec_decode.py:633-651` (uses synthetic `_build_tiny_qwen3_5_text_model`), `tests/test_mtp_lossless.py:180+` (uses `_LosslessTestModule` stub)

## What was ported / added

- `inject_mtp_support(model, mtp_sidecar=None)` — new signature takes an optional sidecar (HF repo id or local path).
- Quantization auto-detected by walking the base model for a `QuantizedLinear` and reading bits/group_size/mode.
- Sidecar weights load via `mx.load(model.safetensors)` -> `mtp.load_weights(strict=False)`.
- `_resolve_inner_text_model` now correctly returns `model.language_model` for the VLM wrapper case.
- `text_config["mtp_num_hidden_layers"]` fallback when the args dataclass strips it.
- `bench/bench_spec_decode_mtp.py` learns `--mtp-sidecar` flag with `_DEFAULT_MTP_SIDECAR` map.
- `cache_patch.patch_gated_delta_net_for_mtp` — chunk-split rollback patch (idempotent, transparent when `n_confirmed=0`).

## Bench evidence (M3 Ultra, mlx-community/Qwen3.5-9B-4bit + sidecar, temp=0, 200 toks)

3 runs each, medians:

| profile | baseline | mtp | speedup | accept |
|---|---|---|---|---|
| chat | 82.4 t/s | 94.5 | 1.15x | 77.7% |
| code | 89.6 | 89.4 | 1.00x | 79.3% |
| math | 86.8 | 91.5 | 1.05x | 90.5% |

**1.5x is unreachable on this hardware/model combo.** Per-cycle probe: baseline 10.05 ms/step, MTP cycle (verify + mtp_forward) 15.4 ms with chunk-split rollback. Theoretical ceiling at α=0.9 is `(1+α)*10/15.4 = 1.24x`. Qwen3.5-9B-4bit is too memory-bandwidth-bound for the 2-token verify forward to amortize against a fast single-token baseline. 27B-w4 (originally named in 0.9-TODO row 43) would amortize toward PR #990's reported 1.57x, but `mlx-community/Qwen3.5-27B-MTP-4bit` does not exist on HF and the only glue script (`scripts/add_mtp_weights.py`) supports Qwen3-Next, not Qwen3.5.

## Lossless verification

Real-weights test `tests/test_mtp_real_weights.py::test_mtp_lossless_byte_equal_against_baseline` — 20 tokens × 3 profiles (chat/code/math) MTP-vs-baseline byte-equal at temp=0. ALL PASS.

## Test plan

- [x] Existing 38 MTP unit tests still pass (`tests/test_mtp_spec_decode.py`, `tests/test_mtp_lossless.py`)
- [x] Existing 8 inject/install tests still pass (`tests/test_mtp_inject_and_install.py`)
- [x] New heavy test `tests/test_mtp_real_weights.py::test_inject_loads_real_sidecar_weights` PASSES — confirms `mtp.fc.weight` equals sidecar byte-for-byte (would catch the random-init defect that PR #918 shipped)
- [x] New heavy test `tests/test_mtp_real_weights.py::test_mtp_lossless_byte_equal_against_baseline` PASSES on chat/code/math at temp=0
- [x] Red->green proof: with this PR's changes stashed, the new tests ERROR with `TypeError: inject_mtp_support() got an unexpected keyword argument 'mtp_sidecar'`; with changes restored, both PASS
- [x] Bench `bench/bench_spec_decode_mtp.py --model mlx-community/Qwen3.5-9B-4bit --runs 1 --prompts 1 --max-tokens 30` returns real speedup (no longer "FAILED: MTP injection failed")
- [x] Per-cycle perf probe confirms 15.4 ms MTP cycle vs 10.05 ms baseline (hardware ceiling 1.24x at alpha=0.9, matches measured ~1.05x with generator Python overhead)

## Known limitations

- Speedup on Qwen3.5-9B-4bit at M3 Ultra is below the 0.9-TODO 1.5x gate (1.05-1.15x median). This is a hardware-level finding — the per-cycle MTP overhead (~5 ms) is too large relative to the 10 ms baseline single-token decode on this small/quantized model. Operators wanting 1.5x need a larger model (27B or 35B) where the verify-of-2 cost amortizes better.
- 27B-MTP-4bit sidecar publishing is a separate follow-up (needs a Qwen3.5-family extension to `scripts/add_mtp_weights.py`).
- `n_confirmed` parameter is accepted on the patched `TextModel.__call__` and threaded into `GatedDeltaNet` via `cache.n_confirmed_for_mtp` (no upstream `mlx-lm` changes needed). The chunk-split fires only for verify cycles where `S > n_confirmed > 0`; the original path is byte-equal in the fast (`n_confirmed=0`) path.

### DIFF
--- a/bench/bench_spec_decode_mtp.py
+++ b/bench/bench_spec_decode_mtp.py
@@ -210,10 +210,14 @@ def _parse_args() -> argparse.Namespace:
 # Map from common base aliases / HF paths to their matching MTP sidecar.
 # The sidecar repo holds the MTP head only (no embed_tokens, no
 # backbone layers) — see ``mlx-community/Qwen3.5-9B-MTP-4bit`` README.
+# Keys are normalized to lowercase; ``_resolve_mtp_sidecar`` lowers
+# the incoming alias before lookup so case variants
+# (``Qwen3.5-9B-4bit`` vs ``qwen3.5-9b-4bit`` vs full
+# ``mlx-community/Qwen3.5-9B-4bit``) all hit the same default.
 _DEFAULT_MTP_SIDECAR: dict[str, str] = {
     "qwen3.5-9b-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
-    "mlx-community/Qwen3.5-9B-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
-    "mlx-community/Qwen3.5-9B-MLX-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
+    "mlx-community/qwen3.5-9b-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
+    "mlx-community/qwen3.5-9b-mlx-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
 }
 
 
@@ -221,13 +225,15 @@ def _resolve_mtp_sidecar(model_alias: str, explicit: str | None) -> str | None:
     """Pick the MTP sidecar for ``model_alias``.
 
     Explicit ``--mtp-sidecar`` always wins. Otherwise look up the
-    alias in ``_DEFAULT_MTP_SIDECAR``. Return ``None`` if no default
-    is known — the inject will then fall back to a no-op load and
-    log a warning.
+    alias in ``_DEFAULT_MTP_SIDECAR`` after lowercasing — codex
+    flagged on PR #954 that without normalization,
+    ``Qwen3.5-9B-4bit`` (case variant of the dict key) missed the
+    documented default. Return ``None`` if no default is known —
+    the inject will then fall back to a no-op load and log a warning.
     """
     if explicit is not None:
         return explicit
-    return _DEFAULT_MTP_SIDECAR.get(model_alias)
+    return _DEFAULT_MTP_SIDECAR.get(model_alias.lower())
 
 
 def _planned_matrix(args: argparse.Namespace) -> dict[str, Any]:
--- a/bench/bench_spec_decode_mtp.py
+++ b/bench/bench_spec_decode_mtp.py
@@ -181,7 +181,53 @@ def _parse_args() -> argparse.Namespace:
             "cost of reduced workload coverage."
         ),
     )
+    parser.add_argument(
+        "--mtp-sidecar",
+        default=None,
+        help=(
+            "MTP head sidecar (HF repo id or local path). The mlx-lm "
+            "0.31.3 ``qwen3_5.py::sanitize`` unconditionally strips "
+            "``mtp.*`` weights during the main load, so MTP weights "
+            "must be loaded from a separate sidecar safetensors blob "
+            "after ``mlx_lm.load(...)`` completes. Default: auto-pick "
+            "based on the base model alias (Qwen3.5-9B-4bit → "
+            "mlx-community/Qwen3.5-9B-MTP-4bit). Pass an explicit "
+            "value to override."
+        ),
+    )
+    parser.add_argument(
+        "--mtp-only",
+        action="store_true",
+        help=(
+            "Skip the baseline (--spec-decode none) condition and run "
+            "only the MTP condition. Useful when comparing against a "
+            "previously-captured baseline number."
+        ),
+    )
     return parser.parse_args()
+
+
+# Map from common base aliases / HF paths to their matching MTP sidecar.
+# The sidecar repo holds the MTP head only (no embed_tokens, no
+# backbone layers) — see ``mlx-community/Qwen3.5-9B-MTP-4bit`` README.
+_DEFAULT_MTP_SIDECAR: dict[str, str] = {
+    "qwen3.5-9b-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
+    "mlx-community/Qwen3.5-9B-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
+    "mlx-community/Qwen3.5-9B-MLX-4bit": "mlx-community/Qwen3.5-9B-MTP-4bit",
+}
+
+
+def _resolve_mtp_sidecar(model_alias: str, explicit: str | None) -> str | None:
+    """Pick the MTP sidecar for ``model_alias``.
+
+    Explicit ``--mtp-sidecar`` always wins. Otherwise look up the
+    alias in ``_DEFAULT_MTP_SIDECAR``. Return ``None`` if no default
+    is known — the inject will then fall back to a no-op load and
+    log a warning.
+    """
+    if explicit is not None:
+        return explicit
+    return _DEFAULT_MTP_SIDECAR.get(model_alias)
 
 
 def _planned_matrix(args: argparse.Namespace) -> dict[str, Any]:
@@ -208,6 +254,7 @@ def _run_once(
     prompt: str,
     max_tokens: int,
     temp: float,
+    mtp_sidecar: str | None = None,
 ) -> RunResult:
     """Run one generation under the requested condition.
 
@@ -235,13 +282,24 @@ def _run_once(
             validate_mtp_support,
         )
 
-        if not inject_mtp_support(model):
+        if not inject_mtp_support(model, mtp_sidecar=mtp_sidecar):
             raise RuntimeError(
                 f"MTP injection failed on model {model_alias!r} — "
-                "checkpoint likely lacks mtp.* weights. Re-convert from "
-                "HF with mlx-lm PR #990's sanitize() path."
+                f"sidecar {mtp_sidecar!r}. Confirm the sidecar repo or "
+                "local path exists, holds model.safetensors with the "
+                "expected MTP head schema, and the base model's "
+                "config carries mtp_num_hidden_layers >= 1."
             )
         assert validate_mtp_support(model)
+        # The patch lands on the inner TextModel (the VLM wrapper's
+        # ``language_model`` field). The generator drives the model
+        # directly, so re-bind ``model`` to the patched inner for the
+        # ``mtp_generate_step`` call. The mlx-lm 0.31.3 Qwen3.5
+        # arch ships only the VLM wrapper; the inner TextModel
+        # carries embed_tokens, lm_head, layers — everything the
+        # generator and ``mtp_forward`` reference.
+        if hasattr(model, "language_model"):
+            model = model.language_model
 
     prompt_ids = mx.array(tokenizer.encode(prompt), mx.uint32)
     counter = MTPAcceptCounter()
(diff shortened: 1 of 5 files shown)


raullenchai:
# Codex round-9 response: defending round-8 BLOCKINGs as misreads

Re-read of HEAD `72e396b8` against the round-8 codex output. All three round-8 BLOCKINGs are misreads of code already addressed in earlier rounds; the line numbers codex cites land inside comment blocks (or off by 60+ lines) rather than on the code being described. Specifics below — each finding is responded to in order.

---

## 1. [BLOCKING #1] `qwen3_5_inject.py:310` — fail-closed default

> `inject_mtp_support()` now returns `False` whenever `mtp_sidecar is None`, so existing callers that relied on the previous public `inject_mtp_support(model)` behavior will silently lose MTP support unless every call site was updated. Fix: either preserve the old checkpoint-load path/default behavior, or update all production call sites and document the new required `mtp_sidecar` contract.

**Misread — already addressed in round-5 (`39598366`) + round-7 docstring (`d4e1b720`).** Two factual problems with the finding:

**(a) Line number is off by ~90 lines.** Line 310 is `logger.info("[mtp.inject] config has no mtp_num_hidden_layers; skipping MTP injection.")` — the early-return for non-MTP configs. The fail-closed default codex is actually describing lives at lines **400-416**, which include both the warning + the documented rationale ("Codex round-5 BLOCKING fix: default is fail-closed…").

**(b) Both prescribed remediations are already in place.** Codex's "Fix:" suggests *either* preserve old behavior *or* update all call sites + document the new contract. The PR did the second one:

- **Call-site audit** (every `inject_mtp_support` caller in the tree):
  - `bench/bench_spec_decode_mtp.py:291` → already passes `mtp_sidecar=mtp_sidecar` ✓
  - `vllm_mlx/utils/tokenizer.py:603` → imports from `..patches.qwen3_next_mtp`, a **different module** with signature `(model, model_path, config)`. Different function, different model family (Qwen3-Next, not Qwen3.5/3.6). Not affected.
  - `vllm_mlx/scheduler.py` → uses the `_install_mtp` wrapper, not this function. Not affected.
  - All `tests/` call sites updated to pass `allow_random_init=True` in round-5.
- **Documentation** at `qwen3_5_inject.py:250-258` is explicit:
  > "The only production caller of this function is `bench/bench_spec_decode_mtp.py` (which already passes `mtp_sidecar`). There are no pre-existing bare `inject_mtp_support(model)` call-sites to break with the new `allow_random_init=False` default."

This function landed in PR #918 (`f28fdb45`); between then and PR #954 nothing modified it. The `mtp_sidecar` signature is **new in this PR** and was added with all in-tree callers updated atomically. There is no out-of-tree caller to worry about — this is an internal helper not part of any public API surface (no `__all__` export, no documentation page).

---

## 2. [BLOCKING #2] `cache_patch.py:199` — rollback_state lifecycle

> the patched `GatedDeltaNet.__call__` clears `cache.rollback_state` at entry for every cached call, including the fast path, so a later ordinary forward can erase a previously captured rollback snapshot before `_rollback_draft` consumes it. Fix: only clear stale rollback state after an accept/commit boundary, or clear it in the generator once the snapshot is no longer needed.

**Misread — already addressed in round-7 (`d4e1b720`).** Two factual problems:

**(a) Line 199 is inside a COMMENT block, not executable code.** Lines 198-202 read:
```python
# Fast path — no MTP boundary signaled, or chunk has 1
# token, or boundary is outside the range, or NO cache at
# all. Defer to the original implementation (byte-equal
# behavior). rollback_state is NOT touched on this path
# (see round-7 ordering fix below).
```
The comment explicitly says "rollback_state is NOT touched on this path." Codex appears to have read the comment and inverted its claim.

**(b) The actual `cache.rollback_state = None` lives at line 238**, immediately before the chunk-split path writes a fresh snapshot at line 306 — exactly the structure codex's "Fix:" prescribes (clear only at a commit boundary). The control-flow proof:

- Lines 211-212 — fast path: `if cache is None or n_conf <= 0 or n_conf >= S or S < 2: return _orig_gated_delta_call(...)`. Returns before any clear.
- Lines 215-222 — TP fallback: `if self.sharding_group is not None: return _orig_gated_delta_call(...)`. Returns before any clear.
- Line 238 — clear runs **only** on the chunk-split path that immediately writes a fresh snapshot at line 306. Single atomic write per chunk-split call.

Round-7 commit message (`d4e1b720`) documents this exact restructuring; the inline comment at lines 224-237 cites it.

---

## 3. [BLOCKING #3] `cache_patch.py:262` — fast-path divergence under mlx-lm version drift

> `cache[1] = state_final` is only written in the chunk-split path, but the fast path delegates to the original method; if the original method updates cache position/state differently across mlx-lm versions, this monkey-patch can diverge from upstream semantics. Fix: keep this code structurally tied to the upstream implementation with a version guard or fall back unless the expected `GatedDeltaNet` shape/version is detected.

**Misread — line number off by 67, concern already addressed across rounds 5/6.** Three issues:

**(a) Line 262 is `conv_snap = mx.contiguous(conv_input[:, n_conf : n_conf + n_keep, :])`**, not `cache[1] = state_final`. The actual `cache[1] = state_final` write is at line 329.

**(b) Fast-path delegation to `_orig_gated_delta_call` is the *intended* behavior, not a bug.** The whole point of the fast path is byte-equal behavior with upstream when `n_confirmed_for_mtp == 0` (i.e., outside the MTP verify boundary). If upstream changes its cache update across mlx-lm versions, fast-path callers get the upstream change for free — that's the whole reason we keep this delegating instead of duplicating.

**(c) Version guarding is already in place.** From `cache_patch.py:33-35`:
> "Safe under future mlx-lm versions that add the slot themselves — the guard checks `"rollback_state" in cls.__dict__` before patching, so once upstream lands the change this becomes a no-op."

And the chunk-split path's `cache.advance(S)` mirrors upstream's `mlx_lm/models/qwen3_5.py` line 196-198 in 0.31.3, documented inline at lines 330-340. `mlx_lm == 0.31.3` is pinned in our requirements, so the "across mlx-lm versions" concern is also bounded.

---

## 4. [NIT #4] `qwen3_5_inject.py:260` — stale "tracked separately" docstring

> the docstring still says `n_confirmed` "does NOT thread through" and is "tracked separately," contradicting the newly added `GatedDeltaNet` patch.

**Partially valid.** The docstring codex is describing actually lives at **line 458** (inside the inner `_Qwen3_5WithMTP.__call__` class docstring), not 260. Line 260 already reads:
> "`n_confirmed` rollback is implemented as of this PR: it threads through to each `ArraysCache` via `n_confirmed_for_mtp` before forward…"

— that text is current. The line-458 docstring is stale ("currently a no-op below this layer — the GatedDeltaNet rollback patch is tracked separately"). Will refresh this in a follow-up nit-fix commit. Acknowledging the finding.

---

## 5. [NIT #5] `tests/test_mtp_real_weights.py:39` — default-CI coverage

> the only real-checkpoint lossless coverage is skipped unless `RAPID_MLX_RUN_HEAVY_TESTS=1`, so default CI can still ship regressions in the real quantized sidecar path. Fix: add a lightweight default test that exercises the quantized synthetic sidecar path…

**Misread — round-5 (`39598366`) already added the prescribed test.** The "Fix:" codex sketches is exactly what round-5 implemented. See `tests/test_mtp_spec_decode.py`:

- **Line 742** — `test_inject_mtp_support_loads_synthetic_sidecar`: builds a synthetic MTP head, persists to temp safetensors, re-injects, asserts byte-equal load. Runs by default.
- **Line 830** — `test_inject_mtp_support_refuses_synthetic_sidecar_missing_tensor`: drops a required tensor and asserts the coverage check refuses. Runs by default.

Both are in the default `pytest` collection (no `RAPID_MLX_RUN_HEAVY_TESTS` gate). The heavy test in `test_mtp_real_weights.py` is the real-Qwen3.5-9B-4bit coverage that gates on the 5 GB download — that gate is intentional. The quantize → load → coverage path codex is worried about IS exercised by default in `test_mtp_spec_decode.py`.

---

## Summary

| Finding | Verdict | Action |
|---|---|---|
| BLOCKING #1 | misread — line off by 90, callers audited + documented | no change |
| BLOCKING #2 | misread — line points at a comment that says the opposite of what codex claims; round-7 ordering fix is in place | no change |
| BLOCKING #3 | misread — line off by 67; fast-path delegation is intended; version guard exists | no change |
| NIT #4 | partially valid — wrong line (458, not 260), real text is stale | will refresh inner-class docstring in follow-up |
| NIT #5 | misread — synthetic-sidecar tests added in round-5 already cover this | no change |

Defense built from worktree HEAD `72e396b8` against codex artifact `/tmp/pr_validate/pr-954/run-20260627T183704Z-53715-1782585424282079000/codex-review.md`. No commits pushed; no force-push. Will re-run `pr_validate` once this is posted so codex gets a fresh round-9 look.

PR #963, opened 2026-06-29
fix(scheduler): correct batch_gen -> batch_generator typo blocking disk-KV writes

raullenchai:
## Why

The disk-backed KV checkpoint hook (`Scheduler._maybe_disk_checkpoint`, added in PR #919) reads the BatchGenerator off the Scheduler with the wrong attribute name. The canonical attribute is `self.batch_generator` — assigned at `vllm_mlx/scheduler.py:1889` and read at `:2549, :2645, :2788, :4074, :4975, ...`. `self.batch_gen` is **never assigned anywhere in the codebase**:

```
$ grep -nE 'self\.batch_gen\s*=' vllm_mlx/scheduler.py
(no matches)
```

So `getattr(self, "batch_gen", None)` always returns `None`, the very next line `if batch is None: return` fires unconditionally, and the hook never reaches `maybe_write_checkpoint`. Result: `rapid_mlx_kv_checkpoint_writes_total` is stuck at 0 on `origin/main` even with `--kv-disk-checkpoint-interval 256`, which is what makes 体感 2 row 4 ("Disk-KV @ 32k reduces RAM >= 50%") measure 0.01% in the 0.9-TODO dashboard.

## What

One-line attribute-name correction:

```diff
-        batch = getattr(self, "batch_gen", None)
+        batch = getattr(self, "batch_generator", None)
```

at `vllm_mlx/scheduler.py:4650`. Same hook block; comments on `:4644-4648` already use the correct name `batch_gen.active_batch` for the *local* variable, so only the `self.<attr>` getattr was wrong.

## Verification

**Static.**
- `grep -nE 'self\.batch_gen\b' vllm_mlx/scheduler.py` -> 0 hits after fix.
- `git grep 'self\.batch_generator' vllm_mlx/scheduler.py | wc -l` -> 21 canonical reads.

**Unit.** `pytest tests/test_disk_kv_checkpoint.py -x -v` -> 27 passed, 0 failed (as expected — these test `write_checkpoint` directly, not the scheduler wiring; they pass on `main` today and still pass with the fix).

**Integration smoke.** Boot `rapid-mlx serve qwen3.5-9b-4bit --kv-disk-checkpoint-interval 256 --port 8773`, send a 400-token completion, scrape `/metrics`:

- **Before fix (`main`):** `rapid_mlx_kv_checkpoint_writes_total = 0`, `~/.cache/rapid-mlx/kv_checkpoints/` directory not created. The early `if batch is None: return` swallows the hook.
- **After fix (this PR):** `rapid_mlx_kv_checkpoint_writes_total = 0`, `~/.cache/rapid-mlx/kv_checkpoints/` directory still not created.

Wait — same number? See the next section. **This typo fix is necessary but not sufficient.**

## Honest disclosure: a second bug surfaces after this fix

Instrumented logging on the smoke (reverted before commit) shows the typo fix correctly progresses past `if batch is None`, but the very *first* line of `_maybe_disk_checkpoint` then raises an unrelated `AttributeError`:

```
WARNING:rapid_mlx.scheduler:[DBG-CKPT-PRE] calling _maybe_disk_checkpoint req=... num_tokens=406
WARNING:rapid_mlx.scheduler:[DBG-CKPT-EX] ...: 'Scheduler' object has no attribute 'scheduler_config'
```

The hook reads `self.scheduler_config.kv_disk_checkpoint_interval` at `:4610` and `self.scheduler_config.kv_cache_dtype` at `:4634`, but the canonical Scheduler attribute is `self.config` (assigned at `:1868` as `self.config = config or SchedulerConfig()` and read at `:1914-1938` and ~40 other sites). This is a **second** wrong-attribute typo from the same PR #919. The hook-wrapper at `:4406-4410` silently catches the `AttributeError`, which is why neither typo was loud.

Per the operator's instruction to ship only the one-line typo, **this PR does NOT fix the `scheduler_config` -> `config` typo**. That should be a follow-up PR: two `getattr(self.scheduler_config, ...)` calls at `:4610` and `:4634` to `getattr(self.config, ...)`. After both PRs land, the smoke counter goes positive.

## Test plan

- Unit: `pytest tests/test_disk_kv_checkpoint.py -x -v` passes (27/27).
- Static: `self.batch_gen` is gone from the file (grep proof above).
- pr_validate scorecard: see linked artifact / comment thread.
- Follow-up: open a separate PR for the `scheduler_config` -> `config` typo so the metric actually moves.

Refs: #919 (introducing PR), 0.9-TODO 体感 2 row 4.

### DIFF
--- a/vllm_mlx/scheduler.py
+++ b/vllm_mlx/scheduler.py
@@ -4647,7 +4647,7 @@ class Scheduler:
         # method directly on ``batch_gen.active_batch``. Walking both
         # surfaces keeps this hook portable across the mlx-lm versions
         # rapid-mlx supports.
-        batch = getattr(self, "batch_gen", None)
+        batch = getattr(self, "batch_generator", None)
         if batch is None:
             return
         gen_batch = getattr(batch, "_generation_batch", None) or getattr(


raullenchai:
## pr_validate scorecard

Three consecutive runs with `PR_VALIDATE_NO_STRESS=1` (stress is skipped only because two of the three eligible golden models in `golden_models.yaml` — Qwen3.5-35B-A3B-8bit and Qwen3.6-27B-MLX-8bit — exceed the 18.6 GB free space on this host; this is a pre-existing host-disk limit, not a PR-introduced regression).

All three runs show identical pass/fail shape:

| step | run 1 | run 2 | run 3 |
|---|:---:|:---:|:---:|
| fetch | PASS | PASS | PASS |
| test_plan_check | PASS | PASS | PASS |
| cl_description_quality | PASS | PASS | PASS |
| supply_chain | PASS | PASS | PASS |
| test_env_check | PASS | PASS | PASS |
| codex_review | PASS (1 NIT) | PASS | PASS |
| lint | PASS | PASS | PASS |
| targeted_tests | PASS (28/28) | PASS (28/28) | PASS (28/28) |
| full_unit | **FAIL** | **FAIL** | **FAIL** |
| stress_e2e_bench | skip (disk) | skip (disk) | skip (disk) |

## full_unit failure is a pre-existing flake, not a regression of this PR

The single failing test in all three runs:

```
FAILED tests/test_signal_observability.py::test_subprocess_sighup_default_disposition_dumps_and_stays_alive
```

Failure mode (identical across runs):

```
tests/test_signal_observability.py:451: ready_line = _read_ready_with_timeout(proc)
tests/test_signal_observability.py:71:  raise AssertionError(...)
E AssertionError: subprocess died before emitting READY; returncode=None,
  stderr='Traceback (most recent call last):
    File "<string>", line 6, in <module>
  AssertionError'
```

The subprocess fails its own `line 6` assertion — `assert signal.getsignal(signal.SIGHUP) == signal.SIG_DFL` — i.e. SIGHUP is not SIG_DFL in the freshly spawned child. This is entirely about signal-handler inheritance from pytest's parent process under load, and has zero coupling to `vllm_mlx/scheduler.py`.

Evidence this is a load-induced flake, not a regression:

1. **Isolated run on the PR branch**: `pytest tests/test_signal_observability.py -v` -> all 10 tests pass.
2. **Isolated run on origin/main code** (scheduler.py reverted, fix removed): same single test passes in isolation.
3. **Subset run** (78 tests across `test_signal_observability.py` + `test_pr_validate_runner.py` + `test_pr_validate_test_env.py`): all pass.
4. **Reproduces only in `pytest tests/`** (10897 tests under load): fails 3/3 consecutive runs.
5. PR #907 already documented this exact test as flaky under load (see `b5825a72`'s commit message — "test_subprocess_sighup_default_disposition_dumps_and_stays_alive flake … ReportCrash contention"). #907 fixed one root cause; the residual flake under high load (`uptime` shows `13.6` 5-min average on this host while operator's qwen3-vl-30b-4bit server saturates Metal) is what's surfacing here.
6. **No coupling**: this PR changes one attribute name in the disk-KV checkpoint hook (`vllm_mlx/scheduler.py:4650`). It does not touch `vllm_mlx/_signal_observability.py`, any test file, any signal handling, or the subprocess-spawn path.

Per [[engine-PRs-must-show-validate-PASS]], full_unit must be clean. Recommend operator-side options:

- **Option A**: re-run pr_validate when host load is lower (operator's qwen3-vl-30b-4bit at port 8801 is the dominant load source — paused, this test passes deterministically).
- **Option B**: address the residual #907 flake in a separate tests-only PR (likely: install `signal.signal(signal.SIGHUP, signal.SIG_DFL)` in the subprocess's program body BEFORE the `assert`, defensively against parent inheritance), then re-run pr_validate here.

This PR does not attempt either — scope is strictly the one-line attribute fix.

raullenchai:
Superseded by #965 (c031b0b1) — same 1-line fix landed there along with 2 more sibling typos in the same hook block and silent-swallow wrapper hardening. Closing to keep the queue clean.

PR #964, opened 2026-06-29
fix(tokenizer): don't swap SentencePiece decoder to ByteLevel (gemma-3 ▁ space leak)

raullenchai:
## Bug

On `gemma3-1b-qat-4bit` (and all gemma-3 sizes — same vocab), every space in generated text leaked as the SentencePiece metaspace marker `▁` — chat content, streaming `delta.*`, and `/v1/completions[0].text` all rendered `The▁search▁results▁mention…` (the user saw it as `The__search__results…`). Reported via the desktop app.

## Root cause

`repair_byte_level_decoder()` (which fixes DeepSeek-R1-distill tokenizers whose **byte-level** vocab was handed a SentencePiece decoder) **false-positived on a genuine SentencePiece vocab**:

1. It scanned the vocab for any token whose pretty form starts with a GPT-2 byte-level marker (`Ġ`/`Ċ`/`ĉ`).
2. It found gemma-3's lone **literal `ĉ` character token** (U+0109, id 240630) — a real character, not byte-level mojibake.
3. `decode([id])` correctly returned `'ĉ'`, but the check "decoded still contains a marker" was satisfied, so it concluded the decoder was a broken byte-level chain.
4. It swapped the decoder to `ByteLevel()`, **destroying the SentencePiece `Replace("▁"," ")` step for the entire tokenizer** → `▁` leaked into every space.

Plain `mlx_lm.load` (no repair) decodes gemma-3 correctly; only our load path broke it.

## Fix

Gate the repair on the vocab's **space-encoding convention**, not "some token's pretty form starts with a marker char":

- A genuine SentencePiece vocab encodes spaces as `▁` on tens of thousands of tokens (gemma-3: **137,541** `▁` vs **1** `Ġ`).
- A byte-level (GPT-2) vocab encodes them as `Ġ`.
- Compare **counts, not presence** (either family can carry an isolated literal `Ġ`/`Ċ`/`ĉ` char token). If metaspace dominates, the decoder is already correct → leave it alone.

This preserves the DeepSeek-R1-distill byte-level repair (those vocabs are `Ġ`-dominant, `▁`≈0).

## Verification

- **End-to-end**: `gemma-3-1b` streaming generation of "why is the sky blue?" now returns proper spaces, **no `▁`/`__`** ("The sky appears blue due to a phenomenon called Rayleigh scattering. …").
- Decoder stays `Sequence` (SPM), `repair_byte_level_decoder()` correctly returns `False`.
- New `TestRepairSkipsGenuineSentencePiece` (4 cases) in the synthetic-fixture suite, incl. a sanity test proving the old marker heuristic *would* have tripped on the fixture. Full byte-level suite **20/20**; related tokenizer/detok suites green.
- Codex review: 0 BLOCKING / 0 MAJOR.

Note: this affects the bundled sidecar — reaches desktop users via a rapid-mlx release + submodule bump.

### DIFF
--- a/tests/test_streaming_detokenizer_bpe.py
+++ b/tests/test_streaming_detokenizer_bpe.py
@@ -258,6 +258,84 @@ class TestRepairByteLevelDecoder:
         )
 
 
+def _build_sentencepiece_with_literal_marker_token() -> PreTrainedTokenizerFast:
+    """A genuine SentencePiece (metaspace) tokenizer that ALSO carries an
+    isolated literal ``ĉ`` (U+0109) character token — the exact gemma-3
+    shape that false-tripped ``repair_byte_level_decoder``.
+
+    Spaces are encoded as the metaspace marker ``▁`` and the decoder is the
+    correct SentencePiece ``Replace("▁"," ")`` chain. The lone ``ĉ`` token
+    is a real character in the vocab, NOT byte-level mojibake — the repair
+    must leave this tokenizer untouched.
+    """
+    vocab = {
+        "<pad>": 0,
+        "<s>": 1,
+        "</s>": 2,
+        "▁the": 3,  # ' the'
+        "▁search": 4,  # ' search'
+        "▁results": 5,  # ' results'
+        "ĉ": 6,  # genuine U+0109 literal char token — the trap
+        "the": 7,  # 'the'
+    }
+    model = models.BPE(vocab=vocab, merges=[])
+    tok = Tokenizer(model)
+    tok.pre_tokenizer = pre_tokenizers.Metaspace()
+    tok.decoder = decoders.Sequence(
+        [
+            decoders.Replace("▁", " "),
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing as superseded by the already-merged #959 (`c2b37c35`).

The gemma-3 `▁`→space leak (the `repair_byte_level_decoder` false-positive on gemma's *literal* `ĉ` U+0109 token) is already fixed on `main` by #959, which uses a behavior-based gate — it observes how a space is actually encoded (byte-level emits `Ġ`/`Ċ`/`ĉ`, SentencePiece emits `▁`) and bails when the convention is metaspace, plus a post-swap revert if `▁` leaks. That is a cleaner, equivalent root-cause fix. This branch was cut from before #959 landed and is now redundant (and conflicts). Verified on `main`: gemma-3-1b decodes spaces correctly and the decoder stays a SentencePiece `Sequence`.

PR #980, opened 2026-07-01
fix(stt): guard Whisper silence hallucination via Silero VAD pre-trim (#961)

### LINKED ISSUE #961: Whisper STT hallucinates text over silence on short clips — no_speech_threshold / logprob_threshold / hallucination_silence_threshold are all inert (no_speech_prob comes back NaN)
### Rapid-MLX version

0.8.0 (installed) — verified against main / v0.9.8 (its stt.py is unchanged here)

_No response_

### Hardware

Mac mini (Mac16,11), Apple M4 Pro, 24 GB

### macOS version

macOS 26.2 (build 25C56)

### Python version

3.12.13

### Model

mlx-community/whisper-large-v3-turbo (also the default whisper-large-v3-mlx)

### Full serve command

Bug is independent of serve flags and needs no server to reproduce (Python API, see below). STT is exposed at POST /v1/audio/transcriptions. Example: rapid-mlx serve mlx-community/whisper-large-v3-turbo --host 127.0.0.1 --port 8000

Actual: short / silent audio produces words that were never spoken — 12 s of pure silence transcribes as "Thank you." On real clips with trailing dead air, the speech is transcribed correctly PLUS an invented tail (one word up to several sentences).
Expected: silent / non-speech audio -> empty transcription.

Root cause: STTEngine.transcribe() calls self.model.generate(audio_path, **kwargs) with no anti-hallucination handling, AND the documented guards (no_speech_threshold / logprob_threshold / hallucination_silence_threshold) are inert because the chunked generate() returns no_speech_prob = NaN, so the skip condition (no_speech_prob > no_speech_threshold AND avg_logprob < logprob_threshold) is never true. See the kwarg sweep in the logs field.

Fix options: (1) populate no_speech_prob / avg_logprob so the existing thresholds take effect (may belong in mlx_audio); or (2) VAD pre-trim in transcribe() using the bundled mlx_audio.vad Silero model — return empty when there's no speech, else trim to the speech span — ideally opt-in and timestamp-safe. I have a working VAD pre-trim locally and can open a PR if welcome.

### Minimal reproduction

# 12 s of pure digital silence
ffmpeg -y -f lavfi -i anullsrc=r=16000:cl=mono -t 12 silence.wav

# No server needed — Python API:
from vllm_mlx.audio.stt import STTEngine
eng = STTEngine("mlx-community/whisper-large-v3-turbo"); eng.load()
print(repr(eng.transcribe("silence.wav").text))   # -> 'Thank you.'   (expected: '')

# Or via the API:
curl -s -X POST http://127.0.0.1:<port>/v1/audio/transcriptions \
  -F file=@silence.wav -F response_format=text -F model=whisper-large-v3-turbo   # -> "Thank you."

```shell
# Sweeping the relevant decode kwargs on pure silence -> byte-identical output,
# and the per-segment probabilities come back NaN (so the gating can never fire):
from mlx_audio.stt.utils import load_model
m = load_model("mlx-community/whisper-large-v3-turbo")
for kw in [{},
           dict(temperature=0.0, condition_on_previous_text=False),
           dict(no_speech_threshold=0.1, logprob_threshold=0.0),
           dict(word_timestamps=True, hallucination_silence_threshold=2.0)]:
    r = m.generate("silence.wav", verbose=False, **kw); s = r.segments[0]
    print(kw, '->', repr(r.text), 'no_speech_prob=', s.no_speech_prob, 'avg_logprob=', s.avg_logprob)

# every line: -> ' Thank you.'  no_speech_prob= nan  avg_logprob= nan
```

raullenchai:
## Summary

Fix for [#961](https://github.com/raullenchai/Rapid-MLX/issues/961) — Whisper STT hallucinates over silence: 12 s of digital silence transcribes as ``"Thank you."``, and real clips with trailing dead-air get transcribed correctly + an invented tail.

## Reporter's exact repro

Generate silence + run the reporter's Python snippet:

```bash
ffmpeg -y -f lavfi -i anullsrc=r=16000:cl=mono -t 12 silence.wav
```

```python
from vllm_mlx.audio.stt import STTEngine
eng = STTEngine("mlx-community/whisper-large-v3-turbo"); eng.load()
print(repr(eng.transcribe("silence.wav").text))
```

| behaviour | text |
|---|---|
| pre-fix (main) | `'Thank you.'` |
| post-fix (this PR) | `''` |

Also verified against the trailing-silence pattern (3 s silence + 2.5 s "the quick brown fox..." + 3 s silence): pre-fix produced a hallucinated tail, post-fix returns a single clean segment with the speech text and an absolute start time of `2.78 s` (leading silence correctly reflected).

## Root cause

`STTEngine.transcribe()` at `vllm_mlx/audio/stt.py` calls `self.model.generate(audio_path, **kwargs)` with no anti-hallucination handling. The documented guards inside `mlx_audio.stt.models.whisper.Model.generate` — `no_speech_threshold=0.6`, `logprob_threshold=-1.0`, `hallucination_silence_threshold` — are inert in practice because on pure silence Whisper returns:

* `no_speech_prob ≈ 9.6e-11` (upstream field: `mlx_audio/stt/models/whisper/whisper.py:1010`)
* `avg_logprob ≈ -0.26` (upstream field: `whisper.py:1008`)

The AND-gate skip condition at `whisper.py:964-970` (`no_speech_prob > no_speech_threshold AND avg_logprob < logprob_threshold`) therefore never fires and the model emits its silence-priors ("Thank you.", "Thanks for watching!", etc.). The reporter observed `NaN` on some chunked paths; either way the guard's precondition never holds on true silence.

Fixing this upstream would require modifying `mlx_audio` itself (not vendored here); monkey-patching it is fragile across upstream bumps.

## Design decision — VAD pre-trim vs upstream fix

| option | assessment |
|---|---|
| PR upstream to populate `no_speech_prob` | slow (gated on maintainer response); not our release |
| monkey-patch upstream `Model.generate` in rapid-mlx | fragile — breaks on next `mlx_audio` bump |
| re-compute `no_speech_prob` from decoder logits in `STTEngine` | reproduces internal decode loop; heavy for a bug fix |
| **VAD pre-trim in `transcribe()`** | **self-contained, systematic, model-agnostic** |

The chosen invariant — *"if the audio has no speech, don't invoke Whisper on it"* — is strictly stronger than post-hoc detection. It also uses a bundled dependency (`mlx_audio.vad.silero_vad`) so the `[audio]` extra needs no changes.

## Implementation

`vllm_mlx/audio/stt.py`:

* `_get_vad_model()` — lazy, once-per-process cached `mlx-community/silero-vad` load. Returns `None` on failure (missing extras, network, weight mismatch) → guard transparently falls back.
* `_maybe_vad_trim(audio_path) -> _VADTrimResult` — 3-shape outcome: `skipped=True` (fallback), `has_speech=False` (return empty), or `has_speech=True` (trim + offset). Never raises.
* `_shift_segment_time(seg, offset)` — mutates segment/word timestamps back to file-relative absolute time.
* `STTEngine.__init__(..., enable_vad_pretrim: bool = True)` — new kwarg.
* `STTEngine.transcribe()` — runs the guard before `self.model.generate(...)`; picks trimmed waveform vs. original path; shifts returned segments.

Gating contract (both must be true for the guard to run):
* engine kwarg `enable_vad_pretrim = True` (default)
* env `RAPID_MLX_STT_VAD_PRETRIM` ∉ `{"0","false","no","off"}` (case-insensitive)

Guard applies only to Whisper engines (`self._is_whisper`). Parakeet/Canary/future STT engines are unaffected.

## Timestamp-preservation contract

When VAD trims leading silence, every returned `segment.start` / `segment.end` (and per-word `word.start` / `word.end` when `word_timestamps=True`) is shifted back by the trim offset so downstream consumers see file-relative absolute times. Verified end-to-end: `3 s silence + speech` → `segment.start == 2.78 s`, not `0.0`.

## Test coverage

`tests/test_stt_vad_pretrim.py` — 30 tests, all passing:

- `TestSilenceReturnsEmpty::test_pure_silence_returns_empty_string` — VAD reports no speech → text=="", segments==[], Whisper NOT invoked.
- `TestRealSpeechStillTranscribes::test_real_speech_still_transcribes` — passes trimmed waveform to Whisper, text unchanged.
- `TestTrailingSilenceTrimmed::test_speech_with_trailing_silence_stops_at_speech_end` — no tail hallucination; last segment end within `speech_end + 0.5 s`.
- `TestVADDisabledViaEnv::test_vad_disabled_via_env_pass_through` — `RAPID_MLX_STT_VAD_PRETRIM=0` fully bypasses guard.
- `TestVADDisabledViaEnv::test_env_disable_accepts_common_falsy_strings` — parametrized over `false / no / off / FALSE / Off`.
- `TestAbsoluteTimestampsPreserved::test_absolute_timestamps_preserved_when_vad_trims_leading_silence` — 3 s silence + speech → segment.start ≥ 2.7 s.
- `TestAbsoluteTimestampsPreserved::test_word_level_timestamps_also_shifted` — per-word timestamps shifted too.
- `TestKwargOverrideDisables::test_enable_vad_pretrim_false_kwarg_disables` — kwarg False disables.
- `TestVADImportFailureFallsBack::test_vad_import_failure_falls_back` — simulated `mlx_audio.vad` `ImportError` → falls back to unmodified path.
- `TestParakeetEngineSkipsVAD::test_parakeet_engine_skips_vad` — Parakeet engine never consults VAD.
- `TestEnvHelper` (12 parametrized) — env-parser sanity.
- `TestShiftHelper` (4) — segment/word/object timestamp shifting.

No regressions in existing STT/audio suites:
- `tests/test_stt_response_format.py` — 18 passed
- `tests/test_stt_corrupted_file.py` — 14 passed
- `tests/test_audio*.py` — 204 passed

## Operator-visible behaviour changes

* **New env**: `RAPID_MLX_STT_VAD_PRETRIM` — set to `0` / `false` / `no` / `off` to disable the guard run-wide.
* **New engine kwarg**: `STTEngine(..., enable_vad_pretrim: bool = True)`.
* **First-request latency** on Whisper engines: adds a one-shot ~2 s VAD model download + load (cached in `~/.cache/huggingface/hub/models--mlx-community--silero-vad`), then ~200 ms VAD inference per call. Net latency is negative for silent clips (skips multi-second Whisper decode).
* **Timestamps in transcription responses**: now file-relative absolute times even when the guard trims leading silence (previously would have been `0.0`-relative to the trimmed span had the trim path existed — it did not).
* **Silence semantics**: pure-silence uploads return `text=""`, `segments=[]`, `duration=0.0`, `language=None` (was: hallucinated tokens like `"Thank you."`).

## Test plan

- [x] Reporter's `ffmpeg silence.wav` + Python snippet returns `''` on this branch.
- [x] Real 2.5 s speech still transcribes correctly.
- [x] `3 s silence + speech + 3 s silence` returns clean segment with absolute `start ≈ 2.78 s` and no tail hallucination.
- [x] `RAPID_MLX_STT_VAD_PRETRIM=0` restores pre-fix behaviour.
- [x] `tests/test_stt_vad_pretrim.py` — 30/30 pass.
- [x] `tests/test_stt_*` + `tests/test_audio*` — no regressions.
- [x] `ruff check` + `ruff format --check` — clean.

Closes #961

### DIFF
--- a/docs/guides/audio.md
+++ b/docs/guides/audio.md
@@ -261,6 +261,35 @@ curl http://localhost:8000/v1/audio/transcriptions \
   -F model=whisper-large-v3
 ```
 
+#### Silence-hallucination guard (Whisper only)
+
+Whisper is known to invent tokens like `"Thank you."` or
+`"Thanks for watching!"` when handed pure-silence input, and to append
+similar tails to legitimate clips that end in dead-air (issue #961).
+Every Whisper request is therefore pre-filtered by Silero VAD
+(bundled with the `[audio]` extra):
+
+- **Pure-silence clips** — VAD detects no speech; the endpoint returns
+  `{"text": "", "segments": [], "language": null}` and skips Whisper
+  entirely (latency win: ~200 ms VAD vs multi-second decode).
+- **Clips with leading / trailing silence** — VAD reports the speech
+  span; Whisper only sees the trimmed waveform, and returned segment
+  timestamps are shifted back to the original file's absolute time.
+- **Sanity check** — if VAD reports no speech but the waveform still
+  has non-trivial energy (RMS > ~-50 dBFS), the guard trusts the
+  audio and falls back to Whisper. Real quiet / whispered speech is
+  never silenced.
+
+Operator controls:
+
+| Knob | Type | Default | Effect |
+| --- | --- | --- | --- |
+| `RAPID_MLX_STT_VAD_PRETRIM` | env var | on | Set to `0`, `false`, `no`, or `off` to disable the guard run-wide and restore the pre-fix pass-through behaviour. |
+| `STTEngine(..., enable_vad_pretrim=True)` | Python API | `True` | Same effect at the engine level; use when embedding `STTEngine` directly. |
+
+The guard applies only to Whisper backends. Parakeet, Canary, and
+other non-Whisper STT engines are never pre-filtered.
+
 ### POST /v1/audio/speech
 
 Generate speech from text (OpenAI TTS API compatible).
--- a/tests/test_no_out_of_band_routing.py
+++ b/tests/test_no_out_of_band_routing.py
@@ -338,6 +338,16 @@ ALLOWED_RAPID_MLX_ENV_VARS: frozenset[str] = frozenset(
         # blow up the FS. Pure wire-level capacity gate — never
         # selects a model, parser, or routing tier.
         "RAPID_MLX_KV_CHECKPOINT_MAX_BYTES",
+        # F-K-WHISPER-961 anti-hallucination VAD pre-trim opt-out
+        # (accepts ``0`` / ``false`` / ``no`` / ``off``). Consumed by
+        # ``vllm_mlx.audio.stt._vad_pretrim_disabled_by_env`` only to
+        # decide whether ``STTEngine.transcribe`` runs Silero VAD
+        # before Whisper for silence-hallucination suppression on
+        # pure-silence and trailing-silence clips (issue #961). Pure
+        # behaviour-policy knob on a single existing STT engine —
+        # never selects a model, parser, or routing tier; identical
+        # semantic shape to ``RAPID_MLX_REASONING_RESCUE`` above.
+        "RAPID_MLX_STT_VAD_PRETRIM",
     }
 )
 
--- a/tests/test_stt_vad_pretrim.py
+++ b/tests/test_stt_vad_pretrim.py
@@ -1,0 +1,529 @@
+# SPDX-License-Identifier: Apache-2.0
+"""F-K-WHISPER-961 — Whisper silence hallucination guard.
+
+kootenayalex reported (issue #961) that ``STTEngine.transcribe`` on
+12 s of digital silence returns ``"Thank you."`` — a canonical Whisper
+silence hallucination. The documented anti-hallucination guards
+(``no_speech_threshold`` / ``logprob_threshold`` /
+``hallucination_silence_threshold``) inside
+``mlx_audio.stt.models.whisper.Model.generate`` are inert in practice:
+on pure silence the model returns ``no_speech_prob ≈ 1e-11`` and
+``avg_logprob ≈ -0.26``, and the AND-gate skip condition never fires.
+
+The fix at ``vllm_mlx/audio/stt.py`` runs the bundled Silero VAD
+(via ``mlx_audio.vad``) before Whisper:
+
+* No speech → return an empty ``TranscriptionResult`` without invoking
+  Whisper.
+* Speech present → trim to the speech span, invoke Whisper on the
+  trimmed waveform, then shift each returned segment (and per-word)
+  timestamp back by the trim offset so callers see absolute times.
+
+These tests stub the VAD + Whisper models to exercise the guard's
+control flow deterministically without downloading multi-hundred-MB
+weights on the CI runner.
+"""
+
+from __future__ import annotations
+
+import importlib
+
+import pytest
+
+# ---------------------------------------------------------------------------
+# Fake VAD + Whisper doubles. Both mimic the shape the production code
+# consumes: ``vad.get_speech_timestamps(...)`` returns a list of
+# ``{start,end}`` dicts (seconds), and ``whisper.generate(audio, ...)``
+# returns an ``STTOutput``-shaped object with ``text``, ``segments``,
+# ``language``.
+# ---------------------------------------------------------------------------
+
+
+class _FakeVAD:
+    """VAD stub that returns a pre-configured speech-timestamp list.
+
+    ``timestamps`` is a list of {"start": s, "end": s} dicts (seconds).
+    The test controls the list to simulate every VAD outcome shape:
+    empty (pure silence), single span (leading/trailing silence trim),
+    multiple spans (mid-silence gap).
+    """
+
+    def __init__(self, timestamps: list[dict]):
+        self._timestamps = timestamps
+        self.call_count = 0
+        self.last_kwargs: dict = {}
+
+    def get_speech_timestamps(self, audio, **kwargs):
+        self.call_count += 1
+        self.last_kwargs = kwargs
+        # We copy so callers can't mutate our fixture in place.
+        return [dict(ts) for ts in self._timestamps]
+
+
+class _FakeWhisperResult:
+    """Mimics ``mlx_audio.stt.models.whisper.STTOutput`` — a dataclass
+    with ``text``, ``segments`` (list of dicts), ``language``."""
+
+    def __init__(self, text: str, segments: list[dict], language: str = "en"):
+        self.text = text
+        self.segments = segments
+        self.language = language
+
+
+class _FakeWhisperModel:
+    """Records every ``generate(...)`` call so the tests can assert both
+    that Whisper was NOT called on pure silence AND that the correct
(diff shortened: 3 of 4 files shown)


kootenayalex:
Thanks for the quick turnaround on this — confirmed the fix logic itself is correct. I pulled `main` (78d3005) and ran the exact repro plus a leading/trailing-silence case against a real endpoint:

- 12s pure silence → `""` (was `"Thank you."`)
- Real speech → transcribed correctly, unchanged
- `3s silence + speech + 3s silence` (`verbose_json`) → single clean segment, no invented tail, `segment.start: 2.778` — absolute timestamp correctly preserved after the leading-silence trim

However, **this ships non-functional on a fresh install**, so #961 is still open in practice for anyone doing a normal `pip install rapid-mlx[audio]` (or `uv tool install`).

### Root cause

`_maybe_vad_trim()` calls `_get_vad_model()` → `mlx_audio.vad.load("mlx-community/silero-vad")`. That HF repo's `config.json` declares `model_type: "silero_vad"`, which only exists as a registered VAD architecture in **mlx-audio >= 0.4.4** (confirmed empirically: `mlx_audio/vad/models/` in 0.4.3 has `sortformer` + `smart_turn` only; 0.4.4 adds `silero_vad` and `fsmn`).

But `pyproject.toml`'s `audio` extra pins:

```
"mlx-audio>=0.2.9,<0.4.4",
```

— the R7-H3 ceiling added to dodge the Kokoro-TTS `SineGen` regression in 0.4.4. That comment says "Cap the upper bound until upstream ships a fix so every fresh `pip install rapid-mlx[audio]` works out of the box" — which was true before #980, but #980 now needs the exact version this ceiling excludes.

On a clean install this resolves to mlx-audio 0.4.3, and every transcription request logs:

```
ERROR:root:Model type silero_vad not supported for vad.
WARNING:rapid_mlx.audio.stt:VAD pre-trim: could not load 'mlx-community/silero-vad' (...). Retrying on next request. Guard falls back to unmodified transcription until load succeeds.
```

...and silently falls back to the pre-fix behavior (silence → `"Thank you."` again).

### Why the test suite didn't catch it

`tests/test_stt_vad_pretrim.py` monkeypatches `_get_vad_model` with a fake stub for all 58 tests (`monkeypatch.setattr(stt_mod, "_get_vad_model", _fake_get_vad_model)`), so none of them exercise the real loader against the real pinned dependency. The manual repro in the PR description was almost certainly run against a dev environment that already had a newer mlx-audio installed, not a fresh extras install.

### Repro

```bash
python -m venv /tmp/v && /tmp/v/bin/pip install -q "mlx-audio==0.4.3"
python -c "from mlx_audio.vad import load; load('mlx-community/silero-vad')"
# RuntimeError: Model type silero_vad not supported for vad.

/tmp/v/bin/pip install -q -U "mlx-audio==0.4.4"
python -c "from mlx_audio.vad import load; load('mlx-community/silero-vad')"
# loads fine
```

### Possible fixes

- Split the shared `audio` extra into separate version ranges for TTS vs. STT/VAD deps (they don't need to move in lockstep), or
- Bump the ceiling once the Kokoro `SineGen` regression is fixed upstream, or
- Add at least one test that calls the real `_get_vad_model()` (network-gated/skippable in CI, but would've caught this before merge).

Happy to open this as a separate issue instead if you'd rather track it apart from #961 — let me know. In the meantime I've worked around it locally by force-installing `mlx-audio==0.4.4` into our my deployment (safe for me since I don't use the TTS lane), but that's not something a typical installer would know to do.

PR #987, opened 2026-07-01
feat(mtp): universal draft-k auto-tune controller (0.9.11 PR-1)

raullenchai:
## Summary

PR-1 of the 0.9.11 Gemma-4 MTP roadmap. Lands the **universal MTP draft-``k`` auto-tune controller** — a runtime feedback loop that adjusts the number of draft tokens per verify pass based on the rolling accept rate. Uses hysteresis thresholds + a cooldown so a noisy accept rate around the boundary doesn't oscillate ``k`` every attempt. Design is future-proof: PR-2/3 (Gemma-4 sidecar, blocked upstream) will consume ``controller.current_k()`` inside the generator hot loop to actually vary the drafting count; PR-1 only lands the controller + feedback wiring.

**Zero behavioral change when off** — the controller is opt-in via ``--mtp-draft-k-auto-tune`` (default False). When off, the vendored ``mtp_generate_step`` runs byte-identically to the pre-PR-1 code (guarded by an ``is None`` check in the hot loop). Regression test locks this in.

## Design

- **Hysteresis (two thresholds).** ``upshift_threshold`` (default 0.80) is the "one more draft is a win" gate; ``downshift_threshold`` (default 0.40) is the "one fewer draft" gate. Between them → hold. A single-threshold controller would ping-pong ``k = 3 ↔ k = 4`` when the rate hovers right at the boundary.
- **Cooldown (default 128 attempts).** Prevents a lucky streak of accepts right after cold start from ratcheting ``k`` up before steady-state kicks in. Every re-evaluation resets the cooldown counter, whether we adjusted or held.
- **Rolling window (default 64).** Bounded ``deque(maxlen=W)``. Adjustment decisions require a fully-populated window — same cold-start guard as above.
- **Bounds clamped.** ``k`` is clamped to ``[1, k_max]``. Degenerate ``k_min == k_max == 1`` still records attempts (for /metrics) but never adjusts.
- **NaN / inf gate at construction time.** Per gotcha ``pydantic-field-does-not-reject-nan``, ``math.isfinite`` is the explicit guard. Inverted thresholds (``upshift <= downshift``) raise loudly — they would silently ratchet ``k`` down on high accept rates.
- **Pure Python, no MLX imports.** Testable without spawning a server. Thread-safe via one lock (held for O(1) work per call).

## Flag surface

| Flag | Type | Default | Notes |
| ---- | ---- | ------- | ----- |
| ``--mtp-draft-k-auto-tune`` | bool | False | Requires ``--spec-decode mtp`` at boot; boot fails otherwise |
| ``--mtp-draft-k-max`` | int | 4 | Range ``[1, 8]``; ignored when auto-tune is off |
| ``--mtp-draft-k-window`` | int | 64 | Rolling window size (attempts) |
| ``--mtp-draft-k-upshift-threshold`` | float | 0.80 | Must be in ``(0, 1]``; NaN/inf rejected |
| ``--mtp-draft-k-downshift-threshold`` | float | 0.40 | Must be in ``[0, 1)`` and strictly less than upshift |
| ``--mtp-draft-k-cooldown`` | int | 128 | Min attempts between adjustment decisions |

When both ``--mtp-num-draft-tokens`` and ``--mtp-draft-k-auto-tune`` are set, the num-draft-tokens value is used as the STARTING ``k`` (must satisfy ``1 <= start <= k_max``).

## Boot-gate scope (post-review tightening)

The auto-tune gate requires ``--spec-decode mtp`` — NOT ``--enable-mtp``. Rationale: the only ``record_attempt()`` feedback wiring in PR-1 lives inside the vendored ``mtp_generate_step`` (mlx-lm PR #990 path), which only runs under ``--spec-decode mtp``. The legacy ``--enable-mtp`` (Qwen3-Next) path uses ``_install_mtp`` in ``scheduler.py`` which does NOT touch the controller — allowing ``--enable-mtp`` alone would install a controller that never receives feedback (silent no-op). A future PR that wires the controller into ``_install_mtp`` can loosen this.

## Prometheus surface

New gauge ``rapid_mlx_spec_decode_mtp_current_draft_k`` (label set ``{family, method="mtp"}``, joined with existing ``rapid_mlx_spec_decode_*`` counters via ``family``). Emitted **only** when a controller has been installed at boot — a placeholder ``0`` would falsely imply the controller ratcheted ``k`` to zero (which the ``k_min >= 1`` bound makes impossible).

## Test plan

- [x] Ramp-up: all-accepts feed ramps ``k`` up to ``k_max`` at cooldown-spaced boundaries, then holds
- [x] Ramp-down: all-rejects feed ramps ``k`` down to ``k_min``, then holds
- [x] In-band accept rate (0.40 < rate < 0.80) → zero adjustments over 1000 attempts
- [x] No oscillation within the same cooldown window (flip mid-window → no downshift until next eval)
- [x] Constructor rejects NaN / +Inf / -Inf on either threshold
- [x] Constructor rejects inverted thresholds (upshift <= downshift)
- [x] Constructor rejects out-of-range thresholds (upshift ∉ (0, 1], downshift ∉ [0, 1))
- [x] Constructor rejects bad integer bounds (k_min < 1, k_max < k_min, window < 1, cooldown < 1)
- [x] Constructor rejects ``bool`` where ``int`` expected (bool is subclass of int in Python)
- [x] Constructor rejects k_start outside [k_min, k_max]
- [x] Degenerate ``k_min == k_max == 1`` never adjusts, never crashes
- [x] Snapshot returns 11 documented keys
- [x] Module-global controller defaults to ``None`` (static-k fast path)
- [x] ``install_global_controller`` + ``clear_global_controller`` round-trip
- [x] Generator regression: static-k path emits byte-identical tokens with no controller installed
- [x] Metrics gauge absent when controller not installed
- [x] Metrics gauge emits current ``k`` when controller installed
- [x] All six new CLI flags advertised in ``rapid-mlx serve --help``
- [x] All 41 pre-existing MTP tests still pass (no regression)

Test suite: ``pytest tests/test_mtp_draft_k_controller.py tests/test_mtp_spec_decode.py tests/test_mtp_lossless.py`` → 65 passed, 0 failed.

## Bench sanity

**SKIPPED** — deliberate. The bench script ``bench/bench_spec_decode_mtp.py`` compares ``--spec-decode mtp`` against ``--spec-decode none`` and doesn't yet accept a ``--mtp-draft-k-auto-tune`` flag. More fundamentally, **PR-1's controller does not vary the drafting count** — the generator's verify pass still drafts exactly one token per step (the vendored PR #990 chain-MTP behavior). The controller only records feedback; consuming ``current_k()`` in the hot loop is PR-2/3. So a bench comparing auto-tune-on vs auto-tune-off would show numbers indistinguishable from noise. Numbers will land with PR-2 when ``current_k()`` actually changes the drafting count.

## Notes for review

- Version file (``pyproject.toml``) untouched — the 0.9.11 bump is a separate follow-up commit, not part of this PR.
- Vendored ``mtp_generate_step`` verify/accept arithmetic is preserved verbatim; only two ``if draft_k_controller is not None: controller.record_attempt(...)`` calls added inside the existing accept/reject branches.
- ``_SUPPORTED_MODEL_TYPES`` in ``detect.py`` untouched — Gemma 4 MTP is PR-2/3 (upstream sidecar parked).
- Test fixture in the new ``test_mtp_draft_k_controller.py`` mirrors the ``_reset_mtp_module_state`` autouse fixture from ``test_mtp_spec_decode.py`` (resets ``mlx_lm.generate.generation_stream``, unpatches ArraysCache, resets the accept counter) so full-sweep runs don't leak state into the other MTP test files.

### DIFF
--- a/tests/test_mtp_draft_k_controller.py
+++ b/tests/test_mtp_draft_k_controller.py
@@ -49,10 +49,47 @@ def _clear_controller_singleton():
 
     Tests that flip on the global for integration coverage must not
     leak into subsequent cases.
+
+    Also resets ``mlx_lm.generate.generation_stream`` and the
+    ``vllm_mlx.spec_decode.mtp`` module singletons — mirrors the
+    :func:`tests.test_mtp_spec_decode._reset_mtp_module_state`
+    fixture so this file is robust to full-sweep ordering. Without
+    it, the integration test at the bottom of this file can leave
+    the ArraysCache patch mutated (or the ``generation_stream``
+    stamped by a prior sweep test's worker executor) and cascade
+    failures into ``tests/test_mtp_lossless.py`` +
+    ``tests/test_mtp_spec_decode.py`` when the whole suite runs
+    together. See the full rationale in the corresponding fixture
+    docstring at ``tests/test_mtp_spec_decode.py:_reset_mtp_module_state``.
     """
+    import sys
+
+    try:
+        import mlx.core as mx
+        import mlx_lm.generate  # noqa: F401
+    except ImportError:
+        mx = None
+
+    from vllm_mlx.spec_decode.mtp.accept_counter import (
+        reset_global_counter_for_tests,
+    )
+    from vllm_mlx.spec_decode.mtp.cache_patch import _unpatch_for_tests
+
     clear_global_controller()
+    _unpatch_for_tests()
+    reset_global_counter_for_tests()
+    if mx is not None:
+        sys.modules["mlx_lm.generate"].generation_stream = mx.default_stream(
+            mx.default_device()
+        )
     yield
     clear_global_controller()
+    _unpatch_for_tests()
+    reset_global_counter_for_tests()
+    if mx is not None:
+        sys.modules["mlx_lm.generate"].generation_stream = mx.default_stream(
+            mx.default_device()
+        )
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


raullenchai:
Closing as superseded. The universal draft-k auto-tune controller in this PR is inert without a chain-of-K generator to consume current_k(), and the full 16-row bench (2026-07-02) revealed the controller's rolling window state leaks across requests, breaking determinism. Chain-of-K + safe auto-tune deferred to 0.9.12 after deeper Ollama comparison. Branch preserved (not deleted) for potential revival of the eligibility gate + skip/tokens counters in the next iteration.

PR #993, opened 2026-07-02
fix(responses): accept Codex namespace tool groups on /v1/responses

wuwangzhang1216:
## Why

Codex CLI can't use the `/v1/responses` lane — every session 400s on the first turn:

```
unsupported_tool_type: 'namespace'
```

Captured from a real `codex exec` request, the `tools` array is:

- **8 × `function`**
- **1 × `namespace`** — Codex's `multi_agent_v1` group, which wraps 5 function tools (`spawn_agent`, `close_agent`, `resume_agent`, `send_input`, `wait_agent`)
- **1 × `web_search`**

`SUPPORTED_RESPONSES_TOOL_TYPES` only accepts `function` / `computer_20251022`, so `validate_responses_tool_types` rejects the whole request before it reaches the engine. The shape is identical from Codex 0.137 through 0.142.5. Codex is our 0.10.1 Tier-1 agent (see 0.10-TODO.md), so this is a mandatory pickup for 0.10.0.

## Fix

In `normalize_responses_tool_types` (which already runs BEFORE the allowlist gate):

- **flatten** each `namespace` group into its contained `function` tools;
- **drop** hosted tool types (`web_search`, `web_search_preview`, `file_search`, `code_interpreter`, `image_generation`) the local engine can't execute — **but only when the request also carries at least one function/namespace tool** (see F13 trade-off below);
- **guard** the namespace `tools` field against non-list shapes so a malformed `{"type":"namespace","tools":1}` falls through to `validate_responses_tool_types` and returns a controlled 400 instead of raising TypeError.

`_convert_tools` still rejects hosted types when called directly, so the F13 explicit-rejection contract for callers that pass them through is preserved.

## F13 trade-off (why the drop-hosted step is conditional)

Yuki F13 (0.8.5 dogfood) fires 400 on `tools=[{"type":"web_search"}]` because a caller who directly asked for a hosted tool must not silently get a 200 that never runs the tool. Codex 0.137's shape adds `web_search` as ambient noise (regardless of user intent) alongside real function tools — 400ing that would break Codex entirely.

Resolution: **drop hosted tools ONLY when the request has function/namespace tools present**. Codex's mixed case (function + hosted noise) proceeds cleanly; F13's direct-user case (hosted-only) still 400s with the supported-types list. All 4 F13 tests pass unchanged.

## Verification

- **End-to-end on Codex 0.142.5** (per original author @wuwangzhang1216): agent drives `/v1/responses` on Qwen3.5-9B (MLX) through a multi-step read → edit → run task. `tools=13` after flattening, every response 200.
- **New tests** in `TestNormalizeResponsesToolTypes` (7 total, was 4):
  - flatten namespace + drop hosted + realistic-Codex-shape passes validation + no-op canonicalisation (author)
  - `test_hosted_only_preserves_f13_400_path` — F13 preservation
  - `test_namespace_flatten_triggers_hosted_drop` — namespace-with-function counts as function for drop-gate
  - `test_malformed_namespace_falls_through_to_validate` — 4 malformed shapes (`tools=1`/`"str"`/`{}`/missing) must not raise
- **F13 regression suite**: `TestF13UnsupportedToolTypes` — 6 tests all pass unchanged.
- **Full targeted suite**: `test_responses_adapter.py` + `TestF13UnsupportedToolTypes` — 86 passed.

## Author's open question — flatten vs drop-namespace

@wuwangzhang1216 asked whether to flatten the multi-agent group into function tools or drop it entirely. **Flatten (keeping author's choice)** — Codex's orchestrator handles the sub-tool calls; dropping the group would break Codex's multi-agent scenarios silently. The model calling `spawn_agent` etc. is Codex's intended flow.

## Attribution

Original PR by @wuwangzhang1216. Follow-up commit adds F13 preservation + malformed-namespace guard from codex adversarial review + PR body rewrite.

Closes #993.

### DIFF
--- a/tests/test_responses_adapter.py
+++ b/tests/test_responses_adapter.py
@@ -224,6 +224,44 @@ class TestNormalizeResponsesToolTypes:
             # will 400 on the ``namespace`` type.
             assert any(t.get("type") == "namespace" for t in tools)
 
+    def test_empty_namespace_with_hosted_does_not_silently_collapse(self):
+        """Codex review round-2 case: an empty ``namespace`` (``tools=[]``)
+        paired with a hosted tool must NOT normalize to ``[]``. The empty
+        namespace is preserved for validation (it 400s on ``type:namespace``),
+        and the hosted tool is preserved because no real function survives
+        flattening — so ``post_flatten_has_function`` stays False and the
+        drop-hosted step is disabled. Both are preserved for validate to
+        400 cleanly instead of silently accepting a zero-tool request.
+        """
+        tools = [
+            {"type": "namespace", "name": "x", "tools": []},
+            {"type": "web_search"},
+        ]
+        normalize_responses_tool_types(tools)
+        # Empty namespace preserved for validate → validate will 400 on
+        # `type:namespace`. Hosted tool also preserved (no function to
+        # anchor the drop). Zero silent-drop.
+        assert any(t.get("type") == "namespace" for t in tools)
+        assert any(t.get("type") == "web_search" for t in tools)
+
+    def test_namespace_with_nondict_children_falls_through(self):
+        """Codex review round-2 case: ``{"type":"namespace","tools":["bad"]}``
+        must NOT silently discard the non-dict child — the whole namespace
+        entry is preserved so validate can 400 it. Otherwise a caller with
+        a typo (list of strings instead of list of dicts) would see their
+        namespace erased and get an ambiguous downstream error.
+        """
+        tools = [
+            {"type": "function", "name": "f"},
+            {"type": "namespace", "name": "x", "tools": ["bad-string-child"]},
+        ]
+        normalize_responses_tool_types(tools)
+        # Namespace with non-dict children preserved for validate. No
+        # silent flatten.
+        types = [t.get("type") for t in tools]
+        assert "namespace" in types
+        assert "function" in types
+
 
 # ---------------------------------------------------------------------------
 # Tool-choice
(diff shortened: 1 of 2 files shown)


raullenchai:
Thanks for the PR — this is a well-scoped fix with a clear problem statement, solid end-to-end verification on Codex 0.142.5, and good test coverage for the happy path.

Overall LGTM, but a few suggestions worth addressing before/after merge:

1. **Namespace flattening safety.** Flattening `multi_agent_v1` into plain `function` tools assumes tool names are globally unique and that no namespace-level metadata matters. If two namespaces contain the same name, or if future clients attach permissions/config at the namespace level, dispatch becomes ambiguous silently. Consider adding collision detection or namespace-prefixing before flattening.

2. **Silent drops vs. explicit errors.** Dropping hosted tools (`web_search`, `file_search`, `code_interpreter`, `image_generation`) and malformed namespace children changes the API contract from "unsupported → 400" to "best effort → partial success". Please add logging so callers can tell their requested tool set was reduced, and document this behavior clearly.

3. **Edge-case test coverage.** Would be great to add tests for:
   - duplicate function names across namespaces,
   - malformed namespace children (`None`, strings, dicts without `type`),
   - nested namespaces,
   - end-to-end dispatch after flattening.

4. **Maintainability.** Consider moving the `_drop_hosted` set to a module-level constant with a short policy comment, rather than an inline local set that can drift as tool types evolve.

5. **Multi-agent semantics.** Since a single-model local backend can't actually orchestrate sub-agents, make it explicit in the docstring/PR description that `spawn_agent` / `close_agent` / etc. are exposed as plain functions and the backend does not implement agent lifecycle.

Happy to merge once logging/docs and a couple of edge-case tests are added.

raullenchai:
Thanks @wuwangzhang1216 — merged! 🎉 Added you to the changelog:

PR #996, opened 2026-07-03
fix(mtp): address codex round-16 blocking findings (post-merge follow-up on PR #990)

raullenchai:
## Summary

Follow-up to #990 (merged as ``aa1762e4`` in 0.9.11). Two BLOCKING findings from codex adversarial review landed AFTER the merge and need to ship in 0.9.12.

## Root cause 1 — short target ``layer_types`` slips past guard

``inject_mtp_support`` verified target's tail ``layer_types`` against the assistant's only when ``len(target_layer_types) >= n_assistant``. If a target published a shorter list, the guard fell through and ``_shared_kv_target_indices`` was later built from an invalid tail mapping (or the positional fallback with negative indices via Python wrap-around). Crashes inside ``mtp_forward`` at ``target_cache[idx]`` or feeds the drafter K/V from a wrong-type target layer.

**Fix:** explicitly refuse when ``len(target_layer_types) < n_assistant`` before wiring the drafter.

## Root cause 2 — sliding drafter layers inherit full-attention offset

``mtp_forward`` snapped ``base_offset = int(shared_kv_slots[-1].offset)`` — the last shared slot's offset — and used it for every drafter layer's RoPE. Under rotating / sliding cache behavior sliding target-cache slots carry a different offset than the full-attention tail slot, so sliding drafter layers rotate their Q with an offset that doesn't match the K/V they read.

**Fix:** snapshot ``layer_offsets: list[int]`` alongside the shared K/V, then use each layer's own ``layer_offsets[i] - (N - 1) + pos`` inside the per-position drafter loop.

## Verification

- ``tests/test_mtp_gemma4_assistant_inject.py::test_inject_refuses_when_target_layer_types_shorter_than_assistant`` locks the new guard.
- All 33 unit tests in the assistant-inject file pass locally.
- Ruff clean.

Per-slot offset behavior is covered by the existing shape/forward tests + the 12B smoke bench downstream. No test added for the offset fix here because the existing suite already exercises the layer-offset code path structurally; the fix is a correctness change under sliding-cache behavior that a unit-level target fake cannot reliably reproduce without also porting the sliding-cache rotation logic.

## Impact assessment

Both bugs are behind additional gates in 0.9.11's real-world hot path — real Gemma 4 12B has 42 target layers vs the assistant's 4, so ``len(target_layer_types) >= n_assistant`` always holds under a well-formed config. The sliding-offset issue only surfaces once the sliding cache rotates (long context, > sliding_window tokens). 0.9.11 users on short-context 12B are unaffected in practice.

## Test plan

- [x] 33/33 assistant-inject tests pass locally in 0.92 s
- [x] Ruff format + check clean
- [x] Cherry-picked from PR #990 branch; no other file touched

### DIFF
--- a/tests/test_mtp_gemma4_assistant_inject.py
+++ b/tests/test_mtp_gemma4_assistant_inject.py
@@ -1079,6 +1079,52 @@ def test_inject_refuses_when_target_tail_layer_types_mismatch(tmp_path):
     assert ok is False, "layer_types-tail mismatch must fail closed"
 
 
+def test_inject_refuses_when_target_layer_types_shorter_than_assistant(tmp_path):
+    """Codex round-16 blocking-fix locked in.
+
+    When the target publishes ``layer_types`` but the list is shorter
+    than the assistant's, the previous revision fell through the guard
+    and ``_shared_kv_target_indices`` was later built from an invalid
+    tail mapping (or the positional fallback with negative indices via
+    Python's wrap-around). That crashes inside ``mtp_forward`` at
+    ``target_cache[idx]`` or feeds the drafter K/V from a wrong-type
+    target layer. Inject must refuse closed before wiring the drafter.
+    """
+    from mlx.utils import tree_flatten
+
+    from vllm_mlx.spec_decode.mtp.gemma4_inject import (
+        _build_assistant_model,
+        _build_assistant_model_args,
+        inject_mtp_support,
+    )
+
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing as superseded. PR #990 is being reverted (see #997); the round-16 fixes will be baked into a fresh re-apply PR that goes through full CLI pr_validate + codex before merge.

PR #1031, opened 2026-07-06
fix(nemotron): fail-open tool-call parsing for degraded wire variants

arthurhaliski:
## Problem

`NemotronToolParser` (`vllm_mlx/tool_parsers/nemotron_tool_parser.py`, registered as `nemotron` / `nemotron3`) keys tool-call extraction on the **full** `<tool_call>...</tool_call>` wrapper:

```python
TOOL_CALL_PATTERN = re.compile(
    r"<tool_call>\s*<function=([^>]+)>(.*?)</function>\s*</tool_call>",
    re.DOTALL,
)
```

Nemotron models — especially at low quantization or after several tool rounds — frequently emit near-miss variants of that shape. Because the wrapper (and its closing tag) are mandatory in the regex, every one of these variants fails to match and the call **leaks through as plain assistant `content`** instead of a structured tool call:

| | Variant | Example |
|---|---|---|
| **(a)** | missing / truncated `</tool_call>` | `<tool_call><function=f><parameter=x>1</parameter></function>` |
| **(b)** | bare `<function=..>..</function>`, no wrapper at all | `<function=f><parameter=x>1</parameter></function>` |
| **(d)** | stray text between `</function>` and `</tool_call>` | `<tool_call><function=f>..</function> ok </tool_call>` |
| **(e)** | prose between `<tool_call>` and `<function=` | `<tool_call>Let me check.<function=f>..</function></tool_call>` |

The streaming path has the matching gap: its sentinel requires `<tool_call>` and it only fires on `</tool_call>`, so a truncated variant that ends at `</function>` never emits a streamed tool call.

## Fix

The load-bearing signature of a call is `<function=NAME>...</function>`; the `<tool_call>` / `</tool_call>` wrapper is decorative. So key on the function signature and treat the wrapper as optional:

- `TOOL_CALL_PATTERN` → `r"<function=([^>]+)>(.*?)</function>"`.
- New `RESIDUAL_WRAPPER_PATTERN = re.compile(r"</?tool_call>")` strips leftover bare `<tool_call>` / `</tool_call>` tags from `content` so they don't leak as text.
- The early-out now returns *not-called* only when **neither** `<tool_call>` **nor** `<function=` is present. Prose that lacks a `<function=..>..</function>` signature still never matches, so this **cannot manufacture a tool call out of plain text** — verified by the zero-false-positive tests below.
- **Streaming** (`extract_tool_calls_streaming`): the guard accepts `<tool_call>` **or** `<function=`, and fires on `</tool_call>` **or** `</function>`. Since `extract_tool_calls` re-parses the full `current_text` and returns *all* calls, the new calls are de-duped against `current_tool_id` (which `reset()` zeroes per request), so a call is emitted exactly once even when `</function>` and `</tool_call>` arrive in separate deltas.
- A diagnostic `WARNING` (with truncated raw `model_output`) is logged on the miss path — a `<tool_call>`/`<function=` marker was present but nothing parsed — so any still-unhandled wire variant can be captured from logs.

The `<parameter=..>` handling and the JSON-body path are **unchanged**.

## Tests

Adds `tests/test_nemotron_tool_parser_fail_open.py` — 15 cases:

- canonical regression (parameter and JSON-body wrappers still parse),
- variants **(a) / (b) / (d) / (e)**,
- surrounding-prose bare call with **no XML leak** into `content`,
- **zero false positives**: plain prose, and prose that mentions the word "function" but has no tag,
- diagnostic-log miss path (marker present, nothing parseable → fails open + warns),
- streaming: passthrough without a marker, **close on `</function>`** for a truncated variant, close on `</function>` for a bare (unwrapped) call, **de-dup** across separate `</function>` / `</tool_call>` deltas, and index increment across two sequential calls.

The test imports only the parser class, so it runs without the MLX model runtime. `tests/test_tool_parsers.py` (156 tests, including the existing `TestNemotronToolParser`) still passes unchanged.

## Note on affected versions

The parser is **byte-identical in 0.9.8 and 0.10.1** (the current `main`), so both releases are affected.

---

Co-authored with Claude Opus 4.8 (the commit carries a `Co-Authored-By` trailer). Happy to adjust naming, log level, or test layout to match project conventions.

### DIFF
--- a/tests/test_nemotron_tool_parser_fail_open.py
+++ b/tests/test_nemotron_tool_parser_fail_open.py
@@ -1,0 +1,250 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Fail-open regression tests for :class:`NemotronToolParser`.
+
+Nemotron's *nominal* tool-call wire shape is fully wrapped::
+
+    <tool_call><function=NAME>...</function></tool_call>
+
+In practice the model (especially at low quantization / after several tool
+rounds) degrades into several near-miss variants. The historical parser
+keyed on the whole ``<tool_call>...</tool_call>`` wrapper, so every one of
+these leaked the call through as plain assistant ``content`` instead of a
+structured tool call:
+
+    (a) a missing / truncated ``</tool_call>``
+    (b) a bare ``<function=..>..</function>`` with no wrapper at all
+    (d) stray text between ``</function>`` and ``</tool_call>``
+    (e) prose between ``<tool_call>`` and ``<function=``
+
+The load-bearing signature of a call is ``<function=NAME>...</function>``, so
+the parser now keys on that and treats the wrapper as optional. Prose that
+does not contain ``<function=..>..</function>`` still never matches, so this
+cannot manufacture a tool call out of plain text.
+
+This module imports only the parser class (no MLX / model runtime), so it is
+runnable in isolation.
+"""
+
+from __future__ import annotations
+
+import json
+
+import pytest
+
+from vllm_mlx.tool_parsers.nemotron_tool_parser import NemotronToolParser
+
+
+@pytest.fixture
+def parser() -> NemotronToolParser:
+    return NemotronToolParser()
+
+
+def _only_call(result):
+    assert result.tools_called
+    assert len(result.tool_calls) == 1
+    return result.tool_calls[0]
+
+
+# ---------------------------------------------------------------------------
+# Canonical regression — the fully-wrapped shape must keep working.
+# ---------------------------------------------------------------------------
+
+
+def test_canonical_parameter_wrapper(parser):
+    text = (
+        "<tool_call><function=get_weather>"
+        "<parameter=city>Paris</parameter>"
+        "</function></tool_call>"
+    )
+    tc = _only_call(parser.extract_tool_calls(text))
+    assert tc["name"] == "get_weather"
+    assert json.loads(tc["arguments"]) == {"city": "Paris"}
+
+
+def test_canonical_json_body_wrapper(parser):
+    text = '<tool_call><function=calculate>{"expression": "2*3"}</function></tool_call>'
+    tc = _only_call(parser.extract_tool_calls(text))
+    assert tc["name"] == "calculate"
+    assert json.loads(tc["arguments"]) == {"expression": "2*3"}
+
+
+# ---------------------------------------------------------------------------
+# Degraded variants that previously leaked as text.
+# ---------------------------------------------------------------------------
+
+
+def test_variant_a_missing_closing_tool_call(parser):
+    """(a) ``</tool_call>`` truncated away — the call must still parse."""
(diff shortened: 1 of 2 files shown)


raullenchai:
Thanks for this @arthurhaliski — the fail-open direction is right and the regression test suite (canonical + degraded variants + zero-false-positive + streaming dedup) is genuinely thorough. 🙏

I ran it through our merge-readiness gate (`pr_validate`). Two of our reviewer's findings are **blocking** and both are about the new code, so I'm leaving this open for a follow-up push rather than merging as-is:

**1. `nemotron_tool_parser.py:150` — raw payload in the warning log (privacy)**
The new diagnostic logs `model_output[:500]` at WARNING level on every malformed marker. That raw slice can contain user prompts, tool arguments, or credentials, and it fires in the normal degraded-wire path (i.e. often, in production). Could you either drop the raw payload from the log, or gate it behind an explicit debug/trace flag with secrets stripped? A structural summary (e.g. "marker present, N `<function=` tags, 0 parseable") captures the diagnostic value without the payload.

**2. `nemotron_tool_parser.py:178` — streaming close-tag split across deltas (correctness)**
The streaming path now fires when `</function>` (or `</tool_call>`) appears **within a single `delta_text`**. If the tokenizer splits the close tag across chunks (`"</fun"` then `"ction>"`), the completed call is never emitted even though the full call is present in `current_text`. Your own `test_streaming_dedup_across_separate_close_deltas` covers *separate* deltas for the two different tags, but not a *single* tag split mid-token. Suggested fix: trigger from the completion state of `current_text` (re-parse and de-dupe against `current_tool_id`, which you already track) rather than from substring presence in `delta_text`, and add a regression test that splits `</function>` across two chunks.

For reference, the rest of the gate was green: `lint` clean, `supply_chain` clean, description quality good. The `full_unit` step reported 2 failures but **both are pre-existing main-branch noise unrelated to this PR** (`test_readme_quickstart_mentions_vision_extra` fails on clean `main`; `test_batched_faster_than_sequential` is a timing flake that passes in isolation) — not a blocker on your side.

Once the two items above are addressed, re-push and I'll re-run the gate. Thanks again!

raullenchai:
Landed via #1073 (squash `4c0b3dc0`) with your original commit cherry-picked — thank you @arthurhaliski 🙏. The two `pr_validate` findings are addressed: the WARNING log now emits a structural summary instead of the raw payload (privacy), and the streaming path triggers on a newly-completed close-tag counted over `current_text` so a `</function>` split across deltas still emits exactly once (with dedupe via `current_tool_id`). Full regression suite green.

raullenchai:
nemotron fail-open tool-call parsing for degraded/malformed wire variants was fixed in #1073 and shipped in **rapid-mlx 0.10.7** (PyPI + Homebrew). The parser now degrades gracefully instead of raising on partial tool-call envelopes. Thanks for the report.

PR #1094, opened 2026-07-11
feat(mtp): HY3 (Tencent Hunyuan 3) native MTP self-speculative decoding

raullenchai:
## Why

HY3 (Tencent Hunyuan 3) is a Tier-1 family in rapid-mlx, but `hy3-preview-4bit` had `supports_spec_decode: false` — no acceleration path. HY3 ships a **native** DeepSeek-V3-style MTP head that the 4-bit MLX convert strips, leaving a ~155 GB model decoding at ~0.4 tok/s with no speculative option. Because the head is native (trained, not a bolt-on drafter), grafting it back gives a lossless self-speculative speedup for free — the highest-ROI acceleration available for this family. This PR re-supplies that head and wires it behind the existing MTP entry point, so a user gets ~1.5–1.7× decode with a single flag and zero extra download management.

## What

Wire HY3's (Tencent Hunyuan 3, `model_type=hy_v3`) **native DeepSeek-V3-style MTP head** as a self-speculative draft head, behind the existing `--speculative-config '{"method":"mtp"}'` entry point. This makes `hy3-preview-4bit` a spec-decode-capable Tier-1 family.

`mlx-community/Hy3-preview-4bit` **strips** the native MTP head at convert time (keeps only backbone layers 0–79). The full-precision `tencent/Hy3-preview` keeps it at layer 80 (a single next-n predict layer, K=1). This PR extracts + quantizes that head into a published sidecar and grafts it at boot.

## User-facing

```bash
rapid-mlx serve hy3-preview-4bit --speculative-config '{"method":"mtp"}'
```

The sidecar (`mlx-community/Hy3-preview-MTP-4bit`, 2.1 GB) is **auto-resolved and downloaded** — no extra flag. Unlike Qwen3.5 (whose MTP head ships inside the base checkpoint), HY3's 4-bit backbone has no baked-in head, so `hy3_inject` defaults `mtp_sidecar` to the published repo.

## Evidence (real weights, M3 Ultra, `mlx-community/Hy3-preview-4bit`)

| Metric | Result |
|---|---|
| Inject + validate on real 155 GB model | **True / True** |
| Draft accept rate (K=1 chain, 8 prompts, pooled) | **64.5%** (98/152); per-prompt 47.6–82.4% |
| Batched-consistent lossless | **5/8 byte-match**; 3 divergences root-caused as near-tie argmax flips (below) |
| MTP-vs-MTP determinism (temp=0) | **byte-identical across runs** — the correctness gate |
| Decode tok/s speedup (MTP-on vs MTP-off) | **~1.62× mean / 1.64× median** (rigorous two-point, prefill-canceling: code 1.47× / reasoning 1.75× / chinese 1.64×) |

### On the 3/8 lossless "divergences"

These are **not** MTP bugs. Per the project's MTP lossless contract (`memory/knowledge/decisions.md`, 2026-07-02), the contract is *batched-consistent* lossless — "a draft token is accepted iff it equals the target model's greedy pick under the same batched forward" — **not** byte-equal vs a single-token `mlx-lm.generate` baseline. MLX SDPA numerics differ between `q_len=1` and `q_len≥2` under 4-bit quant, and bf16 accumulation is non-associative, so a near-tie argmax boundary can flip. Root-cause probe on the earliest divergence (chinese, pos 2):

```
pos 0: top1-top2 gap = 3.84   match=True
pos 1: gap = 1.57             match=True
pos 2: gap = 0.088            match=False   <- razor-thin near-tie flip
pos 3: gap = 6.06             match=True
```

The divergent position has a 0.088 logit gap vs 1.5–6.0 at matching positions — the same upstream-MLX behavior Google mlx-vlm and Ollama ship with. The correctness gate that *does* hold is MTP-vs-MTP determinism (byte-identical), which passes.

## Sidecar

Published to [`mlx-community/Hy3-preview-MTP-4bit`](https://huggingface.co/mlx-community/Hy3-preview-MTP-4bit) (README with full extraction provenance from `tencent/Hy3-preview` layer 80). A from-scratch `snapshot_download` resolve was verified end-to-end (2.111 GB safetensors → standard HF cache). `scripts/extract_hy3_mtp.py` is the reproducible regeneration tool.

## Changes

New:
- `vllm_mlx/spec_decode/mtp/hy3_head.py` — `_HY3MTPModule`: reuses `hy_v3.DecoderLayer` (forced onto the MoE branch) wrapped by enorm/hnorm/eh_proj/norm. Forward = `eh_proj(concat([enorm(embed_next), hnorm(prev_hidden)], -1))` (embedding-first, confirmed vs vLLM `deepseek_mtp.py` + SGLang hunyuan nextn).
- `vllm_mlx/spec_decode/mtp/hy3_inject.py` — inject/validate the four MTP contract surfaces; default sidecar auto-resolution.
- `scripts/extract_hy3_mtp.py` — index-driven shard download, expert stacking, quant-match extraction of the head.
- `tests/test_mtp_hy3_inject.py` — 17 tests mirroring the qwen3_5 path.

Registration (strictly additive):
- `detect.py` — `hy_v3` on the allowlist; `_mtp_num_hidden_layers` reads HY3's `num_nextn_predict_layers` (no cross-contamination of the Qwen `mtp_num_hidden_layers` key; `deepseek_v3` stays off the allowlist).
- `dispatch.py` — `hy_v3` → hy3 inject/validate.
- `scheduler.py` — `hy_v3` in the config-vetted MTP allowlist.
- `aliases.json` — `hy3-preview-4bit` `supports_spec_decode: true`.

## Tests

`198 passed` across the full MTP suite + the new 17 `test_mtp_hy3_inject`. No regression. Ruff clean.

## Codex review (round 1 → converged)

5 blocking findings from `gpt-5.6-sol`, all addressed:
1. **Pinned sidecar revision** — the default sidecar now resolves at an immutable commit SHA (no mutable-HEAD supply-chain window).
2. **detect/inject key-set parity** — the inject layer-count resolver reads the same `num_nextn_predict_layers` + `mtp_num_hidden_layers` set as detection, so a config deemed eligible is never refused at inject.
3. **Default-sidecar quant gate + shape guard** — the default sidecar auto-resolves only when the base matches its 4-bit/gs64 quant; a per-tensor shape check rejects any quant mismatch before load.
4. **extract derives quant from base** — `extract_hy3_mtp.py` reads `config.json::quantization` instead of hardcoding 4-bit, and rejects a base with no quant metadata.
5. **safe cache cleanup** — the script no longer unlinks shared content-addressed HF blobs (which would dangle other snapshots' symlinks); it points at `huggingface-cli delete-cache` instead.

3 new regression tests lock findings #1–#3; findings #4–#5 are script-side. The 7 `supply_chain` `eval()` warnings are false positives (`mx.eval()` is MLX tensor evaluation, not Python `eval`).

### DIFF
--- a/scripts/extract_hy3_mtp.py
+++ b/scripts/extract_hy3_mtp.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 from pathlib import Path
 
 import mlx.core as mx
@@ -296,11 +297,32 @@ def main() -> int:
         )
 
     if all_weights:
-        logger.warning(
-            "UNCONSUMED layer-%d tensors (first 8): %s",
-            mtp_layer,
-            sorted(all_weights)[:8],
-        )
+        # Schema-drift guard (codex NIT): unconsumed layer tensors mean the
+        # upstream head grew params this extractor does not map, so a warning-
+        # only path could publish a silently-incomplete sidecar. Fail hard
+        # unless the operator explicitly acknowledges via
+        # HY3_MTP_ALLOW_UNCONSUMED=1 (e.g. deliberately dropping a known-unused
+        # tensor during a controlled re-extract).
+        leftover = sorted(all_weights)
+        if os.environ.get("HY3_MTP_ALLOW_UNCONSUMED") == "1":
+            logger.warning(
+                "UNCONSUMED layer-%d tensors ignored via HY3_MTP_ALLOW_UNCONSUMED "
+                "(count=%d, first 8): %s",
+                mtp_layer,
+                len(leftover),
+                leftover[:8],
+            )
+        else:
+            logger.error(
+                "UNCONSUMED layer-%d tensors (count=%d, first 8): %s. The upstream "
+                "MTP head carries params this extractor does not map — refusing to "
+                "publish a silently-incomplete sidecar. Update the remap, or set "
+                "HY3_MTP_ALLOW_UNCONSUMED=1 to override deliberately.",
+                mtp_layer,
+                len(leftover),
+                leftover[:8],
+            )
+            return 1
 
     # --- 3. Quantize to match the base checkpoint. ---
     # Norms (1-D) + expert_bias stay FP. eh_proj / attn proj / switch_mlp /
@@ -323,10 +345,25 @@ def main() -> int:
         quantized[k.replace(".weight", ".biases")] = q_b
 
     # --- 4. Save into a DEDICATED sidecar dir (see the module note above). ---
+    # Atomic write (codex NIT): serialize to a temp sibling then os.replace onto
+    # the destination, so an interrupt / disk-full mid-write can't truncate or
+    # destroy a previously-valid sidecar at ``out``.
     out = Path(args.out).expanduser()
     out.parent.mkdir(parents=True, exist_ok=True)
-    logger.info("Saving %d tensors to %s", len(quantized), out)
-    mx.save_safetensors(str(out), quantized)
+    # mx.save_safetensors SILENTLY no-ops unless the path ends in .safetensors,
+    # so the temp sibling must keep that suffix (…​.tmp.safetensors), not a bare
+    # .tmp. os.replace within the same dir is atomic.
+    tmp_out = out.with_name(out.stem + ".tmp" + out.suffix)
+    logger.info("Saving %d tensors to %s (atomic via %s)", len(quantized), out, tmp_out)
+    mx.save_safetensors(str(tmp_out), quantized)
+    if not tmp_out.exists():
+        logger.error(
+            "mx.save_safetensors wrote nothing to %s (path must end in "
+            ".safetensors); aborting before replace.",
+            tmp_out,
+        )
+        return 1
+    os.replace(tmp_out, out)
 
     # --- 5. Spot-checks. ---
     total_bytes = sum(v.nbytes for v in quantized.values())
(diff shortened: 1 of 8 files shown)


raullenchai:
## Merge decision (CC take-over)

codex_review gate is now **PASS / MERGE-SAFE** (gpt-5.6-sol, round 11): 0 blocking, 1 non-blocking nit.

**Convergence** — each round's blocking was a *distinct, legitimate* issue, fixed in turn (not repeated nitpicking); nit count strictly decreased:
- R8 `1B+3N` → `_kind()` dtype guard too coarse → tightened to exact-dtype-unless-both-float.
- R9 `1B+2N` → `_resolve_num_mtp_layers` parsed layer count outside the fail-closed boundary → routed through `detect._safe_int` (malformed config degrades to no-MTP instead of aborting boot).
- R10 `1B+0N` → the R8 guard's test only substituted bool-in-float-slot (which the *coarse* guard already caught) → rewrote to exercise the int32-vs-uint32 branch the tightening actually added.
- R11 **PASS** `0B+1N`.

**Deferred (non-blocking):** the single remaining nit — `extract_hy3_mtp.py` leaves a multi-GB tmp sidecar on disk if `save`/`replace` fails — is in the sidecar *regeneration* dev-tool, not the runtime path. Tracked as a follow-up; does not gate this merge.

**Evidence** (real 155 GB `mlx-community/Hy3-preview-4bit`, M3 Ultra): 64.5% draft accept (K=1, 8 prompts), ~1.62× mean / 1.64× median decode two-point, MTP-vs-MTP determinism byte-identical, sidecar published + from-scratch resolve verified, 27 hy3 inject + 198 MTP-suite tests pass, all CI green.

Ships in **0.10.8** (batched with the bonsai ternary alias + backlog).

PR #1102, opened 2026-07-13
fix gemma4 nested tool arguments

romanbsd:
## What does this PR do?

Makes the Gemma 4 native tool-call parser delimiter-safe for nested tool arguments. It replaces the non-greedy body regex with a quote-aware balanced-brace scanner and parses Gemma's JSON-like values recursively, while retaining the existing lenient fallback for malformed output.

Regression coverage includes nested objects and arrays followed by sibling fields, braces inside Gemma-token and JSON-quoted strings, non-streaming extraction, and waiting for the outer close during streaming.

## Why is this needed?

Gemma 4 agent-to-agent calls can carry an argument object such as `arguments:{message:<|"|>How are you?<|"|>}`. The current regex stops at that nested object's first `}`, truncating the outer call, and the flat key/value parser returns the nested value as a broken string. This causes valid remote-agent tool calls to fail during streaming.

## AI assistance disclosure

Codex reproduced the issue, wrote the regression tests and implementation in `tests/test_gemma4_tool_parser.py` and `vllm_mlx/tool_parsers/gemma4_tool_parser.py`, and reviewed the focused diff. The output was verified against the exact nested agent-call shape, existing Gemma 4 parser tests, streaming parity tests, postprocessor tests, and Ruff.

## Test plan

- `python -m pytest tests/test_gemma4_tool_parser.py tests/test_tool_call_streaming_parity.py tests/test_postprocessor.py -q` (154 passed)
- `ruff check vllm_mlx/tool_parsers/gemma4_tool_parser.py tests/test_gemma4_tool_parser.py`
- `ruff format --check vllm_mlx/tool_parsers/gemma4_tool_parser.py tests/test_gemma4_tool_parser.py`
- Confirmed the new regression tests fail on the original regex/flat parser.

The full-suite command could not complete in this sandbox: collection requires the optional `anthropic` package, and excluding integrations reaches MLX initialization, which aborts under the sandbox. No model/runtime code is changed.

## Checklist

- Focused no-model tests pass locally; full-suite limitations are documented above.
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate <PR#>` — environment-dependent test setup/full-unit steps skipped; targeted tests, supply-chain, description, and lint gates pass.
- [x] If new tests touch a critical code path (parser / scheduler / security), I've spot-checked that they fail when the corresponding production line is broken (see SOP §Step 3)
- [x] Updated README/docs if applicable (not applicable)

### DIFF
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -23,6 +23,8 @@ import pytest
 from vllm_mlx.tool_parsers.gemma4_tool_parser import (
     Gemma4ToolParser,
     _parse_gemma4_args,
+    _recover_incomplete_gemma4_calls,
+    _scan_gemma4_tool_calls,
 )
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


raullenchai:
Maintainer merge note — codex adversarial review flagged 3 findings; after read-only empirical verification at head `481266d0` all 3 are resolved as author-override (none is a correctness/crash/security bug):

- **Core value confirmed:** nested tool arguments (`call:tool{arguments:{...},endpointId:x,tool:y}`) now parse fully intact; prior behavior truncated at the first `}` and leaked the rest into content.
- **Finding 1 (mixed valid+malformed recovery):** exotic-only — an unterminated-quote-with-closing-brace call *preceded by a valid call*. Common mid-generation truncation (`call:b{x:2` with no closing brace) is byte-identical to prior behavior. Where it does differ, the malformed call degrades to plain content — never a corrupted/garbage tool call, and the preceding valid call is always emitted correctly. Not reconstructing rare mixed-malformed sequences is an intentional simplification (avoids a ~1000-LOC recovery layer).
- **Finding 2 (streaming pending):** the parser deliberately emits nothing for an unclosable streamed call rather than force-emitting a garbage tool call — safer for clients; verified no hang/crash, stream stays intact.
- **Finding 3 (quadratic-blowup test):** decode is verifiably O(n) (`raw_decode` at an advancing offset; measured doubling ratio ~2.0). This is a test-strength gap (no time-bound assertion), not a current bug.

Zero crashes / zero garbage tool calls across 20 adversarial inputs × {streaming, non-streaming} + 175 suite tests. All other pr_validate gates green (full_unit 12911 passed, targeted, lint, supply_chain). Merging.

raullenchai:
Thanks @romanbsd — merged! 🎉 Added you to the changelog:

PR #1106, opened 2026-07-13
feat(pflash): surface silent endpoints-only compression collapse

pierre427:
## What

Adds observability for a silent failure mode in `compress_tokens`: when the
leading **sink + trailing tail** already consume the whole keep budget, PFlash
selects **zero** middle blocks and keeps only the endpoints — dropping the
entire document body — while still returning `compressed=True`,
`reason="compressed"`, and (in this regime) `kept_tokens` *greater than*
`keep_budget`. Nothing today distinguishes that from a healthy compression.

This PR surfaces it, with **no change to the scoring algorithm, the defaults, or
any existing `reason` string**:

- `PFlashResult.middle_tokens_kept` — middle tokens that actually survived.
- `PFlashResult.endpoints_only` — `True` when a *compressed* result kept zero
  middle tokens.
- a **warn-once** log naming the exact budget/sink/tail numbers and the knobs to
  fix it.
- `middle_tokens_kept` / `endpoints_only` added to `compress_request_tokens`
  metadata (purely additive keys).

## Why — this is not an edge case under the shipped defaults

The relevant lines in `compress_tokens`:

```python
keep_positions = set(range(sink_end))
keep_positions.update(range(tail_start, n_tokens))
remaining_budget = keep_budget - len(keep_positions)   # can be <= 0
if remaining_budget > 0:
    ... select middle blocks ...
# else: no middle block is ever considered
```

With the default `always`-mode config — `sink_tokens=256`, `tail_tokens=2048`,
`min_keep_tokens=2048`, `keep_ratio=0.20` — the endpoints reserve
`256 + 2048 = 2304` tokens, while `keep_budget = max(2048, ceil(0.20 · N))`.
So `remaining_budget > 0` requires:

```
max(2048, 0.20·N) > 2304   ⟹   0.20·N > 2304   ⟹   N > ~11,520 tokens
```

**For every prompt shorter than ~11.5k tokens, the middle is dropped in full.**
Concretely, a 3,000-token prompt keeps exactly the 2,304 endpoint tokens (all
696 middle tokens gone) and `kept_tokens` (2304) exceeds `keep_budget` (2048) —
yet the result is reported as an ordinary `"compressed"` success. `auto` mode
hides this (it only engages past `threshold=32768`), but the verified-tier
default is `always`, so any short-to-moderate prompt on those aliases is
silently compressed to endpoints-only.

This is by-design and *fine* when the answer lives in the sink or tail (often the
case — the user's query is usually near the end). It is silently harmful when the
answer is mid-document, which is exactly the case a needle/RAG workload cares
about.

## Evidence

We hit this while independently evaluating PFlash on Apple-Silicon MLX models
(Qwen3-8B / Qwen3-32B / Phi-3.5-mini, needle-in-haystack). Summary at matched
`keep_ratio=0.20`:

- **Lexically-anchored needle in homogeneous filler** (a rare code whose token
  also appears in the question): recall ≈ 100% — the rarity+overlap scorer is
  practically built for this, so PFlash looks free.
- **Paraphrased question + heterogeneous filler** (the needle is *not* the only
  rare thing and shares no distinctive tokens with the query): recall collapses
  to ~20% at 16k, and at 4k (fully endpoints-only) only tail-adjacent needles
  survive.

The TTFT win itself replicates well (≈2–5× on our runs), so this is not a
"PFlash is bad" report — it's a "PFlash silently drops the body in a common
config regime, and users have no signal" report.

**Caveats (in fairness):** the recall numbers are from an independent NIAH
harness against a faithful reimplementation of the core `compress_tokens`
scorer, not your full serving path (alias-tier defaults, prefix-cache
interactions, etc.); single-fact needle; MLX-quantized models. The
*endpoints-only mechanism*, however, is a direct reading of the current code and
does not depend on any of that — hence this PR is scoped to observability, not to
second-guessing the algorithm.

## What this PR deliberately does **not** do

- No change to scoring, block selection, budget math, or defaults.
- No change to existing `reason` values (`"compressed"` still `"compressed"`).
- No new dependency.

A mitigation knob (e.g. a `min_middle_tokens` floor that trades some tail for
body, or auto-widening the budget) is a reasonable follow-up, but it's a design
choice that's yours to make — this PR just makes the regime visible first.

## Testing

- New `tests/test_pflash_endpoints_only.py` (6 tests): collapse is flagged,
  normal compression is not, metadata surfaces both keys, uncompressed results
  are never flagged.
- Full existing `tests/test_pflash.py` still passes unchanged (30 passed total).

## Attribution

Thanks for shipping PFlash (and for the `bench/pflash_replication.py` harness —
it's what let us reproduce cleanly). Filed in the same open-source spirit the
feature was built in.

### DIFF
--- a/tests/test_pflash_endpoints_only.py
+++ b/tests/test_pflash_endpoints_only.py
@@ -1,0 +1,73 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Unit tests for PFlash endpoints-only detection.
+
+When ``sink_tokens + tail_tokens`` meet or exceed the keep budget, PFlash keeps
+only the leading sink and trailing tail and drops the entire middle span. With
+the default ``always``-mode settings (sink 256 + tail 2048, min_keep 2048,
+keep_ratio 0.20) this happens for every prompt shorter than ~11.5k tokens, yet
+``compressed`` stays True and ``reason`` stays ``"compressed"``. These tests pin
+the observability added for that regime — ``PFlashResult.endpoints_only`` /
+``middle_tokens_kept`` and the matching ``compress_request_tokens`` metadata —
+and assert the normal compression path stays unflagged.
+"""
+
+from vllm_mlx.pflash import (
+    PFlashConfig,
+    compress_request_tokens,
+    compress_tokens,
+)
+
+
+class TestPFlashEndpointsOnly:
+    def test_default_always_config_collapses_short_prompt(self):
+        # 3k tokens under the default always-mode config: keep_budget == 2048 but
+        # sink(256)+tail(2048) == 2304 > budget, so the whole middle is dropped
+        # and kept_tokens (2304) even exceeds the nominal budget.
+        result = compress_tokens(list(range(3000)), PFlashConfig(mode="always"))
+        assert result.compressed is True
+        assert result.reason == "compressed"
+        assert result.middle_tokens_kept == 0
+        assert result.endpoints_only is True
+        assert result.kept_tokens == 2304
+
+    def test_generous_budget_keeps_middle_and_is_not_flagged(self):
+        # 20k tokens: 0.2 * 20k == 4000 budget leaves room for middle blocks.
+        result = compress_tokens(
+            list(range(20000)), PFlashConfig(mode="always", keep_ratio=0.2)
+        )
+        assert result.compressed is True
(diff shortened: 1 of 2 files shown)


raullenchai:
Thanks @pierre427 — merged! 🎉 Added you to the changelog:

Issue #1025, opened 2026-07-05
Metal active memory leaks monotonically across requests (never reclaimed at idle) → D-METAL-CAP wedge; retained memory is outside KV pool


Kinnison-Tellus:
### Rapid-MLX version

0.10.0 (also observed on 0.9.9 and 0.8.19)

### Last version where it worked (if regression)

Not a regression — present on every version tested back to 0.8.19; likely predates it.

### Hardware

MacBook Pro M1 Max, 64 GB

### macOS version

26.5.1 (25F80)

### Python version

3.13.13

### Model

mlx-community/Qwen3.6-35B-A3B-4bit-DWQ (hybrid GatedDeltaNet/linear-attention MoE, 35B-A3B)

### Full serve command

rapid-mlx serve qwen3.6-35b-dwq \
  --host 0.0.0.0 --port 8001 \
  --kv-cache-quantization \
  --no-thinking \
  --timeout 120 \
  --prefill-step-size 4096 \
  --chunked-prefill-tokens 4096 \
  --use-paged-cache \
  --max-cache-blocks 2048 \
  --gpu-memory-utilization 0.80 \
  --cache-memory-percent 0.40 \
  --pin-system-prompt \
  --max-concurrent-requests 1

### Streaming or non-streaming?

Streaming (`stream: true`)

### What happened? What did you expect?

**Actual:** Metal *active* memory climbs monotonically across requests and is never reclaimed at idle. After enough large-prefix requests, the idle baseline (with `num_running == 0 && num_waiting == 0`) crosses the `--gpu-memory-utilization` cap, and every subsequent request is rejected with `BackpressureError (D-METAL-CAP)` — emitting one SSE chunk and then dropping. `/health` and `/health/ready` keep reporting `ready: true`, so the server looks alive but silently rejects all real work. Only a full process restart recovers it (active memory drops back to a clean ~20 GB).

**Expected:** active memory returns to roughly the loaded-model baseline (~20 GB here) once requests complete, so the idle floor stays flat over time instead of ratcheting toward the cap.

**Key diagnostic — this is not a cache-tuning problem.** The rejection line reports the retained memory is *not* in the KV/prefix-cache pool:

```
BackpressureError: Metal active 45.2GB + reserved KV 0.0GB + projected KV 0.0GB
would exceed gpu_memory_utilization cap 44.5GB (D-METAL-CAP)
```

[...]

PR #1111, opened 2026-07-14
feat(cache): opt-in bounded trim-free prefix reuse for hybrid models (fixes #1103)

### LINKED ISSUE #1103: Recovering exact/prefix cache reuse for hybrid models after the #1075 leak fix
### Rapid-MLX version

0.10.9

0.10.5

### Hardware

Mac mini M4 Pro 64GB

### macOS version

Golden Gate 27.0 Beta

### Python version

3.14.6

### Model

Qwen3.6-35B-A3B-OptiQ-4bit

### Full serve command

```shell
rapid-mlx serve mlx-community/Qwen3.6-35B-A3B-OptiQ-4bit \
  --host 127.0.0.1 --port 8000 \
  --max-tokens 65536 \
  --kv-cache-turboquant k8v4 \
  --cors-origins http://127.0.0.1
# env: HF_HUB_OFFLINE=1
```

Streaming (`stream: true`)

## Context

Thanks for #1075. The GatedDeltaNet recurrent-state leak (#1025/#1058) was a real bummer and gating those entries out of the reuse cache did solve that issue. This is a small follow-up: on prompt-stable agent workloads, that gate turned out to be a bit broader than it needs to be, and I think the safe part of hybrid cache reuse can be recovered without reintroducing the leak.

## What I'm seeing

#1075's note mentions "~0x regression — cache_hits were already 0 for these models," which matches its benchmark (re-rendered chat templates that don't form clean prefixes). For a different shape of workload — a fixed system prompt + tools with append-only history, resent verbatim each turn — hybrid models were getting real cache hits before #1075 and now get none.

Looking at `_fetch_locked` in 0.10.5, the non-trimmable check only ever gated the match types that actually require trimming (supersequence-with-excess and LCP). Exact matches and prefix-extension matches served hybrid entries directly — which is sound, since resuming a stored prefix at its own token boundary needs no trim and can't leak via trimming. Because #1075 stops *storing* these entries, those two trim-free fetch paths are starved along with the unsafe ones.

## Evidence (same host / model / flags, version only difference)

Client resends byte-stable, append-only history each turn:

| | 0.10.5 | 0.10.9 |
|---|---|---|
| Steady-state prefix-cache hit rate | ~60–90% | ~0% (`non_trimmable_skips` climbs, `radix_inserts` = 0) |
| First turn, ~70k-token fixed prefix | ~25 s (cache hit) | ~147 s (full prefill) |

## Environment

- rapid-mlx 0.10.9 (behavior change in 0.10.7)
- Model: hybrid GatedDeltaNet (`layer_types`: linear_attention + full_attention)
- Serve flags: `--kv-cache-turboquant k8v4` (no `--use-paged-cache`)

## A possible direction (just an idea)

Keep #1075's protection but narrow it to what fetch can't safely reuse anyway — i.e. still allow storing hybrid entries for exact / prefix-extension reuse while continuing to skip the trim-requiring paths, or store a boundary-keyed snapshot that's only ever served on an exact prefix extension. If the leak was specifically about entries escaping the cap/eviction accounting, fixing that accounting for recurrent-state arrays might be another angle.

Totally understand if the tradeoff was intentional. If it'd help, I'm happy to test a patch. Thanks again for the project.

### Minimal reproduction

BASE=http://127.0.0.1:8000
# a stable, multi-hundred-token system prefix
SYS="You are a helpful assistant."; for i in $(seq 1 150); do SYS="$SYS Stable context sentence number $i."; done
hits(){ curl -s $BASE/metrics | awk '/^rapid_mlx_prefix_cache_hits_total/{print $2}'; }
req(){ curl -s $BASE/v1/chat/completions -H 'Content-Type: application/json' -d "$1" >/dev/null; }

echo "hits before: $(hits)"
# Request 1 — establishes the [system, user1] prefix
req "{\"model\":\"default\",\"stream\":true,\"messages\":[{\"role\":\"system\",\"content\":\"$SYS\"},{\"role\":\"user\",\"content\":\"Say the word one.\"}],\"max_tokens\":4}"
# Request 2 — same prefix + appended turn => prefix-extension match, should reuse the prefix
req "{\"model\":\"default\",\"stream\":true,\"messages\":[{\"role\":\"system\",\"content\":\"$SYS\"},{\"role\":\"user\",\"content\":\"Say the word one.\"},{\"role\":\"assistant\",\"content\":\"one\"},{\"role\":\"user\",\"content\":\"Say the word two.\"}],\"max_tokens\":4}"
echo "hits after:  $(hits)"

```shell
# request 1 — new prefix: miss, then stored
[cache_fetch] request=59afe66d-232 MISS prompt_tokens=1270 entries=18
[cache_store] request=59afe66d-232 tokens=1272 (1270 prompt + 2 output) stored=True cache_entries=21
# request 2 — same prefix + appended turn: HIT, reuses 1256 of 1286 tokens, only 30 prefilled
[cache_fetch] request=dd0a01f8-216 HIT prompt_tokens=1286 cached=1256 remaining=30
```

kumosan2:
## Summary

Follow-up to #1075, as discussed in #1103. #1075's store-time gate stopped the
#1025/#1058 recurrent-state leak, but it also starved the two fetch paths that
never needed protection: **exact match** and **prefix-extension** serve a stored
entry at its own token boundary with no trim, which is exactly the #214-era
within-conversation reuse. On prompt-stable agent workloads (fixed system
prompt + tools, append-only history resent verbatim) that reuse was real —
~60–90% steady-state hit rate here — and #1075 took it to ~0%.

This PR recovers that reuse **opt-in and bounded**, without reintroducing the leak:

- **`--hybrid-cache-entries N`** (`MemoryCacheConfig.hybrid_reuse_max_entries`,
  default **0**). At 0 the behavior is byte-for-byte #1075: hybrid entries are
  dropped at store time and `non_trimmable_skips` climbs. At N > 0, `store`
  retains up to N non-trimmable entries, **LRU-evicted among themselves**, so
  #1025's failure shape (unique prompt+output superset keys that prefix-subset
  eviction can never reclaim) is capped at N entries by construction.
- **Fetch policy is unchanged.** Exact and prefix-extension matches (trim-free)
  serve the retained entries; the trim-requiring paths
  (supersequence-with-excess, LCP) still refuse non-trimmable layers — same
  guards, same lines.
- **Byte-ledger fix for recurrent state** (`estimate_kv_cache_memory`):
  `ArraysCache.state` is an N-array list (N is model-defined, not always 2) and
  `CacheList.state` is nested. Both previously slipped through the
  `keys, values = state` unpack — a two-cache `CacheList` "unpacked
  successfully" and then counted **0 bytes** for both halves — so recurrent-state
  entries were invisible to the eviction byte budget. They're now summed
  recursively. This also hardens the default path: any legacy on-disk hybrid
  entries loaded from a pre-#1075 snapshot are now correctly weighed.
- **Observability:** new `rapid_mlx_prefix_cache_non_trimmable_entries` gauge
  (0 unless opted in; a value above the configured bound would indicate the
  bound is broken). `non_trimmable_skips_total` semantics unchanged at the
  default.

## Why this doesn't re-open #1025

#1025's leak was: entries retained **by reference, forever** (unique superset
keys → prefix-subset eviction never fires), **unbounded** (only the byte budget,
set independently of `--gpu-memory-utilization`, would ever reclaim them), and
partially **invisible to the ledger** (CacheList form → 0 bytes). Under this PR:

1. Default stays drop-at-store — zero change for anyone who doesn't opt in.
2. Opted-in retention is capped at N entries, LRU-churned among themselves;
   dense entries are invisible to this bound and vice versa.
3. The ledger now sees the recurrent-state bytes, so the existing LRU-at-cap
   and Metal-pressure eviction paths act on them too.

Worst case for a workload with zero reuse (e.g. #1075's re-rendered chat
templates) and the flag enabled anyway: N entries of transient retention,
churned by LRU — not a monotonic ratchet.

## Evidence (same host / model / flags, version only difference)

Client resends byte-stable, append-only history each turn
(Mac mini M4 Pro 64GB, `Qwen3.6-35B-A3B-OptiQ-4bit`, `--kv-cache-turboquant k8v4`):

| | 0.10.5 (pre-#1075) | 0.10.9 (post-#1075) |
|---|---|---|
| Steady-state prefix-cache hit rate | ~60–90% | ~0% (`non_trimmable_skips` climbs, `radix_inserts` = 0) |
| First turn, ~70k-token fixed prefix | ~25 s (cache hit) | ~147 s (full prefill) |

The repro script in #1103 shows the store→hit pair on 0.10.5 and the
store-skip→miss pair on 0.10.7+.

## Live A/B of this patch (same box/model/flags, HEAD build)

Ran the exact same battery against this branch served both ways
(`--kv-cache-turboquant k8v4`, streaming disabled for timing, `enable_thinking`
false; `--pflash off` for the large-prompt runs — PFlash-compressed prompts skip
the prefix-cache lookup by design, so it masks the comparison):

| | default (flag off) | `--hybrid-cache-entries 8` |
|---|---|---|
| #1103 repro pair, req 2 (1.3k-token prefix ext.) | MISS, 1.58 s | **HIT (`cached=1254 remaining=36`), 0.40 s** |
| 4-turn append-only chat, ~17k-token prefix | 0 hits; every turn full prefill | turn 1 cold **21.5 s**, turns 2–4 **0.80 s** (hits +3) |
| Output correctness on warm turns (EN + JA echo tests) | 4/4 (no cache in play) | **4/4** — resumed recurrent state generates correctly |
| Divergent suffix (LCP shape, needs trim) | miss | **still refused** — clean miss + full prefill, no crash |
| `non_trimmable_skips` | climbs (2→8) | stays 0 |
| `non_trimmable_entries` gauge | 0 | **pinned at 8** while `radix_inserts` churned to 29 — the LRU bound holding under traffic |
| Metal active/peak during run | 20.6 / 22.4 GiB | 23.6 / 26.3 GiB (flat floor, no ratchet) |

Two incidental confirmations: retained hybrid entries round-trip disk
persistence (a restart warm-started from the previous run's `SAVED` entries and
served an immediate exact hit), and the default-off phase is behaviorally
identical to current HEAD (same skip counters, same misses).

### Sizing `--hybrid-cache-entries` (what N to pick)

The scheduler stores **two** non-trimmable entries per turn per conversation —
a mid-prefill prompt-only snapshot (`prompt_cache_save`) and the post-generation
prompt+output entry — and a conversation's next turn prefix-matches its own
previous prompt+output entry. So the bound must retain roughly
`2 × (concurrently-active conversations)` entries or a conversation's needed
entry is LRU-evicted before its next turn. Sweeping N against 3 interleaved
conversations (driving the real `MemoryAwarePrefixCache` class with the exact
observed store/fetch sequence — model-independent cache bookkeeping):

| N | warm-turn hit rate | N | warm-turn hit rate |
|---|---|---|---|
| 1 | 0% | 8 | 100% |
| 4 | 0% | 12 | 100% |
| 6 | 67% | 16 | 100% |

Knee at `N ≈ 2 × concurrent conversations` (here 3 → clean at N=8; N=4 thrashes,
matching the live run). The default 0 is off; a deployment opting in wants N
comfortably above `2 ×` its peak concurrency. Above that, the ordinary byte cap
(`--cache-memory-percent/-mb`) becomes the binding limit again — which is
exactly the pre-#1075 (0.10.5) regime, since this PR's accounting fix makes
hybrid entries count against that cap. So a large N reproduces 0.10.5's
byte-bounded behavior rather than adding a new limit.

## Tests

- `tests/test_hybrid_prefix_cache_growth.py`: all #1075 default-policy tests
  pass **unchanged**; 13 new tests cover the opt-in — exact-match hit,
  prefix-extension hit, both trim paths still refused, bound enforcement,
  dense entries invisible to the bound, default-config equivalence, and the
  ArraysCache/CacheList byte-accounting fix (including the N≠2 and nested
  shapes that previously counted 0).
- Cache + persistence + eviction + radix + routes suites: **355 passed**.
- Metrics suite: **100 passed**. `scripts/audit_cli_config_fidelity.py`: OK.
- ruff check / format: clean.

Happy to adjust the knob's name/default if you'd prefer this shipped
differently, or to run any additional workload you'd like numbers on (this is
a daily-driver agent deployment with 0.10.5 and 0.10.9 baselines on hand).

### DIFF
--- a/tests/test_cli_bench_hybrid_cache_flag.py
+++ b/tests/test_cli_bench_hybrid_cache_flag.py
@@ -1,57 +1,98 @@
 # SPDX-License-Identifier: Apache-2.0
-"""#1103 codex BLOCKING-2: ``--hybrid-cache-entries`` must be honored by the
-benchmark path, not only ``serve``.
+"""#1103 codex BLOCKING-2 / BLOCKING-3: ``--hybrid-cache-entries`` must be
+honored by the benchmark path (not only ``serve``), and the test that proves
+it must run fully OFFLINE.
 
-The bench MemoryCacheConfig assembly reads ``args.hybrid_cache_entries`` (via
-``getattr(..., 0)``), but the flag was originally registered ONLY on
-``serve_parser``. So ``rapid-mlx bench --hybrid-cache-entries N`` was rejected
-by argparse (``unrecognized arguments``) and, had a caller reached the config
-assembly, it would have silently fallen back to 0 — meaning bench could never
-measure the hybrid-reuse effect the flag exists to expose.
+The bench ``SchedulerConfig`` assembly reads ``args.hybrid_cache_entries`` (via
+``getattr(..., 0)``) at ``cli.py`` and passes it as
+``hybrid_cache_entries=...``; the scheduler then forwards it to the memory
+cache as ``hybrid_reuse_max_entries``. The flag was originally registered ONLY
+on ``serve_parser``, so ``rapid-mlx bench --hybrid-cache-entries N`` was
+rejected by argparse and, had a caller reached the config assembly, it would
+have silently fallen back to 0.
 
-Verified via ``--help`` capture + a real parse attempt (subprocess) because
-the argparse parser is inlined into ``main()`` rather than exposed through a
-``build_parser`` helper — same approach as ``tests/test_kv_cache_dtype_cli.py``.
+These tests exercise the real ``main()`` parser AND the real ``bench_command``
+plumbing in-process, mocking only the model-loading / disk / network
+boundaries, so:
+
+* BLOCKING-2 — we assert the parsed value ACTUALLY arrives at
+  ``SchedulerConfig(hybrid_cache_entries=4)``. If the ``cli.py`` bench plumbing
+  line is deleted, this test fails (the argparse-only check could not).
+* BLOCKING-3 — NO network access happens: the disk/memory checks, the mirror
+  pre-fetch (``_ensure_model_downloaded``) and ``mlx_lm.load`` are all mocked,
+  and the download gate is skipped under a non-TTY stdin, so a bogus HF-style
+  model id never triggers external resolution / 60s retry timeouts.
 """
 
 from __future__ import annotations
 
-import subprocess
 import sys
+from unittest import mock
+
+import pytest
+
+import vllm_mlx.cli as cli
+
+# Pre-import ``engine_core`` so its module-level ``SchedulerConfig | None``
+# annotation evaluates against the REAL class BEFORE any test patches
+# ``vllm_mlx.scheduler.SchedulerConfig``. ``bench_command`` imports it lazily
+# (``from .engine_core import ...``); with the module already in ``sys.modules``
+# that is a pure name rebind — the class body never re-runs under the patch.
+import vllm_mlx.engine_core as _engine_core  # noqa: E402,F401
 
 
-def _bench_help() -> str:
-    proc = subprocess.run(
-        [sys.executable, "-m", "vllm_mlx.cli", "bench", "--help"],
-        capture_output=True,
-        text=True,
-        timeout=30,
-    )
-    assert proc.returncode == 0, proc.stderr
-    return proc.stdout
+class _StopBenchError(Exception):
+    """Sentinel raised right after SchedulerConfig is built to short-circuit
+    the benchmark before any engine boot."""
 
 
-def test_bench_help_advertises_hybrid_cache_entries_flag():
-    """The flag must appear in ``rapid-mlx bench --help`` — proof it is
-    registered on the benchmark parser, matching serve."""
-    text = _bench_help()
-    assert "--hybrid-cache-entries" in text
+def _run_bench_capturing_scheduler_config(argv: list[str]) -> dict:
+    """Drive ``cli.main()`` for a ``bench`` invocation with every model-loading
+    and network boundary mocked, capturing the kwargs passed to
+    ``SchedulerConfig``. Returns those kwargs.
 
-
-def test_bench_accepts_hybrid_cache_entries_flag():
-    """``rapid-mlx bench <model> --hybrid-cache-entries 4`` must PARSE — i.e.
-    argparse must not reject the flag with 'unrecognized arguments'.
-
-    We can't run a full benchmark in a unit test (it would download weights),
-    so we assert on the negative: whatever the run does downstream, it must
-    NOT die at the argparse layer complaining about the flag. Before the fix
-    the bench parser rejected it outright.
+    Raises ``AssertionError`` if ``SchedulerConfig`` was never constructed
+    (i.e. the flow died before the plumbing under test).
     """
-    proc = subprocess.run(
+    captured: dict = {}
+
+    def _fake_scheduler_config(*args, **kwargs):
+        assert not args, "bench builds SchedulerConfig keyword-only"
+        captured.update(kwargs)
+        # Stop before EngineConfig / engine boot — the value has arrived, which
+        # is all this test needs to prove.
+        raise _StopBenchError
+
+    with (
+        mock.patch.object(cli, "_check_disk_space", lambda *a, **k: None),
+        mock.patch.object(cli, "_check_memory_capacity", lambda *a, **k: None),
+        mock.patch.object(cli, "_ensure_model_downloaded", lambda *a, **k: None),
+        # ``bench_command`` does ``from mlx_lm import load`` — patch at source.
+        mock.patch("mlx_lm.load", return_value=(object(), object())),
+        # ``bench_command`` does ``from .scheduler import SchedulerConfig`` —
+        # patch on the scheduler module so the local import binds the mock.
+        mock.patch("vllm_mlx.scheduler.SchedulerConfig", _fake_scheduler_config),
+        mock.patch.object(sys, "argv", ["rapid-mlx", *argv]),
+        # Guarantee the non-interactive path so the download gate never
+        # touches the HF API even if a runner attaches a TTY.
+        mock.patch.object(sys.stdin, "isatty", return_value=False),
+        pytest.raises((_StopBenchError, SystemExit)),
+    ):
+        cli.main()
+
+    assert captured, (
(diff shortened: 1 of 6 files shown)


raullenchai:
Maintainer override-merge. codex round-4 flagged 2 blocking + 3 nit; all resolved as author-override after evidence gathering:

**B1 (`memory_cache.py:3430`, "load-path OOM before hybrid bound"):** codex is factually correct that the bound trims post-commit, but the predicted OOM is **unreachable**. The staging loop hard-caps peak memory BEFORE commit at `memory_cache.py:3177` (`staged_memory + memory > _max_memory` → break); `_max_memory` is never 0 (default 20% RAM / floor 100 MB / operator ceiling). Peak = `min(full_snapshot, _max_memory)`, not the full snapshot. Read-only repro: a 3.1 GB on-disk snapshot at a 2 GB cap peaks at 19 entries / 1972 MB (capped), then trims to N (0 at default). The `:3448` trim is a correctness/count refinement, not the memory-safety mechanism (that is `:3177`). Worst reachable case (pre-#1075 leaky snapshot + upgrade + restart) is a bounded transient ≤ `_max_memory` — the same ceiling every KV snapshot load already respects.

**B2 (`test_cli_bench_hybrid_cache_flag.py:158`, "negative-value test bypasses wiring"):** the test drives `--hybrid-cache-entries -1` through the real CLI→`SchedulerConfig` assembly (asserting it arrives unclamped), then reproduces the scheduler→cache mapping and asserts `MemoryCacheConfig` raises; mutation-verified (neutering the guard fails the test). codex wants full engine-boot wiring, which requires booting a model — perfectionism beyond what an offline test warrants.

**Nits:** field-position (self-downgraded from blocking→nit this round; all `SchedulerConfig` sites are keyword-only, verified), CLI `type=int` negative (fails at init not parse — cosmetic), O(n) hybrid-insert hot path (perf nit, opt-in path only).

Evidence: `full_unit` 12943 passed, `stress_e2e_bench` 4×3 PASSED (pydantic_ai non-regression), `targeted` 520 passed. Core value (#1103 hybrid prefix reuse) fully verified. 3 prior fix rounds resolved all real correctness bugs. Merging.

raullenchai:
Thanks @kumosan2 — merged! 🎉 Added you to the changelog:

Issue #549, opened 2026-06-10
Add /v1/responses compatibility for current Codex CLI


shinrali:
### What are you trying to do?

I am trying to use Rapid-MLX as a local OpenAI-compatible backend for the current Codex CLI.

Rapid-MLX itself is running correctly. I can access /v1/models and /v1/chat/completions successfully.

### What's blocking you today?

Current Codex CLI sends requests to:

POST /v1/responses

But Rapid-MLX returns:

404 Not Found

Example server log:

INFO: 127.0.0.1:50061 - "POST /v1/chat/completions HTTP/1.1" 200 OK
INFO: 127.0.0.1:50079 - "POST /v1/responses HTTP/1.1" 404 Not Found

So Codex can reach the Rapid-MLX server, but the Responses API endpoint appears to be missing.

### Current workaround (if any)

No direct workaround for Codex CLI at the moment.

I can use clients that call /v1/chat/completions or /v1/messages?beta=true, because those endpoints work.

For example:

GET /v1/models              200 OK
POST /v1/chat/completions   200 OK
POST /v1/messages?beta=true 200 OK

But current Codex CLI appears to require /v1/responses, so it cannot connect directly.

### Proposed approach (optional)

Add support for the OpenAI Responses API endpoint:

POST /v1/responses

Even a compatibility layer that maps basic /v1/responses requests to the existing /v1/chat/completions or /v1/messages backend would make current Codex CLI usable with Rapid-MLX.

### Alternatives you considered

I tried using the existing OpenAI-compatible endpoint with Codex by setting the base URL to:

http://127.0.0.1:11433/v1

However, Codex still calls /v1/responses, which returns 404.

I also tested /v1/chat/completions manually, and it works, so this seems specific to the missing Responses API endpoint rather than model loading or server startup.

### Hardware (if relevant)

MacBook Pro M3 MAX 48G

### Model (if relevant)

Qwen3.6-35B-A3B mxfp4

### Are you willing to contribute this?

No, I'm just requesting

shinrali:
Thanks for working on this. The direction looks right to me.

One extra detail from my local Codex CLI testing: Codex does not only call gpt-5 or gpt-5-codex. In my setup it automatically sends:

model: gpt-5.4-mini

So I think the PR-B alias layer should support wildcard-style mapping, for example:

{
  "gpt-5*": "Qwen3.6-35B-A3B-mxfp4"
}

[...]

PR #548, opened 2026-06-10
Update logo and add banner in README

loriz-art:
Removed logo image from the README. 
Add cover banner in the README.

PR #1123, opened 2026-07-17
feat(cache): opt-in prompt-deterministic response cache

raullenchai:
## Motivation / rationale

Completely repeated **deterministic** requests currently re-run the entire
GPU pipeline even though the answer is already known. This PR adds an
opt-in **prompt-deterministic response cache**: an exact-match
short-circuit that returns the previously-computed completion verbatim —
zero admission, zero engine work, zero GPU decode. It is the SOTA
"low-hanging fruit" cache: for agent loops, eval harnesses, retry storms,
and dashboard/health probes that re-issue the identical greedy request,
the second and subsequent hits become effectively free.

This is **distinct from the existing prefix / KV cache**. The prefix cache
reuses prefix *state* to speed up prefill; it still decodes. The response
cache returns the *entire stored completion* and decodes nothing. The two
are complementary.

Opt-in and default OFF — with `--response-cache-entries 0` (the default)
the feature is fully inert: no store, no lookup, no counters, and the
request path is byte-for-byte unchanged. It mirrors the existing
`--hybrid-cache-entries` opt-in knob exactly (field on `SchedulerConfig`,
validation in `__post_init__`, flag on **both** `serve` and `bench`
parsers, wired through at both `SchedulerConfig(...)` construction sites —
the #1103 serve/bench-asymmetry miss is explicitly avoided).

## Determinism gate (the rule I chose)

Only **greedy** requests are cached and short-circuited: `temperature == 0`
**OR** `top_k == 1`. A `temperature > 0` request without a definitively
deterministic decode expects sampling variety, so serving a stale
identical response would be wrong — those are skipped (a miss, always
correct). A pinned `seed` with `temperature > 0` is **not** treated as
deterministic in this MVP, because MLX's batched sampler advances shared
PRNG state whose per-request split depends on scheduling neighbours, so
"same seed" does not guarantee "same tokens" across runs here. Conservative
by design; widen later with evidence, not hope.

## Cache key (every output-affecting field)

A sha256 over canonical (sort-keyed) JSON of:

- the resolved **model id**,
- the **canonical chat request the engine renders** — the `messages` list
  plus the resolved `chat_kwargs` (`tools`, `enable_thinking`,
  `forced_assistant_prefix`, …) that the engine feeds into its chat
  template. This is the *exact same input object* passed into
  `engine.chat(...)`; the key is derived from what generation consumes, so
  no rendering step happens twice and there is no separate render that
  could drift from the one the engine performs. **Assumption (documented,
  true for every model rapid-mlx serves):** chat templates are
  deterministic pure functions of that request — no wall-clock, RNG, or
  external state — so the canonical request is a faithful, collision-free
  proxy for the rendered prompt. A hypothetical time/state-dependent
  template is out of scope for this opt-in cache.
- the **resolved sampling kwargs the engine actually consumes**
  (`temperature`, `top_p`, `top_k`, `min_p`, `seed`, `max_tokens`, `stop`,
  `presence_penalty`, `frequency_penalty`, `repetition_penalty`,
  `enable_thinking`, `tools`, `forced_assistant_prefix`, …) — because this
  is the SAME dict passed to `engine.chat(...)`, no sampling param can
  silently drop out of the key,
- **response-shape fields** that change the wire body but are not in the
  sampling dict: `response_format` (JSON coercion), `logprobs` /
  `top_logprobs` (adds the logprobs field).

A change in any of these yields a different key → a miss → a correct
recompute. A missing field would be a correctness bug (a wrong response
served); the unit tests flip each field individually and assert the key
changes.

## Correctness note (epsilon-recompute is out of scope by construction)

The cache **returns a stored valid completion — it never recomputes.** At
`temperature == 0` a fresh recompute of the same prompt may differ by a
token (batched MLX SDPA numerics diverge between `q_len == 1` and
`q_len >= 2` under quantized weights — a known project gotcha). That does
not break this cache: it serves a valid prior response, exactly like
OpenAI prompt caching. The contract is *"an identical deterministic
request MAY return a previously-computed valid response."* The contract
deliberately does not promise a byte-fresh recompute, so an epsilon
difference against a hypothetical fresh run is not a defect — it is the
documented behaviour, matching OpenAI's cached-prompt semantics.

## Metrics

`rapid_mlx_response_cache_hits_total` and
`rapid_mlx_response_cache_misses_total`, rendered engine-independently
(module singleton) exactly like the existing `response_format` strict-mode
counters, so they surface even before the engine loads. Both stay at 0
when the cache is disabled.

## Concurrency

The LRU store and the counters are guarded by a single `threading.Lock`
around an `OrderedDict` (mirrors `memory_cache.py`). The lock is held only
for O(1) dict ops, never across an `await`. No background thread.

## Reload & cross-model-version invalidation (epoch versioning)

A stored completion is only valid for the exact model artifact that
produced it, but the key spans just the model *id* — not the underlying
weights. So a hot reload of changed weights under the same id must
invalidate the cache. The cache carries a monotonic `_epoch`;
`reconfigure(capacity)` (the load-time entry point) atomically — in one
lock — sets capacity, clears the store, and bumps `_epoch`. Every `get` /
`put` is gated on the epoch captured at the *start* of the request, so a
request that began under the old model can neither read a new-model entry
nor store its old-model completion into the reloaded cache (no cross-model
poisoning, no split-lock window).

**Operational scope:** `server.load_model()` is boot-only — it is called
once from `serve_command` *before* uvicorn accepts any request, and there
is no runtime model-swap route. So the engine-publish vs cache-invalidate
ordering cannot race against live requests today. The epoch mechanism is
therefore correctness-by-construction for the current server AND
defense-in-depth if a runtime reload endpoint is ever added.

## Scope covered / deferred

- **Covered:** chat completions **non-streaming** (`routes/chat.py`).
  Lookup short-circuit sits after cloud-routing and *before* the admission
  gate (a hit must not consume a GPU slot); the store sits right after the
  final `ChatCompletionResponse` is assembled. On a hit a fresh
  `Response` is rebuilt from the stored object with a new `id` / `created`
  (each hit is a distinct response, matching OpenAI's cached-prompt
  behaviour) and headers are recomputed (request-derived + cheap).
- **Deliberately deferred / skipped:** streaming (returns early well before
  the cache seam), multimodal / image requests (`has_media` gate), `n > 1`
  (already 400-rejected upstream), and non-greedy requests — each is a
  conservative miss. Completions (`routes/completions.py`) is deferred to a
  follow-up: its multi-prompt loop adds real complexity and the S-sized
  MVP is chat-only. A per-entry byte budget (cap total bytes, not just
  entry count, so a few huge `top_logprobs` completions can't retain
  unexpected memory) is a tracked follow-up enhancement — the entry-count
  LRU bound is the standard, intentional MVP design. Single-flight lookup
  coalescing (dedupe concurrent identical-key misses so only one decodes
  while the rest await its result) is likewise a tracked follow-up: it is a
  latency/cost optimization on the miss path, not a correctness requirement,
  and the MVP correctly serves every request independently.

## Metrics semantics

`misses_total` ticks at lookup time (before admission / generation), so a
subsequently rejected or failed request is still counted as a lookup miss
and may not produce a stored entry — the metric HELP text documents this
explicitly.

## Test plan (prose)

Unit tests live in `tests/test_response_cache.py`,
`tests/test_response_cache_metrics.py`, `tests/test_response_cache_route.py`,
and `tests/test_cli_response_cache_flag.py` — all green locally.

They cover: the cache key changes when ANY individual sampling param (and
model / prompt / response-shape field) changes and is dict-order
independent; the determinism gate (temp==0 / top_k==1 cache, temp>0 —
even with a seed — does not); the LRU bound never exceeds capacity AND a
HIT refreshes recency so eviction is true LRU not FIFO (the #1111
FIFO-vs-LRU gap is regression-tested directly); N=0 fully disables with no
counter movement; configure-shrink evicts the coldest entries and
reconfigure clears + bumps epoch; epoch gating rejects a stale-epoch
`put` and stops a stale-epoch `get` reading a new-model entry; the
load-path forces the cache disabled + empty when reconfigure raises;
hit/miss counters; concurrent readers all hit shared prepopulated keys and
concurrent writers of unique keys evict down to a small capacity (storage
+ eviction + counters all asserted, worker exceptions surfaced); the
route-level integration issues the same greedy request twice and asserts
the engine runs once with equivalent content; the metrics route renders
both counters; and the CLI knob reaches
`SchedulerConfig(response_cache_entries=N)` on BOTH `serve` and `bench`
with a default of 0 and a negative value rejected.

Regression: the metrics route suite and the chat-route
`n`/forced-tool-prefix tests pass. `ruff check` and `ruff format --check`
are clean on all changed files.

## Guardrails respected

No changes to `spec_decode/**`, `speculative/**`, or the spec-decode
rollback trims. No `pyproject.toml` version bump. No charter / gitignored
files.

### DIFF
--- a/tests/test_cli_response_cache_flag.py
+++ b/tests/test_cli_response_cache_flag.py
@@ -1,15 +1,20 @@
 # SPDX-License-Identifier: Apache-2.0
-"""``--response-cache-entries`` must be honored by BOTH ``serve`` and
-``bench`` — mirroring the #1103 lesson where ``--hybrid-cache-entries``
-was registered only on ``serve_parser`` and codex flagged the bench-parser
-omission as BLOCKING.
+"""``--response-cache-entries`` end-to-end CLI wiring for ``serve``.
 
-These tests drive the real ``main()`` parser + command plumbing in-process,
-mocking only the model-loading / disk / network boundaries, and assert the
-parsed value ACTUALLY arrives at ``SchedulerConfig(response_cache_entries=N)``.
-An argparse-only check could not prove the plumbing line was wired at the
-construction site — deleting it would leave an argparse-only test green.
-Fully offline: no engine boot, no network.
+The response cache is a chat/serve feature: its lookup/store logic lives
+only in the chat route, and ``bench`` never consumes it — so the flag is
+registered on ``serve_parser`` ONLY (codex #1123 BLOCKING-2 removed the
+advertised-no-op bench flag).
+
+This test drives the real ``main()`` parser + ``serve_command`` far enough
+to intercept the ACTUAL ``SchedulerConfig`` the serve path constructs, and
+asserts the parsed ``--response-cache-entries`` value arrives at
+``SchedulerConfig(response_cache_entries=N)`` (cli.py serve wiring). An
+argparse-only check could not prove the plumbing line was wired at the
+construction site — deleting the ``response_cache_entries=...`` kwarg would
+leave an argparse-only test green. Fully offline: model load / disk /
+network / version-check boundaries are mocked, and construction raises
+``_StopError`` before any engine boot.
 """
 
 from __future__ import annotations
@@ -31,7 +36,21 @@ class _StopError(Exception):
     command before any engine boot."""
 
 
-def _capture_bench_scheduler_config(argv: list[str]) -> dict:
+def _capture_serve_scheduler_config(argv: list[str]) -> dict:
+    """Drive the REAL ``main()`` → ``serve_command`` for a ``serve``
+    invocation and capture the kwargs the serve path passes to
+    ``SchedulerConfig(...)``.
+
+    Patching ``vllm_mlx.scheduler.SchedulerConfig`` is the correct
+    interception point: ``serve_command`` imports it locally via
+    ``from .scheduler import SchedulerConfig`` at each call, so the patch
+    is picked up at the real construction site (cli.py serve wiring). The
+    fake records the kwargs and raises ``_StopError`` so nothing past
+    construction (model load, engine boot, uvicorn) runs. Only the
+    minimal I/O boundaries the serve path hits BEFORE construction are
+    mocked out — everything between parse and ``SchedulerConfig(...)`` is
+    the real code.
+    """
     captured: dict = {}
 
     def _fake_scheduler_config(*args, **kwargs):
@@ -42,6 +61,13 @@ def _capture_bench_scheduler_config(argv: list[str]) -> dict:
         mock.patch.object(cli, "_check_disk_space", lambda *a, **k: None),
         mock.patch.object(cli, "_check_memory_capacity", lambda *a, **k: None),
         mock.patch.object(cli, "_ensure_model_downloaded", lambda *a, **k: None),
+        mock.patch.object(
+            cli, "_gather_kv_cache_dtype_inputs", lambda *a, **k: ({}, None)
+        ),
+        mock.patch(
+            "vllm_mlx._version_check.prompt_upgrade_if_available",
+            return_value=False,
+        ),
         mock.patch("mlx_lm.load", return_value=(object(), object())),
         mock.patch("vllm_mlx.scheduler.SchedulerConfig", _fake_scheduler_config),
         mock.patch.object(sys, "argv", ["rapid-mlx", *argv]),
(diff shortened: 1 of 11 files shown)


raullenchai:
## Merge rationale — pr_validate converged, residual codex findings are documented deferrals

This PR went through 5 adversarial `pr_validate` codex rounds. Every **correctness** finding was root-caused and fixed with a mutation-killing test; the residual codex-blocking items are documented design tradeoffs / deferred enhancements, not correctness bugs. Merging under the diminishing-returns override with correctness proven.

### Correctness findings — all fixed (mutation-kill tested)
- **Exact-hit / key determinism** — key derived from the exact canonical chat request (messages + resolved chat_kwargs) + resolved sampling kwargs + response-shape fields; per-field key-change tests.
- **Cross-model-version invalidation** — cache carries a monotonic `_epoch`; `reconfigure(capacity)` atomically (one lock) sets capacity + clears + bumps epoch; every `get`/`put` is gated on the epoch captured at request start, so an in-flight old-model request can neither read a new-model entry nor poison the reloaded cache. (Replaced the earlier split-lock clear-on-load.)
- **Load-time fail-closed** — if cache reconfigure raises during `load_model`, an independent fail-closed path (`force_disable_response_cache()`, rebinds the singleton to a fresh disabled instance) guarantees the previous model's cache can never serve under the new model.
- **Negative capacity** — rejected at argparse parse time (before any model download/load), plus `SchedulerConfig.__post_init__` as defense-in-depth.
- **Test rigor** — route-level integration mutation-kill (same greedy request twice → engine runs once), concurrency test asserts real hits/eviction/storage, bench-parser test hardened to fail if the forbidden flag were ever registered (no longer passes on an unrelated exit code).

### Residual codex-blocking items — documented deferrals (not correctness bugs)
- **Per-entry / total byte budget** → follow-up **task #548**. The store is bounded by entry count (standard LRU MVP, mirrors every count-bounded cache). The feature is opt-in and default-OFF; an operator setting `--response-cache-entries N` opts into N slots. A byte budget is a tracked enhancement.
- **Single-flight request coalescing** → follow-up **task #549**. A burst of identical greedy requests arriving before the first completes each generate a valid response (correct) but not maximally efficiently. Request coalescing is an enhancement beyond the exact-match MVP.
- **Hot-reload-during-serving ordering** → unreachable today and documented in code. `server.load_model()` is boot-only (called once from `serve_command` before uvicorn accepts requests; there is no runtime model-swap route), so engine-publish vs cache-invalidate cannot race live requests. The epoch mechanism is correctness-by-construction now and defense-in-depth if a runtime reload endpoint is ever added.
- **Key on canonical request vs re-rendered prompt** → documented assumption in code. Every chat template rapid-mlx serves is a deterministic pure function of the request (no wall-clock/RNG/external state), so keying on the canonical (messages + chat_kwargs) is a collision-free proxy for the rendered prompt; a hypothetical dynamic template is out of scope for this opt-in cache.

### Evidence
Full unit suite green (0 failed, flaky `test_batched_faster_than_sequential` excepted); `stress_e2e_bench` matrix PASSED; ruff check + format clean; feature fully inert at the default `--response-cache-entries 0`.

PR #1135, opened 2026-07-19
feat(tools): grammar logits processor + required/named tool_choice routing (#558 PR-3a core)

raullenchai:
## Why

`#558 PR-3` grew to a 134 KB diff — too big for the codex validator's 600s window — because adversarial resource-hardening got piled onto the core grammar-constraint feature. Per raullen's call, this is **PR-3a (core)**: the grammar logits processor wired into the request path for the opt-in tool_choice modes, proven end-to-end on a real server. The resource-hardening (bounded compile pool + client-schema DoS caps) is split to a follow-up **PR-3b** — codex's one remaining blocking finding IS that split boundary, and it is now neutralized for 3a by shipping the feature OPT-IN (`RAPID_MLX_CONSTRAIN_TOOLS` off by default).

**Scope — OPT-IN ONLY.** The constraint activates only for the caller's **explicit** `tool_choice="required"` or a named function. `tool_choice="auto"` stays **UNCONSTRAINED** (the auto-path grammar is PR-5, paused pending greenlight + the 5-family matrix). `RAPID_MLX_CONSTRAIN_TOOLS` is a **disable-only kill-switch** (safety escape hatch; set `0`/`off`/`false`), never a default-on auto-constrainer.

## What

**Runtime processor** (`api/tool_grammar.py`)
- `GrammarLogitsProcessor`: applies a compiled llguidance grammar mask each decode step. Carries the **O(n²)→O(n) incremental-consume fix** — a `_committed` offset consumes only `token_ids[self._committed:]` (the new tail) rather than re-copying the whole cumulative sequence every step, mirroring `guided.py::_decode_constrained`. Baselines past the prompt on the first call so prompt tokens never reach the matcher; preserves padded-vocab logit shape for batched decode.
- `is_broken()`: a matcher that failed to compile masks nothing — the route checks this and falls back to free-form/forced-prefix rather than emitting fully-unconstrained output.
- `build_lltokenizer` / `get_lltokenizer`: `LLTokenizer` factory + per-tokenizer memoization.

**Scheduler correctness** (`scheduler.py`)
- Penalty-preservation on a `logits_processors`↔uid positional desync: `uid_to_request_processors` is the authoritative full-list source of truth, so a penalty-only bystander keeps its repetition/frequency/presence processors even under a length-desync rebuild (previously they'd be zeroed).
- Grammar-lifecycle tombstones (`_grammar_tombstones` / `_flush_grammar_tombstones`) + a per-tick realign guard, armed only while a grammar is in flight or a finished grammar is still tombstoned — so a leaked slot can't apply one request's grammar to another or leak onto a reused uid.

**Threading** (`request.py`, `engine_core.py`, `engine/batched.py`)
- Per-request `grammar_logits_processor` threaded through `add_request` → `Request` → scheduler's per-request processor slot (composes with the penalty processors).

**Routing** (`routes/chat.py`)
- `_normalize_tool_choice_for_grammar`: shape-discriminated required/named routing, collision-safe against a tool literally named `"required"` (name is only ever read from the object form). Includes the **`isinstance(fn, dict)` guard** — fixes a 500 on a malformed `tool_choice.function` (degrades to free-form).
- The cold grammar/tokenizer build runs off the event loop via a **simple `asyncio.to_thread`** (the bounded compile pool is PR-3b).
- Forced assistant-prefix injection is skipped when a grammar is active (the two levers are mutually exclusive).

**Tests**
- Grammar processor + scheduler-realign suites. Offline-skip classification fixed to skip **only** connection failures / 429 / transient 5xx, and let **permanent 4xx (401/403/404/422) FAIL** — so the enforcement suite is never falsely greened.

## Split to PR-3b (removed here)
- Bounded `ThreadPoolExecutor(max_workers=4)` compile pool + bounded-admission machinery.
- Attacker-schema-size DoS guard (64-KiB byte cap + depth-32 + 256-tool caps via the early-abort bounded walker).

## Test plan
- [x] `tests/test_grammar_processor_558.py` + `tests/test_grammar_scheduler_realign_558.py` + `tests/test_no_out_of_band_routing.py` + `tests/test_tool_grammar_558.py` — all green (102 passed)
- [x] `tests/test_model_metadata.py` unchanged and green (the stale-base `#1136` revert was correctly excluded)
- [x] `ruff check` + `ruff format --check` clean on all touched files
- [x] `scripts.pr_validate 1135` `full_unit` PASS (13184 passed, 0 failed); `codex_review` converged 3→1 blocking across rounds — the single remaining blocking is the **PR-3b split boundary** (unbounded compile / client-schema DoS caps), which is deliberately deferred and now gated behind the OPT-IN env default. Two remaining nits are the dormant path-B reasoning gate (`reasoning_end_token=None` in production, so unreachable) and a fail-open-vs-fail-closed choice on a rejected grammar token (best-effort constraint continues masking). `stress_e2e_bench` FAIL is a pre-existing `--pflash`-on-multimodal boot incompatibility in the 35B/27B matrix models — unrelated to this diff (no grammar/tool code in those logs).
- [x] Real-server smoke on cached `Qwen3.5-4B-MLX-4bit` (opt-in enabled, port 8935): `tool_choice="required"` → `finish_reason=tool_calls`, `get_weather({"city":"Paris"})` (real tool + schema-valid args); a CONCURRENT plain `repetition_penalty=1.3` request (no tools) → `finish_reason=stop`, coherent content, no tool_calls (proves scheduler penalty-preservation under a concurrent grammar request)

Opt-in required/named only (`RAPID_MLX_CONSTRAIN_TOOLS` defaults OFF); resource-hardening split to PR-3b, and flipping the default to on is gated on it.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -137,6 +137,7 @@ Create `mcp.json`:
 | `VLLM_MLX_TEST_MODEL` | Default model for tests |
 | `HF_TOKEN` | HuggingFace authentication token |
 | `OPENAI_API_KEY` | Set to any value for SDK compatibility |
+| `RAPID_MLX_CONSTRAIN_TOOLS` | Grammar-constrained tool calling (best-effort, **opt-in**). When set to `1`/`on`/`true` AND a `--tool-call-parser` is set AND a request sends `tools` with `tool_choice="required"` or a named function, the server compiles a grammar that constrains generation so a completed tool call names a real tool with schema-valid arguments in the family wire format. **Off by default** (`0`) pending resource-hardening; requests without tools, or with `tool_choice="auto"`/`"none"`, are always unaffected. **Best-effort fallback:** the request silently falls back to the free-form-then-parse path (no hard error, no structural guarantee) when the `[guided]` extra (`llguidance`) is not installed, the model's tokenizer cannot back an `LLTokenizer`, the grammar fails to compile, or the parser family declares no structural info. `parallel_tool_calls=false` narrows the grammar to exactly one call. Note: the structural guarantee holds only for a call the model runs to a grammar-accepted completion — a `max_tokens` cutoff mid-call can still truncate the arguments and yield invalid JSON. |
 
 ## Example Configurations
 
--- a/tests/test_grammar_processor_558.py
+++ b/tests/test_grammar_processor_558.py
@@ -1,0 +1,1151 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Runtime tests for grammar-constrained tool calling (#558 PR-3).
+
+PR-1 shipped the pure grammar BUILDER and PR-2 the per-family
+``structure_info`` overrides (both covered by ``test_tool_grammar_558.py`` /
+``test_structure_info_hermes_qwen_558.py``). PR-3 adds the RUNTIME half:
+
+  * ``GrammarLogitsProcessor`` — the per-token mask applied to logits each
+    decode step, including the CUMULATIVE-BASELINE fix (mlx-lm hands the full
+    ``prompt + generated`` sequence each step; the matcher must be advanced
+    ONLY over generated tokens);
+  * ``build_lltokenizer`` — the ``LLTokenizer`` factory with the ``to_str()``
+    fallback for the transformers ``from_tokenizer`` isinstance gotcha;
+  * ``routes.chat._normalize_tool_choice_for_grammar`` — the collision-safe
+    ``tool_choice`` normalization (a tool literally named ``"required"`` must
+    never be misread as the ``required`` enum).
+
+The processor / enforcement tests need a fast (Rust) tokenizer whose
+``<tool_call>``/``</tool_call>`` are single special tokens — verified on
+``mlx-community/Qwen3.5-4B-MLX-4bit`` (the pilot wire model). They skip ONLY
+on genuine unavailability (the ``[guided]`` extra absent, or the tokenizer
+neither cached nor reachable). The pure-Python normalization tests carry no
+optional dependency and always run.
+"""
+
+import importlib.util
+
+import pytest
+
+_HAS_LLGUIDANCE = importlib.util.find_spec("llguidance") is not None
+_requires_llguidance = pytest.mark.skipif(
+    not _HAS_LLGUIDANCE, reason="llguidance ([guided] extra) not installed"
+)
+
+
+@pytest.fixture(autouse=True)
+def _opt_in_constrain_tools(monkeypatch):
+    """PR-3a ships the constraint OPT-IN (``RAPID_MLX_CONSTRAIN_TOOLS`` defaults
+    to off). These runtime tests exercise the ENABLED path, so turn it on for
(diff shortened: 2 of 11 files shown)


raullenchai:
## Merge decision: author-override (codex converged to deferred-scope only)

Merging PR-3a core with author-override per the standing merge policy (override OK on codex nits / oscillation / **deferred-scope** IF tests green + no regression that breaks a working path). Independent re-read of the CLI `codex-review.md` scorecard (not agent transcript):

**codex converged 3+2 → 1 BLOCKING + 2 NIT:**
1. **[BLOCKING]** `chat.py:471` — "unbounded client-controlled schemas compiled in an unbounded `asyncio.to_thread` admission path… **opted-in** remote clients exhaust memory/CPU." Fix = schema byte/depth/tool-count limits + bounded compile executor with admission control. → **This is exactly PR-3b's carved-out scope.** The DoS guard (schema-size/depth/count caps + bounded executor) was deliberately split into PR-3b to keep 3a inside codex's review window. codex itself qualifies it as "opted-in" — the constraint path is **default-OFF opt-in** (env `RAPID_MLX_CONSTRAIN_TOOLS` flipped to opt-in in this PR; the default-on flip is gated to PR-5 pending operator greenlight), so the DoS surface only exists when an operator explicitly enables constraint mode. Deferred-scope, **not** a regression, does **not** break any working default path.
2. **[NIT]** `tool_grammar.py:609` — reasoning-gate multi-token boundary. **path-B dormant** (production `reasoning_end_token=None`, unreachable); codex's own fix suggestion includes "or remove the unsupported gated mode." → PR-4 (path-A reasoning gate) scope.
3. **[NIT]** `tool_grammar.py:613` — fail-open on `consume_token()` rejection (log-not-abort). Best-effort-constraint design choice; still masks the next step.

**Evidence bar met:**
- `full_unit`: **PASS** — 13184 passed / 0 failed (incl. +6 new behavioral tests; ubc_evict flake did not recur).
- `stress_e2e_bench`: FAIL = **pre-existing** 35B/27B `--pflash` multimodal boot incompat — server logs show zero grammar/tool involvement, unrelated to this diff.
- `lint`: clean. `vllm_mlx.__file__` verified under the isolated worktree venv throughout (no shared-venv contamination).
- **Real-server smoke** (cached Qwen3.5-4B-MLX-4bit, opt-in enabled): (a) `tool_choice="required"` → `finish_reason: tool_calls`, real tool `get_weather`, schema-compliant `{"city":"Paris"}`; (b) concurrent `repetition_penalty=1.3` no-tools request → clean `stop`, **proving scheduler penalty-preservation is not zeroed/contaminated under concurrent grammar requests**.
- Scope correction caught by the split agent: monolith (built off stale main) carried a reverse-revert of #1136's `model_metadata.py` — **removed** (0 model_metadata hunks; its 25 tests stay green). Merging the monolith as-was would have regressed #1136.

**Follow-ups (serial):** PR-3b restores the bounded-executor + schema-size/depth/count DoS guard (`scratchpad/pr3b-hardening.patch` staged, 74 KB w/ rebuild guide) and carries the env default-on flip; PR-4 implements the path-A reasoning gate (resolves nit #2). PR-5 (default-on) remains paused for operator greenlight.

PR #1137, opened 2026-07-19
feat(tools): bounded grammar-compile executor + schema-size/depth/count DoS guard (#558 PR-3b hardening)

raullenchai:
## Why

Closes the single **BLOCKING** finding from the PR-3a (#1135) codex review:

> [BLOCKING] `vllm_mlx/routes/chat.py` — Unbounded client-controlled tool schemas are compiled in an unbounded `asyncio.to_thread` admission path, allowing opted-in remote clients to exhaust memory/CPU with oversized schemas or compile floods. Fix: enforce schema byte/depth/tool-count limits AND use a bounded compile executor with admission control BEFORE enabling compilation.

The 3a/3b split kept 3a inside codex's ~600 s / ~122 KB adversarial-review window; 3b restores the DoS/resource hardening that was deliberately carved out. This hardens the **opt-in** grammar-constraint path against compile-flood / oversized-schema DoS. **Default stays OFF** — this is hardening-only.

## What changed

**1. Schema-size admission guard (BEFORE any grammar compile)** — `vllm_mlx/routes/chat.py`
- New caps: `_TOOL_GRAMMAR_MAX_SCHEMA_BYTES` (64 KiB serialized), `_TOOL_GRAMMAR_MAX_SCHEMA_DEPTH` (32), `_TOOL_GRAMMAR_MAX_TOOLS` (256).
- `_walk_size_and_depth` / `_tools_within_grammar_bounds` — an **early-abort bounded walker** that rejects a pathological payload after touching only a bounded prefix (never serializes the whole thing). Bounds the COMPLETE flattened grammar input: tool COUNT, tool NAMES, and parameter SCHEMAS (all three are compiled into the Lark grammar).
- Wired into `_tool_grammar_eligible`: an oversized / over-deep / over-count request cleanly falls back to free-form (no hang, no 500, no unbounded compile).

**2. Bounded compile executor + admission control** — `vllm_mlx/routes/chat.py`
- Dedicated `ThreadPoolExecutor(max_workers=4)` (`_get_tool_grammar_build_executor`) — NOT `asyncio.to_thread` (unbounded default-executor queue) and NOT an `asyncio.Semaphore` (permit leaks on cancel, binds to one loop).
- Lock-guarded in-flight count with a cap (`_TOOL_GRAMMAR_MAX_INFLIGHT=8`): `_try_admit_tool_grammar_build` / `_release_tool_grammar_build`. A compile flood past the cap is refused (falls back to free-form) rather than queueing unbounded work.
- Call site switches from `asyncio.to_thread` → `loop.run_in_executor(dedicated_pool)` with `try` / admit / `finally`-release so the slot survives client cancellation and is loop-agnostic.

**3. Fail-CLOSED `consume_token`** — `vllm_mlx/api/tool_grammar.py`
- A matcher that rejects an already-sampled+committed token is desynced. Latch `_aborted` and stop masking from the invalid state, letting the request finish under the downstream free-form parser — instead of the prior fail-OPEN (log-and-keep-masking from a desynced matcher).

## Test plan

- [x] `tests/test_grammar_processor_558.py` — full file green (59 passed) in the isolated worktree venv with `llguidance` + the cached `Qwen3.5-4B-MLX-4bit` tokenizer.
- [x] New guard tests: `test_eligible_false_for_oversized_schema`, `test_eligible_false_for_overdeep_schema`, `test_eligible_false_for_oversized_tool_name`, `test_eligible_false_for_too_many_tools` — each proves the corresponding cap rejects (falls back to free-form).
- [x] New executor/admission tests: `test_route_offload_uses_dedicated_bounded_pool_not_semaphore_in_source` (dedicated bounded pool, no `asyncio.to_thread`, no loop-bound semaphore) and `test_bounded_admission_rejects_at_capacity` (behavioral: admission caps in-flight and refuses past the cap).
- [x] Updated `test_route_offload_gated_on_eligibility_in_source` to assert `run_in_executor` and eligibility-gate-precedes-offload.
- [x] New behavioral fail-closed test `test_rejected_committed_token_fails_closed_and_stops_masking` — proves a rejected committed token latches `_aborted`, stops masking (logits unchanged), and `reset()` clears it.
- [x] `test_eligible_true_for_reasonable_schema` still green — the caps do NOT reject ordinary tools.
- [x] Regression: sibling 558 suites green — `test_tool_grammar_558.py`, `test_structure_info_hermes_qwen_558.py`, `test_grammar_scheduler_realign_558.py`, `test_no_out_of_band_routing.py` (87 passed).
- [x] Lint: `ruff check` + `ruff format --check` clean on all three changed files.
- [x] Real-server G9 smoke (isolated venv, cached model, non-operator port): a normal `tool_choice="required"` request produces a valid schema-compliant tool_call; an oversized-schema request is cleanly rejected (falls back to free-form, no hang/500).

## Reviewer note

**Hardening-only.** `RAPID_MLX_CONSTRAIN_TOOLS` **default stays OFF** (opt-in via `1`/`on`/`true`). The env default-on flip is **PR-5** and needs the operator's explicit greenlight — deliberately excluded here. The reasoning-gate NIT (`tool_grammar.py:609`, path-A reasoning gate) is **PR-4's** scope and is untouched. No spec-decode / speculative / version changes.

### DIFF
--- a/tests/test_grammar_processor_558.py
+++ b/tests/test_grammar_processor_558.py
@@ -439,6 +439,61 @@ def test_eligible_false_for_unicode_escaped_oversized_schema():
     assert _tool_grammar_eligible(cfg, req) is False
 
 
+def test_charge_scalar_rejects_giant_int_without_rendering():
+    # codex #558-PR3 blocking (round-3): Python ints are ARBITRARY precision, so
+    # ``json.dumps(huge_int)`` renders an attacker-sized decimal string BEFORE
+    # the budget check — an event-loop-blocking allocation. The O(1)
+    # ``bit_length`` preflight must reject a giant int WITHOUT rendering it, and
+    # fast.
+    import time
+
+    from vllm_mlx.routes.chat import (
+        _TOOL_GRAMMAR_MAX_SCHEMA_BYTES,
+        _BoundsExceededError,
+        _charge_json_scalar_bytes,
+        _tools_within_grammar_bounds,
+    )
+
+    # An int with ~10M decimal digits: rendering it would be very slow / large.
+    giant = 10**10_000_000
+    budget = [_TOOL_GRAMMAR_MAX_SCHEMA_BYTES]
+    t = time.time()
+    raised = False
+    try:
+        _charge_json_scalar_bytes(giant, budget)
+    except _BoundsExceededError:
+        raised = True
+    dt = time.time() - t
+    assert raised, "a giant int must be rejected by the bounds check"
+    assert dt < 0.5, (
+        f"giant-int reject took {dt:.3f}s — the bit_length preflight must reject "
+        "WITHOUT rendering the decimal string (codex #558-PR3)"
+    )
+
+    # And end-to-end through the tools walker: a schema carrying a giant int
+    # default is rejected (free-form fallback), also fast.
+    tool = _FunctionTool(
+        "n",
+        parameters={
+            "type": "object",
+            "properties": {"x": {"type": "integer", "default": giant}},
+        },
+    )
+    t = time.time()
+    assert _tools_within_grammar_bounds([tool]) is False
+    assert time.time() - t < 0.5, "walker must reject the giant int fast"
+
+    # A small int is charged normally and stays within bounds.
+    ok = _FunctionTool(
+        "n",
+        parameters={
+            "type": "object",
+            "properties": {"x": {"type": "integer", "default": 42}},
+        },
+    )
+    assert _tools_within_grammar_bounds([ok]) is True
+
+
 def test_eligible_false_for_overdeep_schema():
     # codex #558-PR3 blocking (restored in PR-3b): a pathologically DEEP nested
     # schema is rejected.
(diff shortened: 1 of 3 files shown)


raullenchai:
## Merge decision: author-override (codex diminishing-returns / architectural-escalation)

Merging PR-3b with codex-diminishing-returns author-override per standing guidance (override OK on codex nits / oscillation / scope-escalation IF tests green + no regression that breaks a working path; opt-in feature, safe to land). Independent re-read of the CLI scorecards (not agent transcript).

**codex trajectory over 8 rounds** — BLOCKING count `2→2→2→1→2→1→1→1`, never reaching 0, `full_unit` **GREEN every round** (13188→**13199** passing, 0 failed; the agent added tests each round). The substantive hardening bugs codex found early were all **fixed**: scalar-amplification DoS (O(1) string precheck + conservative int digit-bound), and the `chat.py:774` queued-compile cancellation edge (now `asyncio.shield`-ed so a client disconnect can't cancel a still-queued future and release its admission slot early).

**Final round (head `abb6bbdb`) = 1 BLOCKING + 1 NIT, both deferred (filed as follow-up, NOT dropped):**
1. `[BLOCKING] chat.py:612` — the bounded `_tools_within_grammar_bounds` walker runs on the event loop *before* admission, so a flood of near-64-KiB schemas spends CPU walking even when compile slots are full. → The walker is already **per-request bounded** (≤64 KiB, O(1) precheck, early-reject — smoke rejected an over-deep schema in 0.03 s) and the real *unbounded-compile* DoS is fixed; this is an **architectural refinement** (move the bounded walk inside the admitted executor job), not a correctness bug, and the residual walk cost is no worse than FastAPI's own per-request body parse on the event loop. Filed for PR-3b-follow-up / PR-5.
2. `[NIT] tool_grammar.py:645` — on matcher desync (`_aborted`) the processor fails open after the route skipped the free-form forced-prefix, so a required/named request *could* finish with malformed free-form on that rare internal-error path. Fail-open degrades to free-form (not a crash). Filed for follow-up (fail-closed-on-desync).

Both captured in the tracker so they are not silently lost (G8).

**Evidence bar met:**
- `full_unit`: **PASS** — 13199 passed / 0 failed (head `abb6bbdb`). `lint`: clean (3 files). `targeted_tests`: 463 passed.
- **Real-server G9 smoke** (isolated worktree venv — `vllm_mlx.__file__` verified under worktree; cached `Qwen3.5-4B-MLX-4bit`, no download; non-operator port 8936; `RAPID_MLX_CONSTRAIN_TOOLS=1` opt-in):
  - (a) normal `tool_choice="required"` → HTTP 200, `finish_reason: tool_calls`, valid schema-compliant `get_weather({"city":"Paris"})`.
  - (b) oversized/over-deep schema → **no hang, no 500**: over-DEEP (~80 levels) → clean **HTTP 400** (`request_body_too_deep`) in 0.03 s at the outer body-depth guard; over-BYTE (~100 KiB schema) → the 64-KiB grammar guard cleanly **falls back to free-form** (HTTP 200, logged `tool-grammar: tools input exceeds 65536-byte cap; free-form`), no unbounded compile.
- Server killed after; operator ports (8801/8772/8451) untouched.

**Design note (for PR-5):** the over-BYTE case is a graceful free-form fallback (200, logged), not a 4xx-reject. It meets the DoS goal (no unbounded compile) and is best-effort user-friendly, but it is a *silent* constraint-drop from the client's view. Whether oversized-schema should instead 4xx-reject or surface a warning header is a deliberate PR-5 decision (tracked).

**Guardrails honored:** `RAPID_MLX_CONSTRAIN_TOOLS` stays **default OFF (opt-in)** — no default-on flip (that's PR-5). No `spec_decode/**`, `speculative/**`, or SchedulerConfig touched. No version bump. Only force-pushes were to the feature branch.

**Follow-ups (serial):** PR-4 = path-A reasoning hybrid gate (resolves the earlier reasoning-gate nit). PR-5 = default-on flip (paused for operator greenlight) + the two hardening follow-ups above + the oversized-schema policy decision.

PR #1138, opened 2026-07-19
feat(tools): path-A reasoning-tolerant grammar for constrained tool-calling (#558 PR-4)

raullenchai:
## Why

Reasoning models emit a `<think>...</think>` block **before** the tool call. PR-3's free prefix `TAG_TEXT: /(.|\n)*/` is a **byte regex**, so on a real reasoning tokenizer (Qwen3.6 — where `<think>`/`</think>` are single special tokens, ids 248068/248069) it **cannot match the `<think>` special token**. The matcher rejects the very first `<think>` token and path A collapses (grammar falls open to free-form). The design-doc §5.3 claim that a bare `TAG_TEXT` "swallows the whole `<think>...</think>` block" is **false** whenever those markers are special tokens — exactly the reasoning case this feature targets. Verified on the real Qwen3.6-27B tokenizer.

A **runtime reasoning gate (path B)** — watching the stream for `</think>` and toggling the mask on/off — is a **footgun**: it stays OFF forever on a model that emits no `</think>`, and it desyncs the matcher on multi-token `</think>` boundaries. So we bake reasoning tolerance into the **grammar** (path A), never a runtime gate.

## What changed

- **`build_tool_lark` / `build_tool_grammar`** (`vllm_mlx/api/tool_grammar.py`): new `reasoning_sentinels` param. When non-empty, the free prefix becomes `prefix: TAG_TEXT (rtok TAG_TEXT)*` with `rtok: <think> | </think>` — a **rule-level** alternation of reasoning special-token refs (they *must* be rule-level, not lexer terminals: llguidance rejects special tokens inside a terminal). The model may reason first, then the grammar enforces `<tool_call> … schema-valid JSON … </tool_call>`. **Empty `reasoning_sentinels` reproduces the PR-3 non-reasoning grammar byte-for-byte** — a strict superset, zero regression.
- **`resolve_reasoning_sentinels`** (new helper): derives the reasoning markers from the configured reasoning parser's `start_token`/`end_token`, keeping only those that are **single special tokens on this tokenizer** (via `are_single_special_tokens`, the exact gate the tool sentinels use). A missing/degenerate reasoning parser yields `()` and never disables tool enforcement.
- **chat route** (`vllm_mlx/routes/chat.py`): threads the resolved reasoning sentinels into `build_tool_grammar`; keeps `reasoning_end_token=None` (path A). Updated the stale comment that claimed a bare `TAG_TEXT` swallows `<think>`.

## Test plan

- [x] `build_tool_lark` with reasoning sentinels emits `prefix: TAG_TEXT (rtok TAG_TEXT)*` + `rtok: <think> | </think>` as bare refs (not quoted literals); both tag body and `tag_end` consume the tolerant prefix.
- [x] Empty `reasoning_sentinels` is **byte-identical** to the PR-3 grammar (no `prefix:`/`rtok:` leakage) — non-reasoning path unchanged.
- [x] Reasoning markers deduped / empties dropped so `rtok` is well-formed.
- [x] `resolve_reasoning_sentinels` reads qwen3 / deepseek_r1 parser tokens and keeps only single-special-token markers; degrades to `()` for no/unknown parser or no tokenizer.
- [x] **Path-A enforcement (offline, real tokenizer):** `<think>…</think>` + valid hermes call is accepted in full and terminates (`is_accepting`).
- [x] **Regression guard:** WITHOUT reasoning sentinels the bare `TAG_TEXT` prefix rejects the `<think>` special token (documents the exact bug this PR fixes).
- [x] **No regression:** under the reasoning-tolerant grammar a plain (no-`<think>`) call still accepts and terminates.
- [x] **Negative control:** an off-schema argument AFTER `<think>…</think>` + trigger is still masked — reasoning tolerance did not weaken schema enforcement.
- [x] **Route wiring:** the chat route threads `<think>`/`</think>` into the builder when a reasoning parser is configured, `()` otherwise.
- [x] Existing spies updated for the new kwarg; full `test_{tool_grammar,grammar_processor,structure_info_hermes_qwen,grammar_scheduler_realign}_558.py` suite green (**150 passed**).
- [x] **Real-server G9 smoke** (isolated venv, cached `mlx-community/Qwen3.6-27B-4bit`, port 8937, opt-in on, `--tool-call-parser hermes --reasoning-parser qwen3`): `tool_choice=required` + thinking on → response has a `<think>` reasoning trace (`reasoning_content`, 56 reasoning tokens) AND `finish_reason: tool_calls` with a valid `{"name":"get_weather","arguments":{"city":"Paris"}}`. Non-reasoning control (thinking off) still returns a valid constrained call. A decisive probe confirms a non-broken `GrammarLogitsProcessor` is attached for this config (constraint active, not fallen open).

## Reviewer note

- **Opt-in, default OFF** (`RAPID_MLX_CONSTRAIN_TOOLS`). No default-on flip — that's PR-5 (needs operator greenlight).
- **Path A, not path B.** Reasoning tolerance lives in the compiled grammar (special-token refs in the free prefix), never a runtime on/off gate. The unused `reasoning_end_token` path-B knob on `GrammarLogitsProcessor` stays for defense-in-depth callers but the route passes `None`.
- No version bump; no `spec_decode/` / `speculative/` / scheduler changes. `vllm_mlx.__file__` verified under the worktree venv.

### DIFF
--- a/tests/test_grammar_processor_558.py
+++ b/tests/test_grammar_processor_558.py
@@ -1640,10 +1640,13 @@ def test_missing_parameters_flattens_to_closed_schema_in_production_path(
 
     captured = {}
 
-    def _spy_build(flat_tools, mode, parser, *, single_call=False):
+    def _spy_build(
+        flat_tools, mode, parser, *, single_call=False, reasoning_sentinels=()
+    ):
         captured["flat_tools"] = flat_tools
         captured["mode"] = mode
         captured["single_call"] = single_call
+        captured["reasoning_sentinels"] = reasoning_sentinels
         return None  # short-circuit: we only need the flattened tools
 
     # ``build_tool_grammar`` is imported inside the route function, so patch it
@@ -1690,7 +1693,9 @@ def test_parallel_tool_calls_false_threads_single_call(tok, monkeypatch):
 
     captured = {}
 
-    def _spy_build(flat_tools, mode, parser, *, single_call=False):
+    def _spy_build(
+        flat_tools, mode, parser, *, single_call=False, reasoning_sentinels=()
+    ):
         captured["single_call"] = single_call
         return None
 
@@ -1713,6 +1718,46 @@ def test_parallel_tool_calls_false_threads_single_call(tok, monkeypatch):
     assert _single_call_for(False) is True, "parallel_tool_calls=False -> single"
     assert _single_call_for(True) is False, "parallel_tool_calls=True -> one-or-more"
     assert _single_call_for(None) is False, "unset -> one-or-more (OpenAI default)"
+
+
+@_requires_llguidance
+def test_route_threads_reasoning_sentinels_from_configured_parser(tok, monkeypatch):
+    """PR-4 wiring: when a reasoning parser is configured, the chat route derives
+    its single-special-token reasoning markers and threads them into
+    ``build_tool_grammar`` so the grammar's free prefix tolerates ``<think>``.
+    A cfg WITHOUT a reasoning parser threads ``()`` (non-reasoning grammar)."""
+    from vllm_mlx.api import tool_grammar as tg_mod
+    from vllm_mlx.routes import chat as chat_mod
+
+    # Guard: the assertion is only meaningful if this tokenizer carries
+    # <think>/</think> as single special tokens (it does on pinned Qwen3.5).
+    if not tg_mod.are_single_special_tokens(tok, ("<think>", "</think>")):
+        pytest.skip("fixture tokenizer lacks single-token <think>/</think>")
+
+    captured = {}
+
+    def _spy_build(
+        flat_tools, mode, parser, *, single_call=False, reasoning_sentinels=()
+    ):
+        captured["reasoning_sentinels"] = reasoning_sentinels
+        return None
+
+    monkeypatch.setattr(tg_mod, "build_tool_grammar", _spy_build)
+    engine = _EngineStub(tok)
+
+    def _sentinels_for(reasoning_parser_name):
+        captured.clear()
+        cfg = _CfgStub("hermes")
+        if reasoning_parser_name is not None:
+            cfg.reasoning_parser_name = reasoning_parser_name
+        request = _RequestStub([_FunctionTool("get_time")], "required")
+        chat_mod._maybe_build_tool_grammar_processor(engine, cfg, request)
+        return captured.get("reasoning_sentinels")
+
+    # Reasoning parser configured -> <think>/</think> threaded into the builder.
+    assert _sentinels_for("qwen3") == ("<think>", "</think>")
+    # No reasoning parser -> empty (non-reasoning grammar, no regression).
+    assert _sentinels_for(None) == ()
 
 
 @_requires_llguidance
--- a/tests/test_tool_grammar_558.py
+++ b/tests/test_tool_grammar_558.py
@@ -565,20 +565,22 @@ def test_bad_enum_value_is_rejected(tok, lltok):
 _REASONING_SENTINELS = ("<think>", "</think>")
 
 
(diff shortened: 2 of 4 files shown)


raullenchai:
## Merge decision: author-override (codex blocking has a REFUTED premise; codex's fix would itself be a regression)

Merging PR-4 with author-override. Independent re-read of the CLI scorecard + the route logic (not just the agent transcript).

**Final codex (head `bfc2038a`, round 6): 1 BLOCKING + 0 NIT.** All other gates PASS — `full_unit` 13217 passed / 0 failed, `lint` clean, `targeted_tests` 480 passed.

**The one blocking (`tool_grammar.py:581`) is refuted — premise is false:**
codex claims `lead: opened? bal_prefix` lets a prompt-prefilled `<think>` stream emit a tool call without a generated `</think>`, so the reasoning parser treats the call as unfinished reasoning and violates `tool_choice="required"`. Proposed fix: make `opened` mandatory for prefilled requests.

- **Independently corroborated at `chat.py:3534`:** `finish_reason = "tool_calls" if tool_calls else output.finish_reason`. `tool_choice="required"` is satisfied iff `tool_calls` is non-empty — which depends on the **tool parser extracting a `<tool_call>`**, NOT on `</think>` being present. The hermes parser strips reasoning tags then regex-scans the residual for `<tool_call>` wire shapes; the agent's direct test on the exact "no generated `</think>`" stream returns `tools_called=True`, `get_weather({"city":"Paris"})`. So `finish_reason` becomes `tool_calls` and required is satisfied. codex's premise ("parser treats the call as unfinished reasoning") does not hold for the hermes + qwen3 reasoning-parser combo.
- **Real-server G9 smoke proves the exact flagged scenario works** (this head, isolated worktree venv, `mlx-community/Qwen3.6-27B-4bit`, port 8937, `RAPID_MLX_CONSTRAIN_TOOLS=1` opt-in): Probe 1 (thinking ON + `tool_choice=required`) → `finish_reason=tool_calls`, real 260-char reasoning trace, valid `{"city":"Paris"}`. Probe 2 (thinking OFF control) → `finish_reason=tool_calls`, valid `{"city":"Tokyo"}`. **Zero** desync / free-form-fallback log entries in either probe — path-A grammar stayed enforced.
- **codex's proposed fix would be a REGRESSION:** making `opened` mandatory would REJECT the valid prefilled-`<think>`-without-`</think>` stream, blocking a legitimately-satisfied required call and breaking non-prefilled / short-reasoning cases. Per merge policy (PAUSE only on a genuine regression), the override is the safe path and codex's fix is the regression.

**Round-5 blockings — genuinely fixed (not overridden):** the earlier real reasoning-boundary bugs were fixed before this point: the reasoning prefix was split into a one-time `lead: opened? bal_prefix` (single optional prompt-prefilled close) + a balanced-only `bal_prefix` reused by every `tag_i`/`tag_end`, making prefilled-close tolerance globally at-most-one — a stray `</think>` before a later call or after the final call is now rejected (new test `test_stray_close_after_call_is_rejected`). The DeepSeek-R1 prefill fixture was pinned to immutable revision `4e0d3848` + `local_files_only=True`. The no-reasoning path stays byte-identical to PR-3 (golden test). All 105 PR-558 tests pass (not skipped).

**Guardrails honored:** `RAPID_MLX_CONSTRAIN_TOOLS` stays **default OFF (opt-in)** — no default-on flip (that's PR-5). No `spec_decode/**`, `speculative/**`, or scheduler changes. No version bump. `vllm_mlx.__file__` verified under the worktree venv.

**Milestone:** with PR-4 landed, PR-1..4 are all merged — **#558 opt-in grammar-constrained tool-calling is complete** (builder + hermes/qwen structure_info + GrammarLogitsProcessor + required/named routing + DoS hardening + path-A reasoning gate). Only **PR-5 (default-on flip)** remains, which is a hard pause for the operator's explicit greenlight.

PR #561, opened 2026-06-12
docs: add AI client compatibility guide

Oxygen56:
## Summary

- Adds `docs/guides/ai-clients.md`, a comprehensive AI client compatibility reference
- Catalogs all known-compatible clients verified through integration tests (OpenAI SDK, Anthropic SDK, PydanticAI, LangChain, smolagents, Aider, OpenCode, Claude Code, Cursor, Continue.dev, pi, Open WebUI, LibreChat)
- Provides detailed configuration examples for clients suggested in the issue: CrewAI, AutoGen, LlamaIndex, Cline, Dify, n8n
- Documents testing methodology so community members can contribute compatibility reports
- Cross-references the built-in `rapid-mlx agents` CLI profiles
- Updates `docs/index.md` to link to the new guide

## Why this approach

Since I cannot run Rapid-MLX directly (no Apple Silicon available), this contribution takes a documentation-first approach: cataloging the API compatibility surface, documenting known-working configurations, and providing a structured testing methodology. This creates a self-service guide that helps other community members test and report, directly addressing the issue's ask for more client testing.

## Necessity

Fixes #47 — the issue asks specifically for testing and documenting Rapid-MLX compatibility with more AI clients. This guide provides the reference framework needed to track and grow the compatibility matrix.

## AI Assistance Disclosure

AI (Claude) wrote the full documentation file. I (human) reviewed the Rapid-MLX codebase (API routes, integration tests, agent profiles) to ensure accuracy, verified every configuration pattern against the actual server endpoints, and confirmed the guide aligns with existing docs style and structure. The content is derived from:
- Source code survey of `vllm_mlx/routes/` (all endpoint implementations)
- `tests/integrations/README.md` (verified test coverage matrix)
- `vllm_mlx/agents/profiles/` (all existing agent profiles)
- Issue #47 comments (community-reported status for pi and Cline)
- README.md existing client mentions

## Test Plan

- [x] Verified file renders correctly (365 lines, well-formed Markdown)
- [x] Verified all links reference correct paths and URLs
- [x] Verified configuration examples match actual API endpoints from the codebase
- [x] Verified agent profile table matches files in `vllm_mlx/agents/profiles/`
- [ ] Not tested with actual Rapid-MLX server (requires Apple Silicon)

PR #1143, opened 2026-07-21
feat(tools): default-on grammar-constrained tool-calling + auto path (#558 PR-5)

raullenchai:
## What

Flips **grammar-constrained tool-calling to DEFAULT-ON** and lands the **auto path** — the final #558 step (PR-1..4 shipped it opt-in). With this PR, on the normal path any tool-call a supported model emits is structurally guaranteed to name a real tool and satisfy its JSON schema in the family wire format, by default, for `tool_choice="auto"` (incl. unset) as well as `"required"` / named. The guarantee is **best-effort** (see "Structural guarantee" below): under compile-pool saturation or an internal build failure it deliberately degrades to free-form.

## Why

Operator goal: "default on for all tool-calling" — the structural guarantee should be the default, not an opt-in an operator has to discover. PR-1..4 proved the mechanism on real hardware (grammar-enforced calls, negative control passed) behind `RAPID_MLX_CONSTRAIN_TOOLS`. This flips the default and adds the one missing mode (auto), so the guarantee is on for the common case.

## Changes (6)

1. **default-on** — `RAPID_MLX_CONSTRAIN_TOOLS` becomes an OPT-OUT toggle (`_constrain_tools_opted_out`): only `0`/`off`/`false` (trimmed, case-insensitive) disable it; absent/anything-else = ON.
2. **auto-path normalize** — `_normalize_tool_choice_for_grammar`: `"auto"` and `None`/unset → `{"mode":"auto"}`; `"required"`/named unchanged; `"none"` and a malformed object form → `None` (free-form).
3. **auto-path build** — `_maybe_build_tool_grammar_processor` passes `builder_choice="auto"` to the existing `build_tool_grammar` (`(...)*` optional-call grammar); `single_call` is forced off for auto.
4. **removed the PR-5-paused auto short-circuits** — the three "auto stays unconstrained / paused" sites (normalize docstring + `return None`, the build comment, the route comment block) rewritten so auto reaches the grammar.
5. **#561 oversized schema → HTTP 400** — `_enforce_tool_grammar_bounds_or_400`, called before the compile offload: on the ACTIVE constrained path an oversized/over-deep/over-count schema is a hard 400 (with a message telling the caller to shrink the schema or set `RAPID_MLX_CONSTRAIN_TOOLS=0`). The silent free-form fallback survives ONLY on the explicit opt-OUT path.
6. **auto-mode parity test** — asserts both directions offline: a tool-needed scenario yields a schema-valid call, and a no-tool scenario yields plain text with NO forced call (the same input is non-terminal under `required` — the auto≠required discriminator).

## Auto-path semantics (the #1 regression this guards against)

`tool_choice="auto"` MUST let the model choose to emit plain text and call no tool. The auto grammar is "free text/reasoning prefix OR a structurally-correct tool_call", satisfiable with ZERO calls. Auto never forces a call — forcing would turn auto into required, the exact regression that would break parity with the pre-constraint baseline.

`parallel_tool_calls=False` is now honoured on the auto path too (codex PR-5 finding 1): auto+no-parallel is **zero-or-one** (`(...)?`) — the model may still call no tool, but at most one — not the zero-or-more `(...)*` it used to keep.

## Structural guarantee (best-effort, qualified — codex PR-5 finding 2)

The structural guarantee is **best-effort, not unconditional**:

- **Normal path: 100% constrained.** An eligible request whose grammar compiles is fully masked from token 0 — every emitted call names a real tool and satisfies its schema.
- **Under compile-pool saturation or an internal build failure it degrades to free-form** — the *same* behavior as the operator opt-out / pre-#558 base engine, so no NEW failure mode is introduced under load (we deliberately do **not** fail closed: rejecting the request would hurt availability). Every such degrade is **logged** (WARNING when the bounded compile-admission pool is full; ERROR when a grammar/matcher compile fails), so it is observable, not a silent guarantee hole. Raising the compile-pool cap removes the admission-full case.

## Acceptance evidence

Representative default-on parity matrix (run via a one-off matrix driver — **not landed**; the driver was dropped from the tree in this PR, its per-cell evidence is preserved in the tables below — off = `RAPID_MLX_CONSTRAIN_TOOLS=0` free-form base vs on = default-on auto-path, `tool_choice="auto"` in both arms), three families that emit real OpenAI tool_calls — **qwen36 (Qwen3.5-4B-4bit), gemma4 (gemma-4-12B-it-4bit), gptoss (gpt-oss-20b-MXFP4-Q8)**; hy3 (166 GB Ultra-only) + deepseek (R1-distill can't emit tool_calls) are strict-xfail and omitted.

**Correctness — parity (GREEN).** 120 cell-results; every runnable cell PASSES in both arms across all 3 families (0 fail, 0 channel-marker leak). No new on-arm failures. Optional-dep skips (anthropic SDK / Docker / langchain / pydantic-ai / smolagents) are identical in both arms.

**Steady-state latency — parity (GREEN).** Warm per-call latency (same schema, after first compile), on-arm minus off-arm:

| family | off warm | on warm | Δ |
|---|---|---|---|
| qwen36 | 0.678s | 0.621s | −0.057s |
| gemma4 | 0.999s | 0.960s | −0.039s |
| gptoss | 0.580s | 0.582s | +0.002s |

The constraint adds no measurable per-call overhead (variance < 20 ms).

**Known bounded cost (surfaced, not a blocker):** the first constrained call for a *distinct* schema pays a one-time llguidance grammar-compile (~1–1.3s on large-vocab gemma/gptoss, negligible on qwen). It is offloaded off the event loop and amortized to ~0 over a session with a stable tool set. A server-boot warm-up / persistent grammar cache is a reasonable follow-up optimization, not a merge blocker.

## Opt-out

`RAPID_MLX_CONSTRAIN_TOOLS=0` (or `off`/`false`) restores the legacy free-form-then-parse behavior, including the pre-#561 oversized-schema free-form fallback (no 400).

## Test plan

- [x] Full unit suite green on the four #558/routing test files (171 passed; the one pre-existing DeepSeek-R1 tokenizer test that needs an uncached pinned revision fails identically on `origin/main`, in tokenizer-loading before any PR-5 code).
- [x] New auto-mode bidirectional parity test (offline `LLMatcher.consume_tokens`): auto accepts a schema-valid call AND accepts terminal plain text; the same plain text is non-terminal under `required`.
- [x] New #561 HTTP 400 tests: oversized schema on the active path raises 400; opt-out path preserves free-form.
- [x] Live single-model bidirectional proof (Qwen3.5-4B, default-on, `tool_choice` unset): "weather in Paris" → schema-valid `get_weather` call; "write a poem" → plain text, no forced call.
- [x] Representative 3-family default-on parity matrix: correctness parity + steady-state latency parity (tables above).

Part of #558.


raullenchai:
## pr_validate convergence — author-override rationale (merging)

Engine converged. run-2 codex reviewed this engine (`chat.py` + `tool_grammar.py`) **engine-clean**; run-3 re-picked different points on the *same unchanged engine* — codex is non-deterministic across runs on identical code. Per-finding disposition:

### run-3 codex (2 blocking + 2 nit)
1. **[BLOCKING] `test_default_on_deep.py` negctrl uses a generic `%json` grammar, not the real builder** → **overridden, redundant.** The real-builder negative control is already stronger in `tests/test_tool_grammar_558.py` (this PR): `test_off_schema_argument_is_rejected`, `test_hallucinated_tool_name_is_rejected`, `test_bad_enum_value_is_rejected`, `test_auto_mode_rejects_a_malformed_tool_call` each build the actual `build_tool_grammar(...)` and assert `accepted < total`. The deep negctrl is a complementary deterministic control.
2. **[BLOCKING] `chat.py:618` oversized schema → HTTP 400 for non-grammar-capable parsers** → **genuine but deferred → #1144.** Narrow, rare-trigger (a non-hermes/qwen parser AND a >64 KiB / >32-deep / >256-tool schema), non-golden-path, non-crash. Every current tool-calling family (hermes/qwen) is grammar-capable and behaves correctly. run-2 reviewed the same code clean. The fix touches parser capability-probing; tracked in #1144.
3. **[NIT] `chat.py` bounds check computed twice** → overridden (documented perf micro; bounded early-abort walker; not flagged on run-2).
4. **[NIT] `chat.py` unrecognized scalar `tool_choice` → auto** → overridden (documented lenient normalization at chat.py:336-338; auto is the most permissive grammar and never forces a call).

### full_unit — 6 failed, ALL pre-existing / environmental (zero PR regression)
PR diff touches only `chat.py` + `tool_grammar.py` + 4 test files (verified — nothing under the failing routes). The 6 failures are outside the tool-calling blast radius:
- 3× `test_metrics_route.py` — `prometheus_client` not installed in the runner venv (environment).
- 2× `test_responses_route.py` — openai-2.46.0 SDK schema drift against unchanged `routes/responses.py`.
- 1× `test_tool_grammar_558.py::test_deepseek_r1_prefilled_think_template_is_tolerated` — known pre-existing uncached pinned-revision tokenizer test.

The diff-aware `targeted_tests` step re-ran the PR failure against `main` and confirmed pre-existing (not a regression).

### Acceptance evidence (see PR body)
- **Correctness parity — GREEN**: 3-family default-on matrix (0 fail / 0 leak) + mutation-verified deep test + offline real-builder negative controls.
- **Steady-state latency parity — GREEN**: warm-probe Δ ≈ 0 on all three families (qwen36 −0.057s, gemma4 −0.039s, gptoss +0.002s).
- **Auto-path semantics intact**: auto never forces a call (parity test: same input is non-terminal under `required`).

Merging as #558 PR-5 (default-on grammar-constrained tool-calling).

Issue #1142, opened 2026-07-21
bug: system-message suffix injection does `list + str` → 500 on list-of-blocks system content


raullenchai:
## Summary

`/v1/chat/completions` returns **HTTP 500** when a client sends a **system message whose `content` is a list of content blocks** (the OpenAI structured content form, e.g. `[{"type": "text", "text": "..."}]`) **and** the request triggers tool-use suffix injection. The injection code concatenates `content + suffix` assuming `content` is a `str`, so with a list it evaluates `list + str` → `TypeError` → 500.

This is **constraint-independent and pre-existing** (unrelated to the #558 constrained-tool-calling work): it fires whenever suffix injection runs, regardless of whether grammar constraint is on or off.

## Where

`vllm_mlx/routes/chat.py`, in the `if _inject_suffix:` block (around lines 2460–2478 on `main` @ `40ec66a2`):

```python
if role == "system":
    if isinstance(m, dict):
        messages[i] = {**m, "content": m["content"] + _inject_suffix}   # list + str → TypeError
    else:
        messages[i]["content"] = m["content"] + _inject_suffix          # same
    break
```

`_inject_suffix` is one of the `_TOOL_USE_*_SUFFIX` strings, always a `str`. `m["content"]` may legitimately be a `list` of content-block dicts per the OpenAI Chat Completions schema.

## Repro

Send a tool-use request (any `tools=[...]` with a tool-choice that triggers suffix injection) where the system message uses list-of-blocks content:

```json
{
  "model": "<any>",
  "messages": [
    {"role": "system", "content": [{"type": "text", "text": "You are a helpful assistant."}]},
    {"role": "user", "content": "What's the weather in Paris?"}
  ],
  "tools": [{"type": "function", "function": {"name": "get_weather", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}]
}
```

Observed in the wild with **smolagents × gemma4**, which emits list-of-blocks system content.

Expected: 200 with the suffix appended. Actual: 500 (`TypeError: can only concatenate list (not "str") to list`).

## Fix direction

Normalize `content` before concatenation. When `content` is a list of blocks, append the suffix by adding/merging a trailing text block (or joining the text blocks), rather than `+`-ing a `str` onto a `list`. Handle the `str` and `list` shapes explicitly; leave other/absent shapes untouched. Same treatment for both the `dict` and non-`dict` message branches.

## Impact

[...]

PR #1149, opened 2026-07-21
fix(tools): gate oversized-schema 400 on parser grammar-capability, fall back to free-form for non-grammar parsers (#1144)

### LINKED ISSUE #1144: bug: oversized tool schema returns HTTP 400 for non-grammar-capable parsers instead of free-form (#558 PR-5 deferred)
## Summary

Under default-on constrained tool-calling (#558 PR-5), an **oversized** tool schema (`> _TOOL_GRAMMAR_MAX_SCHEMA_BYTES` / `> _TOOL_GRAMMAR_MAX_SCHEMA_DEPTH` / `> _TOOL_GRAMMAR_MAX_TOOLS`) is rejected with **HTTP 400** for **any configured `tool_call_parser`** — including parsers that are NOT grammar-capable (`structure_info()` returns `None`). For a non-grammar-capable parser the request would never have been grammar-constrained anyway, so it should fall back to **free-form**, not 400.

## Where

`vllm_mlx/routes/chat.py`: `_tool_grammar_constraint_active` (~line 618) gates on `cfg.tool_call_parser` being set, but does NOT check whether that parser is grammar-capable. `_enforce_tool_grammar_bounds_or_400` then raises 400 on the "active" path for an oversized schema — even when `_maybe_build_tool_grammar_processor` would ultimately return `None` (free-form) because the parser's `structure_info()` is `None`.

## Blast radius

Only two parsers currently implement `structure_info` (grammar-capable): `hermes_tool_parser.py`, `qwen_tool_parser.py` (`abstract_tool_parser.py` is the ABC default → `None`). So any family configured with a different `tool_call_parser` (e.g. deepseek/qwen3coder/harmony-style parsers that parse tool calls but have no `structure_info` yet) hits this. Trigger additionally requires an **oversized** schema (`>64 KiB` / `>32 deep` / `>256 tools`), which is rare in practice. Consequence is a 400 (with a clear message) instead of a free-form response — not a crash, not a data-correctness error, not on the golden path.

## Fix direction

Gate the oversized→400 decision on parser **grammar-capability**, not merely on `tool_call_parser` being set: a non-grammar-capable parser (`structure_info()` is `None`) with an oversized schema should fall back to free-form (it was never going to be constrained), while a grammar-capable parser keeps the #561 400. Needs a cheap capability probe (e.g. a `supports_grammar()`/class-level marker on the parser ABC, or resolving `structure_info` before the 400 check) so the cheap synchronous route gate isn't made heavy.

## Provenance

Surfaced by codex (gpt-5.6-sol) on #558 PR-5 (PR #1143) run-3. The same engine was reviewed engine-clean on run-2 (codex non-determinism across runs). Deferred at merge time as a narrow, rare-trigger, non-golden-path edge; the engine's core auto-path + #561 behavior on grammar-capable parsers (hermes/qwen — all current tool-calling families) is correct and covered.

raullenchai:
## Summary

Under default-on constrained tool-calling (#558 PR-5), an **oversized** tool schema (`> _TOOL_GRAMMAR_MAX_SCHEMA_BYTES` / `> _TOOL_GRAMMAR_MAX_SCHEMA_DEPTH` / `> _TOOL_GRAMMAR_MAX_TOOLS`) was rejected with **HTTP 400** for **any** configured `tool_call_parser` — including parsers that are **not grammar-capable** (their `structure_info()` returns `None`). A non-grammar-capable parser is never decoder-constrained (`_maybe_build_tool_grammar_processor` returns `None` → free-form-then-parse), so an oversized schema for it should fall back to **free-form**, not 400.

`Closes #1144`

## The wrong 400

`vllm_mlx/routes/chat.py::_tool_grammar_constraint_active` decided the constrained path was "active" whenever `cfg.tool_call_parser` was merely **set**. `_enforce_tool_grammar_bounds_or_400` then raised 400 on that "active" path for an oversized schema — even for a parser (e.g. `deepseek_v3`, `llama`, `qwen3_coder_xml`, `glm47`, …) whose `structure_info()` is the ABC default (`None`), which would have produced a free-form response anyway.

## Root cause

`_tool_grammar_constraint_active` never checked parser **grammar-capability** — only that a parser name was configured.

## The fix (capability gate)

- Add `supports_grammar()` + a `SUPPORTS_GRAMMAR` marker on the tool-parser ABC. `supports_grammar()` returns True iff the parser overrides `structure_info` (inferred structurally, so an **out-of-tree** parser that overrides without knowing the marker still gets grammar + #561 enforcement — no silent regression) **OR** explicitly sets `SUPPORTS_GRAMMAR = True` (kept on hermes/qwen for a grep-able discoverability list). Cheap: class-attribute + class-level function-identity check — **no instantiation, no tokenizer**.
- `_tool_grammar_constraint_active` now additionally requires the parser to be grammar-capable via `_tool_parser_supports_grammar(cfg)`: a cached-dict class lookup then `supports_grammar()`. An **unknown** parser name (`KeyError`) → not capable (free-form); any other import/registry failure for a genuinely-configured parser **propagates** (surfaces the bug) rather than silently failing open.

Result: a non-grammar-capable parser + oversized schema → **free-form** (path inactive, no 400, no wasted compile-offload); a grammar-capable parser + oversized schema → **still 400** (#561 preserved).

## Blast radius

Narrow. Only **non-hermes/qwen** parsers **with an oversized schema** change behavior (400 → free-form). #561's grammar-capable 400 for hermes/qwen (all current tool-calling families) is preserved unchanged. Golden path (in-bounds schemas) is untouched.

## Tests

Added to `tests/test_grammar_processor_558.py`:
- `test_supports_grammar_probe_matches_capability` — route probe True for hermes/qwen, False for `deepseek_v3` / unknown / None.
- `test_oversized_schema_still_400_for_grammar_capable_parser` (hermes+qwen × required/auto/None) — **#561 regression guard**: oversized still 400.
- `test_oversized_schema_falls_back_for_non_grammar_parser` — non-grammar + oversized → path inactive, **NOT 400** (free-form).
- `test_normal_schema_free_form_for_non_grammar_parser` — non-grammar + normal schema → not eligible, no 400.
- `test_supports_grammar_marker_declared_for_in_tree_parsers` — in-tree discoverability guard: `SUPPORTS_GRAMMAR` marker itself matches the `structure_info` override for every registered parser.
- `test_supports_grammar_infers_capability_without_marker` — runtime robustness net: an override-without-marker parser is still capable via inference; a plain parser is not.

## Test plan

- [x] `test_supports_grammar_probe_matches_capability` passes
- [x] grammar-capable (hermes/qwen) + oversized schema → still 400 across required/auto/None
- [x] non-grammar-capable (`deepseek_v3`) + oversized schema → free-form (NOT 400)
- [x] non-grammar-capable + normal schema → free-form (not eligible, no 400)
- [x] `SUPPORTS_GRAMMAR` marker declared to match `structure_info` override for all registered parsers
- [x] override-without-marker parser inferred grammar-capable; plain parser not
- [x] `tests/test_grammar_processor_558.py` full file green
- [x] `test_structure_info_hermes_qwen_558.py`, `test_tool_grammar_558.py`, `test_tool_parser_wire_formats.py`, `test_tool_parser_coverage.py`, `test_native_tool_format.py` green (1 network-only tokenizer-download skip unrelated to this change)

### DIFF
--- a/tests/test_grammar_processor_558.py
+++ b/tests/test_grammar_processor_558.py
@@ -794,6 +794,110 @@ def test_reasonable_schema_no_400_on_active_path(monkeypatch):
     assert _enforce_tool_grammar_bounds_or_400(cfg, req) is None
 
 
+# --------------------------------------------------------------------------
+# #1144: the oversized -> 400 decision is gated on parser grammar-CAPABILITY,
+# not merely on ``tool_call_parser`` being set. A grammar-capable parser
+# (hermes/qwen, ``structure_info`` -> wire triple) keeps the #561 400; a
+# non-grammar-capable parser (ABC default ``structure_info`` -> None) was never
+# going to be decoder-constrained, so an oversized schema falls back to FREE-FORM
+# instead of 400.
+# --------------------------------------------------------------------------
+# A registered parser that does NOT override ``structure_info`` (grammar-incapable).
+_NON_GRAMMAR_PARSER = "deepseek_v3"
+
+
+def test_supports_grammar_probe_matches_capability():
+    # The cheap route probe reports True only for grammar-capable parsers.
+    from vllm_mlx.routes.chat import _tool_parser_supports_grammar
+
+    assert _tool_parser_supports_grammar(_CfgStub("hermes")) is True
+    assert _tool_parser_supports_grammar(_CfgStub("qwen")) is True
+    assert _tool_parser_supports_grammar(_CfgStub(_NON_GRAMMAR_PARSER)) is False
+    # Unknown / unset parser name -> not capable (free-form), never raises.
+    assert _tool_parser_supports_grammar(_CfgStub("no_such_parser_xyz")) is False
+    assert _tool_parser_supports_grammar(_CfgStub(None)) is False
+
+
+@pytest.mark.parametrize("parser", ["hermes", "qwen"])
+@pytest.mark.parametrize("choice", ["required", "auto", None])
+def test_oversized_schema_still_400_for_grammar_capable_parser(
+    monkeypatch, parser, choice
+):
+    # #561 regression guard (#1144): a grammar-CAPABLE parser keeps the hard 400
+    # on an oversized schema across every constrainable tool_choice.
+    from fastapi import HTTPException
+
+    from vllm_mlx.routes.chat import _enforce_tool_grammar_bounds_or_400
+
+    monkeypatch.setenv("RAPID_MLX_CONSTRAIN_TOOLS", "1")
+    cfg = _CfgStub(parser)
+    req = _RequestStub(tools=[_oversized_tool()], tool_choice=choice)
+    with pytest.raises(HTTPException) as ei:
+        _enforce_tool_grammar_bounds_or_400(cfg, req)
+    assert ei.value.status_code == 400
+    assert "grammar-compile bounds" in str(ei.value.detail)
+
+
+@pytest.mark.parametrize("choice", ["required", "auto", None])
+def test_oversized_schema_falls_back_for_non_grammar_parser(monkeypatch, choice):
+    # #1144 core fix: a non-grammar-capable parser (structure_info -> None) with
+    # an oversized schema must NOT 400 — it was never going to be constrained, so
+    # it falls back to free-form exactly like the pre-#558 behavior.
+    from vllm_mlx.routes.chat import (
+        _enforce_tool_grammar_bounds_or_400,
+        _tool_grammar_constraint_active,
+    )
+
+    monkeypatch.setenv("RAPID_MLX_CONSTRAIN_TOOLS", "1")
+    cfg = _CfgStub(_NON_GRAMMAR_PARSER)
+    req = _RequestStub(tools=[_oversized_tool()], tool_choice=choice)
+    # Path is inactive for a non-grammar parser, so no 400 (free-form fallback).
+    assert _tool_grammar_constraint_active(cfg, req) is False
+    assert _enforce_tool_grammar_bounds_or_400(cfg, req) is None
+
+
+@pytest.mark.parametrize("choice", ["required", "auto", None])
+def test_normal_schema_free_form_for_non_grammar_parser(monkeypatch, choice):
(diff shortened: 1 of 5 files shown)


raullenchai:
## pr_validate convergence — merging (#1144 fix)

Engine-clean on all mechanical gates; `mergeable=CLEAN`; version untouched (0.10.15); diff clear of spec_decode/speculative and of #1142's site.

**codex convergence over 4 runs (classic non-determinism — one different single blocking per run):**
- run1 (2 findings) **fixed** — added the runtime inference net so out-of-tree parsers that override `structure_info` aren't silently regressed; narrowed the `except` to `KeyError`.
- run2 (test tautology) **fixed** — split into a marker-guard test + an inference-net test.
- run3 **OVERRIDDEN (justified)** — codex demanded gating on the *runtime* `structure_info()` result, i.e. instantiate parser+tokenizer on the synchronous per-request route gate. That is exactly the **heavy probe #1144 forbids**. The edge it cites (hermes on a non-Qwen tokenizer where `structure_info()` runtime-gates to `None` + an oversized schema) is **pre-existing #561 behavior this PR does not change**, and #1144 says to *preserve* the 400 for grammar-capable families (escape hatch: `RAPID_MLX_CONSTRAIN_TOOLS=0`). Design intent documented in the `supports_grammar` docstring.
- run4 (marker test iterated ALL registered parsers → would false-fail a supported out-of-tree parser) **fixed in d1a390a4** — the test now `continue`s on `not cls.__module__.startswith('vllm_mlx.tool_parsers')`, scoping the marker assertion to in-tree classes exactly as codex asked. The latest `codex-review.md` re-flagging this is a **pre-fix stale run** (verified against the committed test source).

**full_unit:** the 4 confirmed pre-existing/environmental failures only (3× `prometheus_client` missing, 1× deepseek-r1 offline tokenizer) — none touch this diff. targeted_tests 471 passed; 96/96 local grammar tests.

Merging as the #1144 fix (oversized-schema 400 gated on parser grammar-capability; non-grammar parsers fall back to free-form; #561 preserved for hermes/qwen).

PR #1153, opened 2026-07-21
feat(tools): harmony structure_info — grammar-constrained tool-calling for gpt-oss (#558)

raullenchai:
## What & why

Extends #558 grammar-constrained tool-calling from **{qwen, hermes}** to **+gpt-oss (harmony)**. Today `HarmonyToolParser.structure_info()` inherits the ABC default (`None`), so gpt-oss tool calls fall back to free-form-then-parse — no structural guarantee that the emitted `arguments` satisfy the tool's JSON Schema. This PR gives gpt-oss the **same genuine llguidance decoder-level constraint** qwen/hermes already get, for the modes where it is sound.

## The harmony wire (verified, not assumed)

The model emits (OpenAI harmony spec):

```
<|channel|>commentary to=functions.NAME <|constrain|>json<|message|>{args}<|call|>
```

The tool NAME lives in the **header** (`to=functions.NAME`), not the body; the body is a pure JSON object constrained by `%json <schema>`. I pinned every byte against the **real gpt-oss tokenizer** (`mlx-community/gpt-oss-20b-MXFP4-Q8`, cached) and the **openai-harmony** library's own renderer/parser:

- `are_single_special_tokens()` → **True** for all four control tokens: `<|channel|>` (200005), `<|constrain|>` (200003), `<|message|>` (200008), `<|call|>` (200012). So they are declared `sentinels` (Lark special-token refs); the header text between them (`commentary to=functions.NAME `, `json`) is byte-literal.
- The header **round-trips byte-exact** on the real tokenizer.
- The **space before `<|constrain|>` is mandatory** — openai-harmony's own parser rejects the no-space header (`unexpected tokens remaining in message header`). The grammar preserves it, and a no-space header is rejected (test).

**Port source:** openai-harmony canonical rendering + the XGrammar-2 GPT-OSS structural tag. Explicitly **NOT** SGLang's `gpt_oss_detector` (it `raise NotImplementedError` for `structure_info`).

## Honest scope: required/named constrained, auto stays free-form

Harmony's only single-special-token trigger is `<|channel|>` — but that token **also** prefixes `final` and `analysis` (non-tool) responses. The #558 builder is a *"free byte prefix + single-special-token trigger"* model: the free prefix is a **byte** regex that cannot contain a special token, so once the model emits `<|channel|>` it is committed to the tool tag. On `auto` that **forces a tool call** — turning `auto` into `required`, the #1 regression #558 guards against.

Verified offline: under an `auto` harmony grammar a plain `<|channel|>final<|message|>...` reply is **rejected at the first token**. Harmony's real tool-call trigger is a **text** pattern (`to=functions.`), which the builder can't express yet (text-trigger support is design §7 open-Q1, deferred). The reasoning-tolerant prefix (PR-4) doesn't help either — `resolve_reasoning_sentinels("harmony")` is `()` (harmony reasoning is channel-based, not `<think>`).

**Resolution:** a new `ToolParser.TOOL_GRAMMAR_AUTO_SAFE` flag (defaults `True`). Harmony sets it `False`; `build_tool_grammar` then declines the **auto** path (free-form fallback — fully non-regressive vs today) while still building the grammar for **required** and **named**, where driving the model straight into a schema-valid tool call is exactly what the caller asked for. hermes/qwen (`<tool_call>` is an unambiguous single-token trigger) keep `True` and are untouched.

A fully-sound harmony **auto** grammar needs a harmony-specific multi-channel grammar (model every channel explicitly) — a distinct design effort, not this "port structure_info" PR. Flagged as follow-up rather than invented here (charter: never invent a tool-call grammar).

## Composes with #1149 (parser-capability-gated oversized-schema 400)

Rebased on top of #1149. Because harmony now overrides `structure_info`, it declares `SUPPORTS_GRAMMAR = True` (the #1149 in-tree discoverability guard asserts the marker matches the override). But that would make `_tool_grammar_constraint_active` True for harmony on **every** mode — so a harmony **auto** request with an oversized schema would newly 400 (was free-form), and the route would pay a grammar-build offload that only declines. Fix: a new `_tool_parser_auto_safe` route probe makes `_tool_grammar_constraint_active` report **harmony + auto as inactive** — oversized schema stays free-form (no spurious #561 400), offload skipped. `required`/named keep the #561 400. Every auto-safe family (hermes/qwen) is unchanged. Covered by new routing tests.

## Proof on real hardware (gpt-oss-20b-MXFP4-Q8, non-operator port 8133)

**Offline matcher enforcement** (`LLMatcher.consume_tokens`, real tokenizer):

| input | result |
|---|---|
| valid call `{"city":"Paris"}` | accepted, terminal ✅ |
| valid enum `{"city":"P","unit":"celsius"}` | accepted, terminal ✅ |
| off-schema `{"city": 4` | rejected mid-stream ✅ |
| bad enum `"unit":"kelvin` | rejected mid-stream ✅ |
| hallucinated `to=functions.get_stock` | rejected ✅ |
| missing space before `<|constrain|>` | rejected ✅ |

`GrammarLogitsProcessor.is_broken()` → **False** for required + named (grammar compiles clean). `build_tool_grammar(auto)` → **None** (declined).

**Live end-to-end** (`/v1/models` shows `tool_call_parser: harmony`, no "Guided generation disabled"):

- `tool_choice=required` → `get_weather({"city":"Paris","unit":"fahrenheit"})` — schema-valid.
- **Adversarial enum**: schema `unit ∈ {celsius, fahrenheit}`, user demanded *"unit must be kelvin"* → emitted `unit="celsius"` — **stayed in-enum** (grammar actively constraining, mirrors the 0.10.15 dogfood).
- `tool_choice=auto` + "write a poem" → **zero tool calls**, wrote a poem (zero-call invariant holds).
- `tool_choice=auto` + weather ask → free-form call still works (auto not blocked).

## Changes

- `vllm_mlx/tool_parsers/harmony_tool_parser.py` — `structure_info()` + `SUPPORTS_GRAMMAR = True` + `TOOL_GRAMMAR_AUTO_SAFE = False` + `_GRAMMAR_SENTINELS`.
- `vllm_mlx/tool_parsers/abstract_tool_parser.py` — `TOOL_GRAMMAR_AUTO_SAFE: bool = True` (ABC default; non-breaking).
- `vllm_mlx/api/tool_grammar.py` — `build_tool_grammar` auto-soundness gate keyed on the flag.
- `vllm_mlx/routes/chat.py` — `_tool_parser_auto_safe` probe + `_tool_grammar_constraint_active` composes it so harmony+auto is inactive (free-form), keeping the #1149 oversized-schema 400 to required/named.
- `tests/test_harmony_tool_grammar_558.py` — 13 offline tests (mirrors `test_tool_grammar_558.py`).
- `tests/test_grammar_processor_558.py` — 5 harmony-auto routing tests.

Does not touch `spec_decode/**` or `speculative/**`.

## Test plan

- [x] All four harmony control tokens confirmed single special tokens on the real gpt-oss tokenizer.
- [x] Header byte-alignment verified against openai-harmony renderer/parser (incl. mandatory pre-`<|constrain|>` space).
- [x] Offline matcher: valid call + valid enum accepted & terminal; off-schema / bad-enum / hallucinated-name / missing-space rejected mid-stream.
- [x] `GrammarLogitsProcessor.is_broken()` == False for required + named; `build_tool_grammar(auto)` == None (harmony opts out).
- [x] Named `tool_choice` narrows to the single tool (other tool rejected).
- [x] #1149 composition: harmony declares `SUPPORTS_GRAMMAR = True` (in-tree marker guard passes); `_tool_grammar_constraint_active(harmony, auto)` == False and `(harmony, required/named)` == True; oversized harmony+auto → no 400, oversized harmony+required → 400.
- [x] Live pilot on gpt-oss-20b: required valid call; adversarial enum stays in-enum; auto poem = zero calls; auto weather = free-form call works.
- [x] New suite `tests/test_harmony_tool_grammar_558.py` green (13 passed).
- [x] Regression: `test_tool_grammar_558.py` + `test_harmony_parsers.py` + `test_harmony_stop_final_channel_only.py` + `test_tool_parser_wire_formats.py` pass offline (only the pre-existing network-gated `test_deepseek_r1_prefilled_think_template_is_tolerated` fails — identical on origin/main, needs an uncached 32B model).
- [x] `ruff check` clean on all changed files.
- [x] `TOOL_GRAMMAR_AUTO_SAFE` defaults True → hermes/qwen auto path unchanged (verified: an auto-safe stub still builds an auto grammar).

## codex review disposition (round 1)

- **[FIXED] lltok fixture skip → assert** — once the real cached tokenizer loads, a `None` from `build_lltokenizer` is a resolver regression, not a skip; the fixture now asserts non-`None` so the enforcement suite can't go green while harmony constraint is silently disabled.
- **[FIXED] tok fixture OSError over/under-catch** — `local_files_only=True` on an uncached box raises a bare `OSError`, not `LocalEntryNotFoundError`. Replaced the except-list with a `try_to_load_from_cache` preflight (no network); load-time corruption errors now propagate.
- **[OVERRIDE] chat.py:693 "required/named oversized→400 when structure_info() could be None"** — this is the **deliberate #1144/#1149 design**: `supports_grammar()` is class-level *by mandate* ("DO NOT invent a heavy probe"), and a grammar-capable family whose `structure_info()` is tokenizer-gated to `None` keeping the #561 400 is exactly what the #1149 docstring documents (identical to hermes-on-a-non-Qwen-tokenizer; remedy `RAPID_MLX_CONSTRAIN_TOOLS=0`). In practice harmony only runs on gpt-oss tokenizers where the four control tokens *are* single special tokens, so `structure_info()` never returns `None`. The genuinely-new case (auto) IS handled — `_tool_parser_auto_safe` makes harmony+auto inactive. Adopting codex's "runtime per-request availability" fix would require the exact heavy per-request tokenizer probe #1144 forbids.

## codex review disposition (round 2)

codex re-ran after the round-1 fixes and returned **1 blocking + 1 nit** (down from 3 blocking):

- **[OVERRIDE — re-raise of round 1] chat.py:693 / harmony_tool_parser.py:160 "required/named classified constrained from the parser class → oversized schema 400 even when structure_info() returns None"** — same finding as round-1 #3, re-surfaced verbatim (codex oscillation on a refuted point). Disposition unchanged and now hardened: class-level capability is the mandated #1144/#1149 contract; the proposed remedy ("base bounds enforcement on cached runtime grammar availability, or reject incompatible configs before serving") is precisely the per-request heavy probe #1144 forbids. On the only tokenizers harmony ever loads (gpt-oss), all four control tokens are single special tokens, so `structure_info()` never returns `None` in practice; the sole genuinely-new mode (auto) is already routed inactive via `_tool_parser_auto_safe`.
- **[DEFER — nit] test_harmony_tool_grammar_558.py:179 "real-tokenizer tests skip when the pinned tokenizer is absent; add a small checked-in tokenizer fixture for always-on CI coverage"** — convention-consistent with the entire repo's #558 enforcement suite (`test_tool_grammar_558.py` gates the same way on the real Qwen tokenizer). A synthetic small tokenizer can't reproduce the load-bearing property under test (the four gpt-oss control tokens being *single special tokens* with byte-exact header round-trip), so it would assert against a fiction. The offline pure-Python routing/flag tests (`TOOL_GRAMMAR_AUTO_SAFE` gating, build_tool_grammar decline) DO run unconditionally in CI; only the real-grammar enforcement layer is cache-gated, matching precedent. Left as a follow-up if the repo later adopts a shared checked-in tokenizer fixture convention.

### DIFF
--- a/tests/test_grammar_processor_558.py
+++ b/tests/test_grammar_processor_558.py
@@ -876,6 +876,82 @@ def test_normal_schema_free_form_for_non_grammar_parser(monkeypatch, choice):
     assert _enforce_tool_grammar_bounds_or_400(cfg, req) is None
 
 
+# --------------------------------------------------------------------------
+# Harmony AUTO opt-out (#558 +gpt-oss) — auto path stays free-form even though
+# harmony is now grammar-CAPABLE (structure_info overridden), because its
+# <|channel|> trigger is shared with non-tool responses (TOOL_GRAMMAR_AUTO_SAFE
+# = False). required/named remain constrained.
+# --------------------------------------------------------------------------
+def test_auto_safe_probe_matches_capability():
+    from vllm_mlx.routes.chat import _tool_parser_auto_safe
+
+    # Single-special-token-trigger families are auto-safe.
+    assert _tool_parser_auto_safe(_CfgStub("hermes")) is True
+    assert _tool_parser_auto_safe(_CfgStub("qwen")) is True
+    # Harmony opts out (shared <|channel|> trigger).
+    assert _tool_parser_auto_safe(_CfgStub("harmony")) is False
+    # Unknown / unset -> permissive default (not constrained anyway).
+    assert _tool_parser_auto_safe(_CfgStub("no_such_parser_xyz")) is True
+    assert _tool_parser_auto_safe(_CfgStub(None)) is True
+
+
+@pytest.mark.parametrize("choice", ["auto", None])
+def test_harmony_auto_path_inactive_free_form(choice):
+    # Harmony is grammar-capable, but on AUTO the constraint path is INACTIVE
+    # (declines the grammar) — so it composes like a free-form path: no grammar,
+    # and (below) no #561 400 on an oversized schema.
+    from vllm_mlx.routes.chat import _tool_grammar_constraint_active
+
+    cfg = _CfgStub("harmony")
+    ok = _FunctionTool(
+        "get_weather",
+        parameters={"type": "object", "properties": {"city": {"type": "string"}}},
+    )
+    assert _tool_grammar_constraint_active(cfg, _RequestStub([ok], choice)) is False
+
+
+def test_harmony_required_and_named_paths_active():
+    # required/named ARE constrained for harmony (a forced call is what the
+    # caller asked for).
+    from vllm_mlx.routes.chat import _tool_grammar_constraint_active
+
+    cfg = _CfgStub("harmony")
+    ok = _FunctionTool(
+        "get_weather",
+        parameters={"type": "object", "properties": {"city": {"type": "string"}}},
+    )
+    assert _tool_grammar_constraint_active(cfg, _RequestStub([ok], "required")) is True
+    named = {"type": "function", "function": {"name": "get_weather"}}
+    assert _tool_grammar_constraint_active(cfg, _RequestStub([ok], named)) is True
+
+
+@pytest.mark.parametrize("choice", ["auto", None])
+def test_harmony_auto_oversized_schema_no_400(monkeypatch, choice):
+    # AUTO non-regression: harmony gaining required/named grammar support must
+    # NOT start 400-ing oversized schemas on the auto path (it declines the auto
+    # grammar, so an oversized schema stays free-form — exactly as before).
+    from vllm_mlx.routes.chat import _enforce_tool_grammar_bounds_or_400
+
+    monkeypatch.setenv("RAPID_MLX_CONSTRAIN_TOOLS", "1")
+    cfg = _CfgStub("harmony")
+    req = _RequestStub(tools=[_oversized_tool()], tool_choice=choice)
+    assert _enforce_tool_grammar_bounds_or_400(cfg, req) is None
+
+
+def test_harmony_required_oversized_schema_400(monkeypatch):
+    # But required DOES keep the #561 hard 400 for harmony (it is constrained).
+    from fastapi import HTTPException
+
+    from vllm_mlx.routes.chat import _enforce_tool_grammar_bounds_or_400
+
+    monkeypatch.setenv("RAPID_MLX_CONSTRAIN_TOOLS", "1")
+    cfg = _CfgStub("harmony")
+    req = _RequestStub(tools=[_oversized_tool()], tool_choice="required")
+    with pytest.raises(HTTPException) as ei:
+        _enforce_tool_grammar_bounds_or_400(cfg, req)
+    assert ei.value.status_code == 400
+
+
 def test_supports_grammar_marker_declared_for_in_tree_parsers():
     # In-tree DISCOVERABILITY guard (#1149 codex): the explicit
     # ``SUPPORTS_GRAMMAR`` marker itself must match the ``structure_info``
--- a/tests/test_harmony_tool_grammar_558.py
+++ b/tests/test_harmony_tool_grammar_558.py
@@ -1,0 +1,400 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Offline tests for harmony (gpt-oss) grammar-constrained tool calling (#558).
+
+Extends the #558 constraint coverage from {qwen, hermes} to +gpt-oss. These
+validate ``HarmonyToolParser.structure_info()`` and its composition with the
+grammar builder WITHOUT a decode loop:
+
+  * the AUTO-path opt-out (``TOOL_GRAMMAR_AUTO_SAFE``): harmony's only single-
+    special-token trigger (``<|channel|>``) is SHARED with non-tool responses
+    (``final``/``analysis``), so the auto grammar would force a call — harmony
+    declines auto (free-form fallback) while constraining ``required``/named;
+  * the harmony wire triple's structural invariants (begin starts with the
+    ``<|channel|>`` trigger; the four control tokens are declared sentinels);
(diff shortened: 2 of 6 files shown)


raullenchai:
Merged with author-override on the single codex BLOCKING finding (chat.py:693).

**Same argument as #1149**: grammar-capability is deliberately judged at the **class level** (cheap, cached `supports_grammar()`) rather than a per-request runtime tokenizer probe. `HarmonyToolParser`'s tokenizer structurally always carries the harmony control tokens (channel/constrain/message/call), so `structure_info()` never returns `None` for any real gpt-oss harmony tokenizer — the None-path codex describes is unreachable in practice. The theoretical edge (harmony + non-standard tokenizer + oversized schema + required → 400 instead of free-form) is tracked under #1144.

**full_unit's 2 fails were environment artifacts, not regressions** (#1153 touches no signal code): (1) `test_subprocess_sighup_default_disposition...` failed only because the validate run was launched under `nohup`, whose inherited `SIG_IGN` disposition pollutes the SIGHUP default-disposition assertion; (2) `test_batched_faster_than_sequential` is a known timing flake. Both **PASS on a clean foreground rerun** (2 passed in 7.09s).

Hardware pilot on gpt-oss-20b-MXFP4-Q8 confirmed: required/named schema-valid, adversarial-enum stayed in-enum, auto → zero tool-call + free-form preserved (the #558 auto-must-allow-zero invariant holds via `TOOL_GRAMMAR_AUTO_SAFE=False`). NIT (checked-in tokenizer fixture) deferred.

PR #1160, opened 2026-07-21
feat(tools): grammar-constrain Qwen3-Coder XML tool-calls (#558 E3, PR-A)

raullenchai:
## Why

Grammar-constrained tool calling (#558) currently only covers JSON-object arg
families (hermes/qwen/harmony). The Qwen3-Coder family serializes tool arguments
as **XML**, so today it gets no decoder-level guarantee — a low-quant model can
emit an off-schema or malformed call that only fails at parse time. This PR
closes that gap for Qwen3-Coder **because** it is the one XML family with a
locally-available checkpoint, so the XML-args machinery can be proven
end-to-end on real hardware before extending to the other XML families.

## What

Extends the #558 grammar builder beyond the JSON-object body
(`%json <schema>`) to the **XML-argument** wire the Qwen3-Coder family emits.
This is **PR-A** of the E1/E3 rollout (serial; E1 section-wrapper + the other
XML families follow in later PRs since they all edit `tool_grammar.py`).

Constrained wire (per-property XML args):

```
<tool_call>
<function=NAME>
<parameter=KEY>
VALUE
</parameter>
</function>
</tool_call>
```

## Copied from upstream (not invented)

The wire and per-property XML value rendering are copied from the **XGrammar-2
`qwen_3_coder` built-in** (`style="qwen_xml"`) and verified **byte-for-byte**
against the Qwen3-Coder chat template's `message.tool_calls` rendering
(`args_value if string else args_value | tojson`) and our existing parsing-side
`qwen3coder_tool_parser` — so the generation grammar and the parser agree.

## Changes

**`vllm_mlx/api/tool_grammar.py`**
- `StructureInfo` gains `arg_format: str = "json"`. The default reproduces
  today's `%json` body **byte-for-byte** — zero regression for the existing
  JSON-body families (hermes / qwen / harmony).
- `arg_format="qwen_xml"` renders one `<parameter=key>…</parameter>` block per
  schema property with the value grammar-typed per its JSON-Schema type:
  enum → literal alternation, integer → `INTVAL`, number → `NUMVAL`,
  boolean → `"true"|"false"`, object/array → inline `%json <subschema>`,
  string/unknown → `STRVAL` (`[^<]*`). Required props mandatory (declared
  order); optionals wrapped `( … )?`. Shared value terminals emitted once.

**`vllm_mlx/tool_parsers/qwen3coder_tool_parser.py`**
- `Qwen3CoderToolParser.structure_info()` emits the wire triple with
  `arg_format="qwen_xml"`, gated on the tokenizer proving
  `<tool_call>`/`</tool_call>` are single special tokens (opts out to free-form
  otherwise), mirroring hermes/qwen. `SUPPORTS_GRAMMAR` /
  `TOOL_GRAMMAR_AUTO_SAFE` set for route-gate discoverability.

## Testing

`tests/test_tool_grammar_e3_qwen3coder_558.py` (19 tests):
- pure-Python builder tests (parameter blocks, shared terminals, enum
  alternation, typed values, optional wrapping, no-arg body, **JSON-path
  regression guard**);
- llguidance **enforcement** on the pinned Qwen tokenizer: a valid XML call is
  accepted-in-full and reaches a terminal state; missing close / JSON body /
  undeclared key / bad enum / missing required / hallucinated name /
  non-integer are all rejected mid-stream.

Existing 558 suites all pass unchanged (JSON path byte-identical).

## E2E (real model)

Served `mlx-community/Qwen3-Coder-Next-4bit` with `--tool-call-parser
qwen3_coder_xml`:
- `tool_choice=required` produced valid schema-matching XML args parsed to
  `{"city":"Paris","unit":"celsius"}`;
- **3 parallel** tool calls emitted+parsed correctly;
- **constraint provably active**: a nonsense required enum `["korlan","vprix"]`
  on a normal weather question was *forced* to `mode="korlan"` — a value the
  model would never emit unconstrained;
- rough speed ~10.6 tok/s wall (80B-A3B MoE @ 4-bit, incl. prefill+HTTP).

## Known limitation

A string parameter value containing a literal `<` cannot be represented under
the conservative `STRVAL` (`[^<]*`) bound (documented inline); `any_order`
parameter ordering is a deliberate later-cut item (the constraint guides the
model to declared order, which stays schema-valid).

### DIFF
--- a/tests/test_tool_grammar_e3_qwen3coder_558.py
+++ b/tests/test_tool_grammar_e3_qwen3coder_558.py
@@ -1,0 +1,428 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Offline tests for the Qwen3-Coder XML-args grammar body (#558 E3).
+
+E3 extends the #558 grammar builder beyond the JSON-object body
+(``%json <schema>``, used by hermes/qwen/harmony/deepseek-JSON) to the
+XML-argument wire the Qwen3-Coder family emits::
+
+    <tool_call>
+    <function=NAME>
+    <parameter=KEY>
+    VALUE
+    </parameter>
+    </function>
+    </tool_call>
+
+The wire and the per-property XML value rendering are COPIED from the XGrammar
+``qwen_3_coder`` built-in (``style="qwen_xml"``) and verified byte-for-byte
+against the Qwen3-Coder chat template's ``message.tool_calls`` rendering and our
+own ``qwen3coder_tool_parser`` (generation grammar and parser agree).
+
+Two layers:
+
+  * pure-Python BUILDER tests (never skip): ``build_tool_lark`` emits the
+    ``<parameter=...>`` blocks, the shared ``STRVAL``/``INTVAL``/``NUMVAL``
+    terminals, enum alternation, ``%json`` for object/array values, and wraps
+    optional properties ``( ... )?`` — while the JSON path stays byte-identical;
+  * grammar ENFORCEMENT (skip only on genuine tokenizer/llguidance
+    unavailability): a well-formed Qwen3-Coder XML call is ACCEPTED in full and
+    reaches an accepting/terminal state, while a hallucinated function name, an
+    off-schema/undeclared parameter, a bad enum value, a non-integer for an int
+    param, a JSON body instead of XML, and a missing required parameter are all
+    REJECTED mid-stream.
+
+The enforcement tests use the same pinned Qwen tokenizer as
+``tests/test_tool_grammar_558.py`` — its ``<tool_call>``/``</tool_call>`` are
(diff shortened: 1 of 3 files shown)


raullenchai:
Superseded by #1170 (merged as 4a43396b). #1170 is the canonical E3 (Qwen3-Coder XML grammar-constrained tool-calls): it branched on current main (incl. #1164 forced-tool fix), solved gap #1 (`<` in string values, via llguidance's `[lazy]` idiom), added a strict-allowlist schema-representability guard through 4 codex rounds, and ships with real-tokenizer enforcement tests. This branch was based on a pre-#1164 main + CONFLICTING and did not solve gap #1. Closing to avoid a duplicate E3.

PR #1164, opened 2026-07-21
fix(tools): forced non-reasoning tool-grammar starts at the trigger (#558)

raullenchai:
## Problem

`#558` constrained tool-calling promised schema-valid arguments for forced tool calls (`tool_choice=required`/named), but on real hardware **hermes/qwen leaked empty `{}` args ~40-60% of the time**. Root-caused by decode trace + grammar dump on Hermes-3-Llama-8B (constrain on):

`build_tool_lark` gave every tag a leading `TAG_TEXT: /(.|\n)*/` **unbounded free-text prefix**. That prefix belongs to the AUTO grammar (a model may decline to call), but it was reused for **forced** modes too — so a weak-tool-prior model (Hermes-on-Llama natively emits `get_weather --city "Paris"` CLI / prose) fills the prefix and **never reaches the `<tool_call>` trigger before `max_tokens`**. The parser extracts nothing and the route synthesises an empty-arg call. harmony/gpt-oss escaped only because its prior emits the trigger immediately.

The grammar was correctly built + masking (no desync); the defect was the grammar **shape**, not llguidance, not the forced-prefix path, not special-token encoding.

## Fix

`build_tool_lark`, **forced + non-reasoning only**; **AUTO and forced + reasoning are unchanged vs `main`**:

- **Forced + non-reasoning (the reported leak):** the **first** forced call starts **directly at the `<tool_call>` trigger** (no free-text prefix at all — this is what forces the mandatory call), and repeated parallel calls are separated by a whitespace-only run — `start: (tag_0 | tag_1) (SEP (tag_0 | tag_1))* tag_end`, `SEP: /[ \t\r\n]*/`. `SEP` admits the real `</tool_call>\n<tool_call>` gap between parallel forced calls (**codex round-3**: dropping the prefix entirely wrongly forced calls to be *adjacent*); it is **unbounded** whitespace because the first call is already forced at token 0, so inter-call whitespace cannot defer anything (**codex round-5** nit: capping it would wrongly reject a real `\n\n` gap while `tag_end: TAG_TEXT` allows unbounded trailing text anyway). Mirrors vLLM `xgrammar` / SGLang forced-tool constraint (forced ⇒ constrain from the trigger).

- **Forced + reasoning (opts out — honest interim):** `build_tool_grammar` returns **`None`** (gated on the *normalized* reasoning pair), so the route forces the call via the pre-#558 `forced_assistant_prefix` injection. Neither static grammar shape is correct, so `None` is the honest interim (**codex round-5**): the shared reasoning grammar (`lead: opened? bal_prefix`) keeps an **unbounded free prefix** → a weak-tool-prior reasoning model can still exhaust `max_tokens` and reproduce the `{}` leak; and forcing the trigger at token 0 leaves the prompt-level `<think>` unclosed, so with `enable_thinking=True` the qwen3 reasoning parser buries the **entire** output — tool call included — in `reasoning_content` (`content=None`), **losing the call** (verified against the real `qwen3_parser.py` no-`</think>` branch). A **correct bounded forced-reasoning grammar** needs the prompt's reasoning-open (prefill) state — a prefilled `<think>` must be allowed to close before the call, while a non-prefilled model must be forbidden that same production or it defers the call — and that signal is **not available at grammar-build time** (grammar built before the prompt is rendered; runtime-gating on `</think>` was already rejected in-tree as a footgun that a no-`</think>` model defeats). This is an **architectural decision (prefill-state plumbing)** left to a follow-up. codex oscillated here across rounds (r4: *don't* return `None`, build a bounded grammar → r5: *do* return `None` until a prefill-aware grammar exists), which is itself the evidence that no static grammar resolves it.

- **AUTO (both):** unchanged — keeps `lead: opened? bal_prefix` (may decline, so an unbounded / prefill-tolerant prefix is safe).

## Verification

- **Real machine** (Hermes-3-Llama-8B = non-reasoning / `reasoning_parser=None`, `RAPID_MLX_CONSTRAIN_TOOLS=1`): `required` + `named` across easy + blind-spot schemas (nested `$ref`, `minItems`) → **15/15 valid, 0 empty** (was ~9/15). The `SEP` variant is *stricter* than the tested shape (first call now sits at the trigger with no prefix), and grammar-level enforcement is proven on the real tokenizer + `LLMatcher` (prose-before-call rejected at token 0; a valid call accepted + terminal).
- **harmony/gpt-oss no-regression**: `named` × 3 schemas → **12/12 valid, 0 invalid**.
- **Tests**: `#558` grammar suite **46 passed** + 5 adjacent 558 files **201 passed**; `ruff check` + `ruff format` clean; `full_unit` **13365 passed / 0 failed**. Includes: bare prose before a forced call rejected (`accepted==0`); the **round-3 regression guard** (two `</tool_call>\n<tool_call>` newline-separated forced calls accepted in full); forced + reasoning **opts out** (`build_tool_grammar` → `None`), gated on the *normalized* pair so a single/malformed sentinel still constrains; AUTO still accepts prefilled-`<think>` streams and prose-then-call.

## Scope

Fixes the reported bug: **non-reasoning** forced calls on the 3 grammar-capable parser families (harmony / hermes / qwen). AUTO is byte-unchanged vs `main`; the non-grammar-capable families (gemma4, qwen3_coder_xml = Qwen3.6-27B, deepseek, llama, …) remain free-form. **Forced + reasoning constrained decoding is a tracked follow-up** — it requires prompt-prefill-state plumbing into the grammar builder, an architectural change worth deciding deliberately rather than shipping a static grammar that codex correctly showed either leaks `{}` or drops a reasoning model's call (until then forced + reasoning uses the proven pre-#558 `forced_assistant_prefix` forcing).

### DIFF
--- a/tests/test_tool_grammar_558.py
+++ b/tests/test_tool_grammar_558.py
@@ -163,8 +163,9 @@ def test_build_tool_grammar_named_choice_narrows_to_requested_tool():
     # Only the requested tool's tag survives — get_time is not reachable.
     assert "get_weather" in grammar
     assert "get_time" not in grammar
-    # Single forced alternative (named => exactly one tag, forced).
-    assert "start: (tag_0)+ tag_end" in grammar
+    # Single forced alternative (named => one tag, forced): the first call sits at
+    # the trigger, repeats separated by the unbounded whitespace-only ``SEP``.
+    assert "start: (tag_0) (SEP (tag_0))* tag_end" in grammar
 
 
 @_requires_llguidance
@@ -238,10 +239,12 @@ def test_lark_quantifier_tracks_tool_choice():
     from vllm_mlx.api.tool_grammar import build_tool_lark
 
     infos = [_hermes_structure_info()(t["name"]) for t in TOOLS]
-    # auto -> may emit zero calls -> (...)*
+    # auto -> may emit zero calls -> (...)* (keeps the free prefix; unchanged).
     assert "start: (tag_0 | tag_1)* tag_end" in build_tool_lark(TOOLS, "auto", infos)
-    # required (and any non-auto choice) -> at least one call -> (...)+
-    assert "start: (tag_0 | tag_1)+ tag_end" in build_tool_lark(
+    # required (non-reasoning) -> at least one call, the FIRST at the trigger and
+    # repeats separated by the unbounded whitespace-only ``SEP`` (#558 forced-leak fix replaced the
+    # old ``(...)+`` free-prefix shape).
+    assert "start: (tag_0 | tag_1) (SEP (tag_0 | tag_1))* tag_end" in build_tool_lark(
         TOOLS, "required", infos
     )
 
@@ -278,7 +281,7 @@ def test_named_choice_narrows_to_single_forced_tag():
     only = [TOOLS[0]]  # caller pre-filtered to the requested function
     info = [_hermes_structure_info()(only[0]["name"])]
     lark = build_tool_lark(only, "get_weather", info)
-    assert "start: (tag_0)+ tag_end" in lark
+    assert "start: (tag_0) (SEP (tag_0))* tag_end" in lark
     assert "tag_1" not in lark  # no other tool's alternative present
     assert "get_time" not in lark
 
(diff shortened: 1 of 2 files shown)


raullenchai:
## Status: reported bug fixed + proven; forced+reasoning is an architectural fork awaiting a decision

**`pr_validate` (commit `5cee8552`):** fetch / test_plan / cl_description / supply_chain / test_env / **lint** / targeted_tests (75) all PASS.

- **`full_unit`**: the 5 failures are all in `tests/test_serve_listen_fd.py` — **unrelated to this diff** (this PR touches only `tool_grammar.py` + its test). The file passes **21/21 in isolation**; it's full-suite test-pollution that surfaced after rebasing past #1167 (request-logging middleware), not this change.
- **`codex`**: 1 nit (SEP "bounded" wording — fixed) + 1 blocking on forced+reasoning.

### The blocking is an architectural fork, not a fixable nit

codex has **oscillated across three rounds** on how forced+reasoning should behave:

| round | reviewing | demand |
|---|---|---|
| r4 | `return None` | **don't** return None — it removes `%json` arg constraint; build a bounded grammar |
| r5 | shared `bal_prefix` grammar | `bal_prefix` keeps an unbounded prefix → still leaks `{}`; **return None** until a prefill-aware grammar exists |
| r6 | `return None` (per r5) | **don't** return None — it removes `%json` arg constraint (≡ r4) |

Both static shapes are correctly rejected, in opposite directions, because **neither is right**: a prompt-prefilled `<think>` model must be *allowed* an unbounded `reasoning </think>` lead before the trigger, while a non-prefilled model must be *forbidden* that same production or it defers the forced call. The two require opposite grammars for the same construct, so a correct bounded forced-reasoning grammar needs the prompt's **prefill state** — which isn't available at grammar-build time (grammar is built before the prompt is rendered; a runtime `</think>`-gate was already rejected in-tree as a footgun a no-`</think>` model defeats).

### What this PR does / doesn't do

- **Fixes the reported bug**: forced **non-reasoning** `{}` leak → Hermes-3-Llama-8B `required`+`named` across easy + blind-spot schemas is **15/15 valid, 0 empty** (was ~9/15); harmony `12/12`.
- **forced + reasoning**: opts out to the pre-#558 `forced_assistant_prefix` forcing (the shipped mechanism before #558's grammar) — **no regression vs pre-#558**, and strictly better than the leaky `bal_prefix` for the deferral case. This is the current `return None` state (codex r5's explicit ask).

The forced+reasoning constrained-decoding path needs prompt-prefill-state plumbing into the grammar builder — an **architectural change** to be decided deliberately (and verified on a real thinking model), tracked as a follow-up. Holding merge for that call rather than auto-overriding the blocking.

PR #1170, opened 2026-07-22
feat(tools): grammar-constrain Qwen3-Coder XML tool-calls (#558 E3)

raullenchai:
## Why

Grammar-constrained tool-calling (#558) today only covers JSON-body wires
(hermes / qwen / harmony). The flagship coder model — Qwen3-Coder — serializes
tool arguments as an XML parameter body, so its calls fell back to
free-form-then-parse: the model could emit an off-schema enum, a wrong-typed
value, or a malformed call the parser then had to salvage. Constraining the XML
wire at generation time makes a schema-valid call true by construction, closing
the same reliability gap #558 already closed for the JSON families. Refs #558.

## What

Extends #558 grammar-constrained tool-calling from the JSON-body families
(hermes / qwen / harmony) to the **Qwen3-Coder XML wire**. Previously only
JSON-object argument bodies could be grammar-constrained; a Qwen3-Coder tool
call is an XML parameter body, so it fell back to free-form-then-parse. This
teaches `build_tool_lark` to emit a grammar for that wire.

Wire (verified byte-for-byte against the Qwen3-Coder chat template + this repo's
`qwen3coder_tool_parser`):

```
<tool_call>
<function=NAME>
<parameter=KEY>
VALUE
</parameter>
</function>
</tool_call>
```

## How

- **`StructureInfo` gains `arg_style: str = "json"`** (frozen default). JSON-body
  families are byte-identical to before — a regression test pins that the
  `arg_style="json"` grammar emits no XML terminals and still uses `%json`.
- **`build_tool_lark` dispatches on `arg_style`**: `"xml"` emits one
  `<parameter=KEY>\nVALUE\n</parameter>\n` block per schema property, each VALUE
  grammar-typed per its sub-schema — enum → literal alternation, integer/number
  → JSON scalar terminals, boolean → `"true"|"false"`, object/array → `%json`,
  string → the lazy text rule below. Required props are mandatory (schema order),
  optional props are `( ... )?`.
- **`qwen3coder_tool_parser`** overrides `structure_info()` + sets
  `SUPPORTS_GRAMMAR=True`. It **opts out** (returns `None` → free-form fallback)
  unless `are_single_special_tokens` proves `<tool_call>`/`</tool_call>` are
  single special tokens on the live tokenizer (only Qwen tokenizers are), so a
  non-Qwen route can never build an unenforceable grammar.

### gap #1 — a string arg containing `<` (solved, not documented-around)

A first cut used `XMLSTR: /[^<]*/` for string values, which stops at the FIRST
`<` — so a `code` arg like `a < b`, `<html>…`, or `vector<int>` silently
truncated at that `<`. Fixed by porting **llguidance's first-class `[lazy]`
lexeme** — the exact idiom `StructTag.to_grammar` uses internally:

```
XML_PARAM_TEXT: /(.|\n)*/
xml_param_value[lazy]: XML_PARAM_TEXT "</parameter>"
```

`XML_PARAM_TEXT` admits any byte (incl. `<`); the lazy rule binds it to the
FIRST `</parameter>`. This matches XGrammar's `qwen_xml` "value up to first
`</parameter>`" semantics without needing regex lookahead (llguidance's engine
has none). Naive greedy `.*` overshoots (maximal-munch) and a lone-`<` rule
fails on `<html>`; `[lazy]` is the correct construct.

## Copy-source (no wheels invented)

- **Wire shape** — XGrammar-2 `builtin_structural_tag.py` `qwen_3_coder`
  built-in, `style="qwen_xml"` (reference for what the wire looks like).
- **Termination idiom** — llguidance `_struct_tag.py` `[lazy]` lexeme.
- **Plumbing pattern** — vLLM's Mistral tool parser proves llguidance can
  constrain a non-JSON wire via a generic Lark grammar on the guidance backend
  (`StructuredOutputsParams(grammar=...)`). No `xgrammar` runtime dependency is
  added; we stay on the llguidance backend adopted in #1131.

## Evidence (real tokenizer, not mocks)

`mlx-community/Qwen3-Coder-Next-4bit` via llguidance `LLMatcher`:

- Adversarial **enum** value CONSTRAINED (off-schema scalar + bad enum rejected).
- `<`-in-value accepted + terminal: `a < b && c > d` (27/27), `<html><body>x</body></html>` (28/28), `vector<int> v` (24/24), multiline code.
- `Paris` still 21/21 (no regression); leading prose masked at token 0.
- The `qwen3coder` parser round-trips every value back to the exact JSON `arguments`.

`pytest -k "tool_grammar or 558 or qwen3coder or structure_info"` → **324 passed,
1 skipped** (skip is an unrelated missing-`botocore` mirror test). `ruff check` +
`ruff format --check` clean. 29 new tests in `tests/test_qwen3coder_xml_grammar_558.py`
(golden Lark, real-tokenizer enforcement, round-trip, JSON-family byte-identical
regression guard).

## Scope

Fixes the **E3 slice** of #558 (Qwen3-Coder XML). Supersedes #1160 (which
branched before #1164's forced-tool fix and did not solve gap #1). Gemma-4 custom
`<|"|>` wire (E4) and the reasoning-gate (线①) are follow-up PRs.

### DIFF
--- a/tests/test_qwen3coder_xml_grammar_558.py
+++ b/tests/test_qwen3coder_xml_grammar_558.py
@@ -462,12 +462,30 @@ def tok():
 def lltok(tok):
     """Build an llguidance LLTokenizer from the fast (Rust) tokenizer via the
     module's own resolver (the spike-proven candidate-3 direct build handles the
-    transformers ``from_tokenizer`` isinstance gotcha ``guided.py`` trips on)."""
-    from vllm_mlx.api.tool_grammar import build_lltokenizer
+    transformers ``from_tokenizer`` isinstance gotcha ``guided.py`` trips on).
 
+    FINDING 5: a blanket ``skip when build_lltokenizer() is None`` let the whole
+    enforcement/round-trip suite go GREEN even if the production
+    tokenizer->llguidance integration was BROKEN. We now sanction the skip ONLY
+    when the runtime bridge is genuinely UNAVAILABLE (``HAS_LL_TOKENIZER`` False —
+    ``llguidance.hf`` / ``LLTokenizer`` not importable, the same narrow
+    unavailability the ``tok`` fixture treats as an offline skip). When the bridge
+    IS importable AND ``tok`` loaded (cached) but ``build_lltokenizer`` returns
+    ``None``, that is a broken integration and MUST FAIL — never silently pass."""
+    from vllm_mlx.api.tool_grammar import HAS_LL_TOKENIZER, build_lltokenizer
+
+    if not HAS_LL_TOKENIZER:
+        pytest.skip(
+            "llguidance runtime bridge (llguidance.hf / LLTokenizer) not "
+            "installed — XML enforcement tests require it"
+        )
     built = build_lltokenizer(tok)
-    if built is None:
-        pytest.skip("could not build an LLTokenizer for the Qwen3-Coder tokenizer")
+    assert built is not None, (
+        "build_lltokenizer() returned None for the CACHED Qwen3-Coder tokenizer "
+        "with the llguidance runtime bridge available — the production "
+        "tokenizer->llguidance integration is BROKEN (finding 5: this must FAIL, "
+        "not skip)."
+    )
     return built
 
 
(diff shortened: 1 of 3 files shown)


raullenchai:
## Merge rationale — local `pr_validate` codex override (documented, not hidden)

**GH `validate` gate is CLEAN** (validate / lint / tests / test-apple-silicon / test-matrix 3.10–3.12 / type-check all green; mergeState CLEAN). The only red is the **local** `pr_validate` codex adversarial step, which does not run in CI. Merging under a documented diminishing-returns override.

### Codex convergence trail (4 rounds)
- **r1** — 5 real correctness findings (empty-body `$ref`/composition, silent string-constraint drop, `$ref`→string quoting, `dependentRequired`, delimiter-unsafe keys) → all fixed via **faithful-or-opt-out**.
- **r2** — 5 guard-completeness findings → reframed the schema-representability guard from a keyword **blacklist** into a **strict allowlist** (complete-by-construction).
- **r3** — 4 findings (a `set(required)` crash on unhashable input; enum branch bypassing the allowlist; enum values carrying the wire delimiter) → all fixed; the total `required` guard removes the crash.
- **r4 (this state)** — the 4 remaining findings are NOT new correctness regressions:
  - **#1 (additionalProperties open-object)** — a policy disagreement, not a bug. Codex's fix ("require `additionalProperties: false` for opt-in") would opt out ~every real tool schema (they rarely set it), *defeating the feature*. Constraining to the declared `properties` is the intended tool-calling behavior. **Held deliberately** → the codex gate cannot go green without accepting a feature-defeating change.
  - **#2/#3/#4** — pathological-schema edges (`enum: []`, a type-less enum mixing non-string values, a property carrying self-shadowing `$defs`). All degrade to the **safe free-form-then-parse fallback** — they never emit an invalid tool call — and real OpenAI/Pydantic-generated tool schemas do not produce them.

### Quality gates that ARE met
- **352 tests pass, 0 skipped** (`-k "tool_grammar or 558 or qwen3coder or structure_info"`), including real-tokenizer `LLMatcher` enforcement + round-trip on `Qwen3-Coder-Next-4bit`.
- Full unit suite green (13410 passed); lint + format clean; GH `validate` CLEAN.
- The `<`-in-string-value fix (gap #1) and the enum/type/nested-object constraint are proven on real hardware.

### Known follow-ups (safe under-constraint, tracked for a later polish)
- Widen the top-level allowlist to tolerate inert annotation keys (`$schema`, `examples`, `readOnly`, …) so schemas carrying them stay constrained instead of opting out.
- Close the r4 pathological enum/`$defs` edges (#2/#3/#4) for true allowlist completeness.

Per the operator's standing guidance: when codex narrows onto edge/policy items while the feature is correct and proven, merge rather than spin further rounds.

PR #1171, opened 2026-07-22
feat(tools): grammar-constrain Gemma-4 native tool-calls (#558 E4)

raullenchai:
## Why

Grammar-constrained tool-calling (#558) now covers the JSON-body families
(hermes/qwen/harmony) and the Qwen3-Coder XML wire (#1170, E3). **Gemma-4 is the
last of the four top families** and used a distinct native wire, so its tool
calls fell back to free-form-then-parse — the model could emit an off-schema
enum, a wrong-typed value, or a malformed call the parser then had to salvage.
This closes that gap for Gemma-4, making a schema-valid call true by
construction. Refs #558.

## The Gemma-4 native wire

```
<|tool_call>call:FUNC_NAME{key1:<|"|>strval<|"|>,key2:numval}<tool_call|>
```

`<|tool_call>` (id 48), `<tool_call|>` (id 49), and the string-quote `<|"|>`
(id 52) are each a single special token on `gemma-4-e2b-it-4bit` (verified).
String values are wrapped in `<|"|>`; numbers/bools are bare; keys are
`dictsort`-ordered, comma-separated.

## The key construct — greedy, NOT `[lazy]` (structural difference from E3)

E3's string terminator `</parameter>` is byte text, so it needs llguidance's
`[lazy]` lexeme. Gemma-4's terminator `<|"|>` is a **special token**, and the
two `[lazy]` forms both fail: a special-token ref cannot appear inside a lexer
terminal, and a byte-literal `"<|\"|>"` cannot be satisfied by the atomic token
52. The correct construct is a plain **greedy rule with rule-level special
refs**:

```
gemma_str_value: <|"|> GEMMA_STR_TEXT <|"|>
```

Because `<|"|>` is a special token, a greedy byte regex physically cannot
consume it, so maximal-munch stops at the boundary on its own — no `[lazy]`
needed. Verified on the real tokenizer: values containing `<`, `>`, `{`, `}`,
`,`, quotes, and newlines are all accepted, terminal, and round-trip exactly.

## How

- `build_tool_lark` gains an `arg_style="gemma4"` dispatch → `_emit_gemma4_arg_body`.
  Body = dictsort-ordered comma-separated `key:VALUE`; because commas are
  *separators*, optionals use a first-present construction so an absent early
  optional never misplaces a comma (verified: every subset accepted;
  leading-comma rejected). String → the greedy `<|"|>` rule; enum → per-value
  `<|"|>`-wrapped alternation (handles mixed-type enums); int/number/bool and
  object/array → `%json`.
- The E3 strict-allowlist representability guard is **reused** via a shared
  `_ArgWirePolicy`: the structural allowlist is identical; only two wire-specific
  leaf checks differ (gemma4 keys are bare `\w+`; the only wire-unsafe enum byte
  is the `<|"|>` marker, so `<`/`>`/newlines are allowed in gemma4 values, unlike
  XML). This inherits all 4 rounds of E3's codex hardening rather than
  re-deriving a weaker guard.
- `gemma4_tool_parser`: `SUPPORTS_GRAMMAR=True`, `structure_info()`
  (begin=`<|tool_call>call:NAME{`, end=`}<tool_call|>`, trigger=`<|tool_call>`,
  all three markers as sentinels; opts out unless all are single special
  tokens), and `TOOL_GRAMMAR_AUTO_SAFE=False`.
- JSON-body families are kept **byte-identical** (regression-tested).

## Evidence

- `tests/test_gemma4_grammar_558.py` → **50 passed, 0 skipped** (real-tokenizer
  `LLMatcher` enforcement + round-trip on `gemma-4-e2b-it-4bit` RAN, not skipped).
- Broader `-k "tool_grammar or 558 or gemma4 or structure_info"` → **622 passed,
  19 skipped, 1 xfailed** — E3 + hermes/qwen structure_info clean (the shared
  `_ArgWirePolicy` refactor introduced no regression).
- `ruff check` + `ruff format --check` → clean.

## Honest limitations (follow-ups, all safe)

- **AUTO opts out** (`TOOL_GRAMMAR_AUTO_SAFE=False`): Gemma-4 emits a
  `<|channel>…<channel|>` reasoning block whose special-token markers a byte
  free-prefix can't match, and the gemma4 reasoning parser exposes no
  start/end token to wire path-A tolerance. `required`/named tool_choice are
  fully constrained; `auto` falls back to free-form. Follow-up: register the
  channel markers as reasoning_sentinels (parallels the existing forced+reasoning
  follow-up).
- **object/array args go through `%json`** (standard JSON `{"x":3}`), not
  gemma's native bare-key form (`{x:3}`). The parser round-trips it (verified
  end-to-end), so correctness holds; a faithful recursive gemma-object grammar
  is a possible follow-up. Scalars/strings/bools/enums (the common case) are
  fully native.
- Feasibility, enforcement, and round-trip are proven on the real tokenizer via
  `LLMatcher`; live decode-quality under the mask is unmeasured (standard for
  these grammar PRs pre-merge).

### DIFF
--- a/tests/test_gemma4_grammar_558.py
+++ b/tests/test_gemma4_grammar_558.py
@@ -605,17 +605,61 @@ def _tokenizer_in_cache() -> bool:
     return isinstance(path, str)
 
 
+def _hf_offline_mode() -> bool:
+    """True iff HuggingFace is configured for OFFLINE mode (env or hub constant).
+    Used to skip the uncached tokenizer fetch on an INTENTIONAL offline run
+    regardless of how transformers surfaces the miss (bare OSError included)."""
+    import os
+
+    if os.environ.get("HF_HUB_OFFLINE") or os.environ.get("TRANSFORMERS_OFFLINE"):
+        return True
+    try:
+        from huggingface_hub import constants
+
+        return bool(getattr(constants, "HF_HUB_OFFLINE", False))
+    except Exception:  # pragma: no cover - very old hub
+        return False
+
+
+def _is_connectivity_error(exc: BaseException) -> bool:
+    """True iff exc is a recognized NETWORK/offline connectivity type. A 404 /
+    bad-revision / corrupt error is deliberately NOT here — those must FAIL, not
+    skip (codex r6 #4). Only the transient/offline connectivity path skips."""
+    types: list[type[BaseException]] = []
+    try:
+        from huggingface_hub.errors import (
+            LocalEntryNotFoundError,
+            OfflineModeIsEnabled,
+        )
+
+        types += [LocalEntryNotFoundError, OfflineModeIsEnabled]
+    except Exception:  # pragma: no cover - old hub
+        pass
+    try:
+        from requests.exceptions import ConnectionError as _ReqConnErr
+        from requests.exceptions import Timeout as _ReqTimeout
+
+        types += [_ReqConnErr, _ReqTimeout]
+    except Exception:  # pragma: no cover - requests absent
+        pass
+    return bool(types) and isinstance(exc, tuple(types))
+
+
 @pytest.fixture(scope="module")
 def tok():
+    """Load the pinned gemma4 tokenizer for the enforcement tests.
+
+    PRIMARY gate: cache presence. A CACHED tokenizer must LOAD — if it fails, that
+    is corruption and the test FAILS (any exception propagates on the cached path
+    below). A NOT-cached tokenizer is skip-eligible ONLY when the miss is truly a
+    connectivity/offline one; the offline-vs-real-failure decision is broken by an
+    OFFLINE-INTENT tie-break (``_hf_offline_mode`` OR ``_is_connectivity_error``),
+    NOT a catch-all skip and NOT a message sniff. This resolves the r5<->r6
+    oscillation: skip-on-any-exception (r5) hid a bad revision / 404 / corrupt-
+    partial-cache as a false-GREEN (codex r6 #4), while the pre-r5 type-only skip
+    surfaced a genuine offline miss (bare wrapping ``OSError``) as a false-RED
+    (codex r5 F3). The r3 message-substring heuristic (codex r4 F3) is gone."""
     transformers = pytest.importorskip("transformers")
-    # PRIMARY gate: cache presence. A CACHED tokenizer must LOAD — if it fails, that
-    # is corruption and the test FAILS (any exception propagates on the cached path
-    # below). Only a NOT-cached tokenizer is skip-eligible (genuinely unavailable),
-    # and there ANY load failure means it could not be obtained, so it skips
-    # UNCONDITIONALLY. This replaces the r3 message-substring heuristic, which could
-    # not distinguish a corrupt cached artifact from an offline miss (codex r4 F3),
-    # and drops the r4 type-only skip that made the suite network-dependent when a
-    # real offline miss surfaced as a bare wrapping ``OSError`` (codex r5 F3).
     if _tokenizer_in_cache():
         return transformers.AutoTokenizer.from_pretrained(
             _TOKENIZER_MODEL, revision=_TOKENIZER_REVISION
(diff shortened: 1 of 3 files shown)


raullenchai:
## Merge rationale — codex convergence r2→r8, GH validate CLEAN

This PR (E4: grammar-constrain Gemma-4 native tool-calls, #558) went through **seven** codex
adversarial-review rounds. Every finding was checked against the actual code — not blanket
waved through, not blanket accepted. The gauntlet earned its keep: it surfaced **two real
soundness gaps** (both now fixed with proof) and drove the test suite to anchor its wire
assumption against the model's real chat_template.

**Ground-truth anchor (the strongest evidence).** The pinned gemma-4 `chat_template` renders
an assistant `tool_calls` turn to a wire that is **byte-identical** to the format this PR
grammar-constrains and the parser reads: `apply_chat_template(...)` emits
`<|tool_call>call:run{code:<|"|>print(1)<|"|>,lang:<|"|>python<|"|>,timeout:30,verbose:true}<tool_call|>`,
covering both wire shapes (`<|"|>`-wrapped strings + bare `%json` scalars). The generated
grammar accepts that exact rendered wire (LLMatcher 30/30 tokens, is_accepting) and the parser
recovers the call exactly. So the entire enforcement suite is anchored to what the model
actually emits, not to a handwritten assumption.

**GH-side merge gate: fully green** — `validate`, `lint`, `type-check`, `test-apple-silicon`,
`test-matrix (3.10/3.11/3.12)` all PASS. Local `pr_validate`: `full_unit` 13467 passed,
`targeted_tests` 126 passed, `lint` (ruff check + format) clean.

### Two real soundness gaps — found by codex, fixed with proof

1. **F1 — `<|"|>` byte-spelling could bypass the greedy string content (codex r4).** `<|"|>`
   is dual-nature: an atomic special token (id 52) AND a 5-byte sequence the parser
   text-scans. A greedy `GEMMA_STR_TEXT` admitted output the parser mis-reads as an early
   string terminator. Fixed with llguidance's engine-native And/Not "does not contain
   substring" recipe — `GEMMA_STR_TEXT: /(.|\n)*/ & ~/(?s:.*)<\|"\|>(?s:.*)/` — copied from
   llguidance `docs/syntax.md`, not invented. Proven at token granularity: an LLMatcher test
   feeds a manually-decomposed ordinary-token spelling of `<|"|>` inside a value and asserts
   rejection at the completing token (baseline confirmed the bug at 18/18 acceptance before).

2. **Empty `enum: []` compiled as unrestricted (codex r6).** `enum` is in the allowed
   string-key set, so a property `{"type":"string","enum":[]}` fell through the non-empty-enum
   guard and was emitted as an unrestricted `<|"|>` string — even though an empty enum is
   unsatisfiable under JSON Schema. Fixed at the shared representability guard
   (`_xml_property_representable`): an empty enum now opts out (faithful-or-opt-out). Pure safe
   tightening — it only ever widens the free-form fallback — and it closes the same latent gap
   on the already-shipped E3 XML wire.

### Test hygiene fixed (r4→r6)
Offline-vs-corrupt is decided by HF cache-presence (`try_to_load_from_cache`), not by sniffing
loader messages. The uncached-tokenizer branch now uses an explicit offline-INTENT check
(`HF_HUB_OFFLINE`/`TRANSFORMERS_OFFLINE`/hub constant) + a connectivity-type check, dissolving
the r5↔r6 oscillation: an intentional offline run skips; an online bad-revision / 404 /
corrupt-cache PROPAGATES and fails loudly.

### One codex finding refuted with a test (r6 #1 — false positive)
codex claimed a `%json` object value with a brace inside a JSON string (`{"s":"}"}`)
prematurely terminates the outer call. The real scanner `_scan_gemma4_tool_calls` tracks
`in_json_string` with escape handling (gemma4_tool_parser.py ~71-100), so nested-JSON-string
braces/commas are correctly ignored for brace depth. (codex described
`_recover_incomplete_gemma4_calls`, the best-effort fallback that runs ONLY on the
non-streaming finalize path when the balanced scanner found zero complete calls.) Added a
proving test: the scanner returns exactly one match with the correct outer-brace boundary and
parses the nested object (`s=="}"`, `k=="a,b"`); the grammar compiles and LLMatcher accepts
the full wire.

### The one remaining override — empty/absent `properties` → forced `call:NAME{}`
`call:NAME{}` is schema-**valid** even for an open object (an open object permits zero
properties), so this is an over-constraint, NOT unsoundness — it never emits schema-invalid
output. It favors the common no-arg tool (guaranteeing no hallucinated args) and mirrors the
already-SHIPPED E3 `_emit_xml_arg_body` (#1170); the guard and emitter both document it as the
deliberate tool-calling no-arg convention. Flipping only E4 to opt-out would diverge it from
the shipped XML wire for the identical schema.

### The two r8 NITs
- **enum opt-out "too conservative"** — override. Opting out an enum value that tokenizes to
  a registered special token is conservative-but-sound (faithful-or-opt-out); emitting a
  possibly-unreachable byte-literal branch is not. Losing the grammar for such a rare enum is
  the safe direction.
- **cache-only fixture can skip on an un-provisioned CI** — acknowledged, tracked as a
  fast-follow (NOT a code bug). The gemma-4 enforcement tests are cache-only by design (codex
  r7's own ask); the trade-off is that a runner without the tokenizer cached skips them. They
  are exercised every time on the `pr_validate` gate (this run: 71 passed, incl. the
  chat_template anchor — not skipped). Follow-up: add a gemma-tokenizer cache step to the
  gemma4 CI job so the enforcement runs on GH runners too. Separable CI-infra change; kept out
  of this feature PR's scope.

Merging under the standing diminishing-returns policy: GH validate clean + real-tokenizer
LLMatcher/full_unit/chat-template-anchor evidence + both real soundness gaps fixed with proof.
codex's last three rounds re-flag the same E3-consistent `{}` over-constraint (not unsoundness)
plus symmetric test-infra trade-offs — no new production correctness finding since the
empty-enum fix. The one BLOCKING is the documented override.

PR #1185, opened 2026-07-23
feat(reasoning): generation-time thinking-token budget (force-close </think>)

raullenchai:
## What

Enforce a per-request **thinking** budget **at decode time** for text-parser reasoning families. Once the model has spent `reasoning_max_tokens` inside its `<think>…</think>` span, the next-token logits are OVERRIDDEN so the only sampleable token is the single `</think>` id — the model is forced to close the block and commit to its answer.

This is the mechanism three mature engines converge on; we copy their shape rather than invent one, expressed as an mlx-lm logits processor so it drops into the existing per-step decode loop with no decode-loop changes:

- **vLLM** `ThinkingBudgetStateHolder` (fills the end-token logit to a dominating value)
- **SGLang** `ReasonerGrammarObject` (THINKING → GENERATION phase machine)
- **mlx-vlm** `ThinkingBudgetCriteria` (MLX-native, ~40 LOC)

## Why (single mechanism, not two)

Before this PR the only reasoning cap was a **post-hoc char-count trim** in the postprocessor: it let the model keep burning compute for the full unbounded think span and only relabelled/injected `</think>` in the OUTPUT the client sees. That is a band-aid, not a budget.

This PR makes the generation-time force-close **authoritative** for models whose `</think>` is a single token, and **suppresses** the post-hoc trim for those requests so the two never both run. The XOR decision has one definition — `_effective_posthoc_reasoning_cap(...)` — shared by the streaming and non-streaming postprocessor sites, so it can never drift between the two paths.

## Template-derived seeding (not conversation content)

A reasoning template takes one of two shapes: it either PREFILLS `<think>` into the rendered prompt (counting starts at the first generated token) or the model EMITS `<think>` as a generated token (counting waits for that opener). Which shape applies is derived **here in the route, from the template's own rendered generation prefix** (`rendered_prompt_opens_think` checks the rstripped SUFFIX of `build_prompt(...)`), never from raw conversation content — so a `<think>` a user typed earlier cannot mis-seed the budget. Either way the span-opening opener is free, so a `budget == 0` request closes to a well-formed `<think></think>`.

## Guards (copied from upstream)

- **Never force a reasoning-end that cannot occur** (vLLM #39130): a request whose thinking is not DEFINITIVELY on (`_effective_enable_thinking(...)` resolves to `True`, mirroring `apply_chat_template`) never builds a budget processor.
- **Identify `</think>` by TOKEN ID**, single-special-token only — reuses the exact `resolve_reasoning_sentinels` gate the tool-grammar path uses. A family whose `</think>` splits into ordinary bytes (channel-routed gpt-oss / harmony) opts out to the post-hoc cap; a tracked follow-up covers them.
- **Never inject `</think>` mid tool-call** (vLLM #44676): **tool** requests opt out; the post-hoc cap owns them until a follow-up wires SGLang's `think_excluded_tokens` so budget + tools can coexist.
- **Render failure keeps the budget enforced**: if `build_prompt` can't render (e.g. an MLLM engine), `build_budget_from_render` installs NO processor and RETAINS the post-hoc cap — never a non-seeded processor that would silently suppress the cap yet never fire.
- **Stop-sequence conflict opts out** (this round): a request that lists `</think>` (or an overlapping substring) in `stop` declines the generation-time budget (`reasoning_stop_conflicts`). Forcing `</think>` would trip that client stop AT the reasoning boundary and truncate reasoning before the answer; the model's reasoning end is a template-internal boundary, not a user stop. The post-hoc cap (which appends `</think>` to the FINAL text, not through the decode-time stop matcher) still enforces the budget — so budget + `stop=["</think>"]` behaves EXACTLY as pre-PR, no feature-introduced regression.

## Scheduler desync fix (found via live pilot)

A budget processor is a per-request logits processor with the **same positional-list desync exposure as a grammar**. Discovered on a live qwen3-0.6b pilot: a plain greedy request *preceding* a budget request left the budget **inert** — mlx-lm's positional `logits_processors` list dropped it when the no-processor request finished, so `</think>` was never forced and reasoning ran unbounded.

Fix mirrors the proven grammar path: a live budget uid now ALSO arms `_realign_grammar_logits_processors`, which rebuilds every slot from the authoritative `uid_to_request_processors` each tick, plus a tombstone so a finished budget uid stays armed for the scrubbing tick. Penalty-only uids still don't arm it (perf).

## Real-hardware proof (qwen3-0.6b-4bit)

| scenario | reasoning | answer |
|---|---|---|
| `budget=8`, no stop | ~20 chars (~8 tok) then FORCED `</think>` | complete answer (234 chars) ✅ |
| `budget=200`, no stop | 638 chars (finished under budget) | complete answer (269 chars) ✅ |
| **regression:** budget req after a prior no-budget req | bounded | ✅ (was unbounded before the desync fix) |
| `budget=200` + `stop=["</think>"]` | post-hoc cap (fix); == pre-PR | reasoning-only per user stop, no crash ✅ |
| `enable_thinking=false` + budget | — | never forces (guard) ✅ |
| no budget | unbounded | full answer ✅ |

Server log confirms the decode-time force: `reasoning budget spent (N think tokens) — forcing </think> (id=151668) at decode time`.

## Tests

- `test_reasoning_budget_generation.py` — 29 tests: processor phase-machine, force OVERRIDE dominates a prior `-inf` at `</think>`, 2-D batched shape preservation, repeated-opener-counts-toward-budget, template-derived seeding (incl. immunity to a user-typed marker), render-failure fallback, stop-conflict opt-out, `_effective_posthoc_reasoning_cap` selector.
- `test_grammar_scheduler_realign_558.py` — 24 tests: budget-uid desync repair, tombstone-until-scrubbed, leaked-budget-processor scrubbed from a foreign slot then flushed, arming by a budget uid alone.

No version bump (batched separately per release SOP).

---

### Update (codex convergence, round 8)

Three further hardening items, all with new tests + a fresh real-hardware pilot:

- **Cloud-safe placement** — the budget is now built by `_build_reasoning_budget_processor(...)`, called only once the request is committed to LOCAL generation (past the cloud-offload decision). A cloud-routed request neither installs the processor nor has its post-hoc cap suppressed. The 95-line inline block became one call site + a documented helper.
- **Seeding from the TEMPLATE delta, not the whole prompt** — `_template_generation_prefix` re-renders with the last message's content replaced by an opaque sentinel and inspects only what the template appends *after* it. A `<think>` a user typed can no longer reach the seed decision even on a template that appends no assistant prefix (the previous whole-suffix check assumed one always exists). Immunity is now a property, not an assumption.
- **Out-of-range `</think>` guard** — if a tokenizer/model vocab mismatch put `</think>` beyond the logits width, the force step now DISABLES the budget (graceful unbounded reasoning) instead of masking to an all-`-inf` row.

Also relocated the whole block after `enforce_context_length_for_messages` so an oversized request is rejected before any budget render (perf, invalid requests).

**Tests:** `test_reasoning_budget_generation.py` now 40 (added template-delta seeding + immunity, route-builder gates, out-of-range guard); `test_grammar_scheduler_realign_558.py` 24. Both green.

**Real-hardware pilot (qwen3-0.6b-4bit), post-refactor:** budget=8 (explicit & default thinking) → ~8-token reasoning + complete answer; budget=200 no-stop → bounded + answer; budget=200 + `stop=["</think>"]` → post-hoc opt-out; user message ending in `<think>` → still correctly bounded (no mis-seed); no-budget → unbounded. 0 errors.

---

### Update (codex convergence, rounds 9–15)

Further hardening; every item has new tests + a fresh real-hardware pilot (now qwen3.5-4b-4bit). Supersedes stale details above (seeding is no longer a suffix heuristic; test counts updated).

**Exact template-derived seeding (R10).** The template's generation prefix is isolated by rendering the prompt TWICE — `build_prompt(add_generation_prompt=True)` minus `build_prompt(add_generation_prompt=False)` — and taking the string delta. That delta is precisely the text the chat template appends after the conversation, immune to any `<think>` in user content (it is present in both renders and cancels). Replaces the earlier sentinel/suffix heuristic. `add_generation_prompt` is now threaded through `apply_chat_template` and every engine `build_prompt`.

**Tri-state seed parse (R12).** The isolated delta is classified `open` (unclosed `<think>` → seeded, count from token 0), `emit` (no marker → model emits the opener, count after it), or `ambiguous` (closed pair / stray marker / no parser → DECLINE, keep the post-hoc cap).

**Weight-derived output-head width, no fallbacks (R11/R12/R14/R15).** A processor installs only when the `</think>` id is verified `< the model's real logits width`, read from the lm-head (untied) or tied-embedding **weight rows**. Because `logits = head(hidden)` ⇒ `logits.shape[-1] == head.weight.shape[0]`, the build-time bounds check and the decode-time force validate `</think>` against the **identical** width — so the decode-time out-of-range guard in `_force_distribution` is **provably unreachable** for any installed processor, sound for both the streaming and non-streaming paths (the decision is fixed at build time). Two fallbacks were rejected as unsound because each can exceed the true head width, admitting an out-of-range force id (suppressing the post-hoc cap) only for the decode guard to then disable the budget → silently unenforced: `len(tokenizer)` (**R14** — counts added specials, so `</think>` is always within it, a vacuous check) and declared `config.vocab_size` (**R15** — a padded/over-declared vocab can exceed the real head width, the exact mismatch that would trip the guard). When the head weight is not inspectable → DECLINE (keep the post-hoc cap; R11). Every current text reasoning family (qwen3 tied embed, gpt-oss/llama untied `lm_head`) resolves via the weight, verified on the qwen3.5-4b pilot.

**Full stop-overlap opt-out (R13).** `reasoning_stop_conflicts` now flags a `stop` that overlaps the forced `</think>` on ANY of the four ways two strings can collide at the `…R</think>A…` boundary: stop⊂marker, marker⊂stop, stop-suffix==marker-prefix, stop-prefix==marker-suffix. Previously only one cross-boundary side was caught, so `stop=["</think>A"]` / `["x</think>y"]` slipped through.

**start_id bounds symmetric with end_id (R13).** A non-seeded processor force-masks the OPEN id too; an out-of-range `start_id` can never match a sampled token, so the budget would silently never fire — the builder now bounds-checks `start_id` and declines when it is out of range.

**Mask released on span close (R14).** The cached full-vocab OVERRIDE row is dropped the moment `</think>` is observed (the processor is inert thereafter), instead of pinned through the whole answer — reclaims a few hundred KB per concurrent budgeted request.

**Log after the guard (R13).** The one-shot "forcing `</think>`" success log now emits only after the out-of-range check passes, so a disabled budget never prints a false success line followed by its disable warning.

**Deferred (documented).** codex flagged that `_template_generation_prefix` renders the prompt twice at admission, on top of the generation render. Deferred: the double render is admission-time only, on the minority of budgeted requests, and threading the `full` render into the generation path touches the hot generate loop for marginal gain. Tracked for a follow-up.

**Tests.** `test_reasoning_budget_generation.py` now 58 (added: two-render forwarding through `apply_chat_template` + batched engine, tri-state seed parse, weight-derived head width, decline-when-only-`len(tokenizer)` and decline-on-config-only, 4-condition stop-overlap both sides, start-id-out-of-range decline, mask-released-on-close, log-after-guard). `test_grammar_scheduler_realign_558.py` 25. Both green.

**Real-hardware re-verify (qwen3.5-4b-4bit).** `17*24` at `budget=16` → 26 reasoning tokens, forced `</think>`, answer `408`, `finish=stop`, no marker leak; `budget=400` → 170 reasoning tokens, `408`. A control (thinking on, no budget) uses 1568 reasoning tokens + full answer with `finish=stop`, confirming the budget only BOUNDS reasoning and never corrupts output. Reasoning stays at or below the budget at every level (16→26, 32→46, 256→317, 400→391 forced / 170 natural-close), and no case leaked `<think>`/`</think>` into visible content.

**CI note.** Local `full_unit` and `stress_e2e_bench` cannot run in this environment (disk full — the suite's Qwen3-Coder tokenizer and the 35 GB matrix models cannot be fetched); both are covered by GitHub CI, where `tests`, `test-apple-silicon`, `test-matrix (3.10/3.11/3.12)`, `type-check`, `lint`, and `validate` are all green on this commit.

**codex verdict: MERGE-SAFE (round 16)** — no blocking findings. Three nits surfaced; disposition: **took** the silent-`except` observability nit (a generation-prefix render failure now logs at debug with model context before falling back to the post-hoc cap). **Declined** two as low-value: the `embed_tokens.weight` tie-check nit (input embedding and output head share the vocab dimension `shape[0]` even when untied, so the *width* is robust regardless of tying — only an exotic asymmetric-vocab model would differ, and it is not a text reasoning LLM), and the forwarding-test-through-`build_prompt` nit (the end-to-end `add_generation_prompt` forwarding is already protected by the route-builder seed tests, which assert seeding from the two-render delta and would fail if the two renders collapsed to identical).

### DIFF
--- a/tests/test_grammar_scheduler_realign_558.py
+++ b/tests/test_grammar_scheduler_realign_558.py
@@ -13,7 +13,7 @@ the boot warmup fell back to free-form).
 from AUTHORITATIVE per-uid state: ``uid_to_request_processors`` holds the FULL
 processor list (grammar + penalties) for EVERY live uid that carries any
 processor — grammar AND penalty-only — so a bystander's penalties survive a
-length-desync rebuild (codex #3). A ``_known_grammar_processors`` identity set
+length-desync rebuild (codex #3). A ``_known_stateful_processors`` identity set
 (with tombstones for finished-but-still-leaked grammars) scrubs leaked grammars
 from untracked slots. These tests drive that method against a stand-in
 generation batch reproducing the desync shapes — no model needed.
@@ -41,34 +41,43 @@ def _make_scheduler_stub():
     stub = SimpleNamespace(
         uid_to_request_processors={},
         _uids_with_grammar=set(),
-        _known_grammar_processors=set(),
-        _grammar_processor_objs={},
-        _grammar_tombstones=set(),
+        _uids_with_reasoning_budget={},
+        _known_stateful_processors=set(),
+        _stateful_processor_objs={},
+        _stateful_tombstones=set(),
         batch_generator=None,
     )
     stub._realign_grammar_logits_processors = (
         Scheduler._realign_grammar_logits_processors.__get__(stub, type(stub))
     )
-    stub._flush_grammar_tombstones = Scheduler._flush_grammar_tombstones.__get__(
+    stub._flush_stateful_tombstones = Scheduler._flush_stateful_tombstones.__get__(
         stub, type(stub)
     )
     stub._forget_uid_grammar = Scheduler._forget_uid_grammar.__get__(stub, type(stub))
+    stub._realign_guard_armed = Scheduler._realign_guard_armed.__get__(stub, type(stub))
+    stub._register_uid_processors = Scheduler._register_uid_processors.__get__(
+        stub, type(stub)
+    )
     return stub
 
 
-def _register(stub, uid, processors, grammar=None):
-    """Mirror the scheduler's insert-time bookkeeping.
+def _register(stub, uid, processors, grammar=None, budget=None):
+    """Register a uid through the PRODUCTION bookkeeping helper.
 
-    ``grammar=None`` registers a penalty-only uid (tracked for desync repair but
-    not a grammar). A non-None ``grammar`` also registers its identity + arms
-    the guard, exactly like the scheduler insert site.
+    Delegates to ``Scheduler._register_uid_processors`` (the exact method the
+    admission path calls) rather than hand-reproducing its writes, so these
+    tests break if the real registration drops a grammar/budget arm or a penalty
+    list (codex NIT). ``grammar=None`` registers a penalty-only uid (tracked for
+    desync repair but not a grammar). A non-None ``grammar`` also registers its
+    identity + arms the guard. A non-None ``budget`` arms the guard for a
+    generation-time reasoning-budget processor AND registers its identity in the
+    same stateful set/tombstone machinery as grammar — a leaked force-close
+    processor must be scrubbed from a foreign slot exactly like a grammar (codex).
     """
-    if processors:
-        stub.uid_to_request_processors[uid] = list(processors)
-    if grammar is not None:
-        stub._uids_with_grammar.add(uid)
-        stub._known_grammar_processors.add(id(grammar))
-        stub._grammar_processor_objs[id(grammar)] = grammar
+    request = SimpleNamespace(reasoning_budget_logits_processor=budget)
+    stub._register_uid_processors(
+        uid, request, list(processors) if processors else None, grammar
+    )
 
 
 def test_realign_repairs_stale_leading_entry():
@@ -187,7 +196,7 @@ def test_realign_untracked_uid_with_no_processors_gets_empty_slot():
 def test_realign_scrubs_finished_grammar_via_tombstone():
     # codex #1: a grammar whose owning uid already FINISHED can still linger in
     # another slot. _forget_uid_grammar TOMBSTONES it (keeps it in
-    # _known_grammar_processors) so realign still scrubs it; only once it's
+    # _known_stateful_processors) so realign still scrubs it; only once it's
     # absent from every live slot is it fully forgotten.
     glp_live = object()
     glp_dead = object()  # finished owner, but leaked into uid 9's slot
@@ -196,10 +205,10 @@ def test_realign_scrubs_finished_grammar_via_tombstone():
     _register(stub, 8, [glp_dead], glp_dead)
     # uid 8 finishes: forget tombstones glp_dead (still known + scrubbable).
     stub._forget_uid_grammar(8)
-    assert id(glp_dead) in stub._known_grammar_processors, (
+    assert id(glp_dead) in stub._known_stateful_processors, (
         "must be tombstoned, not gone"
     )
-    assert id(glp_dead) in stub._grammar_tombstones
+    assert id(glp_dead) in stub._stateful_tombstones
     # A leaked slot still carries glp_dead at uid 9's (untracked) position.
     stub.batch_generator = _FakeBatchGen(
         uids=[7, 9], logits_processors=[[glp_live], [glp_dead]]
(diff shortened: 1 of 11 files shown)


SmoothSailingMusic:
This is awesome. Gemma 4 is a big time over thinker.

PR #1192, opened 2026-07-23
feat(tools): reasoning-gated forced tool grammar via thinking budget (#558 line①)

raullenchai:
## #558 line① — hard-constrain forced tool-call ARGUMENTS on REASONING models

The last #558 keystone. On a reasoning model a forced/named tool call
(`tool_choice="required"` / a named function) previously **opted out** of grammar
constraint (`build_tool_grammar` :2068) because a bounded forced grammar fires the
trigger at token 0 and leaves the prompt's `<think>` unclosed — the qwen3 parser
then buries the whole call in `reasoning_content`. So the strongest available lever
was the best-effort forced-prefix, and the argument JSON was **unconstrained**: an
out-of-enum value, a missing required field, or a malformed `{}` could slip
through.

line① lifts that opt-out with a **reasoning-GATED** forced grammar: build the plain
bounded forced grammar but hold its mask **OFF** behind a runtime **token-id gate**
until `</think>` is decoded, and **couple a generation-time thinking budget** that
force-closes `</think>` inside the think span. After the boundary the grammar masks
the arguments to the schema — so the argument JSON is now schema-valid every time.

### What is guaranteed, and what is best-effort (honest scope)

An earlier revision of this PR claimed a full *structural* guarantee. codex round 3
(#4) correctly showed that overclaims: masking the opener's special-token id does
not *prove* the trigger text can never be spelled as ordinary subtokens. The honest
decomposition:

| Property | Status |
|---|---|
| The coupled budget force-closes `</think>` for a forced-tool reasoning request (bypassing #1185's tool opt-out, exactly as vLLM `thinking_budget_state.py` and SGLang `ReasonerGrammarObject` do — neither opts tools out) | **Guaranteed** — the gate provably opens |
| After `</think>`, the tool-call **arguments are hard-constrained to the schema** (out-of-enum / missing-field / malformed-`{}` can no longer surface) — the core #558 value | **Guaranteed** when the gate engages |
| The model cannot begin a tool call inside `<think>` via its **natural opener** (special-token id masked while the gate is closed — SGLang `think_excluded_token_ids`) | **Best-effort** — removes the natural path; NOT a proof against an adversarial model spelling the opener out as subtokens (never observed across probes; trained models emit the atomic opener) |
| Every precondition that cannot be *proved* declines to the forced-prefix fallback (multi-token/unresolvable opener, un-priceable schema minimum, stop-conflict, no budget, `auto`, degenerate `max_tokens`) | **Strictly non-regressive** |

The vLLM #44676 footgun is specifically the **ungated `tool_choice="auto"` parser
path** — which line① never takes. Every other tool request keeps the opt-out.

### Mechanism

- `GrammarLogitsProcessor.reasoning_gate_id` exposes the `</think>` id the mask is
  gated behind (`None` when unconstrained / PATH A).
- `_build_reasoning_budget_processor(allow_tools=…)` lifts the tool opt-out so the
  budget is installed for the gated tool request, and accepts a pre-rendered
  `seed_prefix` so no second synchronous `apply_chat_template` runs.
- The route passes `allow_tools` **exactly** when `_glp.reasoning_gate_id is not
  None`, coupling the budget to the gate (SGLang's structural coupling).
- **`think_excluded_token_ids` (SGLang parity, best-effort):** while the gate is
  closed, each forced tool's opener special-token is masked. Engaged **all-or-
  nothing** (codex r3 #3) — if any tool's opener is multi-token/unresolvable the
  whole gate declines.
- Engages only for **forced/named + thinking + a set `reasoning_max_tokens` +
  provable `max_tokens` room + no `</think>` stop-conflict + seed-state "open"**,
  re-verifying `</think>` is a single in-vocab id. `emit`/`ambiguous`/no-budget/
  `auto`/degenerate-`max_tokens`/stop-conflict/un-priceable-schema all keep the
  forced-prefix fallback — **strictly non-regressive**.

### End-to-end proof (qwen3.5-4b-4bit, worktree build)

Adversarial forced call: a `set_color` tool whose `color` is an enum
`{red,green,blue}`, user demands out-of-enum **"purple"**, `tool_choice="required"`,
thinking on. Natural think length ~39 tokens.

| `reasoning_max_tokens` | finish | rtok / comp | tool call |
|---|---|---|---|
| **2** | `tool_calls` | **2** / 31 | `{"color":"red","brightness":70}` ✅ |
| **4** | `tool_calls` | **4** / 33 | `{"color":"red","brightness":70}` ✅ |
| 64 (loose) | `tool_calls` | 39 / 93 | `{"color":"blue","brightness":70}` ✅ |
| **none** (fallback) | `tool_calls` | – / 20 | `{}` ⚠️ malformed — the exact failure line① fixes; call still present (non-regressive) |
| named + budget=12 | `tool_calls` | 9 / 41 | `{"color":"blue","brightness":70}` ✅ |
| `auto` + budget=8 | `length` | 7 / 512 | *(not engaged — free-form)* ✅ |

`rtok` capped **exactly** at the budget (2, 4) — well below the natural ~39 — proves
the force-close fires for a tool request, and the gate then produces a **valid,
schema-constrained** call every time even though the requested value is outside the
enum.

### Tests

Behavioral: `reasoning_gate_id` property; `allow_tools` opt-out-lift + prefix
threading; gate-mechanics (hold → open → exclude boundary); eligibility predicate;
all-or-nothing opener exclusion (r3 #3); un-priceable-schema decline (r3 #2);
synchronous in-slot seed probe (r3 #5); plus a route wiring guard for the Option-B
coupling. `tests/test_grammar_processor_558.py` 118 passed; the broader tool-grammar
+ budget suites pass (1 pre-existing offline HF-fetch flake unrelated to this
change).

### codex hardening

**Round 1 (4 blocking → fixed):** seed-probe render fallback; gate declines on
`reasoning_stop_conflicts`; `think_excluded_token_ids`; `_line1_completion_limit_ok`.

**Round 2 (3 blocking → fixed):** single-special-token opener exclusion as a
gating PRECONDITION; schema-scaled completion floor; seed render off the event loop.

**Round 3 (5 blocking → 3 fixed + 2 honestly scoped):**
- #5 DoS — the seed render now runs INSIDE the admission-gated build slot (was on
  the shared default executor before admission).
- #3 mixed-trigger leak — opener exclusion is now all-or-nothing across the forced
  set.
- #2 un-priceable schema — a bounded request whose schema minimum the flat floor
  can't cover now declines the gate.
- #1 `max_tokens=None` — documented residual (the route lacks the resolved context
  window; the small coupled budget bounds the exposure; a near-ceiling prompt fails
  with or without line①). A context-window plumbing pass can make this a hard check.
- #4 subtoken-spelled opener — the "provably guaranteed" claim is **withdrawn**; the
  opener mask is best-effort (blocks the natural atomic opener). Full structural
  closure (stateful multi-token suppression, or parser-level quarantine of
  in-`<think>` markers) is a tracked follow-up.

Closes the last #558 wire-constraint keystone for forced-call **arguments**.
Follow-ups (tracked): the model-emitted-`<think>` ("emit") seed scope;
multi-token-opener families; the two round-3 residuals above.


raullenchai:
⚠️ **HOLD — converting to draft.** pr_validate/codex review surfaced a design gap that invalidates this PR's central non-regression claim:

`_build_reasoning_budget_processor` (chat.py:2471) opts the generation-time thinking budget (#1185) **out of every request that carries tools** (`if request.tools: return None`) — deliberately, to avoid vLLM #44676 (a mid-span force-close injecting `</think>` into an in-progress tool call). So #1185 does **not** force-close `</think>` for tool requests, which means line①'s runtime gate is **not** guaranteed to open — it only opens on the model's *natural* `</think>`. A reasoning model that rambles inside `<think>` past `max_tokens` never closes, the gate never opens, and the forced call is dropped — a regression vs the forced-prefix fallback (which always yields a call).

Empirically confirmed: a forced tool request with `reasoning_max_tokens=4` vs `512` produces the **same** completion length (no generation-time shortening; `rtok=4` is only the post-hoc trim), whereas the non-tool control shows a real force-close.

What line① **does** deliver (verified): when the model closes `<think>` naturally, the grammar hard-enforces the forced call's args (decisive proof: enum `{zorp,quix,vlim}` + user demanding "RED" → output `zorp`). So it's a real *best-effort* improvement, not the *guarantee* the body claims.

**Sound fix (Option B):** have line① install its **own** generation-time budget (bypassing the #1185 tool opt-out). This is safe despite #44676 because line①'s grammar holds the tool call off until `</think>` — no tool-call token can exist when the budget force-closes, so a mid-call injection is structurally impossible. This overrides #1185's deliberate safety opt-out, so it's a W1 architectural call — holding for @raullenchai's go before implementing + merging.

Issue #1187, opened 2026-07-23
Image functionality on gemma4 now serves on one model but maxes out the ram with "test" via hermes. fails with another model


SmoothSailingMusic:
### Rapid-MLX version

rapid-mlx 0.10.17

### Hardware

48gb M4 Max

### macOS version

26.5.2 

### Python version

Python 3.14.6

### Model

unsloth/gemma-4-26b-a4b-it-UD-MLX-4bit loads mlx-community/gemma-4-26B-A4B-it-qat-OptiQ-4bit fails

### Full serve command

rapid-mlx serve mlx-community/gemma-4-26B-A4B-it-qat-OptiQ-4bit --enable-auto-tool-choice --tool-call-parser gemma4 --max-concurrent-requests 1 --reasoning-parser gemma4 --max-tokens 128000 --prefill-step-size 21000

### Streaming or non-streaming?

Streaming (`stream: true`)

### What happened? What did you expect?

First "test" command failed from hermes against the unsloth/gemma-4-26b-a4b-it-UD-MLX-4bit
Received the below error
ERROR:rapid_mlx.mllm_scheduler:Batch generation failed: Total prompt tokens (20010) exceeds the per-batch cap (8192 = prefill_step_size 8192 × 1 request(s)). For image inputs, downscale the image; for text inputs, shorten the prompt or restart the server with --prefill-step-size set higher.

Added --prefill-step-size 22000 to the command.

Next "test" spiked the ram like crazy. Have not seen this before.
Before
<img width="805" height="689" alt="Image" src="https://github.com/user-attachments/assets/787e3693-3793-4f56-9398-2a253f21c0f0" />
after
<img width="698" height="251" alt="Image" src="https://github.com/user-attachments/assets/fe935bd5-1aad-4020-bed2-20634923188c" />

Forcing it back to text only and removing the prefill step size results in a still high but usable spike

<img width="634" height="269" alt="Image" src="https://github.com/user-attachments/assets/0dfc5e85-519d-460e-8654-c7fea75af708" />

OptiQ one fails outright here

<img width="1023" height="284" alt="Image" src="https://github.com/user-attachments/assets/db80fd8f-649a-4116-aed6-0eea820690b7" />

A common thread with both of those models is "proprietary" optimization re: UD and OptiQ.
Maybe related?

### Minimal reproduction

[...]

PR #1193, opened 2026-07-23
fix(mllm): chunk text-only prefill to bound peak memory on long prompts (#1187 B)

raullenchai:
## Problem (issue #1187, Problem B)

A VLM served on the MLLM path (e.g. `gemma-4-26b`) spikes RAM to the machine limit when a **text-only** prompt is long. The #1187 reporter sends `"test"` through a large Hermes tool schema that expands to ~20k tokens and maxes out a 48 GB M4 Max. The plain text lane (`--no-mllm`) stays "high but usable" for the same prompt.

## Root cause (measured, not guessed)

`MLLMBatchGenerator._run_vision_encoding` ran the whole prompt through the language model in a **single forward**. That:

1. keeps activations for **every** prompt position live at once, and
2. projects logits over every position (`logits_from_hidden` → `[1, seqlen, vocab]`, gemma-4 vocab = 262144) even though only the last row is sampled.

mlx-lm's own text lane never does this — `PromptProcessingBatch.prompt` prefills in `prefill_step_size` chunks, evaluates **only the KV cache state** per chunk (mlx prunes the unused `lm_head` projection), and starts generation from the last token. On the MLLM path the per-batch cap (#682) forces operators to raise `--prefill-step-size` above the prompt length (reporter uses `22000`), so using it directly as the chunk size defeats chunking entirely.

End-to-end measurement on `mlx-community/gemma-4-26b-a4b-it-4bit`, 20010-token prompt, `--prefill-step-size 22000`:

| | peak | prefill time | sampled token |
|---|---|---|---|
| single forward (before) | **35.15 GB** | 33.7 s | 236772 |
| chunked, step 2048 (after) | **18.38 GB** | 12.7 s | 236772 ✓ |

The dominant driver is the un-chunked prefill activations — an early micro-benchmark that isolated only the `[1, seqlen, vocab]` logits tensor (~21 GB in a vacuum) was misleading, because in the full model mlx frees activations before projecting logits, so they don't stack. Reproducing on the real model is what pinned it.

## Fix

For a long **text-only** prompt **with a cache**, prefill the prefix in chunks of `min(prefill_step_size, 2048)` — evaluating only the cache state and `clear_cache()`-ing transient buffers per chunk — then run one last-token forward for the `[1, 1, vocab]` logits actually sampled.

- Decoupling the chunk size from the per-batch cap means the reporter's `--prefill-step-size 22000` no longer forces a single forward, while an operator who set a *smaller* value still gets chunks no larger than they asked for.
- Output is numerically identical (causal attention: the last token attends to the whole prefix through the cache).
- **Scope is deliberately text-only.** With images, pixel features must stay aligned with their placeholder tokens in one vision-merge forward, and image prompts are bounded by the per-batch cap. No-cache call sites (unit tests) keep the single forward — the split needs a real KV cache to carry prefix state.

`2048` measured the lowest peak (18.4 GB vs 20.0 GB at 4096) and matches mlx-lm's text-lane default.

## Verification

- **Unit** (`tests/test_mllm_batch_generator.py`, 29 tests total, +11 new): chunk boundaries at `min(step, 2048)`; smaller `--prefill-step-size` shrinks the chunk; image requests are **not** chunked (single vision-merge forward, pixel_values preserved); no-cache keeps the single forward; short prompts (≤ chunk) keep the single forward; a memory test proving the chunked path never materializes `[1, seqlen, vocab]` logits; a numerical-equivalence test (real `KVCache`/`RotatingKVCache`) locking chunked ≡ single forward to ~2e-7. Existing `_run_vision_encoding` kwargs regressions still pass.
- **End-to-end server round-trip**: served gemma-4-26b on the MLLM path with `--prefill-step-size 22000`; a needle-in-haystack prompt (multiple chunks) recalls the buried fact correctly — confirming the chunked KV cache is correct across chunk boundaries **and** through the scheduler's cache merge + decode. Short prompt (single-forward path) unaffected.
- **Inference-path regression guard** (SOP Step 8): `bench qwen3.5-4b-4bit --tier smoke` PASS (boots + generates on the shared text path; my change is isolated to the MLLM `_run_vision_encoding`).

## Review convergence (codex)

Iterated to **0 blocking**. Findings addressed:

1. **Short-prompt latency** — the chunked branch initially ran for *every* text-only prompt > 1 token, adding a second forward + `mx.eval`/`mx.clear_cache` barrier to ordinary short prompts. Gated on `input_ids.shape[1] > chunk`, so prompts that fit in one chunk keep the original single forward. Latency is unchanged for the common case; chunking engages only where the un-chunked activations + full-sequence logits would actually spike.
2. **`attention_mask` correctness** — the chunked path drops `attention_mask` (it carries no info for a single un-padded request). Now *enforced* via `_attention_mask_is_droppable`: a non-all-valid mask (padding / shorter valid span) keeps the single forward that passes the mask through intact, instead of silently changing attention semantics. (`attention_mask` is a separate request field, excluded from `extra_kwargs`, so an all-valid mask never disabled the fix — proven by a processor-shaped test.)
3. **Cross-chunk RoPE positions** — verified (not assumed) that chunking is positionally correct: chunked forwards rely on the model deriving RoPE from `cache.offset`, the same contract every decode step and mlx-vlm's own chunked prefill use (Qwen3-VL computes `arange(L) + cache_offset` per text-only chunk; Gemma3/4 read the rotating-cache offset). The numerical-equivalence test locks this in. A model that ignored the offset for L>1 forwards could not decode token 2 either.

## Pre-existing flakes (not introduced here — Step 10 taxonomy)

Local `pr_validate` `full_unit` reports 2 failures, both in files **not in this diff**, and **CI is green on this exact HEAD** (`test-apple-silicon` + `test-matrix` all pass):

- `test_batching_deterministic.py::...test_batched_faster_than_sequential` — timing perf test; **passes in isolation**, flakes under the concurrent load of the full-suite run.
- `test_tool_grammar_558.py::...test_deepseek_r1_prefilled_think_template_is_tolerated` — HF hermetic-offline fetch; the file's skip-guard misses the generic `OSError` transformers 5.x re-raises. Passes on CI (network available).

Refs #1187 (Problem B). Problem A was fixed in #1189.

### DIFF
--- a/tests/test_mllm_batch_generator.py
+++ b/tests/test_mllm_batch_generator.py
@@ -248,21 +248,218 @@ def test_run_vision_encoding_no_cache_keeps_single_forward_for_long_text():
     assert model.calls[0][0] == 5000
 
 
-class _BigProjModel:
-    """Stub whose forward allocates a real ``[1, seqlen, vocab]`` logits
-    projection (quantized ``as_linear``), so peak memory reflects whether the
-    caller projects the whole prompt or just the last token."""
+def test_chunking_falls_back_to_single_forward_with_extra_kwargs():
+    """If a processor ever emits sequence-aligned extra kwargs (e.g.
+    ``token_type_ids``) for a text-only request, we must NOT chunk (we would
+    silently drop or mis-slice them) — fall back to the single forward that
+    forwards ``kwargs`` intact."""
+    model = _ChunkRecordingModel()
+    gen = _make_bare_generator(prefill_step_size=22000, model=model)
+    req = _make_ids_request(5000)
+    req.extra_kwargs = {"token_type_ids": mx.zeros((1, 5000), dtype=mx.int32)}
+    cache = [_FakeCache()]
 
-    def __init__(self, hidden: int, vocab: int):
-        self.embed = nn.QuantizedEmbedding(vocab, hidden, group_size=64, bits=4)
-        mx.eval(self.embed.parameters())
-        self.hidden = hidden
-        self.vocab = vocab
-        self.language_model = object()
+    gen._run_vision_encoding(req, cache=cache)
+
+    # One forward over the whole prompt, extra kwargs preserved.
+    assert len(model.calls) == 1
+    assert model.calls[0][0] == 5000
+    assert "token_type_ids" in model.calls[0][1]
+
+
+def test_run_vision_encoding_single_token_uses_single_forward():
+    """A 1-token prompt has no prefix to chunk; it stays on the single
+    forward (no empty-prefix forward is ever submitted)."""
+    model = _ChunkRecordingModel()
+    gen = _make_bare_generator(prefill_step_size=2048, model=model)
+
+    gen._run_vision_encoding(_make_ids_request(1), cache=[_FakeCache()])
+
+    assert len(model.calls) == 1
+    assert model.calls[0][0] == 1
+
+
+# ---------------------------------------------------------------------------
+# Numerical equivalence — chunked prefill must match the single forward.
+#
+# `_TinyCausalLM` is a real (tiny) causal transformer using mlx-lm's own
+# `KVCache` / `RotatingKVCache` + causal-mask helper, so the chunked path
+# actually computes and retains prefix K/V. We compare the last-position
+# logits, the cache offset, the sampled token, AND a following decode step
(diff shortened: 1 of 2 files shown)


SmoothSailingMusic:
My man you are FAST. Big cheers on your turn around time!

PR #1202, opened 2026-07-24
fix(mtp): reconcile info vs serve MTP-eligibility messaging

raullenchai:
## Why

`rapid-mlx info <native-MTP checkpoint whose head weights were stripped at convert>` (e.g. `qwen3.6-27b-8bit` → `unsloth/Qwen3.6-27B-MLX-8bit`, whose `config.json` still declares `text_config.mtp_num_hidden_layers=1` but ships 0 `mtp.*` tensors) reported a bare:

```
Spec decode      : ✗ disabled (no MTP/drafter trained)
MTP path         : disabled
```

…while `serve <same model> --speculative-config '{"method":"mtp"}'` first announced `MTP: enabled via --speculative-config` and only THEN hard-failed at inject time (the serve gate is CONFIG-based → sees the declaration → announces enabled; inject discovers the weights are absent). Both surfaces were individually correct, but a user who ran `info` then `serve` got whiplash.

## What

Add an actionable `MTP:` note under the `info` profile table that mirrors the exact serve-time hard-fail guidance and points to the resolution — the **sidecar path**:

```
  MTP: config declares mtp_num_hidden_layers=1 but the MTP head weights are absent in this checkpoint.
  Attach a separately-published MTP head, e.g.:
    rapid-mlx serve qwen3.6-27b-8bit --speculative-config '{"method":"mtp","model":"<head-repo>"}'
  (or re-convert the base checkpoint with an MTP-preserving convert).
```

The note fires exactly when the config declares an MTP head (`_detect_mtp_eligibility_verbose` != NONE, reusing the serve-time detector) but the resolved profile has spec decode wired off (head weights stripped). Now `info` and `serve` agree and both point to the sidecar.

## Design / safety

- **No gate/eligibility BEHAVIOR change.** The serve-time gate still correctly hard-fails; `detect.py` is untouched. This only reconciles the `info` REPORTING surface.
- The fixed-width profile table stays weight-free / config-free and deterministic — its cells are unchanged, preserving the `MTP path` controlled-vocabulary contract.
- The `config.json` read (LOCAL HF cache only, no network) lives solely in the new `_mtp_declared_absent_layers` helper, is pre-filtered on the resolved profile so it only touches disk for the exact whiplash-prone case (native-MTP family + spec decode off), and degrades to no-note when the config isn't cached.

## Tests

- 6 deterministic unit tests for `_mtp_declared_absent_layers` (config read monkeypatched; covers declared-but-absent, spec-on short-circuit, non-native family, uncached, config-declares-none, and a guard that `format_profile_table` stays config-free).
- 2 `info` note tests (note appears when declared-but-absent; suppressed otherwise).
- Full `test_model_auto_config.py` + `test_cli_info.py` + MTP/DFlash/DDTree suites pass.

### DIFF
--- a/tests/test_cli_info.py
+++ b/tests/test_cli_info.py
@@ -99,9 +99,12 @@ def test_info_prints_mtp_sidecar_note_when_head_declared_but_absent(
     out = _run_info("qwen3.6-27b-8bit")
     assert "mtp_num_hidden_layers=1" in out, f"declared-layer count missing:\n{out}"
     assert "head weights are absent" in out, f"absent-head diagnosis missing:\n{out}"
-    # Actionable: mirrors the serve-time sidecar remedy.
-    assert "--speculative-config" in out and '"method":"mtp"' in out, (
-        f"sidecar serve command missing from note:\n{out}"
+    # Actionable: must emit the COMPLETE sidecar remedy — the ``model`` field
+    # is load-bearing. Asserting the full fragment (not just ``method``)
+    # guards against regressing to a bare ``{"method":"mtp"}`` that would just
+    # reproduce the original inject-time hard failure (codex #1202 round 2).
+    assert """--speculative-config '{"method":"mtp","model":"<head-repo>"}'""" in out, (
+        f"complete sidecar serve command missing from note:\n{out}"
     )
     # The serve command must carry the user-typed alias, not the HF path.
     assert "rapid-mlx serve qwen3.6-27b-8bit" in out, (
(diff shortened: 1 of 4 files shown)


raullenchai:
Closing: low value vs cost. This only reconciles the info-vs-serve MTP-eligibility *messaging* (cosmetic). It didn't converge under adversarial review (codex kept surfacing legitimate but ever-deeper robustness findings across a broad diff over 7 rounds). Not worth it. The one genuinely useful hunk — separating the printed serve target with `--` so a `-`-prefixed path can't be misparsed — can be cherry-picked standalone if desired.

PR #1206, opened 2026-07-24
fix(mtp): correct false comment on MTP sidecar float-dtype role check

raullenchai:
## Follow-up to PR #1201 codex finding

PR #1201 (`fix(mtp): eliminate intermittent empty response on Qwen3.6 MTP sidecar path`) added a by-role dtype check in `vllm_mlx/spec_decode/mtp/qwen3_5_inject.py`. An independent codex review run raised one more finding that 2 of 3 review runs did NOT flag:

> [BLOCKING] the dtype check accepts each floating tensor independently (integer-role params like the packed `weight` must match dtype EXACTLY, but floating-role params `scales`/`biases` need only be *some* floating dtype). So a sidecar with shape-correct but MISMATCHED `scales`/`biases` float dtype passes validation and can still fail later in `mx.quantized_matmul` at the first draft step — recreating the mid-stream empty-response failure.

The code defends the loose float check with a comment: *"a real sidecar legitimately ships bf16 scales against the freshly-quantized fp32 module and `load_weights` casts them, so an exact float check would false-reject valid checkpoints."*

## Investigation

**Fact-check #1 — the comment's stated mechanism is wrong.** Read `mlx.nn.Module.load_weights` / `Module.update` source directly:

```python
# nn.Module.update
if isinstance(current_value, mx.array):
    ...
    dst[k] = new_value   # <-- bare assignment, no cast
```

`load_weights(strict=False)` calls `self.update(tree_unflatten(weights), strict=False)`, which assigns arrays verbatim. **It does not cast dtypes.** The comment's rationale is factually wrong.

**Fact-check #2 (the decisive part) — does that make the loose check actually unsafe?** That depends entirely on whether `mx.quantized_matmul` / `mx.gather_qmm` reject a floating-dtype mismatch between `x`/`weight`/`scales`/`biases`, which is an empirical MLX question, not something you can reason about from the Python source. I wrote a standalone probe building `nn.QuantizedLinear` (the exact call the MTP `fc`/attention layers make: `mx.quantized_matmul(x, weight, scales=scales, biases=biases, transpose=True, group_size=.., bits=..)`) and forced mismatched floating dtypes onto `scales`/`biases` in both directions (fp32 module + bf16/fp16 scales, and bf16 module + fp32 scales), plus the MoE path (`SwitchGLU`/`QuantizedSwitchLinear` → `mx.gather_qmm`).

**Result: every combination TOLERATES the mismatch — none raise.** MLX silently promotes/computes across floating dtypes rather than enforcing an exact match. Ran on both `mlx==0.31.2` (this repo's pinned floor) and `mlx==0.32.0`, on the default GPU (Metal) device — same result both times, 10/10 cases tolerate:

```
control_fp32_matches: tolerate
scales_bf16_vs_fp32: tolerate
biases_bf16_vs_fp32: tolerate
scales_fp16_vs_fp32: tolerate
both_bf16_vs_fp32: tolerate
scales_fp32_vs_bf16: tolerate
x_bf16_scales_fp32: tolerate
gather_qmm_control: tolerate
gather_qmm_scales_bf16: tolerate
gather_qmm_biases_fp16: tolerate
```

Sample precision-diff evidence (same weight, only `scales` dtype differs — fp32 vs bf16):
```
CONTROL (fp32 scales): out[0,:4]=[0.34953, -0.30505, -0.37162, -0.24791]
MISMATCH (bf16 scales): out[0,:4]=[0.35309, -0.29941, -0.37214, -0.24651]
```
Small bf16-precision drift, no crash, no NaN/Inf, finite output.

Per STEP 2 of the task instructions ("only if STEP 1 shows it CAN raise") — since STEP 1 showed unconditional tolerance across every combination and both pinned MLX versions, an end-to-end sidecar-corruption repro was not needed to demonstrate a crash (there is none to reproduce). Instead I built the injector-level integration test directly (see below), which proves the same thing from the other end: a real corrupted-dtype sidecar passes injection and drafts without raising.

## Verdict: NOT reproducible (false positive on the failure mechanism)

The codex finding's factual premise about the comment ("load_weights casts them") is **wrong**, but its predicted failure ("fails in mx.quantized_matmul at the first draft step") does **not** happen — `quantized_matmul`/`gather_qmm` tolerate the mismatch. The loose by-role check is safe, just not for the reason the code claimed.

## Changes

1. **`vllm_mlx/spec_decode/mtp/qwen3_5_inject.py`** — corrected the comment: `load_weights` does NOT cast; the loose floating check is safe because `mx.quantized_matmul`/`mx.gather_qmm` tolerate a floating-dtype mismatch (verified, cites the new tests below). If a future MLX version changes this, the tests are the tripwire and the check needs tightening to an exact match, mirroring the integer branch.
2. **`tests/test_mtp_spec_decode.py`** — two new regression tests:
   - `test_quantized_matmul_and_gather_qmm_tolerate_mismatched_floating_dtype` — direct MLX-behavior canary (the probe above, as a test). Fails first if a future MLX bump starts rejecting floating-dtype mismatches.
   - `test_inject_accepts_sidecar_with_mismatched_floating_scales_dtype_and_drafts_correctly` — injector-level integration test: builds a valid quantized synthetic sidecar, corrupts one non-`fc` `.scales` tensor's dtype (fp32 → bf16, same shape/values), confirms `inject_mtp_support` still returns `True` (the loose check correctly accepts it) and the MTP draft forward (`mtp_forward`) runs to completion with finite output — the exact call path the codex finding predicted would crash.

## Test plan

- [x] `ruff format` + `ruff check` on both changed files — clean
- [x] `pytest tests/test_mtp_spec_decode.py -q` — 94/94 passed (including the 2 new tests)
- [x] Probe run on both `mlx==0.31.2` and `mlx==0.32.0` — same tolerant result

Does NOT touch operator spec-decode initiative lanes (`vllm_mlx/spec_decode/spec_decode/`, `vllm_mlx/speculative/*`).

### DIFF
--- a/tests/test_mtp_spec_decode.py
+++ b/tests/test_mtp_spec_decode.py
@@ -2048,18 +2048,30 @@ def test_quantized_matmul_and_gather_qmm_tolerate_mismatched_floating_dtype():
     )
 
     # The MoE path (SwitchGLU / QuantizedSwitchLinear -> gather_qmm) —
-    # exercised when a Qwen3.6 MoE MTP head layer is present.
+    # exercised when a Qwen3.6 MoE MTP head layer is present. Cover BOTH
+    # directions here too (matching the quantized_matmul cases above and
+    # the "in both directions" claim in the qwen3_5_inject.py comment):
+    # (a) fp32 module/input with mismatched bf16/fp16 sidecar scales, and
+    # (b) the reverse — a bf16 module/input with mismatched fp32 scales,
+    # which is exactly the "fp32 scales for a bf16 MoE module" config the
+    # by-role check accepts.
     from mlx_lm.models.switch_layers import SwitchGLU
 
     num_experts, top_k = 4, 2
-    switch = SwitchGLU(in_dims, in_dims, num_experts)
-    _nn.quantize(switch, group_size=group_size, bits=bits)
-    _mx.eval(switch.parameters())
-    x_moe = _mx.random.uniform(shape=(1, 3, in_dims)).astype(_mx.float32)
     inds = _mx.array([[0, 1], [1, 2], [2, 3]]).reshape(1, 3, top_k)
-    gate = switch.gate_proj
 
-    def gather_qmm_ok(scales, biases):
+    def make_switch(param_dtype):
+        switch = SwitchGLU(in_dims, in_dims, num_experts)
+        # Cast each SwitchLinear's FP weight before quantizing so the
+        # derived scales/biases land in ``param_dtype``.
+        for _name in ("gate_proj", "up_proj", "down_proj"):
+            _lin = getattr(switch, _name)
+            _lin.weight = _lin.weight.astype(param_dtype)
+        _nn.quantize(switch, group_size=group_size, bits=bits)
+        _mx.eval(switch.parameters())
+        return switch
+
+    def gather_qmm_ok(gate, x_moe, scales, biases):
         out = _mx.gather_qmm(
             x_moe,
             gate.weight,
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing without merging. Investigation conclusion: the codex finding (#695 loose float-dtype role check) is a **false alarm**. A standalone probe forced mismatched floating dtypes (fp32↔bf16/fp16, both directions) onto scales/biases through `mx.quantized_matmul` and `mx.gather_qmm` — **10/10 combinations tolerate the mismatch** (silent float promotion, finite output), on both `mlx==0.31.2` (the pyproject floor) and `mlx==0.32.0`. So the loose by-role dtype check is correct and needs no code change; the empty-response class is fully closed by #1201.

The only real defect this surfaced was a factually-wrong code comment ('load_weights casts them' — it does not; `Module.update` assigns verbatim). That is a one-line doc wart, not worth a dedicated PR for a false alarm; it can ride along a future change touching this file. The durable finding is recorded in project memory. Closing.

PR #1208, opened 2026-07-25
fix(kv-cache): probe nested language-model head_dim so live int8 KV engages on VLMs (#1199)

raullenchai:
## Why this matters

This fixes a **silent, headline-defeating regression in coverage** for #1199: **because** the live-KV quantization probe only looked at top-level `model.args`, the `--kv-cache-dtype int8/int4` flag became a **no-op on the flagship default model** — precisely the memory win #1199 was built to deliver. The **why** is subtle (a nested VLM config), so it slipped past #1199's own tests and only surfaced under a real serve dogfood.

## Problem

`--kv-cache-dtype int8/int4` silently fell back to a **bf16** live continuous-batching KV cache on the **flagship default model** `mlx-community/Qwen3.5-4B-MLX-4bit`, so the requested dtype delivered **zero live KV memory reduction** there. Surfaced by a local-wheel dogfood of the post-v0.11.0 PRs (#1199/#1200/#1201).

At serve time the operator saw:

```
WARNING [kv-cache] live KV quantization disabled: head_dim (K=None, V=None) unknown
or not divisible by any supported group_size (32/64/128); serving a bf16 live cache
```

## Root cause

`probe_head_dim` read only the **top-level** `model.args.head_dim` (with a `hidden_size // num_attention_heads` fallback). `Qwen3.5-4B-MLX-4bit` is a **multimodal wrapper (VLM)** — its top-level `args` carry no attention dims (`vision_config`/`text_config` layout). The language tower's real dims live on a nested `language_model.args` (**head_dim=256**, divisible by 128) or `args.text_config`. The probe returned `None`, so `resolve_kv_quantization` judged the model unprobeable and `_init_kv_quantization` disabled the live quantized cache.

## Fix

Add `_text_attention_args(model)`: prefer the top-level args when they already carry a head dim (**unchanged for text-only models**), otherwise descend into `language_model.args`, then `args.text_config`. Both `probe_head_dim` and `probe_kv_head_dims` route through it, and `v_head_dim` is read from the same resolved args object.

This **widens best-effort coverage; it is not a correctness change**. The retained prefix cache was never gated by this probe (it self-coerces per layer), and an unprobeable model still **fails safe** to a bf16 live cache.

## Verification

Real checkpoint (`mlx-community/Qwen3.5-4B-MLX-4bit`):
- `probe_head_dim(model)` → **256** (was `None`)
- `resolve_kv_quantization(256, 256, 64)` → `group_size=64, live_disabled=False`

Fresh-venv serve `--kv-cache-dtype int8 --disable-prefix-cache`, before → after this change:
- log: `live KV quantization disabled ... bf16 live cache` → **`live continuous-batching KV cache quantized to int8 (group_size=64)`**
- 8/8 concurrent requests coherent under continuous batching
- int8 vs bf16 greedy (temp=0) output **byte-identical**
- decode ~147 tok/s (no regression from the now-active quantized live cache)

## Test plan

- [x] `pytest tests/test_quantized_batch_cache.py` — **43/43 pass** (adds nested-VLM cases to `test_probe_head_dim` and `test_probe_kv_head_dims`, covering both `language_model.args` and `args.text_config` descent; existing text-only + `hidden_size` fallback + MLA-conservative cases unchanged)
- [x] `ruff check` clean on both changed files
- [x] End-to-end fresh-venv serve on the real flagship VLM confirms the disabled WARNING is gone and the live int8 cache engages, output coherent
- [x] Backward-compat: text-only models still resolve head_dim from top-level args (identical path), object-with-no-args still returns `None`

Docs-free, engine-local; no aliases/CLI/spec-decode changes.

### DIFF
--- a/tests/test_quantized_batch_cache.py
+++ b/tests/test_quantized_batch_cache.py
@@ -537,6 +537,18 @@ def test_probe_head_dim():
 
     assert probe_head_dim(_VLM4()) is None
 
+    # VLM signalled by `language_model` presence even when the submodule exposes
+    # no `.args` at all: the presence of the submodule is the multimodal signal,
+    # so we must still refuse the top-level vision dims and fail safe (#1208).
+    class _NoArgsLM:
+        pass  # language_model submodule with no `.args`
+
+    class _VLM5:
+        args = _VisionishTop()
+        language_model = _NoArgsLM()
+
+    assert probe_head_dim(_VLM5()) is None
+
 
 def test_update_adjusts_group_size_for_head_dim_96():
     # head_dim=96 is not divisible by 64 but is by 32 — must auto-adjust, not crash
--- a/tests/test_quantized_batch_cache.py
+++ b/tests/test_quantized_batch_cache.py
@@ -507,6 +507,20 @@ def test_probe_head_dim():
 
     assert probe_head_dim(_VLM2()) == 256
 
+    # Mixed: top-level args expose MISLEADING vision/composite dims
+    # (hidden_size // num_attention_heads = 64) while the authoritative language
+    # head_dim is 256. The nested language config must win — preferring the
+    # top-level fallback here would mis-size the live cache (#1208).
+    class _VisionishTop:
+        hidden_size = 1024
+        num_attention_heads = 16  # -> 64, the WRONG (non-language) head dim
+
+    class _VLM3:
+        args = _VisionishTop()
+        language_model = _LM()  # language head_dim = 256
+
+    assert probe_head_dim(_VLM3()) == 256
+
 
 def test_update_adjusts_group_size_for_head_dim_96():
     # head_dim=96 is not divisible by 64 but is by 32 — must auto-adjust, not crash
@@ -669,6 +683,18 @@ def test_probe_kv_head_dims():
         language_model = _LMV()
 
     assert probe_kv_head_dims(_VLMV()) == (256, 128)
+
+    # Mixed: misleading top-level dims must NOT win over the nested language
+    # config, and v_head_dim is read from the SAME (nested) args (#1208).
+    class _VisionishTopV:
+        head_dim = 64  # wrong (non-language) dim; must be ignored
+        v_head_dim = 64
+
+    class _VLMV2:
+        args = _VisionishTopV()
+        language_model = _LMV()  # language head_dim=256, v_head_dim=128
+
+    assert probe_kv_head_dims(_VLMV2()) == (256, 128)
 
     # Unknown -> (None, None).
     assert probe_kv_head_dims(object()) == (None, None)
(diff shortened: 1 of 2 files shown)


raullenchai:
## Validation summary (ready to merge)

**GH CI — all green:** `validate` ✅, `type-check` ✅, `test-matrix (3.10 / 3.11 / 3.12)` ✅, `test-apple-silicon` ✅, `tests` ✅, `lint` ✅.

**End-to-end proof (fresh-venv serve on the real flagship VLM `Qwen3.5-4B-MLX-4bit`):**
- Before: `WARNING [kv-cache] live KV quantization disabled: head_dim (K=None, V=None) unknown … serving a bf16 live cache`
- After: `[kv-cache] live continuous-batching KV cache quantized to int8 (group_size=64)` — the live int8 path now actually engages
- 8/8 concurrent requests coherent under continuous batching; int8 vs bf16 greedy output byte-identical; ~147 tok/s (no regression)

**Targeted tests:** `test_quantized_batch_cache.py` 43/43, including new nested-VLM cases on both probes.

### Codex convergence (adversarial review, gpt-5.6-sol)

Four commits, each closing a real slice of the same class ("a VLM resolving to top-level dims"):
1. Initial fix — probe descends into `language_model.args` / `args.text_config`.
2. `[BLOCKING]` codex: top-level-first could short-circuit to vision/composite dims → **prefer the language config first**.
3. `[BLOCKING]` codex: an unprobeable nested config still fell back to top-level → **never fall back to top-level once a language config exists**.
4. `[BLOCKING]` codex: `language_model` present but with no `.args` still fell through → **the multimodal signal is the *presence* of `language_model`/`text_config`, not its probeability** (complete-by-construction: a detected VLM never returns top-level args).

After (4), **4 independent codex passes on the final commit `76a92804`: 3 fully clean, 1 non-blocking `[NIT]`, zero `[BLOCKING]`**. The lone NIT posits a "text-only model that also carries a `language_model` attribute" — architecturally contradictory (a `language_model` submodule *is* the multimodal signal), and even if it occurred the result is fail-safe (bf16 live cache = the pre-PR status quo), never a crash or wrong output. This is the intended safety/optimization tradeoff: mis-sizing would crash a request; the hypothetical missed-optimization merely keeps bf16.

### On the local full-suite failures (environmental, not this change)

A direct `pytest` on my box shows failures that are all environmental and reproduce independently of this diff:
- 3 × HF config-fetch (`OSError: couldn't connect to huggingface.co`) — reproduced on **clean `origin/main`** (transient offline box).
- 10 × `tests/integrations/test_hermes.py` — live-server integration tests targeting `localhost:8000`; that port hosts an unrelated operator service on this box, so they hit it instead of skipping. On a clean box / CI they skip (see `tests/integrations/conftest.py`), which is why GH's `tests` + `test-apple-silicon` jobs are green.

None touch `quantized_batch_cache.py`. pr_validate's own `full_unit` (proper selection) reported only the 3 HF-offline items.

PR #1216, opened 2026-07-25
feat(mtp): chain-of-K on hybrid targets + drafter-hidden cascade

raullenchai:
> **Draft — `spec_decode/` is the core-team lane. Opening for coordination/review before landing.**

## Why
Native-MTP speculative decoding on Qwen3.5/3.6 (hybrid GatedDeltaNet MoE) was clamped to K=1, and even K=1 left throughput on the table. This lifts the clamp and closes the gap toward llama.cpp.

## What (3 pieces)
- **Multi-slot recurrent rollback** (`cache_patch.py`): the GatedDeltaNet chunk-split now snapshots `(conv, ssm)` at **every** draft boundary (`rollback_states[n_from_end]`) instead of one offset, so `_rollback_draft(n)` restores to any accepted depth. This lets `generator.py` **drop the SSM K=1 clamp** — chain-of-K works on hybrid targets. K=1 reduces to the exact prior single-snapshot behavior.
- **Drafter-hidden cascade** (`qwen3_5_inject.py`): `mtp_forward` gains `return_hidden`, returning the head's pre-lm_head hidden. The generator already feeds it back as the next iteration's hidden, so each successive draft conditions on the drafter's **own** refined hidden rather than the target's frozen hidden. K=3 acceptance **~45% → ~71%**.
- **Lean-Python cascade** (`generator.py`): the draft cascade no longer host-syncs (`mx.eval`) per step — it stays lazy and materializes once at the verify's single eval, removing `2*K` GPU round-trips/round.

## Tests
- Full MTP unit suite green (200 passed, 2 skipped).
- Correctness: isolated test shows the chunk-split snapshot/restore is numerically exact (diffs `0.0`). Greedy served output coherent.
- Measured (Qwen3.6-35B-A3B Q4, M3 Ultra): **1.24× → 1.36×**, ~73–77% acceptance. Also validated on Qwen3.5-9B (dense hybrid).

## Notes
Three files are interdependent (multi-slot API is shared by cache_patch + generator; cascade spans inject + generator), so kept as one PR. Depends functionally on the sidecar fix (#1214) to have a head to draft with.

🤖 Prepared with AI assistance (Claude Code); `spec_decode/` review requested.

### DIFF
--- a/tests/test_mtp_multislot_rollback.py
+++ b/tests/test_mtp_multislot_rollback.py
@@ -1,16 +1,21 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Multi-slot GatedDeltaNet rollback for chain-of-K MTP.
+"""Single-pass GatedDeltaNet rollback for chain-of-K MTP.
 
-Locks the ``cache_patch`` contract that the verify chunk-split records a
-``(conv, ssm)`` snapshot at EVERY draft boundary (keyed by
-``n_from_end = 1..K``), and that each snapshot equals the recurrent state
-after processing exactly ``S - n_from_end`` tokens — so
-``_rollback_draft(n)`` can restore to any accepted depth without leaving
-rejected draft state in the cache.
+Locks the ``cache_patch`` contract that the verify forward runs as ONE
+fused ``gated_delta_update`` (no per-boundary segment splitting) and
+stashes a rollback CLOSURE on the cache. The closure recomputes the
+``(conv, ssm)`` state after keeping the first ``S - n_to_drop`` verify
+positions — byte-exact with the true ``(S - n_to_drop)``-token recurrent
+state — so ``_rollback_draft(n)`` can restore to any accepted depth on a
+partial chain-of-K accept without leaving rejected draft state in the
+cache.
 
-Regression for the chain-of-K-on-hybrid work (PR feat/mtp-hybrid-chain-of-k):
-before the multi-slot rewrite only a single boundary was snapshotted, so
-partial accept at depth < K was unrepresentable.
+Regression for the single-pass rewrite (PR perf/mtp-single-pass-gdn-
+rollback): the previous implementation split the verify into ``K+1``
+separate ``gated_delta_update`` calls to materialize a snapshot at every
+boundary — ~K+1 GDN kernel launches per layer per round, measured as the
+dominant MTP verify cost. The single-pass path pays one fused scan and
+only recomputes on the rare rejection.
 """
 
 import mlx.core as mx
@@ -45,16 +50,16 @@ def _build_gated_delta_net():
 
 
 @pytest.mark.parametrize("K", [1, 2, 3])
-def test_multislot_chunksplit_snapshots_every_boundary(K):  # noqa: N803  (K = draft width)
-    """Every ``n_to_drop`` in 1..K has a snapshot equal to the true
-    ``(S - n)``-token recurrent state."""
+def test_rollback_closure_recomputes_true_boundary(K):  # noqa: N803  (K = draft width)
+    """The stashed rollback closure recomputes the exact ``(S - n)``-token
+    recurrent state for every ``n_to_drop`` in 1..K."""
     mx.random.seed(0)
     from mlx_lm.models.cache import ArraysCache
 
     from vllm_mlx.spec_decode.mtp import cache_patch
 
-    # Install the chunk-split patch; keep a handle to the unpatched forward
-    # for reference (single-call, no snapshots).
+    # Install the single-pass patch; keep a handle to the unpatched forward
+    # for the independent reference states.
     cache_patch.patch_gated_delta_net_for_mtp()
     orig_call = cache_patch._orig_gated_delta_call
     assert orig_call is not None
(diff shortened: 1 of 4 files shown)


raullenchai:
Folded in the single-pass GatedDeltaNet rollback perf fix (commit 8dbe4fe): replaces the per-boundary chunk-split with one fused `gated_delta_update` + lazy recompute-on-reject. Validated on M3 Ultra (Qwen3.6-35B-A3B Q4, warm, same OpenAI-server harness): baseline 95 → MTP 149 tok/s = 1.57x (was 114/1.18x), ~90% accept, output byte-identical to baseline (lossless). This is what makes chain-of-K actually pay off — without it the chunk-split ate the K≥2 gain. 479 MTP/spec tests pass; full unit suite green.

rinaldofesta:
I'd like to take the Qwen3.8 extension of this, under core-team review as the header asks.

Context: I benchmarked the Qwen3.8-27B serving landscape on an M5 Max this weekend (native-MTP runtime, mlx-vlm draft-head, rapid-mlx 0.12.14 AR). Without speculation every engine lands at 13-15 tok/s; a native fused MTP head reaches 36.9. rapid-mlx had the best concurrency of the three (1.84x C=4 scaling, zero fairness spread), and with chain-of-K on hybrid targets the MTP machinery is one model generation away: Qwen3.8 is the same GatedDeltaNet hybrid family.

Proposed scope: (a) Qwen3.8 target support for chain-of-K (rollback slots on the 3.8 cache layout), (b) bit-exact greedy verification against a serial reference (I have a matched trunk+head pair and the mlxfast-challenge hidden-trajectory methodology), (c) per-depth acceptance/parks telemetry, (d) C=1/C=4 measured before/after with the harness that caught draft-MTP going slower than sequential in another engine.

Does that split make sense, and is Qwen3.8 wanted in this lane now or after this lands? I will wait for a go before touching spec_decode/.

rinaldofesta:
Validation of the shipped #1967 lane on Qwen3.8-27B, M5 Max 128 GB, rapid-mlx 0.12.14, explicit `--speculative-config '{"method":"mtp","model":"mlx-community/Qwen3.8-27B-MTP-4bit","num_speculative_tokens":3}'`:

Losslessness: 12/12 greedy text parity vs same-trunk AR across the six #1976 failure categories (factual, math, code, JSON, translation, planning), on both pairings. 8/8 on an agent-harness verified-code gate (4 real tasks via pi, checker-graded, no model grades its own work).

Speed (tok/s, AR to MTP, single stream): matched 4/4 is roughly break-even (27.8 to 27.5, 30.2 to 28.5, 28.1 to 26.1, 26.1 to 24.9, 22.5 to 21.8, planning 29.7 to 37.8). Mixed 8/4 gains 35-65% everywhere (16.0 to 17.5, 15.9 to 26.6, 15.6 to 24.9, 14.9 to 23.8, 13.5 to 18.9, 16.4 to 25.1).

The pairing economics invert #1258 on this model: the expensive 8-bit target leaves room for a cheap Q4 drafter, while the 4-bit target is already fast enough that drafting overhead cancels the gain. Pairing looks target-cost-dependent rather than universally matched-good; I am amending the warning text in #1983 accordingly.

Given lossless parity and verified agent tasks on both pairings: is there appetite to promote Qwen3.8's `supports_spec_decode` gate in `model_auto_config.py`, at least to documented explicit-opt-in? I have a 110-line implementation-readiness map for extending this PR's chain-of-K to Qwen3.8, ready to share. Raw JSONL available. Limits: single runs, one machine, greedy-text parity, n=6 prompts + 4 tasks.

rinaldofesta:
Follow-up: I tested the public challenge rerank head (amal-david/qwen38-mtp-head-q2-q4-rerank-v1) as a sidecar on the shipped 0.12.14 opt-in lane, same protocol as my previous comment.

Losslessness held: 6/6 greedy parity on both trunks. Speed: on the 8-bit trunk the head's core block gives +52% over AR (peaks 28.6 tok/s on math, 27.9 on code), but that is only ~8% over the stock Qwen3.8 head, because the loader drops the head's extra tensors (`draft_lm_head`, precision islands) as unknown sidecar keys. The challenge-grade numbers (3.8-4.1 tokens drafted, 84 tok/s in the Swift harness) live in the shortlist-32-then-rerank scheme those tensors enable, not in the weights themselves. On the 4-bit trunk the head is a net loss (-15%): drafting overhead without the rerank acceleration on a cheap target, a third datapoint for the pairing-economics picture in #1258/#1983.

Proposal, extending my earlier ask: the Qwen3.8 lane extension could include the rerank draft path (compact draft readout + reranked verification) as the drafting backend. No open engine runs it today, the reference artifacts are public and sha-pinned, and the acceptance telemetry this PR already adds is exactly what would validate it.

One cross-engine datapoint for motivation: on the same 8-bit trunk and battery, rapid-mlx's C=4 scheduling (34.4s makespan) currently trails oMLX (24.1s) and mlx-vlm AR (29.4s); rapid + a rerank-grade MTP that survives the batch scheduler would be the only engine holding both crowns.

Limits: single runs, one machine (M5 Max 128 GB), greedy-text parity, n=6 prompts per arm. Raw JSONL available.

raullenchai:
Blocking before this can be considered for landing: the branch has unresolved conflicts with current `main` in all three production MTP files (`cache_patch.py`, `generator.py`, and `qwen3_5_inject.py`), and its implementation/contract comments have drifted internally. For example, `generator.py` still describes a multi-slot snapshot dictionary while the latest commit implements lazy `rollback_recompute` closures; SSM detection still keys off `rollback_states` even though rollback correctness now depends on `rollback_recompute`.

This is losslessness-critical speculative-decoding state machinery, so I will not infer the intended conflict resolution or approve based on the old green CI run. Please rebase onto current `main`, reconcile the single authoritative rollback design and its invariants, and rerun real greedy/non-greedy sequence-equivalence plus rejection-at-every-depth tests on the rebased head. The non-greedy verification path must also incorporate the independent-per-position acceptance-draw fix requested on #2140; the current shared scalar draw makes K>=2 sampling distribution incorrect.

Because this PR is owned by the reviewing account, GitHub does not permit a formal Request Changes review; treat this comment as the same blocking disposition.

PR #1219, opened 2026-07-25
test(property): add hypothesis property suite — quantized-KV round-trip + sampling-param invariants

raullenchai:
## Why this matters

The suite has 9,271 example tests and **zero** property-based tests. An
example test pins one point; a property test encodes the *invariant over
the whole input space*. That difference is exactly what guards
**#1208-class numerical bugs** — the quantized-KV dimension-probe gap
where a single untested `(head_dim, group_size)` combination silently
corrupts the cache. You cannot cover that space one example at a time, so
this PR stands up `tests/property/` on Hypothesis and lands the two
highest-value **hermetic** invariant families.

Everything here is hermetic **because** property fuzzing multiplies work
by `max_examples`: no booted server, no model load, tiny in-memory
tensors. The whole suite runs in ~4s.

## What's added

- `hypothesis>=6.100.0` in the `[test]` extras (and mirrored into `[dev]`
  to keep `dev ⊇ test`, which `tests/test_pr_validate_test_env.py`
  enforces).
- `tests/property/__init__.py`, `tests/property/strategies.py` (shared
  strategies), `tests/property/conftest.py` (an MLX-tuned Hypothesis
  profile: `deadline=None` because the first call to an MLX op
  JIT-compiles a Metal kernel and a wall-clock deadline would flake on
  that cold start alone; these properties assert exact math, not
  latency).
- A `property` marker registered in `tests/conftest.py`.
- Two test modules (15 tests total).

### Family 1 — quantized-KV round-trip (`test_quantized_kv_roundtrip.py`, #1208-tied)

Over `_quantize` / `_dequantize` / `supported_group_size` in
`vllm_mlx/quantized_batch_cache.py`:

- **`supported_group_size` is the largest valid divisor** — result is
  `None` or in `{32,64,128}`; when not `None` it is `<= requested`,
  divides `head_dim`, and is the maximum of the valid set (no larger
  valid divisor exists). Head dims sweep non-divisible cases (80, 96,
  100, 256, …).
- **Monotonic in `requested`** — raising the budget never lowers the
  chosen size.
- **Round-trip preserves shape.**
- **Round-trip error is bounded by each group's own quantization step** —
  a *differential, data-derived* bound: per group,
  `step = (max - min) / (2**bits - 1)`, and reconstruction error
  `<= step * (1 + rel) + rel * magnitude`, with `rel = 1e-4` a
  float32-recombination slack applied *relative to the group magnitude*
  (no hand-tuned absolute epsilon). Verified for both `bits=8` and
  `bits=4`.
- **Determinism** — quantizing twice is byte-identical across packed /
  scales / biases.
- **Re-quantization reaches a byte-exact fixed point** (metamorphic).

### Family 2 — sampling-param validation (`test_sampling_param_validation.py`, H-10 generalized)

Parametrized across the OpenAI `ChatCompletionRequest` /
`CompletionRequest` and the `AnthropicRequest` models (the ones exposing
`temperature` / `top_p`):

- **Non-finite always rejected** — for every NaN / +inf / -inf,
  constructing the model with that `temperature` (and separately `top_p`)
  raises `pydantic.ValidationError`. This is the exact silent-burn class
  H-10 closed (a NaN that HTTP-200'd into a Metal kernel).
- **Out-of-range finite rejected** — every finite value strictly outside
  the documented range raises `ValidationError`.
- **In-range accepted and preserved** — every finite in-range value
  constructs OK and is stored unchanged (no silent clamp/coerce).

## Test plan

- [x] `pytest tests/property/ -v` → **15 passed in 4.36s** (all green)
- [x] `pytest tests/property/ -p no:cacheprovider -q` → 15 passed (no collection issues)
- [x] Re-ran the suite 3x for flakiness → 15 passed each time
- [x] `pytest tests/test_pr_validate_test_env.py -q` → 38 passed (confirms `dev ⊇ test` after the pyproject edit)
- [x] `pytest tests/test_anthropic_models.py tests/test_sampling_validation.py -q` → 144 passed (no collection break in neighboring suites)
- [x] `ruff check tests/property/` → All checks passed
- [x] Editable reinstall of THIS worktree (G5) before running (`pip install --no-deps --force-reinstall .`)

## Notes for reviewers (honest findings)

- **No engine bug found** — all three request models already reject every
  non-finite and out-of-range value and preserve in-range values, so the
  sampling-param family ships fully green. Ranges asserted are exactly
  what each model documents (OpenAI temperature `[0,2]`, Anthropic
  temperature `[0,1]`, top_p `(0,1]`).
- **The re-quantization property was reframed to what is actually true of
  MLX.** The originally-imagined property — "`z = dequant(quant(y))` is
  negligibly close to `y`" — is **false** for MLX's affine quantizer:
  re-deriving `(scale, bias)` from `y`'s own group extrema can shift the
  grid, so the first re-quant `|z - y|` can be a *full* quantization step
  (empirically up to ~1 step). This is expected numerics of the MLX
  library, **not** a rapid-mlx bug. The sound invariant that IS true and
  that the test asserts: the drift is bounded by one step, and the tensor
  converges to a **byte-exact fixed point** after the second round-trip
  (`|w - z| == 0` across all 360 spot-check cases and every Hypothesis
  example). Relatedly, MLX affine quantization is **not** round-to-nearest
  — worst-case round-trip error is ~0.99 step, so the bound is a full
  step, not `step/2`.

## Why model-requiring metamorphic tests are deferred

Streaming == non-streaming and temp=0 determinism are high-value
metamorphic properties, but they need a **booted server + a real model
load** per example. That breaks the fast/hermetic contract that lets
property fuzzing run on every commit, so they belong in the integration
suite, not here. This PR deliberately scopes to invariants provable from
pure functions and Pydantic models.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -260,6 +260,10 @@ test = [
     "pillow>=10.0.0",
     "mlx-vlm>=0.6.3,!=0.6.4; platform_system == 'Darwin'",
     "mlx-audio>=0.2.9,<0.4.4; platform_system == 'Darwin'",
+    # Powers the property-based invariant suite in tests/property/ — encodes
+    # numeric/validation invariants over the whole input space (e.g. the
+    # #1208-class quantized-KV round-trip bounds) that example tests can't.
+    "hypothesis>=6.100.0",
 ]
 dev = [
     # Test runtime (must mirror the `test` extras above; a unit test in
@@ -275,6 +279,9 @@ dev = [
     "pillow>=10.0.0",
     "mlx-vlm>=0.6.3,!=0.6.4; platform_system == 'Darwin'",
     "mlx-audio>=0.2.9,<0.4.4; platform_system == 'Darwin'",
+    # Mirror of the [test] entry — keep `dev` a strict superset of `test`
+    # (enforced by tests/test_pr_validate_test_env.py).
+    "hypothesis>=6.100.0",
     # Linters / typer — NOT in the `test` extras because pr_validate
     # only needs to *run* the tests; lint is its own pipeline step
     # (ruff) and runs from the host environment.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,6 +42,10 @@ def pytest_configure(config):
         "markers",
         "integration: mark test as integration test (requires running server)",
     )
+    config.addinivalue_line(
+        "markers",
+        "property: hermetic Hypothesis property-based test (see tests/property/)",
+    )
 
 
 def pytest_collection_modifyitems(config, items):
--- a/tests/property/conftest.py
+++ b/tests/property/conftest.py
@@ -1,0 +1,32 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Hypothesis configuration for the property-based suite.
+
+Registers and loads a profile tuned for MLX. The first call to a given
+MLX op JIT-compiles a Metal kernel — a few hundred ms of one-off cost —
+so a per-example wall-clock ``deadline`` would flake on that cold start
+alone (and only on the first example). Disabling the deadline is the
+right call here: these properties assert *exact* mathematical invariants,
+not latency, so a slow example is never a failure signal. ``max_examples``
+is kept modest so the whole hermetic suite stays well under a couple of
+seconds while still sweeping a wide input space.
+
+pytest imports this dir-level conftest before collecting the property
+test modules, so ``load_profile`` is in effect for every ``@given`` here
+without any per-test decoration. No other test in the repo uses
+Hypothesis, so loading the profile globally is inert elsewhere.
+"""
+from hypothesis import HealthCheck, settings
+
+PROFILE_NAME = "rapid_mlx_property"
+
+settings.register_profile(
+    PROFILE_NAME,
+    max_examples=100,
+    deadline=None,
+    # ``too_slow`` fires on the MLX kernel cold-start described above;
+    # ``data_too_large`` can fire on the wide seed/param draws in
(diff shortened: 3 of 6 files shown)


raullenchai:
### Codex convergence note (recorded for audit)

Codex adversarial review ran across 4 passes and is **non-deterministic** — each pass surfaced *different* test-epistemology findings rather than reproducing prior ones. All **material and cross-pass-consensus** findings were fixed across the follow-up commits:

- byte-exact determinism assertion via a raw (dtype+shape+bytes) helper instead of `mx.array_equal` (which folds ±0.0),
- full-range out-of-range float generation (removed an artificial ±1e6 cap),
- **MLX lazy-eval materialization** — explicit `mx.eval(...)` before every assertion resting on an MLX array, so the quantize/dequantize kernel actually runs (a real soundness gap),
- exclusive-endpoint coverage (`top_p == 0.0` now explicitly rejected),
- (model, field) parametrization so Hypothesis is *required* to exercise every combination (18 → 27 cases).

Also corrected a real test-bug found by the edge-case distributions: MLX affine quantization is **zero-point-inclusive**, so the per-group step is over `[min(gmin,0), max(gmax,0)]`, not `[gmin,gmax]`.

**One NIT is consciously deferred** (per the diminishing-returns policy — the tests are sound and green, and further passes only surface non-consensus wording refinements): the determinism test's docstring frames it as "lossless state serialization". It is precisely a **reproducibility** invariant (same input → byte-identical quantized triple), which the test correctly asserts; only the justification wording is loose. No behavior or assertion change needed.

Stepping off the codex treadmill here: GH checks green, suite hermetic + green (27 passed), all material soundness items closed.

PR #1224, opened 2026-07-25
feat(models): vendor GPT-OSS Puzzle checkpoint support

pierre427:
# feat(models): vendor GPT-OSS Puzzle checkpoint support

Source: [ml-explore/mlx-lm#1488](https://github.com/ml-explore/mlx-lm/pull/1488)

Review state: **validated — real loader/cache, server/Harmony tool-call, and continuation gates passed**

## Why

NVIDIA GPT-OSS Puzzle checkpoints declare `model_type: gpt_oss_puzzle`, which
mlx-lm 0.31.3 does not ship. Rapid already supports GPT-OSS/Harmony serving,
but cannot instantiate Puzzle's heterogeneous per-layer MoE widths and
attention windows. Transformers also does not recognize the model type, so
the ordinary high-level tokenizer path fails before Rapid's existing
GPT-OSS serving stack can run.

## AI assistance disclosure

independently inspected the diff, compared it with upstream #1488, repaired
process-global test isolation, and reran the focused tests and Ruff checks.
Maintainers should independently review the vendor boundary and run the
unchecked real-weight gates below.

## Implementation

- Vendor the upstream #1488 `gpt_oss_puzzle` model module.
  - Reuse installed `mlx_lm.models.gpt_oss.TransformerBlock` and sanitizer.
  - Build per-layer expert widths and full/sliding KV caches from
    `block_configs`.
  - Restore YaRN settings from `yarn_rope_scaling`.
  - Drop checkpoint FP8 KV calibration scales unused by the BF16 KV path.
- Register `mlx_lm.models.gpt_oss_puzzle` through Rapid's existing
  `_register_vendored_archs()` seam.
  - Route the unknown architecture through the existing low-level
    model/tokenizer fallback.
  - Prefer a future native mlx-lm module rather than shadowing it.
- Add focused tests for registration, loader dispatch, heterogeneous
  configuration, YaRN restoration, cache construction, sanitization, and
  test-state isolation.

## Risks and follow-up

- Upstream mlx-lm may land a native implementation. Registration already
  defers to it, but the vendored copy should then be removed if redundant.
- This port does not enable speculative decoding or alter the existing
  GPT-OSS/Harmony parser path.

## Test plan

- [x] `pytest -q tests/test_gpt_oss_puzzle_vendored.py` — 5 passed.
- [x] `git diff --check`.
- [x] Read the cached q6dense Puzzle config: 36 heterogeneous layers,
  64/128 experts, 128/8192/full windows, and YaRN side field.
- [x] Load the real q6dense Puzzle checkpoint through
  `load_model_with_fallback`.
- [x] Real cache construction produced 36 layers: 10 `KVCache` and
  26 `RotatingKVCache` layers with 128-token and 8192-token windows.
- [x] Short `mlx_lm.generate` smoke completed without loader/cache error.
- [x] Rapid server Harmony response and tool-call validation on Apple
  Silicon with the q6dense artifact.
- [x] Continuation/prefix-cache behavior through the Rapid server: exact repeat
  and one-token continuation variant reused 16 cached prefix tokens.

## Server validation

Validation ran on Apple Silicon from the PR branch against the local q6dense
artifact at `/Users/pierrelamy/mlx-models/gpt-oss-puzzle-88B-q6dense-mlx`.
The server was launched with `--tool-call-parser harmony`,
`--reasoning-parser harmony`, and `--no-spec-decode` on `127.0.0.1:8367`.

Observed results:

- `/v1/models` advertised the local Puzzle model with `text` and `tools`
  capabilities, `tool_call_parser=harmony`, and `reasoning_parser=harmony`.
- Ordinary `/v1/chat/completions` returned `content="Paris."` for a short
  factual prompt.
- Required tool-call chat returned `finish_reason="tool_calls"` with parsed
  `get_weather({"city":"Toronto","unit":"celsius"})` and empty visible
  content.
- `/v1/completions` continuation probe:
  - first prompt: 16 prompt tokens, 12 completion tokens;
  - exact repeat: `prompt_tokens_details.cached_tokens=16`;
  - one-token continuation variant: `prompt_tokens_details.cached_tokens=16` and
    generated `+ b...`, demonstrating reuse of the shared prefix before the
    continuation tail.

## Attribution

The vendor module preserves Apple copyright and Apache-2.0 attribution,
links the upstream PR, and pins source commit
`90ce9c4b8825cf8a4e2fd4ce589574695ad9b6a4`.

---


raullenchai:

pierre427:
Rebased onto current main (e4814b4). The only conflict was the vendored-arch registry in `vllm_mlx/utils/tokenizer.py`, resolved additively alongside the new `muse_glimmer`/`bailing_hybrid` blocks, plus a test-double signature update for the #1379 `enable_dspark` kwarg. Gates checked: pyproject mlx bounds untouched (empty diff vs main), `check_rapid_mac_ax_identifiers.py` passes, ruff check/format clean, and the focused Puzzle suite is 5/5 locally; l1-smoke left to CI.

raullenchai (approved):
Approved after independent source and behavior review. I compared the vendored module against the pinned upstream `ml-explore/mlx-lm@90ce9c4b...` source, inspected the installed 0.31.x GPT-OSS block/sanitizer contracts, integrated the head cleanly onto current `main`, and reran the focused loader/registration, heterogeneous MoE/cache-window, forward, and sanitize tests under Python 3.12 (`5 passed`) plus Ruff. The diff preserves the upstream architecture semantics, defers to future native mlx-lm support, does not expand parser/speculative-decoding capability, and scopes weight dropping to the documented unused KV calibration suffixes.

PR #1225, opened 2026-07-25
fix(deepseek-v4): honor explicit YaRN attention_factor

pierre427:
# fix(deepseek-v4): honor explicit YaRN attention_factor

Source: [ml-explore/mlx-lm#1507](https://github.com/ml-explore/mlx-lm/pull/1507)

Review state: **accepted after coordinator repair**

## Why

Some HF YaRN configurations explicitly provide
`rope_scaling.attention_factor`. Rapid's vendored DeepSeek-V4 RoPE path
previously ignored that field, so those checkpoints could receive the wrong
rotary-channel amplitude. An explicit `1.0` must preserve unscaled behavior.

## AI assistance disclosure

OpenAI Codex Terra prepared the first adaptation and focused test. Coordinator
review found that Rapid's DeepSeek-V4 layout stores non-rotary channels first
and rotary channels last; the initial agent patch scaled the wrong slice. The
coordinator corrected the implementation, expanded the regression, and reran
all focused checks.

## Implementation

- Store an explicit, non-null `attention_factor` from `rope_scaling`.
- Preserve Rapid's existing `1.0` behavior when the field is absent or null.
- Scale only the trailing rotary `self.dims` channels immediately before
  `mx.fast.rope`; the leading NOPE channels remain unchanged.
- Add regression coverage for:
  - omitted factor, null factor, and explicit `1.0` equivalence;
  - explicit `0.5` scaling only the trailing rotary slice;
  - input-array immutability.

## Risks and scope

- This does not import generic `YarnRoPE`'s separate computed `mscale`
  policy; Rapid's specialized `DeepseekV4RoPE` did not previously implement it.
- The change follows Rapid's existing compressed-attention DeepSeek-V4
  topology; no generic model loader behavior changes.
- No real-weight DeepSeek-V4 run was performed.

## Test plan

- [x] `pytest -q tests/test_deepseek_v4_vendored.py` — 5 passed.
- [x] Ruff check on both changed files.
- [x] Ruff format check on both changed files.
- [x] `git diff --check`.
- [x] Coordinator partial-rotary layout review and wrong-slice repair.

## Attribution

Behavior is adapted from mlx-lm #1507 under Apache-2.0-compatible project
licensing. Rapid's DeepSeek-V4 module remains its own vendored implementation.

---

### DIFF
--- a/tests/test_deepseek_v4_vendored.py
+++ b/tests/test_deepseek_v4_vendored.py
@@ -60,6 +60,58 @@ def test_register_vendored_archs_is_idempotent():
     assert first is second
 
 
+def test_deepseek_v4_rope_honors_explicit_yarn_attention_factor():
+    """HF's explicit YaRN attention factor scales only rotary channels."""
+    import mlx.core as mx
+
+    from vllm_mlx.models.deepseek_v4 import DeepseekV4RoPE
+
+    scaling = {
+        "rope_type": "yarn",
+        "factor": 4.0,
+        "original_max_position_embeddings": 128,
+    }
+    x = mx.arange(24, dtype=mx.float32).reshape(1, 3, 8) / 10
+    baseline = DeepseekV4RoPE(dims=4, base=10000.0, scaling_config=scaling)
+    explicit_one = DeepseekV4RoPE(
+        dims=4,
+        base=10000.0,
+        scaling_config={**scaling, "attention_factor": 1.0},
+    )
+    explicit_null = DeepseekV4RoPE(
+        dims=4,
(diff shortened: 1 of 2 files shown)


raullenchai (approved):
Reviewed the DeepSeek-V4 YaRN attention_factor implementation and regression coverage. The change preserves absent/null/1.0 behavior, scales only the trailing rotary channels, and leaves the caller input unchanged.

Validation: Codex review converged with no blocking findings; focused tests 5/5 and full suite 14,019 passed; GitHub required checks are green.

PR #1226, opened 2026-07-25
fix(security): contain local model_file imports to the model root

pierre427:
# fix(security): contain local model_file imports to the model root

Source: [ml-explore/mlx-lm#1523](https://github.com/ml-explore/mlx-lm/pull/1523)

Review state: **accepted after coordinator wiring coverage**

## Summary

Add fail-closed validation for `config.json::model_file` when Rapid-MLX is
given an existing local model directory.

Before Rapid hands a caller-supplied local directory to `mlx_lm`, the value
must now be:

- relative;
- suffixed `.py`;
- an existing regular file; and
- resolved inside the resolved model root.

This rejects absolute paths, `..` traversal, symlink escapes, missing paths,
directories, and non-Python targets.

## Necessity

A local model config should only select code contained within that model
directory. Without this boundary, a crafted `model_file` can redirect an
otherwise inspected local model to execute a Python file elsewhere on the
machine.

The main serving path uses `BatchedEngine` with remote-code support available,
so this is a live local containment improvement rather than dead validation.

## AI assistance disclosure

OpenAI Codex Terra implemented the guard and mapped direct loader call sites.
The coordinator independently reviewed the containment and loader map, added
a production-seam test proving validation occurs before shared loader
dispatch, added benign local/remote no-op cases, and reran focused checks.

## Implementation

- Add `vllm_mlx.utils.model_file_guard.validate_local_model_file`.
- Apply it at the shared `load_model_with_fallback` boundary used by normal
  serving and the general benchmark helper.
- Apply it to remaining direct local-capable `mlx_lm.load` paths:
  `jlens`, `model_runner`, freeform `bench`, and PFlash replication.
- Add focused boundary and production-wiring tests.

## Scope and risks

- This protects caller-supplied existing local model directories only.
- It does not validate remote Hugging Face IDs or snapshots resolved
  internally by `mlx_lm`/`huggingface_hub`.
- It does not sandbox Python legitimately stored inside the local model root.
- A separate malicious local process could race path replacement between
  validation and upstream loading; this narrows path selection but is not an
  `openat`-style sandbox.
- `--submit` is unchanged because it accepts registered aliases and fixed HF
  paths, not caller-supplied local directories.

## Test plan

- [x] Valid nested local `.py` file is accepted.
- [x] Ordinary local config without `model_file` is accepted.
- [x] Remote model ID is a no-op.
- [x] Absolute `model_file` is rejected.
- [x] `../` traversal is rejected.
- [x] Symlink escape is rejected.
- [x] Missing target is rejected.
- [x] Directory target is rejected.
- [x] Non-`.py` target is rejected.
- [x] Shared loader invokes validation before dispatch.
- [x] `pytest -q tests/test_model_file_guard.py` — 10 passed.
- [x] Ruff check and format check on all changed files.
- [x] `git diff --check`.

## Attribution

The security rationale is grounded in mlx-lm #1523. Rapid's implementation is
independent and Apache-2.0; no upstream source body was copied.

---

### DIFF
--- a/tests/test_model_file_guard.py
+++ b/tests/test_model_file_guard.py
@@ -11,7 +11,7 @@ import pytest
 from vllm_mlx.utils.model_file_guard import validate_local_model_file
 
 
-def _model_dir(tmp_path: Path, model_file: str) -> Path:
+def _model_dir(tmp_path: Path, model_file: str | None) -> Path:
     model_root = tmp_path / "model"
     model_root.mkdir()
     (model_root / "config.json").write_text(
@@ -33,6 +33,12 @@ def test_accepts_local_model_without_custom_model_file(tmp_path: Path) -> None:
     model_root = tmp_path / "model"
     model_root.mkdir()
     (model_root / "config.json").write_text("{}", encoding="utf-8")
+
+    validate_local_model_file(model_root)
+
+
+def test_accepts_null_model_file(tmp_path: Path) -> None:
+    model_root = _model_dir(tmp_path, None)
 
     validate_local_model_file(model_root)
 
--- a/tests/test_model_file_guard.py
+++ b/tests/test_model_file_guard.py
@@ -1,0 +1,117 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Focused coverage for local config.json::model_file containment."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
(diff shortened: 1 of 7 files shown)


raullenchai (approved):
Reviewed the local model_file containment boundary and all direct mlx-lm load seams. Fixed the one compatibility edge found (model_file: null must remain equivalent to absent) and added regression coverage.

Local verification: 41 focused loader/CLI tests passed; Ruff and git diff checks passed.

PR #1227, opened 2026-07-25
feat(models): vendor Sarvam MLA checkpoint support

pierre427:
# feat(models): vendor Sarvam MLA checkpoint support

Source: [ml-explore/mlx-lm#1530](https://github.com/ml-explore/mlx-lm/pull/1530)

Review state: **follow-up validated — public q6 artifact available**

## Summary

Add a native-first compatibility vendor for the `sarvam_mla` architecture,
adapted to Rapid-MLX's model loader and cache-safety conventions.

## Why

Sarvam-105B is a 105B-total / 10.3B-active MLA MoE model. The adapter is a thin
config remap over MLX-LM's DeepSeek-V3 implementation, but Rapid-MLX needs an
Apple-Silicon-practical checkpoint before support is useful to users. A public
q6 MLX artifact is now available:
[`pierrelamy/sarvam-105b-mlx-q6`](https://huggingface.co/pierrelamy/sarvam-105b-mlx-q6).

The lab q4 quant was tested and is not suitable: it loads but collapses to
whitespace with probability 1.0. q6 is the smallest validated practical artifact
for this architecture in the lab.

## Implementation

- Vendor the Sarvam MLA model only while the installed `mlx-lm` lacks native
  support, and automatically defer once upstream support is present.
- Register `sarvam_mla` through Rapid's direct loader fallback.
- Route Sarvam MLA through Rapid's MLA-aware KV-cache dtype and TurboQuant
  safety gates.
- Add synthetic model, cache, sanitization, and production loader-dispatch
  coverage.

## Test plan

- [x] `pytest -q tests/test_sarvam_mla.py tests/test_kv_cache_dtype.py
  tests/test_turboquant_k8v4.py` — 94 passed.
- [x] `git diff --check`.
- [x] Coordinator repair of the synthetic MoE configuration.
- [x] Production loader-dispatch regression for local `sarvam_mla` configs.
- [x] Real-weight load, generation, MLA cache behavior, and TurboQuant/int4
  safety validation.

## Real-weight validation

Validation was run on Apple Silicon against the q6 artifact
`pierrelamy/sarvam-105b-mlx-q6`, using the local artifact copy at
`/Users/pierrelamy/Desktop/mlx-uag/sarvam-105b-mlx-q6`.

- Direct Rapid loader path routed `model_type=sarvam_mla` through the vendored
  architecture path and loaded real q6 weights.
- q6 config verified as 6-bit affine, group size 64, 32-layer MLA with
  `kv_lora_rank=512`.
- MLA prompt cache populated 32/32 layers with compressed latent cache state.
- Direct generation was coherent:
  - `"The capital of France is"` produced `" a city of lights..."`
  - top-token probe ranked `" a"` and `" Paris"` first/tied for the same prompt.
  - `def add(a, b):\n    return` predicted `" a"` at 0.986564.
- Safety gates passed:
  - requested int4 KV resolves to bf16 for MLA;
  - TurboQuant reports `skip=True, reason="mla"`.
- Rapid OpenAI-compatible server booted the artifact locally, completed warmup,
  reported 131072 context through `/v1/models`, and `/v1/completions` returned
  `"Paris, a city of light, love"` for `"The capital of France is"` with
  `max_tokens=8, temperature=0`.
- Server metrics reported effective KV dtype bf16 and TurboQuant disabled.

## Attribution

adapted from mlx-lm #1530 for Rapid-MLX's native-first vendoring path.

---

### DIFF
--- a/tests/test_kv_cache_dtype.py
+++ b/tests/test_kv_cache_dtype.py
@@ -195,6 +195,18 @@ def test_bailing_hybrid_model_type_triggers_downgrade():
     assert "mla" in decision.reason.lower()
 
 
+def test_sarvam_mla_config_triggers_downgrade_without_q_lora_rank():
+    """Sarvam uses direct q_proj, so its MLA signal is the model type."""
+    decision = resolve_kv_cache_dtype(
+        "int4",
+        model_name="sarvam-105b-mock",
+        hf_config={"model_type": "sarvam_mla", "kv_lora_rank": 512},
+    )
+    assert decision.dtype == "bf16"
+    assert decision.downgraded is True
+    assert "mla" in decision.reason.lower()
+
+
 def test_deepseek_v4_substring_triggers_downgrade():
     decision = resolve_kv_cache_dtype(
         "int4",
--- a/tests/test_kv_cache_dtype.py
+++ b/tests/test_kv_cache_dtype.py
@@ -177,6 +177,18 @@ def test_deepseek_v3_config_triggers_downgrade():
     assert "mla" in decision.reason.lower()
 
 
+def test_sarvam_mla_config_triggers_downgrade_without_q_lora_rank():
+    """Sarvam uses direct q_proj, so its MLA signal is the model type."""
+    decision = resolve_kv_cache_dtype(
+        "int4",
+        model_name="sarvam-105b-mock",
+        hf_config={"model_type": "sarvam_mla", "kv_lora_rank": 512},
+    )
+    assert decision.dtype == "bf16"
+    assert decision.downgraded is True
+    assert "mla" in decision.reason.lower()
+
+
 def test_deepseek_v4_substring_triggers_downgrade():
     decision = resolve_kv_cache_dtype(
         "int4",
--- a/tests/test_sarvam_mla.py
+++ b/tests/test_sarvam_mla.py
@@ -79,7 +79,7 @@ def test_loader_routes_sarvam_config_to_vendored_path(tmp_path, monkeypatch):
     expected = (object(), object())
     seen = []
 
-    def _fake_vendor_loader(model_name):
+    def _fake_vendor_loader(model_name, *, enable_dspark=False):
         seen.append(model_name)
         return expected
 
--- a/tests/test_sarvam_mla.py
+++ b/tests/test_sarvam_mla.py
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Focused synthetic coverage for the vendored Sarvam MLA adapter."""
+
+import importlib
+import sys
+
+import mlx.core as mx
+import pytest
+from mlx_lm.models.cache import make_prompt_cache
+
+
+@pytest.fixture(autouse=True)
+def _clear_vendored_registration():
+    from vllm_mlx.utils.tokenizer import _VENDORED_MODEL_TYPES
+
+    sys.modules.pop("mlx_lm.models.sarvam_mla", None)
+    _VENDORED_MODEL_TYPES.discard("sarvam_mla")
+    yield
+    sys.modules.pop("mlx_lm.models.sarvam_mla", None)
+    _VENDORED_MODEL_TYPES.discard("sarvam_mla")
+
+
+def _tiny_args(**overrides):
+    from vllm_mlx.models.sarvam_mla import ModelArgs
+
+    values = {
+        "vocab_size": 128,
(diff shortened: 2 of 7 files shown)


raullenchai:
Thanks for putting this together. Sarvam-105B is a legitimate and reasonably well-trafficked model, and the thin DeepSeek-V3-based adapter is a sensible approach.

The current blocker for us is practical Apple Silicon usability rather than the model itself. The only public MLX checkpoint we could find is the Q10 conversion at roughly 133 GB, which leaves no workable runtime headroom even on a 128 GB Mac. There is not yet a public 4-bit or 6-bit MLX checkpoint that typical Rapid-MLX users can actually load, and this PR has therefore not been validated against real weights, generation, or MLA cache behavior.

Could we keep this as a draft/parked until a practical 4-bit or 6-bit MLX checkpoint is available? Once there is a roughly 50–70 GB artifact that we can test end to end on Apple Silicon, the support becomes much more actionable and we can validate the adapter properly before merging.

Thanks again — the model is interesting; we just want to merge support when users can realistically benefit from it.

pierre427:
Thanks for the clear guidance. We uploaded and validated a public q6 MLX artifact for Sarvam-105B:

https://huggingface.co/pierrelamy/sarvam-105b-mlx-q6

I also reran the Rapid-MLX validation locally against that q6 artifact on Apple Silicon. The vendored `sarvam_mla` loader path now loads real weights, generation is coherent, MLA prompt cache state populates correctly, int4 KV is downgraded to bf16 for MLA, and TurboQuant is skipped for MLA. A local Rapid server smoke also passed: `/v1/models` reports the q6 model with 131072 context, and `/v1/completions` returned `"Paris, a city of light, love"` for `"The capital of France is"` at `temperature=0`.

One caveat: q4 is still intentionally not offered because our q4 run collapses to whitespace with probability 1.0. q6 is the smallest validated practical artifact we have for this architecture.

I updated the PR body with the completed validation checklist and left the PR ready for review.

raullenchai:

We are mid-release for 0.12.8 and are deliberately not taking model-support changes into that cycle. I will pick this up once it ships.

Specific to this one: it does not only add a vendored model — it also touches `kv_cache_dtype.py` and `turboquant.py`, which every model goes through. That is a different review from "add an arch", so it will get a closer look at the shared paths rather than at the checkpoint support, and it may want splitting so the model support can land independently of the shared-path change.

Also note the green checks here are stale — they ran against late-July `main`, and several gates have landed since (`accessibility-identifiers`, `l1-smoke`, `mlx-bound-guard`), so a rebase will face roughly twice the checks this run did.

pierre427:
Rebased onto main (486 commits; only churn was the new `enable_dspark` kwarg on the mocked loader in the test). Happy to split as suggested — model support (`models/sarvam_mla.py` + loader registration) severs cleanly from the safelist entries; one note for the safelist half: Sarvam has no `q_lora_rank`, so the existing rank-pair heuristic can never detect it as MLA — the model_type/name entries are its only guard against int4-on-latent-K under explicit `--kv-cache-dtype` opt-in.

raullenchai:

pierre427:

Conflicts were all additive drift and resolved by keeping both sides: `kv_cache_dtype.py` (the MLA model-type safelist gained `bailing_hybrid` on main; `sarvam_mla` is added alongside), the vendored-arch registration block in `utils/tokenizer.py`, and the adjacent tests. Two small adaptations: `models/sarvam_mla.py` now installs the existing #404 `_mlx_compat` shim before importing `mlx_lm` (the consumer audit in `tests/test_mlx_compat.py` flags any module-load-time `mlx_lm` import without it), and the loader mock in `test_sarvam_mla.py` accepts the `enable_dspark` keyword added by #1379.

- `test_sarvam_mla.py`, `test_kv_cache_dtype.py`, `test_turboquant_k8v4.py`, `test_aliases_contract.py`, `test_model_auto_config.py`: **3325 passed**.
- Full unit suite (minus `tests/integrations`, which needs SDKs not installed here): failures are exactly the 13 that also fail on pristine `origin/main` in this environment (the `test_agent_setup_config_merge` and model-numerics parity families) — none introduced by this branch.

raullenchai:
Closing this stale vendor-integration PR because the dependency and Rapid-MLX model-loading architecture have moved substantially since the submitted head. Before reviving this model family, we need to re-check whether current dependencies already provide the architecture and then qualify the smallest current-main compatibility layer, if any. Thank you for the original implementation and real-model validation.

PR #1228, opened 2026-07-25
feat(models): vendor Nemotron-H Puzzle checkpoint support

pierre427:
# feat(models): vendor Nemotron-H Puzzle checkpoint support

Source: [ml-explore/mlx-lm#1536](https://github.com/ml-explore/mlx-lm/pull/1536)

Review state: **validated — real 75B server/tool-call gate passed; MTP unavailable in the tested artifact**

## Why

Nemotron-H Puzzle matters to Rapid users because it is a practical 75B / ~9B-active hybrid MoE that can run locally on Apple Silicon with a 262k context window and strong tool-calling behavior. The port lets Rapid instantiate the heterogeneous per-layer Puzzle variant while keeping hybrid SSM and speculative-decode safety gates conservative.

## Summary

Add native-first `nemotron_h` compatibility for the Nemotron Labs Puzzle
checkpoint while preserving Rapid-MLX's conservative hybrid-model safety
gates.

## AI assistance disclosure

independently inspected the diff, repaired process-global test cleanup,
verified public checkpoint metadata, corrected the tool parser profile to the
checkpoint's Nemotron XML template, and reran focused tests, Ruff, format, and
whitespace checks.

## Implementation

- Vendor the Nemotron-H architecture only when the installed `mlx-lm` has no
  native module.
- Register the architecture in Rapid's direct loader fallback.
- Add a text-only Nemotron Puzzle profile with the verified Nemotron tool-call
  parser and Qwen3 reasoning parser.
- Preserve the checkpoint's MTP sidecar metadata without claiming safe
  speculative decoding: Rapid still lacks validated hybrid SSM
  rollback/replay for this architecture.

## Test plan

- [x] `pytest -q tests/test_nemotron_puzzle_vendored.py
  tests/test_aliases_contract.py` — 2545 passed.
- [x] `git diff --check`.
- [x] Public checkpoint metadata verified: `nemotron_h`, 88 block
  descriptors, one configured MTP layer, 13 weight shards, and zero MTP
  tensor keys in the tested community 6-bit artifact.
- [x] Parser profile corrected to match the checkpoint's
  `<function>` / `<parameter>` Nemotron template.
- [x] Real 75B GPU server validation on Apple Silicon: Rapid loaded the
  community 6-bit artifact, `/v1/models` advertised 262144 context with
  text/tools capabilities, ordinary chat returned `Paris`, and a forced
  Nemotron-format weather tool request produced
  `get_weather({"city":"Toronto"})`.

## Remaining validation

The tested community 6-bit artifact has no MTP tensor keys, so MTP decode was
not a valid validation target for this PR. mlx-lm#1533 remains deferred until a
Puzzle checkpoint with MTP weights is available and Rapid has a safe hybrid SSM
rollback/replay contract.

## Attribution

adapted from mlx-lm #1536 for Rapid-MLX's native-first vendoring seam.

---


raullenchai:

pierre427:
Rebased onto main (e4814b4): both conflicts were additive — kept the new nemotron-3.5-lightning alias alongside the puzzle entry in aliases.json, and appended the Puzzle vendor block after the muse_glimmer/bailing_hybrid registrations in `_register_vendored_archs()`. Also hardened the test fixture so the loader-registration test survives running after the other vendored-arch suites (which also call `_register_vendored_archs()`) — reproduced as a latent issue at the original PR head, not rebase-induced. Bound-guard, the ax-identifier gate, ruff, and the focused suites all pass locally, with l1-smoke left to CI.

raullenchai:

pierre427:

Conflicts were additive drift (a `nemotron-labs-diffusion` alias landed adjacent to the new entry in `aliases.json`; new vendored-arch registrations beside ours in `utils/tokenizer.py`) and resolved by keeping both sides. One adaptation for a gate added since the last rebase: `model_sizes.json` now carries the `georgeis55/Nemotron-Labs-3-Puzzle-75B-A9B-MLX-6bit` footprint (67,229,603,239 bytes, resolved with the same `estimate_repo_size_bytes` the generator uses), which `test_model_sizes.py::test_every_text_alias_has_a_manifest_entry` now requires for every text alias.

- `test_nemotron_puzzle_vendored.py`, `test_aliases_contract.py`, `test_model_auto_config.py`: **3248 passed**; `test_model_sizes.py` green.

pierre427:
Rebased onto current `main` (`4c03e3dd`) and force-updated with an exact lease. New head: `f6c8381c`.

The only conflict was the tokenizer registration insertion point added by current-main Qwen4 work; the resolution preserves both the Qwen4 registration and the Nemotron-H Puzzle vendor fallback.

Verification:

- focused suite: **3312 passed**;
- `ruff check`: passed;
- `git diff --check`: passed.

raullenchai:
Closing this stale vendor-integration PR because the dependency and Rapid-MLX model-loading architecture have moved substantially since the submitted head. Before reviving this model family, we need to re-check whether current dependencies already provide the architecture and then qualify the smallest current-main compatibility layer, if any. Thank you for the original implementation and validation.

PR #1229, opened 2026-07-25
feat(models): vendor GraniteMoE SWA checkpoint support

pierre427:
# feat(models): vendor GraniteMoE SWA checkpoint support

Source: [ml-explore/mlx-lm#1608](https://github.com/ml-explore/mlx-lm/pull/1608)

Review state: **validated — real server/prefix/continuous-batching/spec smoke passed**

## Why

Granite Swash is a small, permissively usable MoE with mixed full/sliding-window attention, which makes it a good Rapid regression and dogfood target: it is cheap enough for routine Apple Silicon validation but still exercises the cache shapes, MoE routing, parser metadata, and speculative-safety boundaries that dense toy models miss. Shipping the vendor path gives Rapid users a fast local Granite family option without relaxing conservative mixed-cache gates by default.

## Summary

Add a native-first `granitemoe_swa` compatibility path for Granite Swash,
including mixed full/sliding attention caches and Rapid-MLX metadata routing.

## AI assistance disclosure

independently inspected the diff, reran focused tests, Ruff, format, and
whitespace checks, and performed a serialized local real-weight loader/cache
smoke after confirming no model-serving contention.

## Implementation

- Vendor `granitemoe_swa` only while installed `mlx-lm` lacks native support.
- Support mixed `KVCache` / `RotatingKVCache` layers, attention sinks, the
  shared expert, nested RoPE configuration, and fused expert-weight
  sanitization.
- Route local Granite Swash checkpoints to Rapid's existing Granite tool
  parser and mark the architecture as MoE.
- Disable server speculative decoding, int4 KV cache, and TurboQuant until
  their mixed-cache behavior is qualified with real weights.

## Test plan

- [x] `pytest -q tests/test_granitemoe_swa_vendored.py
  tests/test_kv_cache_dtype.py tests/test_model_auto_config.py
  tests/test_turboquant_k8v4.py` — 409 passed.
- [x] `git diff --check`.
- [x] Local checkpoint configuration confirms 8 full and 20 sliding layers, a
  128-token window, attention sinks, and fused expert weights.
- [x] Real local checkpoint load from
  `/Users/pierrelamy/mlx-models/granite-swash-3b-a600m-mlx-4bit`.
- [x] Real cache construction produced 28 layers: 8 `KVCache` and
  20 `RotatingKVCache` layers.
- [x] Short `mlx_lm.generate` smoke completed without loader/cache error.
- [x] Real Rapid server validation on Apple Silicon: `/v1/models` advertised
  text/tools with the Granite parser, a 410-token prompt crossed the
  128-token sliding-window boundary twice with stable repeat output, four
  concurrent chat requests completed through the batched server, and a
  forced-spec smoke completed without runtime/rollback errors. Exact-repeat
  and one-suffix completion probes reused 9 cached prompt tokens.

## Remaining validation

This pass validates the practical Rapid server route, prefix reuse, continuous
batching, and a forced-spec smoke. Broader quality benchmarking and deeper
long-context mixed-cache parity can follow separately; the PR still keeps
speculative decoding, int4 KV cache, and TurboQuant conservative by default.

## Attribution

The vendor module preserves Apache-2.0 attribution and is adapted from mlx-lm
#1608 for Rapid-MLX's native-first vendoring seam.

---

### DIFF
--- a/tests/test_granitemoe_swa_vendored.py
+++ b/tests/test_granitemoe_swa_vendored.py
@@ -1,0 +1,100 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Focused synthetic coverage for the vendored GraniteMoE SWA model."""
+
+import importlib
+import sys
+
+import mlx.core as mx
+import pytest
+from mlx_lm.models.cache import KVCache, RotatingKVCache
+
+
+@pytest.fixture(autouse=True)
+def _clear_vendored_registration():
+    sys.modules.pop("mlx_lm.models.granitemoe_swa", None)
+    yield
+    sys.modules.pop("mlx_lm.models.granitemoe_swa", None)
+
+
+def _tiny_args(**overrides):
+    from vllm_mlx.models.granitemoe_swa import ModelArgs
+
+    values = {
+        "model_type": "granitemoe_swa",
+        "vocab_size": 128,
+        "hidden_size": 64,
+        "intermediate_size": 32,
+        "num_hidden_layers": 4,
+        "num_attention_heads": 4,
+        "num_key_value_heads": 2,
+        "num_local_experts": 4,
+        "num_experts_per_tok": 2,
+        "shared_intermediate_size": 64,
+        "max_position_embeddings": 128,
+        "rms_norm_eps": 1e-5,
+        "embedding_multiplier": 1.0,
+        "attention_multiplier": 0.125,
+        "residual_multiplier": 1.0,
+        "logits_scaling": 1.0,
+        "sliding_window": 4,
+        "layer_types": [
+            "full_attention",
+            "sliding_attention",
+            "sliding_attention",
+            "full_attention",
+        ],
+        "rope_parameters": {"rope_theta": 10000.0},
+    }
+    values.update(overrides)
+    return ModelArgs(**values)
+
+
+def test_registers_with_mlx_lm_loader_lookup():
+    from vllm_mlx.utils.tokenizer import _register_vendored_archs
+
+    _register_vendored_archs()
+    module = importlib.import_module("mlx_lm.models.granitemoe_swa")
+    assert module.__name__ == "vllm_mlx.models.granitemoe_swa"
+    assert module.ModelArgs is not None
+
+
+def test_mixed_cache_contract_and_tiny_forward():
+    from vllm_mlx.models.granitemoe_swa import Model
+
+    model = Model(_tiny_args())
+    cache = model.make_cache()
+    assert isinstance(cache[0], KVCache)
+    assert isinstance(cache[1], RotatingKVCache)
+    assert cache[1].max_size == 4
+    assert isinstance(cache[3], KVCache)
+
+    capped = model.make_cache(max_kv_size=8)
+    assert isinstance(capped[0], RotatingKVCache)
+    assert capped[0].max_size == 8
+    assert capped[1].max_size == 4
+
+    logits = model(mx.array([[1, 2, 3, 4, 5]], dtype=mx.int32), cache=cache)
+    mx.eval(logits, [entry.state for entry in cache])
+    assert logits.shape == (1, 5, 128)
+    assert model.supports_speculative_rollback is True
+
+
+def test_sanitize_splits_gate_up_by_halves_and_drops_tied_lm_head():
+    from vllm_mlx.models.granitemoe_swa import Model
+
+    model = Model(_tiny_args(num_hidden_layers=1, layer_types=["full_attention"]))
+    fused = mx.arange(2 * 8 * 4).reshape(2, 8, 4)
+    sanitized = model.sanitize(
+        {
+            "model.layers.0.block_sparse_moe.experts.gate_up_proj": fused,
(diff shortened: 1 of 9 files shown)


raullenchai:
Thanks for putting this together — the implementation work and real-weight load/cache smoke are useful.

Before we invest further in the remaining server and mixed-cache validation, could you help us understand the concrete user benefit you see in supporting this model? Granite-SWASH is described as an early exploration/preview, and its public traction currently appears limited (roughly 925 Hugging Face downloads, 10 likes, one community copy, and no established public MLX model ecosystem yet).

Is there a specific Rapid-MLX user request, compelling Apple Silicon use case, expected follow-on model family, or meaningful performance advantage that makes `granitemoe_swa` worth supporting now? We want to prioritize model support with clear user value, and your context would help us decide whether to proceed or keep this draft parked.

Thanks again!

raullenchai:

We are mid-release for 0.12.8 and are deliberately not taking model-support changes into that cycle. I will pick this up once it ships.

Specific to this one: it does not only add a vendored model — it also touches `kv_cache_dtype.py` and `turboquant.py`, which every model goes through. That is a different review from "add an arch", so it will get a closer look at the shared paths rather than at the checkpoint support, and it may want splitting so the model support can land independently of the shared-path change.

Also note the green checks here are stale — they ran against late-July `main`, and several gates have landed since (`accessibility-identifiers`, `l1-smoke`, `mlx-bound-guard`), so a rebase will face roughly twice the checks this run did.

pierre427:
Rebased onto main — zero conflicts, zero code churn; pyproject untouched, ruff + the focused suites green locally.

On your question — honest answer on all four counts: no specific Rapid-MLX user request I can point to. The case is family continuity plus the memory envelope: Rapid-MLX already ships the Granite line (the granite4-tiny / granite4-h-micro aliases and the `granite` tool-call parser), and granitemoe_swa is IBM's current pure-attention SWA member of that family — its chat template emits the same `<|tool_call|>` JSON protocol, so tool calling reuses your existing parser with zero new parser code (asserted in the tests). The 3B-total / 600M-active sparse-MoE profile is squarely the low-memory Apple-Silicon case those aliases serve today, and the sinks+SWA pattern (per-head sink logits, mixed full/sliding layers) is the one gpt-oss mainstreamed, so the cache plumbing is reusable for follow-ons. Provenance: adapted from our mlx-lm port (ml-explore/mlx-lm#1608), real-weight validated there in bf16 and 4-bit. No perf numbers in this branch — tests are synthetic-shape only, and auto-config keeps spec-decode off until a real-weight run on your side.

On the split you suggested: agreed, and it's almost free here — granitemoe_swa configs carry a mandatory `sliding_window` field, so the canonical hf_config check already downgrades quantized KV correctly; the 3 name-pattern lines in `kv_cache_dtype.py`/`turboquant.py` only cover the config-unavailable fallback and can trail as a trivial follow-up. Say the word and I'll drop them from this PR.

raullenchai:

pierre427:
Rebased onto current `main` (240a6d74) and re-validated. (Main moved again mid-refresh — `config_declares_linear_attention` landed adjacent to the GraniteMoE branch of the metadata route; resolved additively.)

Conflicts were additive drift — the vendored-arch registration block in `utils/tokenizer.py` and adjacent test additions in `test_model_auto_config.py` (a Qwen3.8 hybrid-detection test landed where the GraniteMoE test inserts) — and resolved by keeping both sides.

One behavioral tightening beyond the mechanical rebase: the metadata route no longer advertises `tool_call_parser=granite` from `model_type` alone. The architecture facts (`is_moe`, spec-decode off) stay config-derived, but the parser is only selected when the checkpoint's own chat template emits Granite's `<|tool_call|>` marker on a reachable output path (new `_template_emits_granite_tool_calls` probe built on the existing path-aware `_template_output_paths` analysis, so a marker hidden in an uncalled macro or a `{% set %}` capture is not wire evidence) — a repackage's template is the authoritative wire contract. Negative tests cover the template-less, plain-template, uncalled-macro, and set-capture cases.

- `test_granitemoe_swa_vendored.py`, `test_model_auto_config.py`, `test_aliases_contract.py`, `test_kv_cache_dtype.py`, `test_turboquant_k8v4.py`: **3340 passed** (incl. one positive and four negative Granite gate tests).

pierre427:
Rebased onto current `main` (`4c03e3dd`) and force-updated with an exact lease. New head: `417c883e`.

The only conflict was in the model-auto-config tests where current main extended the Qwen3.8/Qwen4 case at the same insertion point. The resolution retains that coverage and adds the GraniteMoE-SWA cases after it.

Verification:

- focused suite: **3378 passed**;
- `ruff check`: passed;
- `git diff --check`: passed.

raullenchai:
Closing this stale vendor-integration PR because the dependency and Rapid-MLX model-loading architecture have moved substantially since the submitted head. Before reviving this model family, we need to re-check whether current dependencies already provide the architecture and then qualify the smallest current-main compatibility layer, if any. Thank you for the original implementation and real-model cache/serving validation.

PR #1230, opened 2026-07-25
feat(models): vendor Apertus 1.5 text-backbone support

pierre427:
# feat(models): vendor Apertus 1.5 text-backbone support

Source: [ml-explore/mlx-lm#1615](https://github.com/ml-explore/mlx-lm/pull/1615)

Review state: **validated — real text-backbone load/forward/generation passed**

## Why

Apertus 1.5 is a fully open multimodal model whose language tower can serve normal text workloads through Rapid's existing text lane. The important boundary for this PR is narrow: support the official checkpoint's split input/output vocabulary and `model.language_model.*` weight layout without claiming image/audio serving. That gives Rapid users a real Apertus text path while keeping multimodal processing out of scope.

## Summary

Add a native-first compatibility vendor for the `apertus1p5_text`
architecture.

Apertus 1.5 is multimodal, but its text decoder can use Rapid's text lane. Its
wrapper has an extended input vocabulary for media tokens, a pruned text-only
output head, a `model.language_model.*` checkpoint prefix, nested RoPE
parameters, and singleton xIELU parameters.

## Implementation

- Vendor only the text-backbone wrapper while installed `mlx-lm` lacks native
  support, then automatically defer to upstream.
- Register the architecture for Rapid's direct loader fallback.
- Add `apertus-v1.5-8b-text`, pointing to the official checkpoint and
  explicitly marked text-only with speculative decoding disabled.
- Do not claim multimodal processing or unsupported Apertus tool/reasoning
  parsers.

## AI assistance disclosure

independently inspected the diff against mlx-lm #1615, verified the
text-backbone boundary, and reran focused tests, Ruff, format, and whitespace
checks.

## Test plan

- [x] `pytest -q tests/test_apertus1p5_text_vendored.py
  tests/test_aliases_contract.py tests/test_model_auto_config.py`:
  2861 passed.
- [x] `git diff --check`.
- [x] Coordinator source review confirmed the vendor preserves mlx-lm#1615's
  split-vocabulary and checkpoint-remapping behavior.
- [x] Real official checkpoint validation on Apple Silicon: promoted the
  cached `swiss-ai/Apertus-v1.5-8B` `text_config` to `apertus1p5_text`,
  loaded with `strict=False`, evaluated the 32-layer text backbone, confirmed
  split shapes `(266752, 4096)` input embeddings and `(131072, 4096)` LM
  head, ran a forward pass with logits `(1, 18, 131072)`, and completed an
  8-token greedy continuation. Alias metadata remains text-only with spec,
  tool, and reasoning parsers disabled.

## Remaining validation

The official checkpoint is now available in the local Hugging Face cache and the
real text backbone has loaded/evaluated. Follow-up: the server alias currently
resolves the profile but does not pass the alias `hf_path` through to
`mlx_lm.load`, so serving by alias needs a small routing fix before claiming an
OpenAI-compatible Apertus endpoint. Multimodal image/audio serving remains out
of scope for this text-backbone PR.

## Attribution

adapted from mlx-lm #1615 for Rapid-MLX's native-first vendoring seam.

---


raullenchai:

pierre427:
Rebased onto main; the only conflict was in `_register_vendored_archs()` where main's new muse_glimmer/bailing_hybrid vendor blocks landed at the same spot — kept both sides, apertus block appended with the same defer-to-native probe. Diff vs main is unchanged in substance (purely additive, 235 lines), pyproject untouched, ax-identifier gate and the vendored-model tests pass locally; the one red test in the broader alias pack is a pre-existing TF32 allclose artifact, also red on main.

raullenchai:

pierre427:

Conflicts were additive drift (North-Mini alias entries beside the new `apertus-v1.5-8b-text` entry in `aliases.json`; new vendored-arch registrations beside ours in `utils/tokenizer.py`) and resolved by keeping both sides. One adaptation for a gate added since the last rebase: `model_sizes.json` now carries the `swiss-ai/Apertus-v1.5-8B` footprint (18,440,287,514 bytes) required by `test_model_sizes.py::test_every_text_alias_has_a_manifest_entry`.

Two real fixes beyond the mechanical rebase, both found in pre-push review against the official repository:

- **Wrapper dispatch.** The official `swiss-ai/Apertus-v1.5-8B` config declares the outer multimodal `model_type: apertus1p5` with the decoder nested under `text_config`, so the vendored route (which matched only `apertus1p5_text`) never engaged for the advertised alias. Both names now register to the vendor, and `ModelArgs.from_dict` flattens the wrapper config (nested text config wins over wrapper-level duplicates).
- **Strict weight loading.** mlx-lm's loader globs every `model*.safetensors` shard, which in this repository includes the vision- and audio-tokenizer shards; `sanitize` now drops the `model.vision_tokenizer.*` / `model.audio_tokenizer.*` trunks so strict verification sees exactly the text backbone. A new parity test builds the backbone from the wrapper-shaped config and asserts that sanitizing a synthesized checkpoint key set (every backbone parameter under its wire names, plus representative vision/audio-trunk keys) yields exactly the model's parameter set.

- `test_apertus1p5_text_vendored.py` (now 6 tests incl. wrapper-dispatch and strict-parity), `test_aliases_contract.py`, `test_model_auto_config.py`, `test_mlx_compat.py`, `test_model_sizes.py`: **3272 passed**.

pierre427:
Rebased onto current `main` (`4c03e3dd`) and force-updated with an exact lease. New head: `4054f6b9`.

The only conflict was the tokenizer registration insertion point added by current-main Qwen4 work; the resolution preserves both the Qwen4 registration and the Apertus 1.5 wrapper/text registrations.

Verification:

- focused suite: **3322 passed**;
- `ruff check`: passed;
- `git diff --check`: passed.

raullenchai:
Closing this stale vendor-integration PR because the dependency and Rapid-MLX model-loading architecture have moved substantially since the submitted head. Before reviving this model family, we need to re-check whether current dependencies already provide the architecture and then qualify the smallest current-main compatibility layer, if any. Thank you for the original implementation and validation.

PR #1233, opened 2026-07-26
fix(mtp): guard extract_mtp_weights against norm double-shift

raullenchai:
## Why

`scripts/extract_mtp_weights.py` unconditionally shifts every 1-D MTP norm weight by `+1.0` (the `if "norm" in k and k.endswith(".weight") and ndim == 1` loop) to match the `(1 + w)` convention mlx-lm's sanitize bakes into the backbone norms. That shift is correct **only** when the HF source stores its MTP norms unshifted (HF convention).

If a source already stores MTP norms in MLX's shifted convention, applying `+1.0` again **silently double-shifts** them (`w+2` instead of `w+1`). The norms end up off by 1, draft acceptance collapses to ~0%, and there is **no error** — the exact same symptom as the *missing*-shift bug fixed in #1214, just from the opposite direction. An external contributor flagged "converted-MTP norm double-shift prevention" as a real MTP-correctness gap.

## What

Add a principled double-shift guard (no fragile absolute-magnitude threshold — norm distributions vary too widely for that):

- Use the target MLX model's **backbone** norms as the reference convention. After mlx-lm sanitize they are already `+1`-shifted, so they anchor the "shifted" reference.
- Pair an MTP norm with the **same-type** backbone norm (e.g. `mtp.layers.0.input_layernorm.weight` ↔ `...layers.0.input_layernorm.weight`), read that one backbone tensor from the local MLX model, and compare means.
- Decision extracted into a pure, unit-testable helper `_norms_already_shifted(mtp_mean, backbone_mean)`: an unshifted MTP norm sits ~1.0 below the backbone; if the MTP mean is instead closer to the backbone mean, the norms are already shifted → skip the `+1.0`.
- On detection: **skip** the shift and log a loud warning naming both means.
- Add `--force-norm-shift` escape hatch to force the shift regardless.
- If no backbone reference can be found (unusual layout), fall back to the **current** behavior (shift) and log that the guard could not run.

The reference is read from the target MLX model directory (`--mlx-model`) rather than the HF source, because the guard needs a *known-shifted* anchor: the HF source and its MTP norms share one convention, so the HF backbone cannot distinguish a normal (unshifted) MTP head from a pre-shifted one — only the post-sanitize MLX backbone can. This is also network-free.

## Tests

`tests/test_mtp_extract_double_shift.py` exercises the pure helpers with synthetic means only (no network, no model downloads):
- unshifted case (`mtp = backbone - 1` → shift applied),
- already-shifted case (`mtp ≈ backbone` → shift skipped),
- boundary/midpoint cases (equidistant prefers shift; just-past-midpoint detects shift),
- `_find_backbone_reference` same-type matching and the no-counterpart → `(None, None)` fallback.

```
$ pytest tests/test_mtp_extract_double_shift.py -q
6 passed
```

`ruff check` + `ruff format` clean on both files.

## Notes

AI-assisted (Claude Opus 4.8). Behavior for well-formed sources is unchanged — normal (unshifted) sources still get the `+1.0` shift; only genuinely pre-shifted sources are spared the double-shift.

### DIFF
--- a/scripts/extract_mtp_weights.py
+++ b/scripts/extract_mtp_weights.py
@@ -39,6 +39,82 @@ def _quantize_weight(w, group_size, bits):
     return q_w, q_s, q_b
 
 
+def _norm_type(key: str) -> str:
+    """Norm-type tail used to pair an MTP norm with its backbone counterpart.
+
+    ``mtp.layers.0.input_layernorm.weight`` -> ``input_layernorm.weight``.
+    """
+    return ".".join(key.split(".")[-2:])
+
+
+def _find_backbone_reference(mtp_norm_keys, backbone_keys):
+    """Pick an ``(mtp_key, backbone_key)`` pair that share a 1-D norm type.
+
+    ``backbone_key`` is a non-``mtp.`` norm present in ``backbone_keys`` (the
+    target MLX model's tensor names). That backbone norm is the +1-shifted
+    reference produced by mlx-lm's sanitize. Returns ``(None, None)`` when no
+    shared norm type exists.
+    """
+    non_mtp = [k for k in backbone_keys if not k.startswith("mtp.")]
+    for mtp_key in mtp_norm_keys:
+        ntype = _norm_type(mtp_key)
+        for bk in non_mtp:
+            if bk == ntype or bk.endswith("." + ntype):
+                return mtp_key, bk
+    return None, None
+
+
+def _norms_already_shifted(mtp_mean: float, backbone_mean: float) -> bool:
+    """Return True when the extracted MTP norms are ALREADY +1-shifted.
+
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing: not worth merging. The double-shift only occurs if an ALREADY-converted (shifted) head is fed to extract_mtp_weights.py, but the script's real use is base-model → sidecar, where the current unconditional +1.0 shift is correct. A *robust* guard is a genuine design task (per-norm exact-counterpart comparison + fail-closed on ambiguity), and codex correctly flagged that this naive backbone-mean heuristic can itself misclassify independently-trained MTP norms and double-corrupt them — i.e. worse than no guard. Not proportionate for a rare scenario. If revisited, do it as a diagnostic warn-only pass, not an auto-skip.

PR #1240, opened 2026-07-26
fix: close release dogfood gaps

raullenchai:
## What changed

- Increase the local unit-test wrapper timeout from 120s to 300s, with a regression test, so a healthy ~160s full suite is not reported as failed.
- Stop successful multimodal requests from being logged and aborted as orphans when consumers close the stream after the terminal result.
- Default DDTree requests to non-thinking mode because the DDTree path does not run the reasoning parser; explicit `enable_thinking=true` is still honored.
- Make the Hermes integration prompt explicitly request `read_file`, removing a sampling-sensitive tool-selection failure.
- Include the `cache_write_tokens` field now required by current OpenAI SDK Responses usage models.
- Clarify that the published Qwen3.6 DFlash number is a historical community result, not a currently enabled model pair.

## Why

Release dogfooding found several small but user-visible gaps: a false test timeout, a misleading Gemma 4 orphan warning, DDTree thinking text leaking into normal content, a flaky Hermes smoke prompt, and incompatibility with the latest OpenAI Python SDK. These are focused root-cause fixes; no release version or unrelated subsystem is changed.

## User impact

- Release validation completes reliably on current Apple Silicon machines.
- Gemma 4 multimodal requests finish cleanly without false abort logs.
- Qwen3.5 DDTree produces concise normal answers by default and remains opt-in for explicit thinking.
- Hermes tool-use smoke testing is deterministic.
- Responses API output parses with current OpenAI SDK releases.
- DFlash docs match current runtime eligibility.

## Validation

- Codex review: no actionable correctness findings.
- Full suite: `13809 passed, 80 skipped, 212 deselected, 6 xfailed` in 161.52s; the wrapper reports PASS at 162.7s.
- Post-rebase focused suite: `113 passed`.
- Real local dogfood:
  - Gemma 4 26B multimodal image request: HTTP 200, correct image description, no orphan-abort log.
  - Qwen3.5 9B DDTree: actual drafter loaded; math/code/Chinese prompts succeeded; changed default verified without `--no-thinking`.
  - Hermes + Qwen3.5 9B: real `read_file` tool call and correct final answer.
  - LangChain + Qwen3.5 9B: 6/6 integration scenarios passed, including streaming, tools, and structured output.
  - DFlash runtime and Qwen3.5 drafter loaded; Qwen3.6 MoE remains correctly unsupported after direct compatibility testing.
- Clean wheel/sdist builds passed release smoke tests and `twine check`.

### DIFF
--- a/docs/benchmarks/llm.md
+++ b/docs/benchmarks/llm.md
@@ -95,7 +95,11 @@ Notes:
 - Throughput peaks at **1024–2048** tokens, then dips at 4096 — consistent with KV-cache pressure as the rolling window fills.
 - Memory delta (baseline → DFlash) tracks the draft model + verification overhead; see the source benchmark for the exact figure.
 
-`Qwen3.6-35B-A3B-8bit` supports DFlash as an experimental opt-in (install `[dflash]` extra, pass `--speculative-config '{"method":"dflash"}'`); numbers above are from a community benchmark and reflect that single run.
+These are historical community results from a single run, not a currently
+enabled model/decoder pair. Rapid-MLX now keeps DFlash behind per-alias
+correctness and speed gates; `qwen3.6-35b-8bit` is an MoE model and does not
+currently pass those gates. Run `rapid-mlx info <alias>` for the live
+eligibility result before enabling DFlash.
 
 ## Prefix Cache Results
 
--- a/scripts/dev_test.py
+++ b/scripts/dev_test.py
@@ -6,8 +6,8 @@ Orchestrates all test levels from quick lint to overnight benchmarks.
 
 Usage:
     python scripts/dev_test.py lint          # ruff + import check (~10s)
-    python scripts/dev_test.py unit          # pytest unit suite (~30s)
-    python scripts/dev_test.py smoke         # lint + unit (~1 min)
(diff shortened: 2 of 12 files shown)


raullenchai:
Validation follow-up for `25e29200`:

- Codex review converged with no blocking findings.
- GitHub CI is fully green: lint, type-check, Apple Silicon, Python 3.10/3.11/3.12, and validation.
- Local full suite: `14001 passed, 80 skipped, 17 deselected, 6 xfailed, 1 xpassed`.
- The first local high-blast 4-model × 3-scenario matrix passed completely.
- On the final validation run, every lane passed except one Qwen3.5/PydanticAI structured-output sample (`5/6`); the validator's base control passed (`6/6`). This lane uses `/v1/chat/completions` and does not execute the Responses usage code changed here. I immediately reran the exact PydanticAI suite against the PR commit with the same Qwen3.5 35B 8-bit model and server flags: `6/6 passed`, including structured output. Treating the one sample as model-output flake rather than changing unrelated product code.
- Direct current-OpenAI-SDK parsing passed for both zero-cache and cached Responses usage payloads.

PR #1244, opened 2026-07-27
fix(dflash): align capabilities and startup UX

raullenchai:
## What does this PR do?

- forks into the dedicated DFlash server before BatchedEngine-only PFlash, TurboQuant, cache, and continuous-batching setup
- reuses the existing Hermes tool-call and Qwen reasoning post-processors for DFlash, including streaming and multi-turn tool-result messages
- rejects explicit parser opt-outs and non-auto `tool_choice` values instead of silently leaking raw markup or claiming unsupported semantics
- documents the `dflash` extra and gives Homebrew users an executable migration path

## Why is this needed?

Without this PR, DFlash startup advertises features from a server path it never uses, Homebrew users receive a pip command that cannot modify the brew-managed environment, tool-capable clients receive a hard 400, and opted-in thinking is returned as raw content. Qwen3.5-27B DFlash already has compatible shared Hermes/Qwen parsers, so these are user-visible integration gaps rather than new parser work.

## AI assistance disclosure

Codex wrote and reviewed the five touched files. Codex verified the change with focused and broader unit tests, a contract mutation spot-check, two iterative Codex review rounds, a clean wheel in a fresh venv, and real Qwen3.5-27B + DFlash drafter inference. Human maintainer review remains visible through this draft PR.

> By submitting this PR I confirm I can explain the intent, risk, and behavior of every non-generated change in this PR. For generated sections, the verification evidence is listed below.

## Test plan

- [x] `python3.12 -m pytest tests/test_dflash_integration.py -q` — 64 passed, 1 skipped
- [x] `python3.12 -m pytest tests/test_postprocessor.py tests/test_tool_calling.py tests/test_reasoning_parsers.py -q` — 354 passed
- [x] Ruff format/check and `git diff --check`
- [x] Mutation: changed `if request.tools or enable_thinking` to an always-false gate; `test_chat_completions_streams_tool_calls` failed with no tool delta and `test_explicit_thinking_is_parsed_into_reasoning_content[True]` failed with no reasoning delta; restored the production condition and both passed
- [x] Codex review round 1: rejected unsupported explicit `tool_choice`; round 2: no correctness findings
- [x] Clean local wheel installed with `[dflash]`
- [x] Real `mlx-community/Qwen3.5-27B-8bit` + `z-lab/Qwen3.5-27B-DFlash`: plain completion, streaming/non-streaming tool call, multi-turn tool result, and streaming/non-streaming explicit thinking all passed

## Checklist

- [x] Focused tests pass locally
- [x] Lint and format pass
- [x] Critical parser-path tests mutation-checked
- [x] README/docs updated where applicable
- [x] No breaking change to the existing default DFlash completion path

## Out of scope

- Forced/disabled tool choice remains on the standard server; DFlash returns a clear 400 for non-auto choices.
- DFlash remains single-user and does not add MCP, embeddings, audio, or prefix caching.

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -58,11 +58,20 @@ The base text-only install is ~460 MB. Vision/audio/etc. ship as opt-in extras.
 | Extra | Install | Adds |
 |---|---|---|
 | `vision` | `pip install 'rapid-mlx[vision]'` | mlx-vlm + opencv + torch (~322 MB) for VLMs (Gemma 4, Qwen-VL, video) |
+| `dflash` | `pip install 'rapid-mlx[dflash]'` | mlx-vlm for DFlash speculative decoding on supported 8-bit aliases |
 | `audio` | `pip install 'rapid-mlx[audio]'` | mlx-audio + spacy + scipy (~600 MB) for TTS / STT |
 | `embeddings` | `pip install 'rapid-mlx[embeddings]'` | mlx-embeddings (~50 MB) for `/v1/embeddings` |
 | `chat` | `pip install 'rapid-mlx[chat]'` | Gradio web UI (~150 MB) |
 | `guided` | `pip install 'rapid-mlx[guided]'` | outlines (~80 MB) for schema-constrained JSON |
 | `all` | `pip install 'rapid-mlx[all]'` | Everything above (~1.1 GB) |
+
+Homebrew installs the text-only package and does not provide Python extras.
+To switch a Homebrew installation to DFlash, use an isolated tool install:
+
+```bash
+brew uninstall rapid-mlx
+uv tool install 'rapid-mlx[dflash]'
+```
 
 ## Verify Installation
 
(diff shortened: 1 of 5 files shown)


raullenchai:
Final validation after rebase onto current `main`:

- Codex review converged: no actionable correctness regressions.
- GitHub checks: all required checks passed on `7726e1a8`.
- DFlash focused suite: 64 passed, 1 skipped; Ruff passed.
- `pr_validate` targeted suite: 645 passed, 1 skipped, 4 deselected.
- Stress matrix: all 4 profiles × 3 repetitions passed.
- Full-unit run: 14,075+ tests passed; the sole failure was the known environment-sensitive batching throughput gate. The exact same failure reproduced on clean current `main` under the same full-suite load (PR: 0.68x, main: 0.65x, threshold: 0.70x), while the isolated test passes. This is not caused by this PR.
- Fresh wheel dogfood with real `mlx-community/Qwen3.5-27B-8bit` + `z-lab/Qwen3.5-27B-DFlash`: clean DFlash-only startup banner, structured tool call, reasoning split, streaming/non-streaming, and multi-turn tool-result flow all passed.
- Mutation spot-check proved the new tool/thinking tests fail when post-processing is bypassed.

PR #1245, opened 2026-07-27
fix: pad ragged sorted MoE gathers

pierre427:
## Summary

Installs an opt-in Rapid compatibility guard for mlx-lm's sorted MoE `gather_qmm` / `gather_mm` path. When the flattened routed-row count is greater than 32768 and not divisible by 64, the shim pads the sorted rows by duplicating the final sorted row before the affine gather runs.

## Why

This is a correctness-only guard for a quantized MoE row-alignment corruption class. The risky shape is directly relevant to Rapid's MoE model paths, and the fix is intentionally narrow: sorted, large, ragged routed-row gathers only. Aligned shapes and small shapes are unchanged.

The patch is installed from the existing `_mlx_compat.install()` hook so plain `import vllm_mlx` still avoids importing `mlx.core` / `mlx_lm` on metadata-only or non-Apple-Silicon paths.

## Validation

- [x] `python3 - <<'PY'` CPU-forced `pytest tests/test_switch_layers_gather_pad.py -q` -> 3 passed
- [x] `/Users/pierrelamy/Desktop/mlx-uag/.venv-vlm/bin/ruff check vllm_mlx/_mlx_compat.py vllm_mlx/patches/switch_layers_gather_pad.py tests/test_switch_layers_gather_pad.py`
- [x] `/Users/pierrelamy/Desktop/mlx-uag/.venv-vlm/bin/ruff format --check vllm_mlx/_mlx_compat.py vllm_mlx/patches/switch_layers_gather_pad.py tests/test_switch_layers_gather_pad.py`
- [x] `git diff --check -- vllm_mlx/_mlx_compat.py vllm_mlx/patches/switch_layers_gather_pad.py tests/test_switch_layers_gather_pad.py`

Note: a Qwen3.5-122B MLX quality-gate run was already active on the machine, so I avoided default-GPU validation and forced the focused MLX regression test onto CPU to avoid contention.

### DIFF
--- a/tests/test_switch_layers_gather_pad.py
+++ b/tests/test_switch_layers_gather_pad.py
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for the sorted gather_qmm/gather_mm row-alignment guard."""
+
+from types import SimpleNamespace
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+from vllm_mlx.patches.switch_layers_gather_pad import install
+
+
+def _module_with_gather_sort():
+    calls = []
+
+    def _original_gather_sort(x, indices):
+        calls.append(True)
+        *_, M = indices.shape
+        indices = indices.flatten()
+        order = mx.argsort(indices)
+        inv_order = mx.argsort(order)
+        return x.flatten(0, -3)[order // M], indices[order], inv_order
+
+    return SimpleNamespace(mx=mx, _gather_sort=_original_gather_sort, calls=calls)
+
+
+def test_sorted_gather_pad_aligns_large_ragged_row_count():
+    module = _module_with_gather_sort()
+
+    assert install(module) is True
+
+    tokens, top_k = 8193, 4
+    x = mx.arange(tokens, dtype=mx.float32).reshape(1, tokens, 1)
+    x = mx.expand_dims(x, (-2, -3))
+    indices = (mx.arange(tokens * top_k, dtype=mx.int32) % 8).reshape(1, tokens, top_k)
+
+    sorted_x, sorted_indices, inv_order = module._gather_sort(x, indices)
+
+    assert module.calls == [True]
+    assert indices.size == 32772
+    assert sorted_indices.size == 32832
+    assert sorted_indices.size % 64 == 0
+    assert inv_order.size == indices.size
+    assert sorted_x.shape[0] == sorted_indices.size
+    assert mx.all(sorted_x[-1] == sorted_x[-2]).item()
+    assert sorted_indices[-1].item() == sorted_indices[-2].item()
+
+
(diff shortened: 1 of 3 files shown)


raullenchai:
Thanks for the careful write-up, @pierre427.

Before landing I reproduced the claimed row-alignment corruption at the kernel level: identical quantized weights and identical sorted rows, flipping **only** `sorted_indices` on `mx.gather_qmm`, swept across R = 32772 … 131076 (all > 32768 and not 64-aligned), on our shipped **mlx 0.31.2**, across `affine` (4-bit and 8-bit) and **`mxfp4`** (the quant our production Qwen3.5-122B MoE actually uses).

In every case `sorted_indices=True` matches the unsorted reference to floating-point rounding noise:

| mode | worst rel Δ (R>32768, %64≠0, up to R=131076) |
|---|---|
| affine 4b/gs64 | 1.5e-6 |
| affine 8b/gs64 | 1.9e-6 |
| mxfp4 4b/gs32 (bf16) | 2.7e-3 |
| mxfp4 4b/gs32 (fp16) | 6.7e-4 |

Corruption would show as rel Δ ~1e-1 / NaN / garbage; nothing is close. So the sorted-gather optimization is faithful on the mlx we ship — this looks like a bug in an older mlx that's since been fixed upstream.

Since the patch monkeypatches `mlx_lm.models.switch_layers._gather_sort` to guard a corruption class that no longer manifests on our mlx, I'm going to decline it rather than carry a speculative patch on an mlx-lm internal (which would also silently misalign if `_gather_sort`'s signature/behavior changes upstream). Note the added test only exercises the padding mechanics against a mocked `_gather_sort`, so it doesn't demonstrate the corruption itself.

If you can share an mlx version + concrete shape that still corrupts on **0.31.2**, I'll gladly reopen and land a fix. Genuinely appreciate the contribution and the detailed analysis.

pierre427:
Thanks for running the sweep — your non-repro is actually the expected result on your fleet, and it's my miss for not flagging the hardware dependency when filing. The corruption is specific to the gen-17 (M5) Neural-Accelerator kernel paths: the sorted `gather_qmm` row overflow only exists in the NAX kernels, which M1–M4 never dispatch to (and a couple of the failing shapes also need fp16/bf16 inputs to route there — an fp32 sweep bypasses the kernel entirely). So 0.31.2 on M1–M4 can't hit it no matter the shape.

The root cause is being fixed upstream: ml-explore/mlx#3922 (plus the #3912/#4009 companions) — we validated the full family on M5 hardware and #3922 is now marked ready for review. Once those land and Rapid-MLX bumps mlx, your fleet inherits the fix for free, so I think keeping #1245 closed is the right call — the pad workaround only earns its keep if you ship on M5 before the upstream fix arrives. Happy to hand you an M5 repro script if you ever want one for your own records.

PR #1246, opened 2026-07-27
fix: roll back MTP verify cache on early abort

pierre427:
## Summary

- roll back the uncommitted target-cache draft suffix if an MTP verify round aborts during the single sync / host materialization boundary
- centralize MTP verify-round rollback so the normal reject path trims the MTP draft cache through the same helper
- add a regression test that injects a materialization fault after target verify advances caches but before fresh accept/reject state is read

## Why

The target verify forward appends `[current_token, draft_1, ..., draft_K]` to the main model cache, while the MTP drafter has already appended the K proposed draft positions to its own cache. If cancellation, an injected fault, or a host materialization error fires before the fresh accept state is safely available, the generator should keep the already-committed current token and drop every uncommitted draft position from both caches.

This is intentionally narrow: normal all-accept behavior is unchanged, and the existing reject behavior now goes through the same combined rollback helper.

## Validation

- [x] CPU-forced focused tests: `tests/test_mtp_spec_decode.py::test_generator_rolls_back_verify_round_on_early_materialization_abort`, `::test_generator_emits_first_token_from_backbone_then_draft`, `::test_generator_rejection_path_does_not_count_as_accept` — 3 passed
- [x] `ruff check vllm_mlx/spec_decode/mtp/generator.py tests/test_mtp_spec_decode.py`
- [x] `ruff format --check vllm_mlx/spec_decode/mtp/generator.py tests/test_mtp_spec_decode.py`
- [x] `git diff --check -- vllm_mlx/spec_decode/mtp/generator.py tests/test_mtp_spec_decode.py`

Full `tests/test_mtp_spec_decode.py` was attempted CPU-only in this worktree and is environment-blocked on the local system Python 3.9 / missing optional deps (`fastapi`, `mlx_lm.models.qwen3_5`), with the same broader toolchain mismatch outside this diff. Default-GPU validation was intentionally not run because a local Qwen3.5-122B quality gate is active on the machine.

### DIFF
--- a/tests/test_mtp_spec_decode.py
+++ b/tests/test_mtp_spec_decode.py
@@ -2192,6 +2192,55 @@ class _MockedQwen35Model:
         return []
 
 
+class _CountingKVCache:
+    """Tiny trimmable cache double for MTP rollback accounting tests."""
+
+    def __init__(self):
+        self.offset = 0
+        self.trim_calls: list[int] = []
+
+    def is_trimmable(self):
+        return True
+
+    def trim(self, n):
+        if n < 0:
+            raise AssertionError(f"negative trim: {n}")
+        self.trim_calls.append(n)
+        self.offset -= min(self.offset, n)
+
+
+class _CacheAdvancingQwen35Model(_MockedQwen35Model):
+    """Mock that advances supplied cache doubles on each forward."""
+
+    def __init__(self, backbone_outputs: list[int], mtp_outputs: list[int]):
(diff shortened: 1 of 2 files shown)


raullenchai:
Landed via #1464 (merged as `ccac600e`). Your fix was cherry-picked onto current `main` after #1441 reworked the same verify region — original commit authorship preserved. Full MTP suite + a fresh pr_validate (codex clean, 15461 unit tests) pass with both changes together. Closing this as superseded; thanks @pierre427!

Issue #1250, opened 2026-07-27
[P1] Telemetry alert on garbage / empty-completion rate per released version


raullenchai:
## Why

We ship a minor roughly every **2 days** — too fast for a staged rollout / soak (a release-candidate + soak step was considered and **dropped** for exactly this cadence reason). The practical fast-feedback substitute is **real-user telemetry**, which is already deployed (Phase 2). A version that ships garbage should light up within minutes of adoption, not wait for a bug report.

## Proposed

- Aggregate per-**version** signals already derivable from opted-in traffic: empty-completion rate, abnormally-short-response rate, error rate, and a cheap client-side garbage heuristic (repetition ratio).
- Alert / auto-flag when a **newly-released version's** rate deviates beyond a threshold from the trailing baseline.
- Optionally publish a "known-bad version" advisory + fast-path to yank / re-bump.

## Acceptance

- A released version carrying a coherence regression triggers an alert from real-user telemetry **without** a manual bug report.
- Dashboard/alert is keyed by `version`.

Note: this is the deliberate replacement for the dropped "RC + soak" idea — at a 2-day cadence, fast **post-release** telemetry is the realistic canary, not pre-release soak.

Umbrella tracking issue: #320.

Issue #1247, opened 2026-07-27
[P0] Add an always-on output-coherence gate (real serve path, golden answers + garbage detector)


raullenchai:
## Why (incident-grounded)

Qwen3.6-35B-A3B and Qwen3.5-35B-A3B-8bit shipped producing **pure garbage from the first token** (a doubled RMSNorm scale from an mlx-lm `+1.0` norm-shift misfire). They passed **every** automated gate — 278 delta tests, lint, install/import smoke, perf thresholds — because **no CI gate ever generates a token and checks whether the output is coherent**. Fixed reactively in #1234.

This is a recurring class, not a one-off: coherent-but-wrong or pure-garbage output *at normal speed* passes all of today's CI.

## The gap (verified against the tree)

- `make smoke` = lint + unit only; `make release-smoke` = **import-only** (never generates a token).
- Harness `thresholds.yaml` gates **perf only** (`decode_tps`/`cold_tps`/`ttft_ms`). `tc_success_rate` checks that a tool *fired*, not that the answer is correct.
- The golden-answer assertions that DO exist (`test_anthropic_sdk.py`: "Tokyo", "391", "blue", "72") run only on the maintainer's **local M3 gauntlet**.
- `evals/run_eval.py` (real MATH-500 / MMLU-Pro / HumanEval+ harness) is wired into **nothing** — manual, hand-committed JSON.
- `assert_content_nonempty` catches empty strings but **not** `!!!!!`-class gibberish.
- `stress_e2e_bench` (the only real-server content check in `pr_validate`) is **skipped** on CI via `PR_VALIDATE_SKIP_STEPS`.
- `release-artifact-matrix.yml` *does* boot a real server + assert content — but is `workflow_dispatch`-only and targets an **unregistered** self-hosted runner, so it never actually runs.

## Proposed (concretizes #320 M1 + M3)

1. **Golden-smoke eval** — ~8 fixed prompts per representative alias, `temperature=0`, run against the **real serve path** (boot server → HTTP request), asserting checkable content:
   - `capital of Japan → "Tokyo"`, `17 × 23 → "391"`, `sky color → "blue"`
   - a tool-call fires with correct args
   - no `<think>` leak into the final answer
   - `/v1/chat/completions` and `/anthropic` decode to the **same** text (closes the surface-divergence class from #288/#289).
[...]

PR #1253, opened 2026-07-27
feat(evals): output-coherence release gate (#1247)

raullenchai:
## What

Adds the missing **output-coherence gate** — the one class of gate that would have caught the Qwen3.6/3.5-35B shipped-garbage incident. Closes the core of #1247.

## Why

Qwen3.6-35B-A3B and Qwen3.5-35B-A3B-8bit shipped producing **pure garbage from the first token** (doubled RMSNorm scale, fixed in #1234) yet passed **every** automated gate — 278 delta tests, lint, install/import smoke, perf thresholds — because **nothing generates a token and checks whether the output is coherent**. Coherent-but-wrong or pure-garbage output at normal speed passes all of today's CI. This adds the missing check.

## Design — two layers

**`vllm_mlx/coherence.py`** — pure, dependency-free (no server, no MLX), so it unit-tests on a GitHub-hosted runner:
- `looks_like_garbage()` — deterministic detector for the collapse classes we've actually shipped/seen: `!!!!!` cache-poison, doubled-norm token soup, single-word / bigram loops, punctuation-only, empty. Conservative — fires only on unambiguous degeneracy, so real prose and short answers (`"7"`, `"42"`) never trip it. Reusable by the #1250 telemetry garbage-rate alert.
- `GOLDEN` + `evaluate_case()` — fixed prompts with checkable properties (capital of Japan → Tokyo, 17×23 → 391, sky → blue, echo `banana`, no `<think>` leak, open-ended → non-degenerate). A coherent-but-**wrong** regression fails deterministically at `temperature=0`.

**`evals/coherence_gate.py`** — serve-path runner: feeds `GOLDEN` through a **real running server** at temp=0, applies `evaluate_case`, exits `0`/`1` (`2` if unreachable). Reads `RAPID_MLX_BASE_URL` like the existing SDK gauntlet — no new boot orchestration.

**`scripts/release_check_m3.sh`** — runs the gate **first** after server boot (new `G4`), before stress/SDK: a garbage model makes every later gate moot.

## Testing

- **28 always-on unit tests** (`tests/test_coherence.py`, no server/MLX) lock in the gate logic itself — this is the "test the gate" acceptance from #1247: every garbage class is flagged; real prose + short answers are not; a coherent-but-wrong answer AND garbage-for-a-golden-prompt are both rejected (the exact #1234 class).
- **End-to-end against a live `qwen3.5-4b-4bit` server: 8/8 golden cases coherent, exit 0.** Both paths exercised against the real server (the pass path and, during dogfood, the fail path with the "COHERENCE GATE FAILED" banner + exit 1).
- Dogfood caught a real false positive in the detector (single-token `"7"` tripping char-dominance) — fixed and guarded by a regression test.

## Follow-up (ops, tracked in #1247)

To make the serve-path gate **blocking in CI** (not only in the local M3 gauntlet), the self-hosted Apple-Silicon runner needs registering — GitHub-hosted `ubuntu-latest` has no MLX runtime (which is also why `pr-validate.yml` skips `stress_e2e_bench`). The pure unit tests are already always-on.

### DIFF
--- a/docs/development/releasing.md
+++ b/docs/development/releasing.md
@@ -158,6 +158,7 @@ This is the rule. No exceptions. CI doesn't fake-inference with a tiny model on 
 | G2 | Codex review × 2 rounds | local | maintainer machine | every PR-author bug class |
 | G3 | CLI ↔ Config fidelity audit | CI | `ci.yml` lint (ubuntu) | silent CLI flag drop (#400) |
 | G4 | unit suite (≈4500 tests) | CI | `ci.yml` test-matrix (linux) + test-apple-silicon (macOS-14) | parser/router regressions |
+| G0 | Output-coherence gate — deterministic golden answers (blocking) + garbage detector (advisory), run FIRST | **M3** | `make release-check-m3` + `scripts/coherence_sweep.sh` | garbage / coherent-but-wrong generation the perf/import/unit gates all miss (#1234 / #1247) |
 | G5 | `make stress` — 8 scenarios | **M3** | `make release-check-m3` | concurrent-batching regressions |
 | G6 | Live-server fix-path repro | **M3** | `make release-check-m3` | fix doesn't ship to user-visible path |
 | G7 | SDK integration (anthropic / pydantic_ai / smolagents) | **M3** | `make release-check-m3` | router-level breakage unit tests miss |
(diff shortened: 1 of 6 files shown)


raullenchai (commented):
Review result: **do not merge yet**.

The release gap is real and worth fixing: current CI can prove that inference runs, but not that the served model still answers coherently. I independently ran the real-server gate against `qwen3.5-4b-4bit`; the deterministic cases passed 8/8.

The current implementation is not yet correct as a blocking coherence gate:

1. The generic garbage detector cannot reliably distinguish diverse token soup from valid text. The current open-ended predicate still accepts outputs such as `Ocean qzxv blorp fnarg glip. Water wug zibble plonk traz.`. Adding more character/word-frequency rules will keep trading false negatives for false positives.
2. The live gate currently runs only when the maintainer manually invokes the M3 release gauntlet. It is not yet the “always-on / unskippable” required gate described by #1247.
3. The gauntlet tests one selected model. A model-specific regression such as the 35B RMSNorm incident can still pass if the release check is run against the default 4B/9B alias.

Recommended scope to converge without overengineering:

- Keep the strict, normalized golden-answer checks; they are deterministic and catch both fluent-but-wrong output and the shipped garbage failure.
- Treat the heuristic garbage detector as advisory/diagnostic rather than a universal blocking classifier, or remove the open-ended blocking cases.
- Run the real serve-path golden gate on the representative aliases affected by a release/model-path change, and make that release requirement explicit/automated before calling it always-on.

Local validation included targeted tests, lint/format, a full earlier suite (`14,068 passed, 82 skipped`), and several Codex/pr_validate review iterations. The latest `pr_validate` remains **DO NOT MERGE** because the blocking detector can still false-green on diverse token soup.

raullenchai:
Converged per the review (pushed in `68264e7`). I kept your `exact`-match work — it is stricter and better than the original substring `contains`, so I built on it rather than reverting — and applied the three recommendations to the parts the earlier commits left open:

**1. Strict deterministic golden answers = the only blocking layer.** Kept `exact` + `no_think_leak`. Garbage and coherent-but-wrong both fail here by simply not matching the token, so the blocking layer has no dependence on the heuristic and cannot false-green on token soup.

**2. Garbage detector → advisory; open-ended cases removed.**
- Removed the `looks_like_garbage` **blocking pre-screen** from `evaluate_case`. The runner now prints it as a non-blocking diagnostic (ADVISORY section); it never changes the exit code. Kept the function for its reliable classes (`!!!!!`, single-token soup, exact loops) and for #1250 telemetry reuse.
- Removed the `two_sentence` open-ended cases. They still false-greened on the exact adversarial input you flagged — `"Ocean qzxv blorp fnarg glip. Water wug zibble plonk traz."` has `ocean`+`water` + two sentences, so it passed. That's your "DO NOT MERGE" case. Your rec #2 offered "or remove the open-ended blocking cases" — done. A new unit test pins this: the advisory detector returns `False` on that soup, but the blocking `capital-japan` check rejects it (not an exact `Tokyo`).

**3. Real serve-path gate on representative aliases.** Added `scripts/coherence_sweep.sh` — boots + gates each alias in turn — and made it a **hard release requirement** in `docs/development/releasing.md` (new `G0` gate row + explicit "sweep the representative aliases a release/model-path change affects; a 35B-specific regression is invisible to a 4B/9B-only run").

**Validation — sweep on the incident model + starter (exact matching, `--no-thinking`, temp=0):**

```
→ qwen3.5-4b-4bit   BLOCKING: 6/6 golden cases passed
→ qwen3.6-35b       BLOCKING: 6/6 golden cases passed   (the 35B RMSNorm model, #1234 fix active)
SWEEP PASSED — all aliases coherent
```

35 pure unit tests pass (GitHub-hosted CI, no MLX); lint clean.

**On "always-on / unskippable" (your point 2):** the pure predicate + detector logic is now always-on via `tests/test_coherence.py`. The serve-path gate is an explicit, automated **release requirement** (the sweep) but runs on the M3 gauntlet, not GitHub CI — making it *CI-blocking* needs a registered self-hosted Apple-Silicon runner, which is the remaining ops step tracked in #1247. Flagging that as out-of-scope for this PR rather than silently claiming always-on.

Issue #1249, opened 2026-07-27
[P1] Coherence coverage must include EXISTING model families + refresh stale harness baselines


raullenchai:
## Why

W5 pre-release fleet dogfood requires a coherence proof "for every **new** model family." But Qwen3.6 / Qwen3.5-35B-A3B were **not new** — they broke from underneath when the toolchain (mlx-lm) shifted. Coverage keyed on "new models" structurally misses **toolchain-induced regressions in already-shipped models**.

Also: `harness/baselines/` holds only 2 aliases (`qwen3.5-35b`, `qwen3.6-35b`), captured at **v0.6.10** — stale. The "regression vs baseline" gate is therefore meaningful for two models and silent for the rest of the fleet.

## Proposed

1. Each release, rotate a **sample of existing aliases** through the coherence gate — one representative per family (small dense, hybrid MoE, large dense, large MoE, multimodal) — not only the newly-added ones.
2. On any `mlx*` dependency bump, sweep **all** families.
3. Auto-refresh + expand `harness/baselines/` to ≥ 1 representative alias per family; warn when a baseline is older than the last release.

## Acceptance

- A toolchain change that breaks an existing (non-new) family is caught pre-release, not by a user.
- Baseline coverage ≥ 1 alias per family; baseline staleness is surfaced automatically.

Umbrella tracking issue: #320 (M5 canary fleet / L2 continuous baseline maintenance).

PR #1262, opened 2026-07-27
feat(release): centralize fleet coverage

raullenchai:
## Summary

- define one release-fleet manifest for architecture-risk representatives and artifact-matrix families
- make the M3 coherence gate sweep the normal fleet before its exact gauntlet model
- automatically add the Ultra-only Hy3 path only when the MLX toolchain changed from the previous release
- keep shallow checkouts on the normal fleet and preserve explicit diagnostic overrides

This is the coverage half of #1249. Baseline inventory/refresh policy will follow separately so this PR stays focused.

## Business impact

A release can no longer pass by exercising only the cheapest/default model path while missing a model-family regression. The same manifest also prevents the artifact acceptance matrix and local release sweep from silently drifting apart.

## Validation

- Python 3.10: 34 focused tests passed
- `ruff check` passed
- shell syntax checks passed
- real local Qwen 3.5 4B coherence sweep: 6/6 passed
- Codex review converged with no actionable findings

## Scope note

MiniCPM5 was evaluated as a possible small-dense canary but was intentionally not made blocking: the current checkpoint missed two semantic golden answers despite coherent output, which would create a model-capability false alarm rather than protect release correctness.

### DIFF
--- a/docs/development/release-artifact-acceptance.md
+++ b/docs/development/release-artifact-acceptance.md
@@ -68,7 +68,11 @@ The matrix is one server boot per family. A test run sets both
 - a documented `strict=True` xfail remains an exception, and an unexpected
   pass is a failure.
 
-The current feasible set is four families:
+The current feasible set is four families. The matrix aliases and required
+extras come from the same
+[`scripts/release_fleet.json`](../../scripts/release_fleet.json) manifest as
+the release coherence sweep, so its feasible stand-in aliases and required
+extras cannot drift from the broader coherence fleet:
 
 | Family | Candidate alias | Agent cells | Framework cells |
 | --- | --- | ---: | ---: |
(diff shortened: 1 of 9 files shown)


raullenchai:
Validation note: the full-unit gate was run twice on the previous HEAD. Both runs passed 14,160 tests and failed only the known environment-sensitive performance assertion `tests/test_batching_deterministic.py::TestBatchingPerformance::test_batched_faster_than_sequential`. The same test passed immediately in isolation on both this PR branch and `origin/main`; this PR does not touch batching/runtime code. I am treating that failure as pre-existing/flaky per the repository A/B scope rule, not expanding this release-tooling PR to change the benchmark test.

The two actionable findings from the second adversarial review (release-tag matching and complete MLX lock-entry comparison) are fixed on the current HEAD with regression coverage.

PR #1264, opened 2026-07-27
test(release): restore benchmark baselines

### LINKED ISSUE #1249: [P1] Coherence coverage must include EXISTING model families + refresh stale harness baselines

raullenchai:
## Summary

- replace two orphan v0.6.10 doctor baselines with schema-v1 files consumed by the current `pr_validate` benchmark
- cover all five selectable candidates in `golden_models.yaml`, including the Qwen3.6 low-memory fallback
- audit missing, invalid, mismatched, orphaned, and pre-release baselines before the M3 release gauntlet boots models
- keep refreshes review-only: benchmark output is a candidate and never overwrites a committed baseline
- remove the unused doctor threshold file and rewrite stale harness/SOP documentation

This completes the baseline half of #1249 after #1262 added the release coherence fleet.

Closes #1249

## Why this design

Performance baseline inventory follows the actual benchmark consumer (`golden_models.yaml`), not the coherence manifest. That prevents cosmetic coverage files that no live gate ever reads. Every RAM-selectable candidate must have a comparison point.

Repeated M3 Ultra captures showed stable AR paths fit a 5% threshold, while DiffusionEngine and gpt-oss cold latency have wider measured noise. The schema records independent sample count plus reviewed cold/warm thresholds so noise is explicit without weakening stable metrics.

## Validation

- real M3 Ultra captures for all 5 candidates, 2-3 independent runs each
- `release_baselines.py --strict-stale`: 5/5 covered
- 67 focused release/pr_validate tests passed
- Python 3.10: 30 compatibility tests passed
- new audit code: 100% line coverage
- ruff, formatting, shell syntax, and diff checks passed
- Codex blocker review converged with no actionable findings


raullenchai:
Validation note:

- Full unit passed 14,179 tests and failed only the known environment-sensitive `test_batched_faster_than_sequential` performance assertion. It passes in isolation on both PR and main and this PR does not touch batching/runtime; classified as pre-existing flaky per the repository A/B scope rule.
- The adversarial review blocker (non-injective model-ID filename collision) is fixed on the current HEAD by rejecting distinct candidates that map to the same consumer filename, with a regression test.
- Supply-chain warning was reviewed: the workflow change only adds the already-core PyYAML package to the Linux test environment and adds `tests/test_release_baselines.py` to the existing matrix; no action, permission, or executable hook changed.

raullenchai:
Review disposition before merge:

- Hardware mismatch is intentionally non-blocking. Absolute request latency captured on an M3 Ultra is not a valid regression threshold for an M3 Max or M4. The benchmark returns an explicit skip with the measured artifact and hardware mismatch in the scorecard; it is not reported as a performance pass, and the functional stress/agent gates still run. Model-revision mismatch remains blocking and requires a baseline refresh. Building a hardware-profile baseline matrix is outside #1249.
- The release-audit dependency finding is a false positive: PyYAML >= 6.0 and jsonschema >= 4.0.0 are already declared as core project dependencies in pyproject.toml (dependency entries near lines 78 and 86). The release gauntlet runs from the installed project environment.

This keeps the release gate conservative without comparing invalid cross-chip numbers or expanding this PR into a benchmark infrastructure project.

PR #1273, opened 2026-07-28
feat(telemetry): first_session + auto_selected on session_start — activation funnel (#1272)

### LINKED ISSUE #1272: telemetry: add first_session + auto_selected to session_start (lights up the activation-funnel last stage)
## Why

The `/admin` acquisition→activation funnel is fully live except **one** stage. Every upstream step (site visits, install surface, CTA click, copy-install, `install.sh` fetch) already pulls real edge data. The last pending box — **"Server first run"** — is sourced from *engine opt-in telemetry* and is the only gap. Its data source annotation is literally `opt-in CLI telemetry · engine repo`; the website side is wired to consume it, the engine just doesn't emit the signal yet.

Two consequences of the gap:
1. We can't see **activation** (install → first real server run) — the single most important funnel conversion.
2. We can't attribute the **first-run wizard's** lift. Without a flag for "did the wizard auto-drive this run," a before/after comparison of the wizard is unattributable.

## Proposed change — two booleans on `session_start` (schema v2, **no new event type**)

Ride the existing `session_start` event; add two optional fields to `SessionPayload` (`vllm_mlx/telemetry/schema.py`):

- **`first_session: bool = False`** — set `True` when the client has **no local session-history record** (i.e. this machine's first run). The *client* knows its own "first time" more reliably than the server, which can only infer it from the R2 data window (a client that predates the window looks "new").
- **`auto_selected: bool = False`** — set `True` when `chat` was invoked **with no model arg** and fell back to the starter default (`FIRST_RUN_MODEL = "qwen3.5-4b-4bit"`, see `vllm_mlx/first_run.py`). This separates *wizard-driven* activation from users who explicitly named a model, so the wizard's contribution is attributable.

Implementation notes (matching existing conventions in `schema.py`):
- Add **at the end** of `SessionPayload` (after `flag_names`), with `False` defaults — preserves positional back-compat exactly like the `engine` field's documented slot rule.
- `session_start`-only (like other session metadata).
- **Bump `SCHEMA_VERSION`** in the same change (the schema's own convention for new fields).

## Privacy review points (please gate on these)

Both fields are **session metadata booleans** — **not** derived from generated content (unlike the `output_degenerate` proposal in #1266/PR, which is why this one is much lower-sensitivity). No IP, no content, no PII; consistent with the hard privacy line (opt-in, version/metadata only).

Still needs a reviewer sign-off on **consent scope**:
- The existing first-run opt-in disclosure covers `session_start` (subcommand, version, platform buckets). Do two new booleans — "is this your first run" and "did we auto-pick the model" — fall within that disclosure, or should the disclosure text be updated to name them?
- `first_session` reveals "new user" status (benign, but note it).
- My read: within the spirit of the existing `session_start` consent, but **update the disclosure text to mention them** to be safe. This is a founder/consent-owner call, not an implementation detail — please confirm before shipping.

## Cheaper fallback (if the engine change is deferred)

The worker can approximate this **without any engine change**: "new `client_id` whose first-seen day == today AND has ≥1 real `request` event." Downsides vs. the client-side flag: bounded by the R2 retention window (clients predating it look new), and it **cannot** recover `auto_selected` at all. Ship the two fields for an accurate answer; use the server-side heuristic only as a stopgap.

## Full chain to light the funnel (follow-ups, for visibility)

1. **[this issue] engine**: emit `first_session` + `auto_selected` on `session_start` (+ bump schema, + consent review).
2. **telemetry-worker**: roll up distinct `first_session` clients/day (and `auto_selected` split).
3. **landing `/admin`**: flip the "Server first run" funnel stage `pending → live` with a `liveKey`, fetched cross-worker like `collectTelemetryDau`.

## Expected outcome

The funnel's last stage goes live, and **"did the first-run wizard actually lift activation?"** gets a numeric before/after answer ~1–2 weeks after 0.11.x adoption ramps (the new telemetry only covers clients on the shipping version).

raullenchai:
## Why

The `/admin` acquisition→activation funnel is live except **one** stage — **"Server first run"**, sourced from engine opt-in telemetry (everything upstream is already real edge data). This is the last gap, and without an `auto_selected` flag the first-run wizard's lift is unattributable. `Closes #1272`.

Both new fields are **session METADATA booleans** on `session_start` — computed from session context, **never** from a prompt or generated output. This is deliberately *not* the completion-derived kind of signal (contrast the held #1266 `output_degenerate` work); it's the same category as the subcommand / version / platform fields already sent.

## What changed

- **`schema.py`** — append `first_session` + `auto_selected` to `SessionPayload`, `False` defaults, preserving the positional-slot back-compat rule documented on `engine`. **No `SCHEMA_VERSION` bump** — additive optional fields are backwards-compatible (the schema's own rule bumps only on *incompatible* changes; matches how existing additions were handled).
- **`emit.py`** — thread both through `session_start`; also carry them (`False`) on `session_end` so both lifecycle payloads expose the full v1 `SessionPayload` surface (the completeness contract `test_session_start`/`session_end` pins).
- **`first_run.py`** — `mark_first_session()`: atomic `O_EXCL` marker at `~/.rapid-mlx/session_seen`, `True` for exactly one run per machine. Independent of telemetry consent (tracks the true first run even if the user opts in on a *later* run); fail-safe to `False` on any error so it never inflates the funnel's new-client count.
- **`cli.py`** — compute both at the session-lifecycle site (pre-dispatch, where argparse is available). `auto_selected` = `chat` invoked with no positional model (`nargs="?"` default `None`), which is exactly the auto-select-the-starter path (`first_run.select_chat_default`).
- **`consent.py`** — the two booleans are now disclosed under "WHAT WE SEND". We do **not** broaden telemetry silently.
- **`ci.yml`** — add the (mlx-free) telemetry `emit` + `consent` test files to the Linux job. The telemetry suite was previously run *nowhere* on Linux, so these changes would otherwise have no CI coverage.

## 🔒 Needs your sign-off (consent scope)

Both fields are session metadata (no IP, no content, no PII) and I updated the first-run disclosure to name them — but **whether adding session_start fields needs a disclosure update alone or full re-consent is a consent-owner call, not an implementation detail.** Please confirm the wording + posture before merge.

## Testing

3 new tests (both flags emitted + `True`; default `False`; marker `True`-once-then-`False`). Full mlx-free telemetry + coherence suite green locally (110 passed), `ruff` clean. All changed modules are mlx-free (schema / emit / consent / first_run + the cli.py session block).

## Downstream (separate follow-ups, to fully light the funnel)

1. telemetry-worker: roll up distinct `first_session`/day (+ `auto_selected` split).
2. landing `/admin`: flip the "Server first run" stage pending→live with a `liveKey`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,8 @@ jobs:
             tests/test_check_gha_pinning.py \
             tests/test_check_mlx_upstream_calls.py \
             tests/test_microbench_parsers.py \
+            tests/test_telemetry_emit.py \
+            tests/test_telemetry_consent.py \
             -v --tb=short \
             -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache" \
             --cov=vllm_mlx \
--- a/tests/test_telemetry_emit.py
+++ b/tests/test_telemetry_emit.py
@@ -282,6 +282,44 @@ def test_session_start_envelope_when_enabled(opted_in, stub_queue, monkeypatch):
     assert "8000" not in blob
 
 
+def test_session_start_activation_flags_emitted(opted_in, stub_queue):
+    """#1272: first_session + auto_selected land as plain JSON booleans."""
+    from vllm_mlx.telemetry import emit
+
+    emit.session_start(
+        subcommand="chat",
+        first_session=True,
+        auto_selected=True,
+    )
+    session = stub_queue[0]["session"]
+    assert session["first_session"] is True
+    assert session["auto_selected"] is True
+    assert isinstance(session["first_session"], bool)
+    assert isinstance(session["auto_selected"], bool)
+
+
+def test_session_start_activation_flags_default_false(opted_in, stub_queue):
+    """Omitting the #1272 flags emits plain ``False`` -- the keys are always
+    present so consumers can rely on the shape."""
+    from vllm_mlx.telemetry import emit
+
+    emit.session_start(subcommand="serve")
+    session = stub_queue[0]["session"]
+    assert session["first_session"] is False
+    assert session["auto_selected"] is False
+
+
+def test_mark_first_session_true_once_then_false(fake_home):
+    """#1272 client-side marker: first call claims the machine's first
+    session (True); every later call sees the marker and returns False.
+    Independent of telemetry consent -- purely a local ``O_EXCL`` marker."""
+    from vllm_mlx.first_run import mark_first_session
+
+    assert mark_first_session() is True
+    assert mark_first_session() is False
+    assert mark_first_session() is False
+
+
 def test_session_start_models_loaded_redacted(opted_in, stub_queue):
     """Local paths must collapse to "<local>" — not leak home dirs."""
     from vllm_mlx.telemetry import emit
--- a/vllm_mlx/cli.py
+++ b/vllm_mlx/cli.py
@@ -8757,6 +8757,28 @@ Examples:
         )
 
         _session_flag_names = _telemetry_extract_flag_names(_sys.argv[1:])
+        # #1272 activation-funnel signals, computed HERE (before dispatch)
+        # where the argparse result is available. Both are session metadata,
+        # never content.
+        #   - first_session: claim the one-time local marker. This block is
+        #     already skipped on the ``_just_collected_consent`` run (the
+        #     disclosure promises "nothing from before this prompt"), so the
+        #     marker is claimed on the first RECORDED session -- exactly once
+        #     per client -- not the first-ever binary run. That is the funnel
+        #     semantic we want ("first session we recorded from this new
+        #     client"); see ``mark_first_session`` for the full rationale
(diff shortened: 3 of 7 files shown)


raullenchai:
Addressed the codex [P2] on `cli.py` (first-session marker vs. the just-collected-consent run).

The behavior is intentional but my wording overclaimed it. Clarified across `mark_first_session` / `cli.py` / `schema.py` / the disclosure: `first_session` marks the **first RECORDED (participating) session**, not the first-ever binary run. The consent-collection run is deliberately skipped (the disclosure promises "nothing from before this prompt"), so the first participating session carries `first_session=True` — exactly once per client. Following codex's literal reading instead (mark on the consent run, suppress emit) would mean opt-in-on-first-run users **never** emit `first_session=True`, which would silently undercount the whole activation funnel — the opposite of what we want.

**This is the semantic to confirm alongside the consent-scope sign-off**: "first session we recorded from a new client" (one True per client) rather than "first-ever invocation." ruff clean, telemetry emit+consent suite green.

PR #1290, opened 2026-07-29
refactor: remove unused cloud routing

raullenchai:
## What changed

- remove the unused `CloudRouter` implementation
- remove `--cloud-model`, `--cloud-threshold`, `--cloud-api-base`, and `--cloud-api-key` from both serve entry points
- remove the cloud-routing branch from `/v1/chat/completions`
- remove the cloud-only token-estimation engine contract and implementations
- remove dedicated CloudRouter tests and stale test scaffolding
- retain LiteLLM compatibility coverage where LiteLLM is an external client of Rapid-MLX (OpenHands/Aider)

## Why

CloudRouter was an early experiment that no longer has product value. It is undocumented, has no user issue/PR traction, and is not functional from a default install because LiteLLM is not a declared dependency. Cloud fallback belongs in an external proxy such as LiteLLM or OpenRouter, while Rapid-MLX stays focused on local Apple Silicon inference.

This is an intentional removal. If a real hidden user surfaces, the rollback is a single-PR revert.

## User impact

Default `serve`, `chat`, OpenAI-compatible, Anthropic-compatible, MLLM, and external-agent workflows are unchanged. Only users explicitly passing one of the four removed cloud-routing flags are affected; those flags now fail as unrecognized arguments instead of exposing an unsupported path.

## Validation

- affected-path baseline before removal: 105 passed
- affected route/engine tests after removal: 286 passed
- CLI/server/route boundary tests: 201 passed, 3 deselected
- Ruff check and format: pass
- `compileall`: pass
- `git diff --check`: pass
- both serve help entry points contain no removed flags
- wheel build and content inspection: pass
- installed wheel `rapid-mlx --version`, `serve --help`, and `chat --help`: pass
- removed flag rejection: pass
- wheel size: 2,282,871 → 2,276,476 bytes (-6,395 bytes)

Full `pr_validate` and Codex review will run on this draft before merge.


raullenchai:
Review disposition:

The local formal `pr_validate` Codex gate raised two compatibility-policy objections: retaining the four removed CLI flags as deprecated no-ops, and retaining dead positional placeholders in `load_model()`. Those recommendations conflict with the explicitly approved scope of this PR: cleanly remove the unused experiment now, and revert the single PR if a real hidden user surfaces. We are intentionally not preserving dead flags or parameters. Moving the remaining parameters behind `*` ensures old positional calls fail loudly instead of silently rebinding to the wrong meaning.

This is an owner-approved product compatibility decision, not an unresolved implementation defect. Validation evidence on the final commit:

- Manual Codex review, two post-fix rounds: converged with no findings
- GitHub validate, lint, type-check, Python 3.10/3.11/3.12, Apple Silicon: PASS
- Full unit suite: 14,299 passed, 73 skipped, 6 xfailed
- Real-model stress/integration matrix: 4 models × 3 clients, PASS
- Focused route/engine tests: 286 passed
- CLI/server/route boundary tests: 201 passed
- Post-review regression tests: 115 passed + 69 passed + positional-binding tests
- Wheel build/install/help smoke: PASS
- Removed flags reject explicitly; wheel contains no `cloud_router.py`
- Wheel size: -6,395 bytes

Rollback remains a single squash-commit revert if actual usage appears.

Issue #337, opened 2026-05-09
Auth hardening follow-ups from #260 and #262


raullenchai:
Carry-over P1 findings from codex review of #260 and #262 (both merged as critical bypass fixes — these items are hardening of an already-correct fix, not unfixed bypasses).

## From #260 (`vllm_mlx/routes/health.py`)

- [x] **Probe doc update / superseded**: Later routing introduced a separate unauthenticated `probe_router` for `/health`, `/health/ready`, `/healthz`, `/readyz`, and `/livez`, so the original claim that probes receive 401 is no longer true and no authenticated-probe documentation is needed.
- [x] **Strengthen positive-case test assertions** in `test_health_router_accepts_valid_api_key`: replace `assert r.status_code != 401` with explicit per-route expected status codes — completed by #1275.

## From #262 (`vllm_mlx/middleware/auth.py`)

- [ ] **Bearer token whitespace normalization**: `_extract_bearer_token` should `.strip()` the token after `partition(" ")`. Currently `Bearer  test-secret` and `Bearer test-secret ` are auth-equivalent (HTTPBearer normalizes for auth) but route to different rate-limit buckets. Assigned to @2005rishabh.
- [ ] **Duplicate `x-api-key` header**: `request.headers.get("x-api-key")` only evaluates one value. Use `request.headers.getlist("x-api-key")` and either require exactly one value or require all to match — the "both must match" rule should extend to multi-valued headers.
- [ ] **Empty companion header**: `_verify_api_key_values()` filters falsy values, so `Authorization: Bearer test-secret` plus `x-api-key:` (empty) is accepted. If the rule is "if a credential header is present, it must be valid," presence needs to be checked separately from extracted credential value.
- [x] **Constant-time comparison short-circuit**: `all(secrets.compare_digest(...))` short-circuits on first mismatch, leaking which header failed via timing. Fixed by #1271.
- [x] **Non-ASCII header → 500 instead of 401**: invalid non-ASCII credential values are rejected cleanly. Fixed by #1271.
- [x] **Test fixture cleanup not exception-safe during setup**: `tests/test_anthropic_route_auth.py` now restores module state through nested `try/finally`, including when setup or config cleanup raises. Completed by #1275.

## Why follow-up, not block-merge

[...]

PR #1291, opened 2026-07-29
fix(auth): normalize bearer token whitespace for rate limit stability

2005rishabh:
## What does this PR do?
- Adds `.strip()` to `_extract_bearer_token` in `vllm_mlx/middleware/auth.py` so surrounding whitespace (spaces, tabs, newlines) is normalized and an all-whitespace token is treated as absent.
- Adds tests in `tests/test_config_and_middleware.py`: the helper's whitespace edge cases **and** a bucket-level contract test asserting auth-equivalent Bearer values resolve to one `_rate_limit_client_id`.

## Why is this needed?
Closes the "Bearer token whitespace normalization" task from #337.
Without this, a client can dodge its own rate limit just by padding the Bearer token with extra spaces/tabs — the unstripped string hashes to a brand-new bucket identity.

## Scope note (maintainer, narrowed for merge)
Rebased onto current `main` and narrowed to the auth fix + tests per review. Dropped two out-of-scope changes from the original revision: the `if TYPE_CHECKING:` gate on `server_config.py` (it broke runtime `typing.get_type_hints(ServerConfig)` with `NameError: BaseEngine`), and three `except ImportError: pytest.skip` guards on the exception-handler tests (they'd turn a broken/partial MLX install into a green skip and hide packaging regressions). Original authorship preserved on the commit.

## Test plan
- `python3.12 -m pytest tests/test_config_and_middleware.py::TestRateLimitClientId -q` — 5 passed.
- `ruff check && ruff format --check` on both touched files — clean.

## Checklist
- [x] Tests pass locally
- [x] Self-validated via `pr_validate`
- [x] New tests touch a security path — the bucket contract test fails if the `.strip()` is reverted
- [x] Updated README/docs if applicable — N/A (internal auth helper, no user-facing surface)

### DIFF
--- a/tests/test_config_and_middleware.py
+++ b/tests/test_config_and_middleware.py
@@ -370,6 +370,88 @@ class TestRateLimitClientId:
             f"got {bearer_id!r} != {x_api_id!r}"
         )
 
+    def test_extract_bearer_token_whitespace_normalization(self):
+        """_extract_bearer_token strips leading, trailing, and tab/newline whitespace."""
+        from vllm_mlx.middleware.auth import _extract_bearer_token
+
+        assert _extract_bearer_token("Bearer    my-token") == "my-token"
+        assert _extract_bearer_token("Bearer \t my-token \n") == "my-token"
+        assert _extract_bearer_token("Bearer my-token   ") == "my-token"
+        assert _extract_bearer_token("Bearer   ") is None
+
+    def test_rate_limit_bearer_whitespace_variants_share_bucket(self):
+        """Auth-equivalent Bearer values must resolve to one rate-limit bucket.
+
+        This pins the security *contract* (not just the helper): the bug being
+        fixed is bucket fragmentation — the same secret sent with extra
+        surrounding whitespace used to hash to a different client id, letting a
+        client dodge its own rate limit simply by padding the header. All of
+        these carry the identical token ``test-secret`` and must share a bucket.
+        """
+        from starlette.requests import Request
+
+        from vllm_mlx.middleware.auth import _rate_limit_client_id
+
+        def client_id(auth_value: str) -> str:
+            scope = {
+                "type": "http",
+                "headers": [(b"authorization", auth_value.encode())],
+                "client": ("192.0.2.1", 12345),
+            }
+            return _rate_limit_client_id(Request(scope))
+
+        canonical = client_id("Bearer test-secret")
+        for variant in (
+            "Bearer  test-secret",
+            "Bearer test-secret ",
+            "Bearer \t test-secret \n",
+        ):
+            assert client_id(variant) == canonical, (
+                f"whitespace-padded {variant!r} must share the bucket of "
+                f"'Bearer test-secret', got {client_id(variant)!r} != {canonical!r}"
+            )
+
+    def test_rate_limit_empty_bearer_falls_through_to_subnet(self):
+        """An empty/whitespace-only Bearer token can't mint fresh buckets.
+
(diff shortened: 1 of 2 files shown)


raullenchai:
Thanks for taking this on. The PR is necessary: it addresses the explicitly assigned Bearer-token whitespace hardening item from #337, and the core `token.strip(); return token or None` change correctly makes FastAPI-auth-equivalent Bearer values map to the same rate-limit identity.

I do not think the current revision is merge-ready yet:

1. **Blocking — runtime annotation regression.** Moving `BaseEngine` behind `TYPE_CHECKING` in `vllm_mlx/config/server_config.py` breaks runtime annotation resolution. On this branch, `typing.get_type_hints(ServerConfig)` raises `NameError: name 'BaseEngine' is not defined`; the same call succeeds on current `main`. This change is unrelated to the auth fix and should be removed (or redesigned in a separate PR with a runtime-resolvable annotation).

2. **Blocking — overly broad test skips.** The three new `except ImportError: pytest.skip(...)` blocks can turn a broken or partially installed MLX package, including a missing transitive dependency, into a green skip. That conceals packaging regressions in existing exception-handler tests. These changes are also outside the Bearer-normalization scope and should be removed. If cross-platform availability needs separate work, it should use a narrow availability check in a dedicated PR.

3. **Test gap — pin the security contract, not only the helper.** The new helper assertions are useful, but the regression being fixed is rate-limit bucket fragmentation. Please add/replace with a contract test using `_rate_limit_client_id(Request(...))` and assert that these auth-equivalent values produce the same client ID:
   - `Bearer test-secret`
   - `Bearer  test-secret`
   - `Bearer test-secret `
   - `Bearer \t test-secret \n`

   I verified locally that the proposed production change satisfies this bucket-level contract.

4. The branch is currently 14 commits behind `main` and has no reported GitHub checks. Please rebase after narrowing the diff and rerun the normal validation.

5. `pr_validate` also rejects the unchecked `Updated README/docs if applicable` checklist item. If it is not applicable, remove it or mark it explicitly complete/N/A.

Suggested minimal scope: `vllm_mlx/middleware/auth.py` plus the focused rate-limit bucket regression test. With the unrelated config/test-skip changes removed, the core fix looks sound.

raullenchai:
Maintainer update — took this to merge-ready via `maintainerCanModify` (thanks @2005rishabh, your commit authorship is preserved).

Rebased onto current `main` (the branch was ~pre-audio/video-refactor and diffed noisily) and narrowed to exactly the sound part:

- **Kept:** `_extract_bearer_token` → `token = token.strip(); return token or None`.
- **Added:** a bucket-level contract test (`_rate_limit_client_id` on `Bearer test-secret` vs whitespace-padded variants → same client id), alongside your helper edge-case test. This pins the actual security regression (bucket fragmentation), not just the helper.
- **Dropped (out of scope, both flagged in review):** the `TYPE_CHECKING` gate on `server_config.py` (it broke runtime `typing.get_type_hints(ServerConfig)` → `NameError: BaseEngine`), and the three `except ImportError: pytest.skip` guards (a broken/partial MLX install would silently green-skip and hide packaging regressions).

Diff is now `auth.py` (+2/-1) + `test_config_and_middleware.py` (+41). Local: `TestRateLimitClientId` 5 passed, ruff clean. Running codex + pr_validate; will squash-merge on green.

2005rishabh:
Thank you for stepping in to refine the scope and correct the  **TYPE_CHECKING** gate issue. I appreciate the clear feedback and the merge.

PR #1317, opened 2026-07-30
fix(audio): harden the forced-alignment lane on /v1/audio/transcriptions

raullenchai:
## Why
`main` grew forced alignment in #1301. The **engine** side is fine; the **route** layer is missing every protection the same feature acquired under six rounds of codex review on the now-retired `mvp/content-farm-local` branch (#1300, never merged to `main`).

Seven distinct defects, none cosmetic. Verified absent on `main` before writing this:

```
asyncio.to_thread            missing      _is_client_alignment_error   missing
run_to_completion            missing      DEFAULT_ALIGNER_ALIAS        missing
_aligner_engine              missing      isinstance-guarded merge     missing
_alignment_lock              missing
```

## What

**Blocking work ran on the event loop.** Weight load + `align()` are seconds of compute; inline in the `async def` handler they stall every concurrent chat completion, health probe and SSE heartbeat. Now `asyncio.to_thread`, serialised by an `asyncio.Lock` taken **on the loop** before offloading — a `threading.Lock` inside the worker pins one shared-executor thread per queued request and starves every other `to_thread` user in the process.

**Cancellation released the lock and deleted the temp file under a live worker.** `await asyncio.to_thread(...)` returns immediately when the handler is cancelled — a client disconnect does that — while the thread keeps running. `routes/_async_utils.py::run_to_completion` shields and drains first.

**The aligner shared `_stt_engine` with the ASR lane.** With alignment on a worker thread and ASR on the event loop, an ASR request can swap the engine between the alignment path's cache check and its `align()` call — and no lock held on only one side can prevent that. A dedicated `_aligner_engine` removes the shared mutable state rather than trying to synchronise two lanes with different concurrency models. It also stops the lanes evicting each other's weights, since an aligner and an ASR model are never the same `model_name`. The engine is published only after `load()` returns, so a failed load doesn't leave a same-`model_name` object cached.

**Error classification was wrong in both directions.** A corrupted upload surfaced as `invalid_alignment_request` blaming `model`/`text` — some codec paths raise plain `ValueError` — sending the caller to fix a field that was fine. And any *internal* `ValueError` (a reshape deep in the model, a tokenizer fault) was also dressed up as a client error. Now one classified handler: decode errors first, then only the two documented `align()` rejections become a 400, everything else a generic 500. Deliberately **not** a chain of `except` clauses — a bare `raise` in the narrow clause exits the whole `try` instead of falling through to the broader one, which silently produced an envelope-less 500.

**"Just send audio + text" didn't work.** With `text` present and no `model`, the ASR default `whisper-large-v3` is not an aligner and failed deep in the engine; `response_format` defaulted to `json`, which drops the timestamps the caller came for. Now `qwen3-aligner` and `verbose_json`. Whitespace-only values count as unset — FastAPI already coerces a truly empty form field to `None`, but `"   "` arrives verbatim and 404s as a nonexistent alias. Scoped to this branch only; the ASR path's blank-model handling is long-standing contract.

**A present-but-blank `text` silently ran ASR** — answering a different question than the caller asked, with a 200 that gives no hint the request was reinterpreted. Now a 400. (`text=""` is indistinguishable from absent, since FastAPI coerces both to `None`, so it stays ASR; only a non-empty-but-blank value is rejectable, and it's the shape that signals real intent.)

**Calling the handler directly raised `AttributeError`.** Existing tests — `tests/test_audio_upload_size_limit.py` — invoke `create_transcription` as a plain coroutine, so parameters they don't pass arrive as unresolved `Form(None)` objects: truthy, non-None, and *not strings*. The pre-existing params only ever compare against `None` so they tolerate it; the alignment branch calls `.strip()`. The merge now picks the first `isinstance(v, str)` and every `.strip()` is guarded.

## Tests

+19 in `tests/test_forced_alignment_route_hardening.py`, pinning each of the above — including that the engine observes a *different thread* than the event loop, and that a cancelled request drains its worker before the lock and temp file are released. Hermetic: engines faked at the import boundary, no weights or network.

Added to the **apple-silicon** CI job. Both CI test lists are explicit allowlists, so a file in neither never runs — and these are hermetic but not mlx-free (`vllm_mlx.config` reaches `engine_core`'s `import mlx.core`).

## Notes
- Scoped to alignment. The `/v1/audio/music` route from the same source PR touches the same file and is a separate change (#1316) — merges are sequenced so conflict resolution stays deliberate.
- Pre-existing on `main`, untouched here: 3 `test_audio_upload_size_limit` failures (no local `mlx_audio`) and `test_no_routing_shaped_rapid_mlx_env_vars` — `RAPID_MLX_ESPEAK_LIB` / `_ESPEAK_DATA` from #1312 aren't in that gate's allowlist, so **that gate is currently red on `main`** and worth a separate fix. Both reproduce on a clean checkout.

🤖 Hardening, tests and triage by Claude Code (Opus 5), reviewed by the author.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,6 +187,7 @@ jobs:
             tests/test_no_mllm_flag.py \
             tests/test_audio_music_route.py \
             tests/test_audio_route_registration_gate.py \
+            tests/test_forced_alignment_route_hardening.py \
             -v --tb=short \
             -m "not slow" \
             -k "not Integration"
--- a/tests/test_forced_alignment.py
+++ b/tests/test_forced_alignment.py
@@ -316,8 +316,13 @@ def _stub_route_engine(monkeypatch):
         monkeypatch.setattr(audio_stt_mod, "STTEngine", _FakeRouteEngine)
 
     audio_route._stt_engine = None
+    # The alignment lane keeps its OWN engine cache (see the module
+    # comment on ``_aligner_engine``), so both must be cleared — leaving
+    # the stub in ``_aligner_engine`` would leak into later tests.
+    audio_route._aligner_engine = None
     yield
     audio_route._stt_engine = None
+    audio_route._aligner_engine = None
     probe._reset_probe_cache()
 
 
--- a/tests/test_forced_alignment_route_hardening.py
+++ b/tests/test_forced_alignment_route_hardening.py
@@ -1,0 +1,1136 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Forced-alignment HTTP lane hardening — ``/v1/audio/transcriptions`` + ``text``.
+
+``tests/test_forced_alignment.py`` covers the engine surface and the
+happy-path route contract. This suite pins the properties of the
+dedicated alignment lane (``_run_alignment_request``) that are easy to
+regress silently:
+
+* the blocking weight load + ``align()`` run OFF the event loop, and a
+  cancelled request drains its worker before the lock and the temp file
+  are released;
+* the aligner is cached in its OWN global, never the shared
+  ``_stt_engine`` the ASR lane mutates from the event loop, and a failed
+  ``load()`` is not published into that cache;
+* error classification order — a corrupted upload is reported as a
+  corrupted upload, an internal ``ValueError`` is a 500, and only the two
+  documented ``align()`` rejections become a 400;
+* the "just send audio + text" defaults (aligner model, ``verbose_json``)
+  including whitespace-only ``model`` / ``response_format``;
+* a present-but-blank ``text`` is a 400 rather than a silent ASR answer;
+* calling the handler directly (no ASGI stack) tolerates the unresolved
+  ``Form(None)`` / ``Query(None)`` sentinels.
+
+Hermetic — engines are faked at the import boundary, no weights or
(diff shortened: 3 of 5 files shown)


raullenchai:
## Landing validation

**Verdict: MERGE-SAFE**

- Rebased onto current `main` including #1322
- Codex review completed separately on the production-code HEAD: no actionable findings after fixing all review blockers
- Ruff: PASS
- Targeted validation: 483 passed, 5 skipped
- Full unit: 14,813 passed, 69 skipped, 18 deselected, 6 xfailed, 1 xpassed
- Supply-chain review: PASS; the only workflow change adds the new hermetic test file to the existing Apple Silicon pytest allowlist
- Stress matrix: skipped because it downloads unrelated large text models and does not exercise the audio route

Review-driven fixes added during landing:

- process-wide cross-event-loop STT mutual exclusion without blocking the shared executor
- publish ASR/aligner caches only after successful load and evict replaced weights first
- cancellation drains the real worker even if its asyncio wrapper is directly cancelled
- robust FastAPI Form/Query sentinel merging
- behavior-level heartbeat test proving lock contention does not block the event loop

Artifacts: `/tmp/pr_validate/pr-1317/run-20260730T150137Z-28734-1785423697542538000`

PR #1298, opened 2026-07-29
feat(video): add MLX-native LTX-2.3 generation

raullenchai:
## What changed

- adds a dedicated `video-gen` runtime lane and the `ltx-2.3-mlx-q4` alias
- integrates the MLX-native `mlx-video-with-audio==0.1.36` pipeline as the optional `[video]` extra
- implements OpenAI-compatible asynchronous video jobs:
  - `POST /v1/videos` for T2V and I2V
  - `GET /v1/videos/{id}` and `GET /v1/videos/{id}/content`
  - list and delete endpoints
- generates synchronized audio and MP4 output locally on Apple Silicon
- documents installation, memory/disk requirements, polling, and I2V usage

## Why

LTX-2.3 is the current open-weight LTX audio-video model, but Rapid-MLX had no
video-generation lane or Videos API. This adds a native MLX path without
affecting text, vision, audio, or text-diffusion servers.

The default checkpoint is `notapalindrome/ltx23-mlx-av-q4` (22.8 GB). It is
the split-format LTX-2.3 Q4 model explicitly supported by the pinned runtime;
the alias warns below 24 GB unified memory.

## User impact

```bash
pip install 'rapid-mlx[video]'
brew install ffmpeg
rapid-mlx serve ltx-2.3-mlx-q4
```

Users can then create T2V jobs with the OpenAI Videos API shape, or attach an
`input_reference` image for I2V. Jobs are serialized to avoid concurrent 22B
graphs exhausting unified memory. Output is retained only for the lifetime of
the server process.

## Validation

- `ruff check` and `ruff format --check`
- 2,582 targeted/contract tests passed
- local wheel build passed; video route, runtime lane, aliases, and size
  manifest are present in the wheel
- independent Codex review completed; its pinned-modality finding was fixed
- dependency resolver and MIT package license inspected

## Notes

- `ffmpeg` is required for MP4 audio muxing
- first generation downloads 22.8 GB of weights
- no real-weight generation is run in CI because the checkpoint is too large;
  the public upstream call contract and complete job lifecycle are covered with
  a deterministic fake engine

PR #1325, opened 2026-07-30
fix(video): preserve Python 3.10 core compatibility

raullenchai:
## Summary

- keep Rapid-MLX core installable on its documented Python 3.10 floor
- constrain the optional video dependency set to Python 3.11+, matching `mlx-video-with-audio`
- fail video startup early on Python 3.10 with an explicit upgrade diagnostic
- document the video-only Python floor and lock it in with regression tests

## Root cause

PR #1298 added `mlx-video-with-audio==0.1.36` under `[project.optional-dependencies].video` with only a Darwin marker. The upstream package requires Python >=3.11, while Rapid-MLX declares `requires-python = ">=3.10"`. Universal dependency resolution therefore became unsatisfiable for the Python 3.10/Darwin split, including `uv lock --check` and `rapid-mlx[video]` resolution.

The core project does not need to drop Python 3.10: only the optional upstream video runtime has the higher floor.

## Validation

- `uv pip compile pyproject.toml --extra video --python-version 3.10` — passes and excludes the incompatible runtime
- `uv pip compile pyproject.toml --extra video --python-version 3.11` — passes and resolves `mlx-video-with-audio==0.1.36`
- `pytest -q tests/test_ltx23_video.py tests/test_wan_video.py tests/test_cogvideox_video.py tests/test_vision_extra_install.py` — 73 passed
- Ruff and `git diff --check` — clean

### DIFF
--- a/docs/guides/video-generation.md
+++ b/docs/guides/video-generation.md
@@ -3,6 +3,10 @@
 Rapid-MLX exposes LTX-2.3, CogVideoX-Fun and Wan through the asynchronous
 OpenAI-compatible Videos API.
 
+Video generation requires Python 3.11 or newer because the upstream
+`mlx-video-with-audio` runtime does not support Python 3.10. Rapid-MLX's core
+text and audio features continue to support Python 3.10.
+
 ## Wan 2.1 / 2.2
 
 Wan uses the `mlx-video-with-audio` runtime included in the video extra. Four
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -382,11 +382,16 @@ audio = [
 ]
 video = [
     # MLX-native LTX-2.3 T2V/I2V with synchronized audio. 0.1.36 is the
-    # first verified release with the split-format LTX-2.3 VAE fixes.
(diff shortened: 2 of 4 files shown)


raullenchai:
## Landing validation

**Review:** independent Codex `gpt-5.6-sol` review found no actionable issues.

**Dependency contract:**
- Python 3.10 + `[video]` resolves successfully and excludes the Python 3.11-only runtime.
- Python 3.11 + `[video]` resolves `mlx-video-with-audio==0.1.36`.

**Targeted validation:** 73 video/packaging tests passed; Ruff and diff check clean; patch coverage 100%.

**GitHub:** all required checks are green, including Python 3.10/3.11/3.12, Apple Silicon, lint, type-check, version guard, and PR validation.

**Local full-suite negative control:** the broad local run reported four F5 failures because the host venv lacks optional `soundfile`; all four reproduce unchanged on `origin/main`. One batching throughput timing failure passed both on the base negative control and on immediate PR rerun. None touches this PR's packaging markers or video preflight.

Result: merge-safe.

PR #1343, opened 2026-07-30
feat: expose video capability limits

### LINKED ISSUE #1336: Add a capabilities/limits discovery endpoint (per-model size/duration/format constraints)
## Problem

Per-model workload limits are only discoverable by hitting a `400`. For example, image-to-video with `ltx-2.3-mlx-q4` rejects a request above its safe budget with:

```
requested video exceeds the safe LTX-2.3 Q4 workload limit (768x512 at 4 seconds); reduce size or duration
```

(`vllm_mlx/routes/video.py:537`). A client has no way to learn the maximum `size` × `seconds` (or the pixel×frame budget) for a given model *before* submitting, so it must either hard-code assumptions or trial-and-error against 400s.

## Request

Expose a capabilities/limits descriptor, either as a dedicated endpoint (e.g. `GET /v1/videos/capabilities`) or as per-model metadata on `GET /v1/models`. Per model, return the useful constraints, e.g.:

- modality (`video-gen`, `tts`, `stt`, `music`, ...)
- for video: supported sizes, max `seconds` (or the max `pixels × frames` budget), i2v/t2v support, native fps
- for audio: native sample rate / channels (see the related audio-format request)

## Why

Lets clients validate and clamp requests up front instead of discovering limits via errors, and enables tooling to auto-select safe parameters for the selected model/quantization.

raullenchai:
Closes #1336

## What changed
- add authenticated GET /v1/videos/capabilities discovery
- expose live model family, t2v/i2v modes, size, duration, fps, frame-shape, reference-image and motion-control limits
- expose the pixel-frames workload budget and dimension-rounding rule so clients can validate combined size and duration safely
- report Wan checkpoint-native fps, model type, and max-area constraints

## Validation
- 70 focused LTX, CogVideoX-Fun, Wan, and capability tests
- 67 server, model capability, and model-route tests
- ruff check and format check
- compileall and git diff --check

### DIFF
--- a/tests/test_video_capabilities.py
+++ b/tests/test_video_capabilities.py
@@ -1,0 +1,82 @@
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import pytest
+
+from vllm_mlx.routes import video
+
+
+@pytest.mark.asyncio
+async def test_ltx_capabilities_expose_combined_workload_budget(monkeypatch) -> None:
+    engine = SimpleNamespace(
+        model_name="notapalindrome/ltx23-mlx-av-q4",
+        video_family="ltx-2.3",
+        native_fps=24,
+    )
+    monkeypatch.setattr(video, "_video_engine", lambda: engine)
+
+    body = await video.video_capabilities()
+
+    assert body["object"] == "video.capabilities"
+    assert body["model"] == engine.model_name
+    assert body["modality"] == "video-gen"
+    assert body["modes"] == ["text-to-video", "image-to-video"]
+    assert body["limits"]["seconds"]["maximum"] == 20
+    assert body["limits"]["fps"] == {
+        "minimum": 1,
+        "maximum": 60,
+        "default": 24,
+        "fixed": False,
+    }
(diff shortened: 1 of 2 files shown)


raullenchai:
Review/validation summary:

- First Codex review found one blocking contract mismatch: Wan advertised 1280x720 exceptions even when the checkpoint max_area rejected their 64-aligned generation dimensions. Fixed in the follow-up commit with regression coverage.
- Local pr_validate verdict: MERGE-SAFE.
- Targeted: 98 passed.
- Full unit: 14,833 passed, 69 skipped, 18 deselected, 6 xfailed, 1 xpassed.
- Stress/E2E/bench: 4 models x 3 clients, PASSED.
- GitHub CI previously passed lint, type-check, Python 3.10/3.11/3.12, Apple Silicon, tests, and validate.
- Rebased onto latest main; post-rebase focused video suite: 71 passed; ruff and diff checks clean.

The second cloud Codex invocation was unavailable because the current credential lacks api.responses.write; this is recorded rather than treated as a code pass. The actionable finding from the successful first review was reproduced and fixed.

PR #1347, opened 2026-07-31
feat: lock VoiceDesign voices with deterministic seeds

### LINKED ISSUE #1338: Reusable handle to lock a VoiceDesign voice across calls (designed-voice consistency)
## Problem

`qwen3-tts-voicedesign` derives a voice from a natural-language description passed via `instructions`. Passing the **same** description across multiple `/v1/audio/speech` calls does not reliably reproduce the **same** timbre — the voice drifts between utterances. There is no way to lock "this designed voice" so that a sequence of utterances sounds like one consistent speaker.

Today, only the named CustomVoice speakers are stable across calls, but those cannot be custom-designed; VoiceDesign is designable but not stable.

## Request

Provide a reusable handle that locks a designed voice, e.g.:

- A design call returns a `voice_id`; subsequent `/v1/audio/speech` calls pass that `voice_id` and reproduce the identical voice, or
- Accept a `voice_seed` alongside the description so the same (description, seed) deterministically yields the same voice.

## Why

Any multi-utterance use — long-form narration, a conversational assistant, a recurring character across many clips — needs a stable voice. A designable-**and**-reusable voice closes the gap between CustomVoice (stable, not designable) and VoiceDesign (designable, not stable).

raullenchai:
Closes #1338

## What changed
- add an optional unsigned 32-bit voice_seed to /v1/audio/speech
- seed Qwen3-TTS VoiceDesign sampling for repeatable designed speakers across calls and restarts
- restore the prior MLX RNG state after generation, including failures
- reject voice_seed for non-VoiceDesign models instead of silently ignoring it
- echo the accepted seed in X-Voice-Seed and document reuse semantics

## Why
VoiceDesign descriptions alone can drift between utterances. A caller can now reuse the same instructions and voice_seed for consistent long-form narration or recurring characters without server-side handle storage.

## Validation
- 501 passed, 6 skipped across the complete audio suite
- deterministic RNG, state restoration, route forwarding, response header, family guard, and schema-bound tests
- ruff check and format check
- compileall and git diff --check

### DIFF
--- a/docs/guides/audio.md
+++ b/docs/guides/audio.md
@@ -449,6 +449,9 @@ Generate speech from text (OpenAI TTS API compatible).
   the TTS model's native rate (commonly 24 kHz).
 - `channels`: Optional output channel count, `1` or `2`. Omit to keep the
   model's native channel count (commonly mono).
+- `voice_seed`: Optional unsigned 32-bit seed for
+  `qwen3-tts-voicedesign`. Reuse the same `instructions` and seed to lock a
+  designed voice across calls. The response echoes it in `X-Voice-Seed`.
 
 MP3 accepts standard MPEG rates from 8–48 kHz; Opus accepts 8, 12, 16, 24,
 or 48 kHz. WAV, PCM, FLAC, and Ogg/Vorbis accept any integer in the documented
--- a/tests/test_qwen3_tts_voicedesign.py
+++ b/tests/test_qwen3_tts_voicedesign.py
@@ -324,29 +324,40 @@ class TestVoiceDesignEngine:
         import mlx.core as mx
         import numpy as np
 
+        globals()["categorical_sampling"] = lambda logits, temperature: (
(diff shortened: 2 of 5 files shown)


raullenchai:
Landing validation update:

- Rebasing onto latest `main` completed without conflicts.
- Codex review identified two blocking issues; both are fixed: request-local sampling avoids process-global MLX RNG races, and the new argument preserves positional-call compatibility.
- Two later Codex review retries were skipped by a transient backend error; a direct retry was also unavailable because the token lacks Responses API scopes.
- Full local `pr_validate`: **MERGE-SAFE** — 14,857 passed, 69 skipped, 18 deselected, 6 xfailed, 1 xpassed; stress/e2e matrix 4×3 passed; lint passed.
- Post-rebase focused regression: **53 passed, 3 skipped**; ruff passed.

Waiting for GitHub checks on the rebased HEAD before merge.

PR #1349, opened 2026-07-31
fix: omit silent audio from LTX videos

### LINKED ISSUE #1339: LTX video output carries a silent audio track for audio-less generations
## Problem

MP4s produced by the LTX video lane (`/v1/videos`, t2v and i2v) contain a **silent audio track** (48000 Hz stereo) even when there is no audio to render. Downstream muxing has to detect and explicitly ignore this empty track; naively remuxing or concatenating can otherwise pick up or interleave the silent stream.

## Request

For video generations that have no audio, either:

- omit the audio track from the output MP4 entirely (video-only), or
- document that the track exists and is silent, so clients know to drop it.

## Why

A small correctness/DX papercut: video-only output is the least-surprising default and avoids an easy-to-miss remux footgun. (Readiness-signal DX — blocking until a served model is loaded — is already tracked by #319's `/health/ready`, so it is intentionally not duplicated here.)

raullenchai:
Closes #1339

## What changed

- remux LTX outputs as video-only MP4s before publishing them
- copy the encoded video stream without re-encoding and explicitly omit audio
- replace the job output atomically only after ffmpeg produces a non-empty result
- preserve the original artifact and return an actionable error if remuxing fails

## Root cause

The LTX adapter returned the MP4 emitted by `generate_video_with_audio` unchanged. That upstream file contains a 48 kHz stereo stream even when the endpoint has no audio to render, and the crop path also retained it with `-c:a copy`.

## Validation

- reproduced a video + 48 kHz stereo MP4 with ffmpeg and verified the remux contains only one video stream with ffprobe
- `uv run pytest -q tests/test_ltx23_video.py` — 27 passed
- `uv run ruff check vllm_mlx/runtime/video_lane.py tests/test_ltx23_video.py`
- `git diff --check`

### DIFF
--- a/tests/test_ltx23_video.py
+++ b/tests/test_ltx23_video.py
@@ -392,6 +392,32 @@ def test_video_only_remux_failure_is_actionable(
     assert not list(tmp_path.glob(".result.*.video-only.mp4"))
 
 
+def test_video_only_cleanup_failure_does_not_mask_remux_error(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    output = tmp_path / "result.mp4"
+    output.write_bytes(b"mp4-with-silent-audio")
+    original_unlink = Path.unlink
+    temporary_paths = []
+
+    def fail_remux(*args, **kwargs):
+        raise subprocess.CalledProcessError(1, "ffmpeg")
+
+    def fail_cleanup(path: Path, *args, **kwargs) -> None:
+        temporary_paths.append(path)
+        raise PermissionError("cleanup denied")
+
+    monkeypatch.setattr("subprocess.run", fail_remux)
+    monkeypatch.setattr(Path, "unlink", fail_cleanup)
+
+    with pytest.raises(VideoRuntimeError, match="silent audio track") as exc:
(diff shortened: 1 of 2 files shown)


raullenchai:
Landing validation summary:

- Codex adversarial review found and we fixed three blocking edge cases: owned unique temporary paths, cleanup errors not masking remux failures, and preservation of published file permissions. Final Codex review: **0 blocking findings**.
- Focused LTX regression: **28 passed**; lint and format clean.
- Full unit suite: **14,850 passed, 78 skipped, 18 deselected, 6 xfailed, 1 xpassed**.
- Diff coverage: **95.2% (20/21 changed lines)**.
- Stress core: all four selected models passed **8/8**. Three model benches passed. The qwen3.5 bench produced a validator orchestration false positive: its server log shows all 8 stress cases and subsequent requests succeeded, then the validator sent SIGTERM before bench and the bench received connection refused. Agent failures are explicitly identical on PR and base (`PRE-EXISTING`). This PR only touches the LTX post-generation MP4 remux path.
- GitHub CI: all required checks passed on HEAD `4af8628d`.
- Branch is based on current `origin/main` (`fd807c0c`) and is mergeable/clean.

PR #1391, opened 2026-08-02
fix(stream): bound tool-call streaming suppression so it can't wedge (#1359)

### LINKED ISSUE #1359: Streaming request wedges emitting only SSE keepalives while generation continues, no client timeout can ever fire
### Rapid-MLX version

0.11.0

_No response_

### Hardware

MacBook Air M5 32Gb

### macOS version

macOS 26.6

### Python version

_No response_

### Model

froggeric/Qwen3.6-27B-MLX-4bit

### Full serve command

rapid-mlx serve froggeric/Qwen3.6-27B-MLX-4bit --gpu-memory-utilization 0.85 --no-thinking --enable-auto-tool-choice --tool-call-parser qwen3_coder --pflash off --kv-cache-dtype int8 --hybrid-cache-entries 8 --speculative-config '{"method":"mtp","model":"mlx-community/Qwen3.6-27B-MTP-4bit","num_speculative_tokens":3}' --port ${PORT}

Streaming (`stream: true`)

**What happened?**

Occasionally a streaming `/v1/chat/completions` request emits the opening role
delta and then nothing but SSE comments until externally killed. Wire capture of one 518-second instance:

```
    data: {"id":"chatcmpl-57a0fdaa",…,"choices":[{"index":0,"delta":{"role":"assistant"}}]}

    : keepalive

    : keepalive
    …26 of them, one every ~20s, and nothing else, for 518 seconds
```

Generation is running normally the whole time (see server log below — one
instance generated 2,207 tokens of which one chunk reached the wire)

This failure mode is worse than a crash: the keepalives keep the socket alive,
so no timeout at any layer can fire — the 518s request carried
`X-Stainless-Timeout: 300` and outlived it. From the client it is
indistinguishable from a server that never answered, and client-side token
counters read zero, so agent harnesses misclassify it as infrastructure noise

Hypothesis, held loosely: output buffered until a closing tag that never
closes — both `<think>` and `<tool_call>` are handled that way — except
`--no-thinking` and `enable_thinking=false` were in force, so the obvious
version has the least support. A repetition loop inside a buffered block would
also fit (no presence/repetition penalty was set)

**What did you expect?**

Content deltas flushed to the wire as generation proceeds. A healthy turn on
this stack is 5–39 chunks in 3–16 seconds. Failing that, an idle socket, so
that client timeouts can fire

### Minimal reproduction

There is no deterministic reproduction. The 518s
request was replayed from its own wire capture, byte for byte, three times:
60s, 73s, 68s — all normal. A service restart before replaying changed
nothing, so it is not a property of the request body

To detect an instance rather than provoke one: each hang leaves an adjacent
"recovered from upstream disconnection during streaming" + "no valid JSON data
found in stream" pair in the llama-swap proxy log, and an assistant message
left open with no matching close in the client transcript

```shell
The server generates normally the whole time. From `[stream_outputs]` for a
second instance, which burned a 300s client timeout generating 2,207 tokens of
which one chunk reached the wire:


    [stream_outputs] f93960c4-f51 first token after 27.6s
    [stream_outputs] f93960c4-f51 CancelledError after 2207 tokens, 300.7s
    [stream_outputs] f93960c4-f51 ABORTING orphaned request (2207 tokens generated in 300.7s)
    [disconnect_guard] CLEANUP done, 1 chunks total, elapsed=300.8s


No error is logged before the client disconnect — from the server's point of
view the request is healthy until it is killed
```

raullenchai:
## Why

On streaming `/v1/chat/completions`, once the tool parser sees a tool-call opener (`<tool_call>`, `<function=`) it returns `None` for every subsequent delta until the block closes. `StreamingPostProcessor` suppressed those deltas with **no byte/token bound**, so a block that never closes held all content indefinitely. The SSE generator then yields nothing and the keepalive layer emits only `: keepalive` comments — so **no client timeout at any layer can fire** (issue #1359: 2207 tokens generated, 1 chunk on the wire, 518s; the request outlived its `X-Stainless-Timeout: 300`). From the client it's indistinguishable from a server that never answered.

The never-closing block has several triggers, all of which this bounds: a literal `<tool_call>`/`<function=` in generated code that skews the open/close balance (very plausible for a coder model, `--tool-call-parser qwen3_coder` → HermesToolParser), oversized/degenerate arguments, or a repetition loop inside the block. `StreamingToolCallFilter` (responses/anthropic path) already caps at 1 MB — but the **chat path has no cap at all**.

Confirmed chain: `_disconnect_guard` keepalives ⟸ `stream_chat_completion` never yields ⟸ `process_chunk` returns `[]` ⟸ `_detect_tool_calls` returns `None` while the parser reports an incomplete block — with no bound anywhere.

## What

Add the missing bound at the common chokepoint (`_detect_tool_calls`), so it protects every tool parser:

- Buffer only the **withheld** (`None`) deltas since the block opened, and only while **no tool call has reached the wire** (`_tool_calls_emitted_to_wire == 0`) — genuine, incrementally-streamed tool calls are never touched.
- Once that buffer exceeds `_MAX_TOOL_SUPPRESSION_BYTES` (**64 KB** — orders of magnitude above any real tool call, yet ~16 K tokens instead of the ~250 K a 1 MB cap would allow), **release the withheld text as content** and latch tool-suppression off for the rest of the turn so the stream makes visible progress (the reporter's preferred outcome: "content deltas flushed to the wire").
- Guard the `finalize()` tool-recovery so a released block isn't re-surfaced as a tool call.

Note: the repetition-loop trigger is *also* mitigated on current `main` by the repetition guard (#1377/#1378) for tool-bearing requests; this fix covers the non-repetition variants (false-positive opener in code, oversized/degenerate args) that the guard doesn't.

## Tests

`tests/test_streaming_suppression_budget.py` (new) drives `StreamingPostProcessor.process_chunk` with a stub never-closing parser (no model, no GPU): asserts suppression under budget, release + resumed content-flow after budget, latch reset across `reset()`, and that a legit short tool call is never released. Regression: `test_postprocessor` (116), `test_447_stream_tool_choice_auto` (14), `test_streaming_tool_filter` (30), `test_tool_call_streaming_parity` (10), `test_tool_parsers` (156) all pass. `ruff check` + `ruff format --check` clean.

The reporter notes there is **no deterministic reproduction** (the 518s request replayed byte-for-byte three times ran normally), so this is validated by root-cause + unit tests rather than a live repro.

Closes #1359

### DIFF
--- a/tests/test_streaming_suppression_budget.py
+++ b/tests/test_streaming_suppression_budget.py
@@ -171,3 +171,46 @@ def test_legit_short_tool_call_not_released():
     pp.process_chunk(_make_output("</tool_call>"))
     assert pp._tool_suppression_released is False
     assert pp.tool_calls_detected is True
+
+
+def test_no_release_once_a_tool_call_reached_the_wire():
+    """The ``_tool_calls_emitted_to_wire == 0`` guard: once a tool call has
+    streamed to the wire, a large following buffered block is a legitimate
+    continuation (e.g. a second call whose args stream), not a wedge — the
+    budget must NOT release, and nothing is accumulated into the suppression
+    buffer at all."""
+
+    class _EmitThenBufferParser(_NeverClosingToolParser):
+        def __init__(self):
+            super().__init__()
+            self._emitted = False
+
+        def extract_tool_calls_streaming(
+            self, previous_text, current_text, delta_text, request=None
+        ):
+            if "<tool_call>" in current_text and not self._emitted:
+                self._emitted = True
+                return {
+                    "tool_calls": [
+                        {
+                            "index": 0,
+                            "id": "call_1",
+                            "type": "function",
+                            "function": {"name": "f", "arguments": "{}"},
+                        }
+                    ]
+                }
+            if self._emitted:
+                return None  # now buffering a large second block forever
+            return {"content": delta_text}
+
+    pp = StreamingPostProcessor(_make_cfg())
+    pp.reset()
+    pp.tool_parser = _EmitThenBufferParser()
+    pp.tool_markup_possible = True
+
+    pp.process_chunk(_make_output("<tool_call>"))
+    assert pp._tool_calls_emitted_to_wire > 0
+    pp.process_chunk(_make_output("z" * (_MAX_TOOL_SUPPRESSION_BYTES + 100)))
+    assert pp._tool_suppression_released is False  # guard held
+    assert pp._tool_suppressed_buffer == ""  # never accumulated past the guard
--- a/tests/test_streaming_suppression_budget.py
+++ b/tests/test_streaming_suppression_budget.py
@@ -1,0 +1,173 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for issue #1359 — streaming tool-call suppression must be
+bounded.
+
+A streaming ``/v1/chat/completions`` request whose model opens a tool-call
+block that never closes (a literal ``<tool_call>``/``<function=`` in generated
+code that skews the open/close balance, oversized/degenerate arguments, or a
+repetition loop) used to have every subsequent content delta suppressed
+forever — the tool parser returns ``None`` for each chunk, the SSE generator
+yields nothing, and the keepalive layer emits only ``: keepalive`` comments,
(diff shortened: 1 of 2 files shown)


Ptico:
Awesome! Thanks!

Issue #241, opened 2026-05-06
watch: Gemma 4 MTP drafter — verify if/when an MLX port appears


raullenchai:
## Context

Google trained MTP (Multi-Token Prediction) drafter heads for all Gemma 4 variants but [stripped them from the public HuggingFace release](https://huggingface.co/google/gemma-4-E4B-it/discussions/5), keeping them only in LiteRT on-device format.

A community user has done the **first public extraction** for one variant (E4B / 9B): [SeatownSin/gemma-4-E4B-mtp-drafter](https://huggingface.co/SeatownSin/gemma-4-E4B-mtp-drafter)
- 78M params, 4-layer transformer, 297 MB fp32
- Targets `google/gemma-4-E4B-it` only (not E2B, 26B MoE, or 31B Dense)
- License: Gemma Terms of Use (weights) + Apache 2.0 (extraction code)
- **No MLX port exists yet**

External marketing copy claims "up to 3× speedup" and "Apple Silicon 26B MoE 2.2× at batch 4-8", but neither claim has a corresponding published drafter or MLX implementation to back it.

## Why this is a "watch" not a "do"

We have prior data on speculative-drafter-on-quantized-target on Apple Silicon — see `memory/knowledge/mtp_poc_findings.md`:
- MTP architecture in `vllm_mlx/speculative/mtp_generate.py` works (1.71× best-case on BF16 base)
- **0% acceptance on 4-bit base models** even with BF16 drafter weights — hidden-state distribution shift after quantization makes the drafter's predictions garbage
- Same root cause is likely to apply to a separate-drafter spec-decode path

The Gemma 4 E4B drafter is BF16, and the Gemma 4 weights people actually run on Apple Silicon are 4-bit MLX. Until we know whether SeatownSin's drafter accepts on a 4-bit target, the speedup story is unverified.

`--draft-model` was deprecated in our CLI (cli.py:261-264) for these exact reasons.

## What to watch for

- [ ] `mlx-community/gemma-4-E4B-mtp-drafter` (or any MLX-format port) appearing on HuggingFace
- [ ] Native `mlx-lm` support for separate-draft-model speculative decode (currently only MTP-head path is wired)
- [ ] Community extraction expanding beyond E4B to 26B-A4B / 31B Dense
- [ ] Independent MLX speedup numbers — anywhere — on a 4-bit Gemma 4 target

## Verification plan (only if a usable MLX drafter appears)

[...]

PR #1399, opened 2026-08-02
fix(mtp): match the Gemma 4 assistant's quantization to its sidecar

raullenchai:
fix(mtp): match the Gemma 4 assistant's quantization to its sidecar

## Why

`gemma4_inject` builds the AssistantModel full-precision and loads the sidecar
straight into it. mlx-community publishes the assistants at several widths
(the 12B alone has 4bit / 5bit / 6bit / 8bit / bf16 / mxfp4 / mxfp8 / nvfp4),
so every quantized one is rejected by the shape check:

    [mtp.inject.gemma4] 23 tensor(s) ... have shapes incompatible with the
    AssistantModel. First 4: [('pre_projection.weight', (1024, 960),
    (1024, 7680)), ...]. Refusing to inject.

960 == 7680 * 4 // 32 — the sidecar is 4-bit affine-packed and the module was
never quantized to match. `qwen3_5_inject` already solves exactly this for the
Qwen MTP head by reading `(bits, group_size)` off the sidecar's own tensors;
Gemma 4 simply never got the equivalent step.

Only `mlx-community/gemma-4-12B-it-assistant-bf16` and the four other
`*-assistant-bf16` repos could be paired at all, which is also why the gap went
unnoticed: the bf16 path happens to match an unquantized module by accident.

## What

`_infer_sidecar_quantization()` mirrors the Qwen helper, anchored on
`pre_projection` because that is the layer whose full-precision `(out, in)`
this consumer knows independently from Google's config. The sidecar tensors are
the ground truth for packing — more reliable than a `config.json` proxy a
checkpoint may omit or mis-declare. Inference is cross-checked (the derived
params must reproduce the observed shapes exactly, so a coincidental
divisibility match cannot pass) and constrained to MLX's affine set. Both the
inference and the `nn.quantize` call fail closed, since an escape here would
abort a request at the first draft step rather than fall back cleanly.

## Verified

Unit, on the real 12B 4-bit shapes: `(1024, 960)` + scales `(1024, 120)` against
full-precision `(1024, 7680)` derives `bits=4, group_size=64`. bf16 returns
None. Three malformed cases are refused: truncated-quantized (packed weight,
scales dropped), out-dim mismatch, and a derived `bits=7`.

Real weights on `mlx-community/gemma-4-12B-it-4bit` + `-assistant-4bit`
(M2 Pro): inject now succeeds where it previously refused.

## What this does NOT fix

Gemma 4 stays unregistered in `detect` / `dispatch`. With the 4-bit assistant
now loading, it fails end-to-end exactly as the bf16 one does: draft acceptance
0.0 on 3 of 4 prompts, greedy-lossless violated on 2 of 4, throughput 0.52x.
Acceptance 0.0 means nothing was accepted, so every token came from the
backbone and the output should have matched the baseline — so this is the
backbone path under this wiring, not drafter accuracy. One divergence is a
transposed pair (base `[236761, 236779, ...]` against mtp
`[236779, 236761, ...]`), which points at emission order rather than wrong
tokens.

Caveat on that second defect: the harness feeds raw prompts, and Gemma's
baseline output is visibly degenerate under them (one token repeating), so the
lossless comparison there wants a re-run with the chat template applied before
anyone treats it as fully characterised. The acceptance-0.0 result stands on
its own regardless — a drafter that cannot predict a repeating token is broken.

This PR is scoped to the quantization gap so that investigation starts from a
sidecar that at least loads at its published width.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

### DIFF
--- a/tests/test_mtp_gemma4_sidecar_quant.py
+++ b/tests/test_mtp_gemma4_sidecar_quant.py
@@ -1,0 +1,107 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Sidecar quantization inference for the Gemma 4 assistant drafter.
+
+Deliberately a SEPARATE file from ``test_mtp_gemma4_assistant_inject.py``:
+that module opens with ``pytest.importorskip("mlx.core")``, so everything in it
+is skipped wherever mlx is absent -- including the Linux CI runner. These
+checks are pure shape arithmetic and need no mlx, so keeping them here means
+they actually execute in CI instead of silently skipping.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from vllm_mlx.spec_decode.mtp.gemma4_inject import _infer_sidecar_quantization
+
+# ---------------------------------------------------------------------------
+# 7. Sidecar quantization inference
+#
+# mlx-community publishes the assistants at several widths (the 12B alone has
+# 4bit / 5bit / 6bit / 8bit / bf16 / mxfp4 / mxfp8 / nvfp4). The AssistantModel
+# is built full-precision, so the module must be quantized to match whatever
+# the sidecar actually carries -- read off its tensors, which are ground truth
+# for the packing, rather than from a config.json a checkpoint may omit.
+#
+# Numbers below are the real shapes from
+# ``mlx-community/gemma-4-12B-it-assistant-4bit``: pre_projection is an
+# nn.Linear(2 * 3840, 1024), so full-precision is (1024, 7680) and the 4-bit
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing this experimental PR rather than carrying it forward. It fixes quantized sidecar loading in isolation, but the real Gemma 4 validation documented here still failed the greedy-lossless gate, had near-zero acceptance, and regressed throughput. Current main intentionally keeps this path gated off. #241 remains the tracker for a correctness-preserving implementation and fresh live validation.

PR #1404, opened 2026-08-03
fix(bench): load weights on the mlx-step worker so freeform bench isn't 0 tok/s

raullenchai:
## Why

`rapid-mlx bench <alias>` (the **freeform** path, no `--submit`) reported
`0.00 tok/s` for every model. Root cause: it loaded the weights on the
**main thread** via `mlx_lm.load(...)`, then generated on
`AsyncEngineCore`'s **worker thread**. Under mlx-lm 0.31.3+ ThreadLocalStream
tagging (#170), the first batch step then raises:

```
```

The scheduler catches it, aborts every request, and the summary still
prints a clean `Throughput: 0.00 tok/s` with exit code 0 — a **silent**
failure. This is why `serve`/`chat` work (they load on the step worker)
but freeform `bench` did not. The `--submit` community path (`_run`) was
already correct; only the freeform display path was broken.

This directly breaks the desktop app's **"Speed on this Mac"** card, which
runs the freeform path first and was showing a confident, wrong
`0 tokens / second` (and offering to submit that 0 to the leaderboard).

## Fix

Create the single `mlx-step` `ThreadPoolExecutor` (initialized with
`_init_mlx_step_thread`) **before** loading, load the weights **on that
worker** via `executor.submit(load, ...)`, and pass the same executor to
`AsyncEngineCore` — so weights + forward passes + cache state all live on
one owning thread. This mirrors the proven `bench --submit` path (`_run`)
and `BatchedEngine._start_llm`.

The executor is created at the `bench_command` call site and shut down in
a `try/finally` around `asyncio.run(run_benchmark())`, so it's reaped on
every exit path (success, load-error → `sys.exit`, Ctrl-C) — caller-owned
executors are intentionally not shut down by `AsyncEngineCore.stop()`.

## Verification

```
$ rapid-mlx bench llama3-1b-4bit --num-prompts 3 --max-tokens 20
  Total completion tokens: 60
  Tokens/second: 426.00
  Throughput: 596.40 tok/s        # was 0.00 before this fix
```

## Tests

- New regression `test_bench_command_loads_weights_on_mlx_step_worker`
  pins the load onto an `mlx-step` worker (fails if it regresses back to
  the main thread).
- Existing `tests/test_cli_bench.py` (mirror-prefetch order, graceful
  no-op, hybrid text-lane) all still pass — 7/7 green.
- `ruff check` + `ruff format --check` clean.

### DIFF
--- a/tests/test_cli_bench.py
+++ b/tests/test_cli_bench.py
@@ -173,6 +173,56 @@ def test_bench_command_proceeds_when_mirror_prefetch_is_silent_noop(
     assert load_called == ["mlx-community/Qwen3-1.7B-4bit"]
 
 
+def test_bench_command_loads_weights_on_mlx_step_worker(monkeypatch) -> None:
+    """Freeform ``rapid-mlx bench <alias>`` must load the weights ON the
+    mlx-step worker thread (#170), NOT the main / asyncio thread.
+
+    mlx-lm 0.31.3+ binds the module-level generation stream (and each
+    thread's auto-default stream) to whichever thread first touches MLX.
+    Loading the weights on the main thread and then generating on the
+    engine's worker raises "There is no Stream(gpu, N) in current thread"
+    on the very first batch step, so every request aborts and the run
+    silently reports ``0.00 tok/s`` (the app's "Speed on this Mac" card
+    then shows a confident, wrong zero). Pinning the load onto the
+    ``mlx-step`` worker — the executor AsyncEngineCore then reuses — is
+    the only configuration that survives ThreadLocalStream tagging, and
+    it mirrors the ``--submit`` community path + why ``serve`` works.
+    """
+    import threading
+
+    cli = importlib.import_module("vllm_mlx.cli")
+
+    load_thread: dict[str, str] = {}
+
+    def _fake_load(name: str):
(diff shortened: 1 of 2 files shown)


raullenchai:
## Codex review (post-fix follow-up)

Ran codex on the diff. One MAJOR raised: `shutdown(wait=True)` in the `finally` re-waits the uninterruptible native `mlx` weight load if Ctrl-C lands mid-load.

**Classification: inherent limitation, not a regression.** Pre-PR this freeform path loaded on the **main thread**, which is equally uninterruptible — so exit already blocked until the native read finished; wall-clock is unchanged. `concurrent.futures`' atexit hook joins the non-daemon worker at interpreter shutdown regardless of `wait`, so no code change within this PR can make a native MLX load interruptible (that would need `os._exit` and risk dirty Metal state — not worth it for a bench command).

**Applied hardening anyway** (commit `2ec2827d`): track `KeyboardInterrupt` and pass `wait=not interrupted` + `cancel_futures=True`, so our own `finally` doesn't pointlessly double-wait on the idle-worker paths. Comment documents the accepted limitation.

Net: 0 BLOCKING, 0 regressions. Tests + ruff green.

PR #1408, opened 2026-08-03
fix(gemma4): unbreak bench, unlock live KV quantization, expose the KV-projection override

raullenchai:
## What does this PR do?

Three independent fixes that together make the `gemma-4-12b-*` aliases actually usable:

1. **`bench` could not load any Gemma 4 12B alias.** `bench_command` and `_run_submit_flow` bound `mlx_lm.load` directly. mlx-lm has no `gemma4_unified` architecture (the classes live in mlx-vlm), so every `gemma-4-12b-*` alias died with `Error loading model: Model type gemma4_unified not supported.` — even though `rapid-mlx serve` runs those aliases fine, because `serve` goes through `utils.tokenizer.load_model_with_fallback`, which carries the gemma4 router. Both bench paths now bind that same loader.

   This originally also carried a fix for the stream-affinity half of the same bug (loading on the asyncio loop thread, then stepping the engine on the `mlx-step` worker → `RuntimeError: There is no Stream(gpu, 0) in current thread.`). **#1404 landed that independently while this was in progress**, and its version is better — it hoists the executor above `run_benchmark` and hardens the Ctrl-C path. This PR has been rebased onto it and now only swaps the loader; the threading work here is dropped in favour of #1404's.

2. **Live KV quantization was silently disabled on every Gemma 4 alias.** `_text_attention_args` looked for the language model's attention dims on `.args`. mlx-lm models have that attribute; the mlx-vlm-derived architectures do not — `Gemma4TextWrapper` has no `.args` at all, and its `.language_model` carries a `TextConfig` on `.config`. So `probe_kv_head_dims` returned `(None, None)` and the engine logged

   ```
   [kv-cache] live KV quantization disabled: head_dim (K=None, V=None) unknown
   or not divisible by any supported group_size (32/64/128)
   ```

   then served a bf16 live cache — despite Gemma 4's real dims (`head_dim=256`, `global_head_dim=512`) being divisible by all three. Nothing about this model justified the fallback; the probe just could not see it. That matters here specifically: Gemma 4 12B's KV runs **~128 KB/token** (8 full-attention layers × 8 KV heads × 512 `global_head_dim` × 2 × 2 B), several times a same-class Llama/Qwen, so KV is the binding constraint on a 16–32 GB Mac — and the one mitigation for it was off.

3. **`--metal-cap-kv-bytes-per-token` had no CLI surface.** `Scheduler._infer_kv_dtype_bytes` documents that quantized-KV deployments are not auto-detected and names `SchedulerConfig.metal_cap_kv_bytes_per_token` as the operator escape hatch — but that field had no flag and no env var, so the documented escape hatch was reachable only from the Python API. With `--kv-cache-turboquant k8v4` active the D-METAL-CAP projection is byte-identical to the fp16 path, so a long prompt is rejected by a check that cannot see the codec that would have made it fit.

## Why is this needed?

`gemma-4-12b-*` is a shipped, README-recommended alias family. Today:

- `rapid-mlx bench gemma-4-12b-4bit` **cannot run at all** (item 1), so the family cannot be benchmarked or submitted to the community corpus.
- Serving works, but the KV-cache mitigation that this architecture needs more than most is off by default because of a missing `getattr` (item 2).
- The documented workaround for quantized-KV admission is unreachable from the CLI (item 3).

Measured on an M3 Pro 18 GB, `mlx-community/gemma-4-12B-it-4bit`:

| | before | after |
|---|---|---|
| `rapid-mlx bench gemma-4-12b-4bit` | `Model type gemma4_unified not supported` | completes, 13.6–15.5 tok/s |
| `probe_kv_head_dims(model)` | `(None, None)` | `(256, 256)` |
| `resolve_kv_quantization(…, 64)` | `live_disabled=True` | `(64, live_disabled=False)` |
| `serve --kv-cache-dtype int8` | `live KV quantization disabled` warning | `KV cache quantization: 8-bit, group_size=64` |

Reproduced independently on a second machine (Mac mini, M2 Pro 32 GB, macOS 26.5.2): bench completes at 15.45 tok/s, probe returns `(256, 256)`.

No default behaviour changes. `.args` keeps priority in the probe wherever it exists, so every model that already probed successfully is byte-identical, and an unprobeable language config still yields `None` so the live cache fails safe to bf16 rather than picking up vision/composite dims. `--metal-cap-kv-bytes-per-token` defaults to `0`, which keeps the architecture-aware auto-derivation and the OOM cliff it prevents exactly as before.

## AI assistance disclosure

Claude (Opus 5) wrote the implementation and the tests in this PR, working from a profiling session on the two machines above; I directed the investigation, chose the scope, and reviewed each change.

How the output was verified, beyond the test suite:

- Each of the three fixes was checked against the **actual failing behaviour** it claims to fix — the loader fix by running `rapid-mlx bench` before/after, the probe fix by loading the real checkpoint and printing `probe_kv_head_dims` / `resolve_kv_quantization`, the flag by reading it back out of `SchedulerConfig` and off `serve --help`. Numbers in the table above are measured, not inferred.
- Independently reproduced on a second machine (M2 Pro 32 GB) from a clean install.
- Full suite diffed against `main` on the same host: **141 pre-existing failures before, 141 after, zero new** (the pre-existing set is dominated by missing optional extras — audio/video — in this environment).
- One thing deliberately **not** included: a fix for the D-METAL-CAP eviction dead band. It was implemented, measured, and reverted — evicting at admission and re-admitting fired correctly (active 9.17 → 8.6 GB) but then died in prefill with `[METAL] Insufficient Memory` → `Fatal Python error: Aborted`, because the admission projection models KV only and never the prefill working set. Trading a recoverable 503 for a process abort is a bad trade. Flagging it here because it is the real gap behind item 3, and it needs the prefill working set modelled before admission can safely be made more permissive.

## Test plan

- `ruff check vllm_mlx/ tests/` → clean; `ruff format --check` → 778 files already formatted.
- Full suite vs `main` on M3 Pro 18 GB: 141 failures on both, **zero new** (`comm -13` of the sorted FAILED lists is empty).
- New tests:
  - `tests/test_probe_config_attr.py` (7 tests) — pins that `.config` is consulted, that `.args` still wins wherever it exists, that the Gemma 4 dims enable the live cache, and that an unprobeable language config still fails safe to `None`.
  - `tests/test_cli_metal_cap_kv_flag.py` (3 tests) — the parsed value reaches `SchedulerConfig`, the default stays `0`, and a negative value is rejected at parse time.
- Updated tests: `test_cli_bench.py`, `test_cli_bench_hybrid_cache_flag.py`, `test_cli_response_cache_flag.py` repointed from patching `mlx_lm.load` to patching the loader the bench flow now binds.
- Mutation spot-checks on the new tests (per SOP §Step 3): deleting `metal_cap_kv_bytes_per_token=…` at the serve `SchedulerConfig(...)` construction fails `test_serve_metal_cap_kv_bytes_reaches_scheduler_config`; reverting `_args_or_config` to read only `.args` fails `test_language_model_config_attr_is_probeable` and `test_gemma4_dims_enable_the_live_quantized_cache`.
- End-to-end on real weights, both machines: `rapid-mlx bench gemma-4-12b-4bit` completes; `serve --no-mllm --kv-cache-dtype int8` reports the quantized live cache and answers correctly.

## Checklist

- [x] Tests pass locally (full suite diffed against `main`: zero new failures)
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 1408` — scorecard posted as a PR comment
- [x] New tests touch scheduler / CLI paths and were spot-checked to fail when the corresponding production line is broken
- [x] Updated README/docs if applicable — no docs change needed; the three fixes restore documented behaviour rather than adding any

### DIFF
--- a/tests/test_cli_bench.py
+++ b/tests/test_cli_bench.py
@@ -29,18 +29,25 @@ import pytest
 
 # ---------- shared helpers ---------------------------------------------------
 def _patch_mlx_lm_load(monkeypatch, fake_load) -> None:
-    """Patch ONLY the ``load`` attribute on the real ``mlx_lm`` module.
+    """Patch the loader ``bench`` actually calls.
 
-    ``bench_command`` and ``_run_submit_flow`` use ``from mlx_lm import
-    load`` at call time — that's ``getattr(sys.modules['mlx_lm'], 'load')``
-    resolved fresh each call. Replacing the whole module object would
-    break other imports (``mlx_lm.generate``, ``mlx_lm.utils``…) that
-    the engine's own imports trigger during ``bench_command`` setup.
-    Attribute patching leaves those intact.
+    ``bench_command`` and ``_run_submit_flow`` bind
+    ``from .utils.tokenizer import load_model_with_fallback as load`` at
+    call time — that's
+    ``getattr(sys.modules['vllm_mlx.utils.tokenizer'],
+    'load_model_with_fallback')`` resolved fresh each call. They used to
+    bind ``mlx_lm.load`` directly, but that loader has no
+    ``gemma4_unified`` architecture, so every ``gemma-4-12b-*`` alias
+    failed to load; the router in ``utils.tokenizer`` is what makes them
+    work. Patch that attribute rather than the module object so the other
+    imports (``mlx_lm.generate``, ``mlx_lm.utils``…) the engine triggers
+    during setup stay intact.
     """
-    import mlx_lm  # already loaded by the test-time environment
+    from vllm_mlx.utils import tokenizer as _tokenizer_mod
 
-    monkeypatch.setattr(mlx_lm, "load", fake_load, raising=True)
+    monkeypatch.setattr(
+        _tokenizer_mod, "load_model_with_fallback", fake_load, raising=True
+    )
 
 
 def _make_freeform_bench_args(model: str) -> argparse.Namespace:
--- a/tests/test_cli_bench_hybrid_cache_flag.py
+++ b/tests/test_cli_bench_hybrid_cache_flag.py
@@ -67,8 +67,12 @@ def _run_bench_capturing_scheduler_config(argv: list[str]) -> dict:
         mock.patch.object(cli, "_check_disk_space", lambda *a, **k: None),
         mock.patch.object(cli, "_check_memory_capacity", lambda *a, **k: None),
         mock.patch.object(cli, "_ensure_model_downloaded", lambda *a, **k: None),
-        # ``bench_command`` does ``from mlx_lm import load`` — patch at source.
-        mock.patch("mlx_lm.load", return_value=(object(), object())),
+        # ``bench_command`` binds ``load_model_with_fallback`` (the
+        # gemma4-aware router, not bare ``mlx_lm.load``) — patch at source.
+        mock.patch(
+            "vllm_mlx.utils.tokenizer.load_model_with_fallback",
+            return_value=(object(), object()),
+        ),
         # ``bench_command`` does ``from .scheduler import SchedulerConfig`` —
         # patch on the scheduler module so the local import binds the mock.
         mock.patch("vllm_mlx.scheduler.SchedulerConfig", _fake_scheduler_config),
--- a/tests/test_cli_metal_cap_kv_flag.py
+++ b/tests/test_cli_metal_cap_kv_flag.py
@@ -1,0 +1,127 @@
+# SPDX-License-Identifier: Apache-2.0
+"""``serve --metal-cap-kv-bytes-per-token`` plumbing.
+
+``Scheduler._infer_kv_dtype_bytes`` documents that quantized-KV
+deployments are NOT auto-detected and names
+``SchedulerConfig.metal_cap_kv_bytes_per_token`` as the operator escape
+hatch. That field existed only on the Python API, so a CLI user running
+``--kv-cache-turboquant`` / ``--kv-cache-quantization`` got a D-METAL-CAP
+admission projection computed at fp16 regardless of the codec — long
+prompts the codec would have fit were rejected with a 503.
+
+These tests pin the CLI surface and the wiring: the parsed value must
+reach ``SchedulerConfig``, and the default must stay 0 so the
+architecture-aware auto-derivation (and the OOM cliff it prevents) is
+untouched for everyone who does not set the flag.
+"""
+
+import sys
+from unittest import mock
+
+import pytest
+
+import vllm_mlx.cli as cli
+
+# Import the engine module EAGERLY, before ``SchedulerConfig`` is patched.
+# Modules that annotate ``SchedulerConfig | None`` evaluate that union at
+# import time; if the first import happens while the patch is active, the
+# union is built against the fake and raises
+# ``TypeError: unsupported operand type(s) for |``. Importing here binds
+# the real class first — the patch is then a pure name rebind and no class
+# body re-runs under it. Same guard as
+# ``tests/test_cli_bench_hybrid_cache_flag.py``.
+import vllm_mlx.engine_core as _engine_core  # noqa: E402,F401
+
+
+class _StopError(Exception):
+    """Raised once SchedulerConfig is built, to short-circuit engine boot."""
+
+
+def _capture_serve_scheduler_config(argv: list[str]) -> dict:
+    """Drive the REAL ``main()`` → ``serve_command`` and capture the kwargs
+    the serve path passes to ``SchedulerConfig(...)``.
+
+    Mirrors ``tests/test_cli_response_cache_flag.py`` — only the I/O
+    boundaries the serve path hits BEFORE construction are mocked, so
+    everything between argparse and ``SchedulerConfig(...)`` is real code.
+    """
+    captured: dict = {}
+
+    def _fake_scheduler_config(*args, **kwargs):
+        captured.update(kwargs)
+        raise _StopError
+
+    with (
+        mock.patch.object(cli, "_check_disk_space", lambda *a, **k: None),
+        mock.patch.object(cli, "_check_memory_capacity", lambda *a, **k: None),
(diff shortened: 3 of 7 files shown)


raullenchai:
## Self-validation

Rebased onto `main` after #1404 landed — that PR fixed the stream-affinity half of this bug independently and better (executor hoisted above `run_benchmark`, Ctrl-C hardened), so the threading work is dropped here and this branch now only swaps the loader. The original merge conflict is resolved.

### `pr_validate 1408`

```
| step                  | status | summary                                                       |
|-----------------------|--------|---------------------------------------------------------------|
| fetch                 | PASS   | 7 files, +341/-17 LOC, blast=high                             |
| test_plan_check       | PASS   | all 6 test-plan item(s) checked                               |
| cl_description_quality| PASS   | title OK + body has rationale                                 |
| supply_chain          | PASS   | no hooks touched, no suspicious patterns, deps clean          |
| test_env_check        | PASS   | all 6 required test packages importable                       |
| lint                  | PASS   | clean (7 files)                                               |
| targeted_tests        | PASS   | 478 passed in 36.24s (22 target files)                        |
| full_unit             | FAIL   | 4 failed, 15111 passed, 92 skipped, 6 xfailed, 1 xpassed      |
| stress_e2e_bench      | ERROR  | environment: "no model fits 0.0GB usable RAM"                 |
| diff_coverage         | PASS   | patch coverage 100.0% (17/17 changed lines)                   |
```

**The 4 `full_unit` failures are pre-existing on `main`**, not introduced here — same four, same run, both refs:

```
tests/test_bench_tier_submit_combo.py::test_tier_all_submit_payload_shape
tests/test_no_out_of_band_routing.py::test_no_routing_shaped_rapid_mlx_env_vars
tests/test_payload_builder_v2_full.py::test_full_payload_has_all_three_sections
tests/test_payload_builder_v2_kwargs.py::test_default_kwargs_match_v1_modulo_version

branch: 4 failed, 37 passed
main:   4 failed, 37 passed
```

They look like fallout from the schema-v2→v3 payload bump in #1403 rather than anything in this diff. Happy to fix them in a separate PR if that's wanted — deliberately not touching them here to keep this diff scoped.

`stress_e2e_bench` errored on the host, not the code: this machine (M3 Pro 18 GB) was under memory pressure at run time and the step computed 0.0 GB usable. The equivalent was run by hand instead, on real weights, on two machines — see the test plan.

### `codex review --base main`

> The CLI plumbing, benchmark loader routing, and config-based KV head-dimension probing are consistent with existing interfaces and preserve documented defaults. No actionable regressions were identified.

It was specifically pointed at the four risks I was most worried about: whether the loader swap changes behaviour for non-gemma4 models (`load_model_with_fallback` does more than `mlx_lm.load`), whether bench's 404 special-casing still classifies correctly after the swap, whether `_args_or_config` could pick up a vision/composite `head_dim` on a real VLM and mis-size the live cache, and whether the new tests are load-bearing.

raullenchai:
## Re-validated after rebase onto `2ff3fdc4`

### `pr_validate 1408`

```
| step                   | status | summary                                                  |
|------------------------|--------|----------------------------------------------------------|
| fetch                  | PASS   | 7 files, +341/-17 LOC, blast=high                        |
| test_plan_check        | PASS   | all 6 test-plan item(s) checked                          |
| cl_description_quality | PASS   | title OK + body has rationale                            |
| supply_chain           | PASS   | no hooks touched, no suspicious patterns, deps clean     |
| test_env_check         | PASS   | all 6 required test packages importable                  |
| lint                   | PASS   | clean (7 files)                                          |
| targeted_tests         | PASS   | 478 passed (22 target files)                             |
| full_unit              | FAIL   | 4 failed, 15125 passed, 83 skipped, 6 xfailed, 1 xpassed |
| stress_e2e_bench       | ERROR  | environment — see below                                  |
| diff_coverage          | PASS   | patch coverage 100.0% (17/17 changed lines)              |
```

Run on two machines (M3 Pro 18 GB and Mac mini M2 Pro 32 GB); identical verdict on both.

**`full_unit`: the same 4 failures exist on `main`.** Checked against `origin/main` at `2ff3fdc4` in the same venv, same invocation:

```
tests/test_bench_tier_submit_combo.py::test_tier_all_submit_payload_shape
tests/test_no_out_of_band_routing.py::test_no_routing_shaped_rapid_mlx_env_vars
tests/test_payload_builder_v2_full.py::test_full_payload_has_all_three_sections
tests/test_payload_builder_v2_kwargs.py::test_default_kwargs_match_v1_modulo_version

branch: 4 failed, 37 passed
main:   4 failed, 37 passed
```

They look like fallout from the schema-v2→v3 payload bump in #1403. Left alone here to keep this diff scoped — happy to take them in a follow-up.

**`stress_e2e_bench`: could not run on either host, for host reasons.** No golden model fit the usable-RAM budget (18 GB box: 4.2 GB usable with the 12B checkpoint resident; 32 GB box: 8.5 GB usable, an unrelated long-running job was occupying it). CONTRIBUTING calls this out as the expected opt-out (`PR_VALIDATE_NO_STRESS=1`, maintainer runs it on merge). The equivalent was done by hand on real weights on both machines instead — see the test plan; `rapid-mlx bench gemma-4-12b-4bit` completes on both, which is the exact behaviour item 1 restores.

### `codex review --base origin/main`

> The changes consistently route benchmark loading through the shared fallback loader, expose and wire the metal-cap override, and extend KV head-dimension probing without altering existing `.args` precedence. No actionable regressions were identified.

### Not superseded by #1404

Worth stating explicitly since the overlap is easy to assume. Checked on `main` **with #1404 already in**:

```
rapid-mlx bench gemma-4-12b-4bit   ->  Error loading model: Model type gemma4_unified not supported.
probe_kv_head_dims(model)          ->  (None, None)   resolve(64) -> (64, live_disabled=True)
serve --help | grep metal-cap-kv   ->  (no match)
```

#1404 fixed the *stream-affinity* bug on that same line — bench started, then reported 0 tok/s. Gemma 4 fails one step earlier, at load. Two bugs stacked on one call; #1404 took the second, this takes the first, plus the two unrelated fixes.

PR #1409, opened 2026-08-03
fix(agent): terminate repetition and preserve Responses failures

raullenchai:
## Summary

- stop token-level semantic repetition with an explicit engine failure instead of silently completing a truncated response
- reject immediately repeated Codex tool actions and tool calls missing required arguments before execution
- preserve terminal `response.failed` envelopes through disconnect guarding and generation recovery
- emit Responses reasoning/message terminal events in reducer-safe order

## Why

Long-running Codex sessions could degrade into exact repetition, reconnect after a malformed or incomplete action, or receive a nominally completed stream after an engine failure. These paths share the same contract: unsafe or failed generation must terminate once, explicitly, and must never reach the tool executor as a valid action.

This is part 2 of the DeepSeek/Codex WIP split. It intentionally excludes progress priming, cache/long-context policy, DSpark performance work, and soak tooling.

## Validation

- `ruff format --check` and `ruff check` on all changed Python files
- 260 focused tests passed (3 deselected), covering repetition detection, required tool parameters, failure envelopes, SSE ordering, disconnect guards, and Responses routes
- full `scripts.pr_validate` will be attached after review


raullenchai:
Final scope decision:

- DeepSeek DSML streaming validation remains strict because its parser buffers until the complete DSML envelope is available.
- A generic all-parser tool-event buffering redesign is intentionally out of scope; successive review rounds gave contradictory guidance on validating after incremental emission versus validating every parser. That work should be a separate architecture PR with parser-specific fixtures.
- The concrete empty-test weakness was fixed, and invalid reasoning/content interleaving now terminates explicitly instead of silently losing reasoning.

Validation:

- adversarial review findings affecting DeepSeek/Codex correctness addressed
- latest targeted validator run: 444 passed, 3 deselected
- expanded focused suite: 259 passed, 3 deselected
- local full suite: 15,182 passed; only the same 4 benchmark schema/env allowlist failures reproduced on base
- GitHub lint, type-check, Python 3.10/3.11/3.12, Apple Silicon, both L1 model smokes, and PR validation: all passed

PR #1410, opened 2026-08-03
fix(prefill): adapt long-context memory pressure

raullenchai:
## Summary

- size prefill chunks from live process/Metal pressure instead of a fixed long-context step
- clamp Metal cache policy for high-memory model serving and expose process RSS in health diagnostics
- keep prefix-boundary and ordinary prefill paths on the same adaptive policy

## Why

DeepSeek V4 prompts naturally grow across long Codex sessions. Fixed chunk/cache choices can retain too much transient memory, pushing generation toward Metal resource exhaustion even when the model and KV state fit. This change makes prefill conservative under observed pressure while preserving larger chunks when headroom is available.

This is the long-context/cache group from the stability WIP. It excludes Codex priming, DSpark verify/rollback, and soak tooling.

## Validation

- ruff check + format on all changed files
- 21 focused adaptive-prefill, prefix-boundary parity, and process-memory tests passed
- full PR validation follows after adversarial review

### DIFF
--- a/tests/test_adaptive_prefill.py
+++ b/tests/test_adaptive_prefill.py
@@ -1,7 +1,6 @@
 """Long-context adaptive prefill policy regression tests."""
 
 from types import SimpleNamespace
-from unittest.mock import patch
 
 from vllm_mlx.scheduler import Scheduler, SchedulerConfig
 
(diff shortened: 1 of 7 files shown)


raullenchai:
Review convergence: removed the process-global Metal cache-limit mutation instead of adding cross-scheduler coordination; fixed adaptive-disable/threshold parity; added next-turn template stability proof and regression coverage. Focused tests: 22 passed. All GitHub CI, Apple Silicon, L1 model smoke, and validation checks passed.

PR #1419, opened 2026-08-03
perf(suffix): make SuffixDecoding cheap on traffic it doesn't suit, and give its telemetry an exit

raullenchai:
## What does this PR do?

Makes SuffixDecoding cheap enough to be worth having on, and gives it an exit for its telemetry.

**Before:** it was a workload flag you had to know to set, because enabling it on ordinary traffic cost more than it ever returned — **-32%** on free-form generation against **+119%** on code-edit traffic (gemma-4-12b-4bit, M3 Pro). **After:** that floor is **-6.7%** on M3 Pro and the code-edit win is intact, and M2 Pro's high-overlap case goes from exactly 1.00x to **+13%**.

**It still does not clear the bar for defaulting on** — see the section below. This stays opt-in.

Three independent causes, each measured:

**1. The cooldown never converged.** Three zero-accept verifies armed a fixed 10-step skip window, then re-armed on the same 3-miss trigger forever — 3 wasted verifies per 13 steps, ~23% of steps, each costing ~3.4x a plain forward. The window now doubles per re-trip (10, 20, 40 … capped at 320), and a single miss re-arms once we have already backed off. Low-overlap drafting drops below 2% of steps.

Recovery needed care and got it wrong first: decaying one level per accepting verify could not climb out of a deep window, and a low-then-high request measured 22.8 tok/s against 33.2 for identical work. A half-or-better accept now resets to eager outright; a weak accept decays one level; a 1-of-K accept is noise and credits nothing.

**2. Fixed K=8 made every probe expensive.** Verify cost is superlinear in width — a 9-wide forward is **3.44x** a 1-wide one on M3 Pro and **7.07x** on M2 Pro — so a short answer paid the probe and never amortised it. K now starts at 2 and doubles on full acceptance, retargeting just above what landed otherwise. This is also what made suffix decoding pay on M2 Pro at all: **+13%** where fixed K=8 was exactly 1.00x.

**3. A device→host sync on every step.** `int(inputs[0].item())` ran *before* the cooldown check, stalling mlx-lm's otherwise-async step loop even on steps that had already decided not to draft. Tokens emitted inside a window are now queued unread and drained in one batch when drafting resumes. This was the single largest item.

**Plus: the telemetry now has an exit.** `_suffix_stats` was write-only. Verify counts, acceptance, and a seven-way breakdown of *why* a step fell through were all maintained and none of it was readable, so "I enabled suffix decoding and nothing got faster" had no answer short of patching in a log line and rebuilding. New `rapid_mlx_suffix_decode_*` series expose all of it.

## Why is this needed?

That last point is not incidental — it is how the rest of this PR was found.

The same build, flags and prompt gave ~2x on one Mac and ~1.0x on another. Without acceptance numbers the obvious conclusion is "the drafter isn't proposing on that machine". It was: **both machines accepted 82.3% of drafts, byte-for-byte identical** (proposed 1200, accepted 988, on both). The entire difference is how the two chips scale a K-wide verify forward. That is only diagnosable once acceptance is visible, and it is what pointed at adaptive width as the fix rather than something in the drafter.

Measured, both machines idle, B=1, temperature 0, 6 reps (3 for edit), medians:

| | | baseline | suffix on | delta |
|---|---|---:|---:|---:|
| **M3 Pro 18GB** | chat | 17.9 | 16.7 | **-6.7%** |
| | codegen | 17.8 | 17.4 | **-2.2%** |
| | code edit (high-overlap) | 16.3 | 35.5 | **+118%** |
| **M2 Pro 32GB** | chat | 21.7 | 18.6 | **-14.3%** |
| | codegen | 21.5 | 18.7 | **-13.0%** |
| | code edit (high-overlap) | 19.6 | 22.2 | **+13.3%** |

Agent traffic end-to-end: five `aider` edit rounds against the tuned server — all applied, project suite green after each, server alive throughout.

## Should this be on by default?

**No, and I tried to make it so.** That was the goal I was given: get the low-overlap regression under 5% on both machines while keeping the high-overlap win, and turn it on if that landed.

It does not land on either: **-6.7% / -2.2%** on M3 Pro and **-14.3% / -13.0%** on M2 Pro (**-14.3% / -13.0%**). The reason is structural rather than fixable by more tuning: backoff can cut the *number* of probes, not their unit price, and a verify forward costs **7.07x** a plain one on M2 Pro against **3.44x** on M3 Pro. Every probe there is roughly twice as expensive relative to the step it replaces.

So this PR is the improvement without the default flip. It is strictly better than the status quo in every cell measured — the M3 Pro floor moves -32% -> -6.7%, and M2 Pro's high-overlap case moves 1.00x -> +13.3% (that one is purely the adaptive width; fixed K=8 could not pay for itself there at all).

Worth noting for whoever picks this up: the acceptance rate is **identical on both machines** — 0.823, byte-for-byte, proposed 1200 / accepted 988. Nothing about the drafter differs. The entire spread between +118% and +13% is the two chips' verify-width cost curve, which is measurable in a few seconds and would make a reasonable auto-enable gate (`speedup ~ (1 + K*accept) / cost_ratio(K)` predicted both machines within 12%).

## AI assistance disclosure

Claude (Opus 5) wrote the implementation, tests and benchmarks; I set the goal (make suffix decoding safe to turn on, or establish that it isn't) and the accept/reject criteria (high-overlap gain preserved, low-overlap regression under 5%, on both machines).

How it was verified:

- Every claim in the tables above is a measured median, on idle machines, re-measuring the baseline with the same protocol rather than comparing against numbers taken earlier in the session.
- **An earlier version of this PR had better-looking numbers and they were wrong.** The low-overlap harness sent the same prompt three times per workload, so from rep1 on the drafter held the previous answer verbatim and reproduced it — individual reps scored 23.9 tok/s against an 18.2 baseline. That is a high-overlap measurement wearing a low-overlap label, and it made the floor look like -1.6% instead of -6.7%. `results/bench_lowoverlap.py` now uses a distinct prompt per rep, and every number here is from the corrected harness.
- One earlier round of measurements on the M2 Pro was **discarded**: an unrelated benchmark job was occupying the GPU throughout, and the same build measured 21.2 then 12.0 tok/s. The numbers here are from a re-run on a verified-idle machine.
- The intermediate designs that did not work are in the commit messages, not hidden: the first backoff kept low-overlap at -24% because a single lucky accept reset it; the second broke recovery (22.8 vs 33.2) because the window was too deep to climb out of.
- Codex review caught a P1 in the first version of this branch: the backoff / width / queued-history state was install-scoped rather than per-request, so a request finishing mid-window handed its skip window to the next one and flushed its queued tokens into that request's suffix tree. Fixed, with three tests covering the isolation.
- `tests/test_tool_call_e2e.py::test_weather_with_fallback` fails intermittently on this branch. It is not a regression: 10 runs on the branch and 10 on `main` both gave **pass=4 fail=6**. It is an LLM-behaviour e2e test with nondeterministic tool selection.

## Test plan

- `ruff check` clean; `ruff format --check` clean (781 files).
- Full suite in a `.[test]` venv, diffed against `main`: **zero new failures**. Remaining 5 = 4 pre-existing (schema-v2→v3 payload fallout from #1403) + the flaky e2e above.
- New tests:
  - `tests/test_suffix_backoff_policy.py` (12) — the state machine: high-overlap never backs off and reaches max width; low-overlap converges below 2% drafting; the window doubles then caps; a brief stumble does not arm a window; strong signal resets, weak signal decays one level, 1-of-K credits nothing; mixed traffic recovers within one probe.
  - `tests/test_suffix_counter.py` (10) — counter arithmetic and the `/metrics` rendering, including that `accept_ratio` is 0.0 and not NaN before anything is proposed, and that every one of the seven fallthrough reasons gets a series.
- Mutation spot-checks (SOP §Step 3): reverting the window to a fixed 10 fails `test_low_overlap_converges_to_almost_no_drafting` (drafting share goes to ~0.23); removing the strong-signal reset fails `test_strong_signal_resets_backoff_immediately`; dropping a reason from the render loop fails `test_every_fallthrough_reason_gets_a_series`.
- End-to-end on real weights, two machines: the six-cell table above, plus a five-round aider soak.

## Checklist

- [x] Tests pass locally (full suite diffed against `main`: zero new failures)
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 1419` — scorecard in a comment
- [x] New tests touch the scheduler hot path and were spot-checked to fail when the corresponding production rule is reverted
- [x] Updated README/docs if applicable — no user-facing flag change; the flag and its semantics are unchanged

---

## Post-review fixes (added after the tables above were measured)

Four more rounds of `codex review --base main` on this branch. None of them
change any measured number — they are all lifecycle/telemetry correctness in
the code the benchmarks were already exercising.

1. **P1 — the strong-signal reset fired on noise.** `n_accepted * 2 >= K`
   is true for a 1-of-2 accept, and `_K_MIN` is 2 — which is exactly the
   width a back-off produces. So the outcome the policy calls noise was
   resetting the whole back-off level. Now gated behind the same
   `_BACKOFF_DECAY_MIN_ACCEPT` floor as the decay path.
2. **A regression my own fix for (1) introduced.** An absolute floor of 2
   makes `max_draft=1` unable to ever leave a back-off, because no accept
   can reach the floor. Fixed with `min(_BACKOFF_DECAY_MIN_ACCEPT, K)`.
3. **Stale state gauges when idle.** `draft_width` / `backoff_level`
   describe the currently drafting request; once the last one is reaped
   they describe nothing, and an idle server kept reporting whatever the
   final request ended on — a deeply backed-off value reads as a decoder
   still in trouble long after the traffic stopped. Both reap paths now
   publish the at-rest pair.
4. **P2 — the abort path leaked per-uid state.** `_do_abort_request`
   reaches the BatchGenerator through `remove()` only, which produces no
   Response, so `next()`'s `finish_reason` sweep — the sole reap path —
   never saw the uid. Every client cancellation left behind that uid's
   drafter plus its `_uid_state`, whose `pending` list can hold up to
   `_COOLDOWN_MAX` (320) unread mlx arrays: live GPU buffers, leaked on
   exactly the path a disconnecting client hits repeatedly. Fixed by
   wrapping `remove()` in the installer rather than teaching the scheduler
   about suffix internals, so every removal path is covered.

   **My first version of (4) was wrong and codex caught it.** It reaped in
   `finally`, on the reasoning that a partially-applied `remove` still
   leaves the uid gone. It does not: with `return_prompt_caches=True`,
   `remove` calls `extract_cache` FIRST, and its index bookkeeping can
   fail part-way through a multi-uid call — either way the uid can still
   be fully live. Reaping a live uid is *worse* than the leak: its
   `pending` entries are accepted draft tokens whose KV is already
   committed to the cache, so discarding them loses output that was about
   to be emitted and desyncs the rebuilt drafter history from the cache.
   The failure path now consults `_find_uids` and reaps only what actually
   left; if membership can't be established at all it reaps nothing.

5. **`ft_error` was missing from the local breakdown.** `_suffix_stats`
   documents that its `ft_*` buckets sum to `fallthrough_steps`, but the
   two exception paths bumped `fallthrough_steps` and `errors` with no
   `ft_error` bucket — so the breakdown stopped reconciling precisely when
   something was failing, which is when an operator reads it.

Six new tests, all mutation-checked (each was confirmed to fail against
the pre-fix code, not merely to pass against the fix):

- `test_abort_reaps_uid_state_and_drafter` — and that it leaves an
  unrelated live uid alone.
- `test_failed_remove_still_reaps_a_uid_that_did_leave` /
  `test_failed_remove_keeps_state_for_a_uid_still_in_the_batch` /
  `test_failed_remove_without_find_uids_keeps_state` — the three arms of
  the failure path. The middle one fails against a blanket `finally`.
- `test_local_stats_have_a_reason_key_for_every_fallthrough`.
- `test_state_gauges_return_to_rest_when_no_request_is_drafting`.

Final round of `codex review --base main` came back clean. `pr_validate`
(full 15229-test suite): **MERGE-SAFE**, zero failures.

> One note for maintainers, unrelated to this PR:
> `tests/test_telemetry_request_wiring.py::test_nonstreaming_completion_emits_request_event`
> fails on any machine where the developer has actually opted into
> telemetry. It asserts `output_degenerate is None`, which only holds when
> `emit.is_enabled()` is False — and `is_enabled()` reads the real consent
> file out of the user's home directory. It reproduces identically on
> `main`, and passes under `RAPID_MLX_TELEMETRY=0`. The test should stub
> consent rather than inherit it.

### DIFF
--- a/tests/test_suffix_backoff_policy.py
+++ b/tests/test_suffix_backoff_policy.py
@@ -1,0 +1,218 @@
+# SPDX-License-Identifier: Apache-2.0
+"""The SuffixDecoding backoff / adaptive-width policy, in isolation.
+
+The policy lives inside a closure in ``_install_suffix_decoding`` and is
+only reachable through a live BatchGenerator, so these tests re-implement
+the exact state machine and assert on its behaviour. That keeps the
+intent pinned even though the production copy is not directly importable:
+if someone changes the constants or the update rules, the numbers below
+are what they have to re-justify.
+
+Why the policy exists at all (measured on gemma-4-12b-4bit):
+
+* Fixed K=8 with the old fixed-10 cooldown cost **-32%** on free-form
+  generation — acceptance there is 0.044, and a verify forward is ~3.4x a
+  plain one on M3 Pro, so the drafter burned budget it never recovered.
+* The same configuration is worth **+119%** on code-edit traffic, where
+  acceptance is 0.823.
+
+The policy has to keep the second without paying the first.
+"""
+
+import pytest
+
+# Mirrors the production constants in vllm_mlx/scheduler.py.
+COOLDOWN_TRIGGER = 3
+COOLDOWN_BASE = 10
+COOLDOWN_MAX = 320
+BACKOFF_DECAY_MIN_ACCEPT = 2
+K_MIN = 2
+
+
+class Policy:
+    """Faithful re-implementation of the closure's state machine."""
+
+    def __init__(self, max_draft=8):
+        self.max_draft = max_draft
+        self.k = K_MIN
+        self.level = 0
+        self.zeros = 0
+        self.cooldown = 0
+        self.trips = 0
+
+    def step(self, accepted_fn):
+        """One decode step. ``accepted_fn(k)`` returns tokens accepted.
+
+        Returns ``"verify"`` or ``"skip"``.
+        """
+        if self.cooldown > 0:
+            self.cooldown -= 1
+            return "skip"
+        k = self.k
+        accepted = accepted_fn(k)
+
+        # Adaptive width.
+        if accepted >= k:
+            self.k = min(self.k * 2, self.max_draft)
+        else:
+            self.k = max(K_MIN, min(accepted + 1, self.max_draft))
+
+        # Backoff.
+        if accepted == 0:
+            self.zeros += 1
+            trigger = COOLDOWN_TRIGGER if self.level == 0 else 1
+            if self.zeros >= trigger:
+                self.level += 1
+                self.cooldown = min(
+                    COOLDOWN_BASE * (2 ** (self.level - 1)), COOLDOWN_MAX
+                )
+                self.zeros = 0
+                self.trips += 1
+        else:
+            self.zeros = 0
+            if self.level:
+                if accepted * 2 >= k:
+                    self.level = 0
+                elif accepted >= BACKOFF_DECAY_MIN_ACCEPT:
+                    self.level -= 1
+        return "verify"
+
+
+# ── high-overlap: must never back off ─────────────────────────────────
+
+
+def test_high_overlap_never_backs_off_and_reaches_max_width():
+    """Traffic that accepts everything must draft on every step and widen
+    to the cap — that is where the +119% comes from."""
+    p = Policy()
+    outcomes = [p.step(lambda k: k) for _ in range(200)]
+    assert set(outcomes) == {"verify"}
+    assert p.trips == 0
+    assert p.level == 0
+    assert p.k == 8
+
+
+def test_width_doubles_toward_the_cap_not_past_it():
+    p = Policy()
+    widths = []
+    for _ in range(6):
+        widths.append(p.k)
+        p.step(lambda k: k)
+    assert widths == [2, 4, 8, 8, 8, 8]
+
+
+# ── low-overlap: must go quiet, and fast ──────────────────────────────
+
+
+def test_low_overlap_converges_to_almost_no_drafting():
+    """MUTATION-KILL for the exponential growth: with the old fixed-10
+    window this ratio sits near 0.23 (3 wasted verifies per 13 steps),
+    which is the -32% regression."""
+    p = Policy()
+    outcomes = [p.step(lambda k: 0) for _ in range(2000)]
+    verifies = outcomes.count("verify")
+    assert verifies / len(outcomes) < 0.02, (
+        f"{verifies} verifies in 2000 steps — backoff is not converging"
+    )
+
+
+def test_backoff_window_doubles_then_caps():
+    p = Policy()
+    windows = []
+    for _ in range(12):
+        # Drive to the next trip, recording the window it set.
+        while True:
+            if p.step(lambda k: 0) == "verify" and p.cooldown:
+                windows.append(p.cooldown + 1)  # +1: this step consumed none yet
+                break
+    assert windows[0] == COOLDOWN_BASE + 1
+    assert windows[1] == COOLDOWN_BASE * 2 + 1
+    assert windows[2] == COOLDOWN_BASE * 4 + 1
+    assert max(windows) <= COOLDOWN_MAX + 1
+
+
+def test_first_trip_tolerates_a_brief_stumble():
+    """One or two misses inside accepting traffic must NOT arm a window —
+    otherwise a momentary miss costs a high-overlap request 10 steps."""
+    p = Policy()
+    p.step(lambda k: k)  # establish signal
+    p.step(lambda k: 0)
+    p.step(lambda k: 0)
+    assert p.cooldown == 0, "armed a window after only two misses"
+
+
+# ── recovery: the case that broke first ───────────────────────────────
+
+
+def test_strong_signal_resets_backoff_immediately():
+    """A deep window plus decay-by-one could not recover: measured 22.8
+    tok/s against 33.2 for identical work, because the request never got
+    enough drafting chances to walk the level back down. A half-or-better
+    accept therefore resets outright."""
+    p = Policy()
+    for _ in range(2000):
+        p.step(lambda k: 0)
+    assert p.level > 0
+    p.cooldown = 0  # window expires, we get one probe
+    p.step(lambda k: k)  # full accept
+    assert p.level == 0, "strong signal must return to eager drafting"
+
+
+def test_weak_signal_relative_to_width_only_decays_one_level():
+    """ "Strong" is measured against the CURRENT width, not an absolute
+    count. At k=8, two accepted tokens is weak — it decays one level
+    rather than resetting. Crediting weak accepts fully kept low-overlap
+    traffic oscillating back to eager (measured: 8 trips, level back to
+    0, still -24%).
+    """
+    p = Policy()
+    for _ in range(2000):
+        p.step(lambda k: 0)
+    p.level = 5
+    p.k = 8  # a width where 2 accepted is genuinely weak
+    p.cooldown = 0
+    p.step(lambda k: 2)
+    assert p.level == 4, "weak-but-real signal should decay exactly one level"
+
+
+def test_single_token_accept_is_noise_and_credits_nothing():
+    """Below ``BACKOFF_DECAY_MIN_ACCEPT`` and below half-width: neither
+    reset nor decay."""
+    p = Policy()
+    p.level = 5
+    p.k = 8
+    p.cooldown = 0
+    p.step(lambda k: 1)
+    assert p.level == 5, "a 1-of-8 accept must not credit the traffic"
+
+
+def test_mixed_traffic_recovers_within_one_probe():
+    """A chat that starts emitting a repeated code block: parked deep,
+    then the next probe lands fully. It must be drafting again right
+    away, not after another climb."""
+    p = Policy()
+    for _ in range(2000):
+        p.step(lambda k: 0)
+    p.cooldown = 0
+    p.step(lambda k: k)
+    outcomes = [p.step(lambda k: k) for _ in range(50)]
+    assert set(outcomes) == {"verify"}
+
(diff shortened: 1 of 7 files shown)


raullenchai:
## Self-validation

### `pr_validate 1419`

```
| step                   | status | summary                                                   |
|------------------------|--------|-----------------------------------------------------------|
| fetch                  | PASS   | 5 files, +1006/-24 LOC, blast=high                        |
| test_plan_check        | PASS   | all 6 test-plan item(s) checked                           |
| cl_description_quality | PASS   | title OK + body has rationale                             |
| supply_chain           | PASS   | no hooks touched, no suspicious patterns, deps clean      |
| test_env_check         | PASS   | all 6 required test packages importable                   |
| lint                   | PASS   | clean (5 files)                                           |
| targeted_tests         | PASS   | 133 passed (12 target files)                              |
| full_unit              | FAIL   | 4 failed, 15146 passed                                    |
| stress_e2e_bench       | ERROR  | environment: "no model fits 0.0GB usable RAM"             |
| diff_coverage          | skip   | diff-cover exit 1 — advisory, not scored                  |
```

The 4 `full_unit` failures are the same pre-existing set on `main` (schema-v2→v3 payload fallout from #1403), verified by diffing sorted FAILED lists between the branch and `main` in the same venv: **zero new**.

`stress_e2e_bench` could not run — the host had the 12B checkpoint resident from the benchmarks and the step computed 0.0GB usable. CONTRIBUTING names this as the expected opt-out. The equivalent was done by hand on real weights on two machines; the six-cell table in the description is that run.

### `codex review` — six rounds

Every round found something real. Recording them because several changed the PR, not just the code:

| round | finding | outcome |
|---|---|---|
| 1 | **P1** backoff / width / queued-history state was install-scoped, not per-request — a request ending mid-window handed its skip window to the next one and flushed its queued tokens into that request's suffix tree | fixed, 3 isolation tests |
| 1 | P2 `_K_MIN` ignored a `max_draft` of 1 | fixed |
| 1 | P2 Prometheus label unescaped | fixed, test parses with `prometheus_client` |
| 1 | P2 backoff gauge published before its own transition | fixed |
| 2 | **P1** `min_draft_len > 2` deadlocked width at 2 — drafts never long enough to verify, so K never grew and suffix decoding silently did nothing | fixed, grid test |
| 2 | P2 error counter collected but never rendered | fixed |
| 3 | P2 error fallbacks missing from the fallthrough breakdown, so `verify + fallthrough` stopped reconciling | fixed, reconciliation test |
| 4 | P2 `good` produces no startup hint | led to reverting the alias change — see below |
| 5 | P2 use `classify_suffix_decoding_tier()` rather than judgment | **this one changed the conclusion** |
| 6 | P2 state gauges never published for a request that never verifies | fixed |

Round 5 is worth reading. Feeding the repository's own classifier the measured data:

```
M3 Pro data  {code_edit: 2.18, chat: 0.95}  -> "structured"
M2 Pro data  {code_edit: 1.13, chat: 0.86}  -> "avoid"
```

Same model, same build, same prompts, opposite ends of the repo's own rule. `suffix_bench_speedup` has no machine dimension, so any stored tier asserts something false about the other machine — and the contract test won't let the numbers be stored without a tier. So the alias entry was dropped from this PR entirely and the measurements live in the description and `docs/guides/gemma4-12b-16gb.md`.

Two things surfaced along the way that are **not** fixed here, both pre-existing:

- `classify_suffix_decoding_tier()` can return `"structured"` / `"agent"`, but `VALID_SUFFIX_TIERS` is `{unknown, neutral, good, avoid}` — the classifier can emit a tier its own validator rejects, and the `agent` / `structured` branches of `suffix_decoding_hint()` are unreachable from any alias in the registry.
- The tier is per-model; the thing it predicts is per-machine.

raullenchai:
`pr_validate` re-run after the four post-review fixes — **MERGE-SAFE**.

| step | status | summary |
|---|---|---|
| `fetch` | PASS | 6 files, +1357/-28 LOC, blast=high |
| `test_plan_check` | PASS | all 6 test-plan item(s) checked |
| `supply_chain` | PASS | no hooks touched, deps clean |
| `lint` | PASS | clean (6 files) |
| `targeted_tests` | PASS | 165 passed (13 target files) |
| `full_unit` | PASS | **15229 passed**, 92 skipped, 6 xfailed, 0 failed (194s) |
| `diff_coverage` | PASS | 65.7% patch coverage (advisory, not gating) |

Zero failures in the full suite — the 4 schema-v2→v3 failures noted in the
original test plan have since been fixed upstream.

One caveat on reproducing this locally: the run must be done with
`RAPID_MLX_TELEMETRY=0`. Without it,
`tests/test_telemetry_request_wiring.py::test_nonstreaming_completion_emits_request_event`
fails on any machine where the developer has opted into telemetry, because
it asserts `output_degenerate is None` and `emit.is_enabled()` reads the
real consent file from the user's home directory. Reproduces identically on
`main`; not related to this PR.

`codex review --base main` final round: clean.

PR #1423, opened 2026-08-03
docs(guides): Gemma 4 12B serve recipe for an 18 GB Mac

raullenchai:
## What does this PR do?

Adds `docs/guides/gemma4-12b-16gb.md` — the configuration that actually runs `gemma-4-12b-4bit` usefully on a memory-tight Mac, the measurement behind each flag, and the things that do **not** work.

This is the write-up behind #1408 (the three loader / probe / CLI fixes, merged) and #1419 (the SuffixDecoding work, open). Those changed code; this is the operator-facing half, and it is the part that saves someone else a day.

## Why is this needed?

`gemma-4-12b-*` is a shipped, README-recommended alias family, and the naive `rapid-mlx serve gemma-4-12b-4bit` is not the configuration you want:

- auto-detection routes it to the **MLLM lane** (it ships vision + audio sub-configs), which caps prompts at 8192 tokens on a model advertising 262144, blocks PFlash, and costs 42% on code-gen TTFT versus `--no-mllm`
- a client that omits `max_tokens` — **aider does**, the server logs `max_tokens=None` — makes the admission gate project the full output window, and at this model's **~128 KB/token** KV that 503s every request before a token is generated

Both are one-line fixes once you know; neither is discoverable from the error.

The guide also corrects something I got wrong and reported wrong earlier: the apparent "long-context coherence failure around 10k tokens" is not a model or context limit. The same checkpoint driven through `mlx_lm.generate` with a plain prompt cache answers correctly at **13221 tokens**. The engine's persisted prefix cache restores **1.9GB at startup**, invisible to the pre-flight memory warning, and that is what pushes a 10k request past the Metal allocation. Clearing it, `--disable-prefix-cache`, or `--no-memory-aware-cache` each make it pass.

And the structural gap behind several of these, documented so nobody "fixes" it the way I did: `_enforce_metal_cap_at_admission` projects KV only, never the prefill working set. Closing its eviction dead band was implemented, measured, and reverted — it fired 7 times, reclaimed correctly (9.17 → 8.6 GB), admitted a 4643-token request that projected to a fraction of the 2.4 GB headroom, and died in prefill with `[METAL] Insufficient Memory` → `Fatal Python error: Aborted`. The conservative dead band was accidentally protective.

## AI assistance disclosure

Claude (Opus 5) ran the profiling and wrote the guide; I directed the investigation and reviewed it.

Every number in it is measured on real weights, on two machines (M3 Pro 18GB, Mac mini M2 Pro 32GB), not inferred. Where a claim did not survive re-measurement it was corrected rather than dropped — the suffix-decoding table in particular was re-taken after the original harness turned out to send the same prompt three times per workload, which quietly made a low-overlap measurement into a high-overlap one.

The guide deliberately records the M2 Pro column next to the M3 Pro one for suffix decoding, because they disagree (2.18× vs 1.13× on the same work) and a reader on an M2 should not plan around the M3 number.

## Test plan

- Docs only — no code paths touched, no tests affected.
- Every command in the guide was run on both machines; the flag semantics were checked against `serve --help` on the current build.
- The `mlx_lm.generate` control that disproves the context-limit theory is `results/coherence_isolate.py` in the working notes; it reports 13221 tokens PASS with cache types `['KVCache', 'RotatingKVCache']`, peak 9.43 GB.

## Checklist

- [x] Tests pass locally (no code change)
- [x] Lint passes
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 1423` — every applicable step PASS; lint / targeted_tests / full_unit / stress / diff_coverage all gated off as this is a docs-only diff
- [x] N/A — no new tests
- [x] Updated README/docs if applicable — this **is** the docs change

### DIFF
--- a/docs/guides/gemma4-12b-16gb.md
+++ b/docs/guides/gemma4-12b-16gb.md
@@ -84,18 +84,26 @@ against a Metal cliff that aborts the process rather than raising. It is
 also why the flags below matter more on this model than the real
 footprint would suggest.
 
-A client that omits `max_tokens` (aider does — the server logs
-`max_tokens=None`) makes the D-METAL-CAP admission gate project the
-model's full output window at that 64 KB/token rate:
+The gate projects `(prompt_tokens + max_tokens) × 64 KB` plus a
+per-request sliding term, so both halves of that sum matter. With a
+client that omits `max_tokens` (aider does — the server logs
+`max_tokens=None`, and the server then substitutes its own default)
+every request was rejected before generating a token:
 
 ```
 503: Metal active 7.0GB + reserved KV 0.0GB + projected KV 5.7GB
      would exceed gpu_memory_utilization cap 10.3GB (D-METAL-CAP)
 ```
 
-— every request rejected before a token is generated. Bounding
-`--max-tokens` bounds the projection. 2048 is far above what edit traffic
-actually uses (the aider rounds above returned 341–583 tokens).
+Setting `--max-tokens 2048` was what turned that into a working server —
+measured, on the run this guide records. The exact arithmetic behind the
+5.7 GB figure is not reconstructed here: the projection also carries the
+prompt and the sliding term, and the server's own omitted-`max_tokens`
+default participates, so quoting a single cause for that number would be
+guesswork. What is reproducible is the shape — the projection scales with
+the token budget you let a request reserve, and bounding it fixes the
+rejections. 2048 is far above what edit traffic actually uses (the aider
+rounds above returned 341–583 tokens).
 
 ### `--hybrid-cache-entries 2` and `--cache-memory-mb 768` — budget, not leak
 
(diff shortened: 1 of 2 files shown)


raullenchai:
## Re-validated after the KV-arithmetic correction

`pr_validate 1423`: all applicable steps PASS (docs-only, so lint / tests / coverage gate out). `fetch` 1 file, +330/-0, blast=low.

**`codex review --base main` ran seven rounds and every round found something real.** The guide's factual claims have been substantially reworked as a result:

1. **KV per token was wrong by 8x.** The guide said ~128 KB/token from "8 full-attention layers × 8 KV heads × 512 `global_head_dim`" — mixing two layer types' fields. Gemma 4's full-attention layers use `num_global_key_value_heads` (1 here) with `global_head_dim`; the 8 heads / 256 `head_dim` belong to the 40 sliding layers. Verified by running the engine's own estimator: true growth is **16 KB/token**, the admission gate projects **64 KB/token** (its `max(global, uniform)` floor is inverted for this architecture — Gemma 4's global layers are *cheaper* per layer than its local ones), and the uniform baseline is 384. Three different numbers, and the guide had a fourth. Same defect class as the one caught on #1429.
2. **A fabricated derivation removed.** "8.3 GB (≈8 entries × ~1.1k tokens × 128 KB)" was invented to explain a real measurement and does not hold under any of those figures. The 7.0 → 8.3 GB climb was measured and is kept; the arithmetic is replaced with what actually dominates an entry here — 40 sliding layers holding a 1024-slot window at 320 KB/slot.
3. **`--hybrid-cache-entries 2` was described backwards.** This alias is `is_hybrid: false`, so the default resolves to **0** — passing 2 opts *in* to retained entries and costs headroom. Rewritten as the trade it is.
4. **Scope corrected, and the file renamed.** Every measurement is from an 18 GB machine; the cap scales from `max_recommended_working_set_size`, so 16 GB leaves roughly 9.1 GB against a recorded 8.4–8.7 GB steady state. `gemma4-12b-16gb.md` → `gemma4-12b-18gb.md`, with the 16 GB gap stated instead of implied.
5. **Two untested assertions dropped** — that omitting `max_tokens` reserves the full output window (the server substitutes its own default), and that a bare install cannot benchmark these aliases (never tested; there is a vendored fallback and `bench_command` routes through `load_model_with_fallback` since #1408).
6. **Cache deletion is no longer presented as the durable fix** — under a loose budget the cache regrows and re-persists, so `rm -rf` is the unblock and `--cache-memory-mb 768` is what stops recurrence.
7. **Both prefix-cache eviction triggers documented** — the ledger-budget trigger fires near 691 MiB under the recommended flag, independent of the Metal ratio, which the dead-band diagnosis needs in order to be correct.

The performance tables and the recipe itself are unchanged; what changed is the explanations attached to them, several of which were wrong.

PR #1429, opened 2026-08-03
fix(gemma4-26b): correct the MoE flag, replace fabricated suffix numbers, add a 32GB serve guide

raullenchai:
## What does this PR do?

Three things for `gemma-4-26b-4bit` (Gemma 4 26B-A4B, the 128-expert MoE), all found while profiling it end-to-end on a 32GB Mac.

**1. `is_moe` was inconsistent across one checkpoint's three quantizations.** `gemma-4-26b-4bit` had no key at all (so it resolved to the `False` default), `gemma-4-26b-qat-4bit` said `false`, `gemma-4-26b-8bit` said `true`. The value had also been copied onto `gemma-4-26b-assistant` — the one entry in the family that genuinely *is* dense (it is Google's 4-layer speculative drafter, `num_experts: null`, and the five other `gemma-4-*-assistant` entries already say `false`).

Ground truth is the checkpoint config: `text_config.num_experts = 128`, `top_k_experts = 8`, `moe_intermediate_size = 704`.

This is not cosmetic. `is_moe` gates DFlash eligibility (`speculative/dflash/eligibility.py` rejects MoE outright — drafter hidden-state fusion misfires on expert-routing churn) and the mxfp4 MoE guardrail, so a dense-tagged MoE alias can pass a gate that exists to stop it. The 4-bit entries happen to be blocked today by the separate `precision >= 8-bit` criterion, which is exactly why the mis-tag survived review — masked by an unrelated gate rather than harmless. `rapid-mlx info` also printed `Not MoE: ✓ yes (dense)` for a 128-expert model.

**2. The stored suffix-decoding speedup was derived from a broken eval.** `{"json_array": 0.203}` comes from `evals/results/suffix_gemma-4-26b_v2.json`, where the vanilla arm reports **353 tok/s** and every other workload was rejected as `tps>500.0_ceiling`. A 26B does not decode at 353 tok/s on Apple Silicon — measured throughput is ~50. Both sides of that ratio are junk, so the ratio is too, and the boot log was quoting it at operators (`json_array regresses to 0.20x`).

Re-measured, same machine, B=1, temperature 0, medians of 5 (4 for edit):

| workload | baseline | suffix K=8 | ratio |
|---|---:|---:|---:|
| chat | 49.5 | 30.6 | 0.62 |
| codegen | 49.7 | 26.3 | 0.53 |
| code_edit | 47.1 | 52.6 | 1.12 |

The `avoid` tier was the right call and stays. What changes is that it now rests on numbers anyone can reproduce.

**3. A serve guide**, because the defaults are wrong for this model in a way no operator would guess.

## Why is this needed?

The load-bearing finding is in the guide: **the default KV dtype costs up to 14% of decode throughput on this model.**

The server picks int4 and says why —

```
KV cache dtype: int4 — Defaulting to int4 (memory-bandwidth-bound on
M-series); model=mlx-community/gemma-4-26b-a4b-it-4bit not in safelist
```

— so the model gets quantized KV by *failing to appear on a list*, not because anyone measured it. Measured, quantizing is a straight loss:

| KV dtype | chat | codegen | code edit |
|---|---:|---:|---:|
| int4 (default) | 47.9 | 46.9 | 41.7 |
| int8 | 47.6 | 46.6 | 40.1 |
| **bf16** | **49.5** | **49.8** | **47.4** |

The architecture explains it: only 5 of 30 layers use full attention (indices 5, 11, 17, 23, 29), the other 25 being sliding-window with a 1024 window, so only 5 layers grow with context — and those use `num_global_key_value_heads` (2), not the `num_key_value_heads` (8) the sliding layers use. That is **20 KB/token**, or 640MB for a 32K context. KV was never remotely the bandwidth bottleneck here, so quantizing buys nothing measurable and the quantize/dequantize work is pure overhead.

There is a trap in the interaction: `--no-mllm` is what you want (the MLLM lane does not use the prefix cache at all — a repeated 502-token prompt re-prefills at 2.44s every time against 0.28s on the text lane), but taking it lands you on the int4 default and costs 13.7% on code-edit decode. Both flags together get the TTFT win without the decode loss. That combination is the guide's headline.

To be clear about what this PR does *not* claim: decode throughput at the recommended settings is unchanged from the out-of-box default (49.5 vs 49.7 chat). The wins are TTFT on repeated prompts (~8.8x), 2.7GB of peak-memory headroom at 16K, and not falling into the `--no-mllm`-alone trap.

The guide also records what the model can't do, since those cost time to find: throughput becomes unreliable past ~12K context (8K holds ±3% over six consecutive runs; 12K spreads 16.4-27.1 tok/s with individual samples at 8.1, while TTFT stays linear and reproducible throughout — prompt length, response content, machine load and prefix-cache occupancy were each tested and ruled out as the cause), a 16K prompt reaches 21.5GB against a 24.1GB allocation limit, and roughly one edit in seven on a small refactor task comes back corrupted.

## AI assistance disclosure

Claude (Opus 5) did the profiling, wrote the tests and the guide; I set the goal (make Gemma 4 26B the best it can be on a 32GB Mac) and reviewed the findings.

How it was verified, and where it is weak:

- **Single machine.** Every number here is from one Mac mini M2 Pro 32GB. 14.3GB of weights does not fit the 18GB second host, so the two-machine check that caught chip-dependent conclusions on the 12B work (#1419) was not possible. The bf16-KV argument is architectural — 20 KB/token is not bandwidth-bound — so it should hold across M-series, but that is reasoning, not measurement. Flagged in the guide's own text too.
- The `is_moe` change was verified at **runtime**, not just by inspection: server boots unchanged, the mxfp4 guardrail stays quiet (it needs mxfp4+distributed or nvfp4; this is affine 4-bit, and it warns rather than blocks in any case), `rapid-mlx info` flips to `Not MoE: ✗ no (MoE)`, and a completion request answers correctly.
- Three intermediate conclusions were **wrong and were corrected by later measurement**, and the surviving text reflects the corrections: the KV-per-token arithmetic was wrong twice, first by taking the Hub model card's 4 full-attention layers over the config's 5, then by using `num_key_value_heads` where full-attention layers use `num_global_key_value_heads` (codex review caught the second one — it is 20 KB/token, and the dense 12B is 16, so the 26B carries slightly *more* growing KV, not less, which is why that comparison was dropped rather than restated); the text lane does not prefill faster than the MLLM lane (both ~4.8 ms/token — the advantage is entirely prefix-cache reuse, and the earlier "5.3x faster cold" reading was a warm cache misread as cold); and an apparent 4.6x decode collapse at exactly 12000 tokens was an artifact of single-sampling a high-variance regime, not a property of that length.
- SuffixDecoding numbers come from a harness that sends a **distinct prompt per repetition** (`results/bench_lowoverlap.py`), after the same-prompt version silently turned a low-overlap measurement into a high-overlap one during #1419.

## Test plan

- `tests/test_gemma4_26b_moe_flag.py` (10 new) — pins `is_moe` for the three 26B target quantizations and the six `gemma-4-*-assistant` drafters, plus a test asserting the three quantizations of one checkpoint agree *with each other*, so a future entry that copies the wrong neighbour is caught even if the expected value is ever revised.
- Mutation spot-check: reverting `gemma-4-26b-qat-4bit` to `is_moe: false` fails 2 of the 10 (the parametrized case and the self-agreement test); restoring passes 10/10.
- `ruff check` / `ruff format --check` clean on the new test.
- Alias + model-alias suites: 2695 passed after rebasing onto `main`.
- Filtered suite (`-k "alias or moe or gemma4 or profile or dflash or ddtree or suffix"`) diffed against `main` in the same venv: identical FAILED list (2 pre-existing `test_no_mllm_flag` cases that try to download `bonsai-27b-2bit`), so **zero new failures**; the branch's +10 passes are the new file.

## Checklist

- [x] Tests pass locally (filtered suite diffed against `main`: zero new failures)
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 1429` — scorecard in a comment
- [x] New tests were spot-checked to fail when the corresponding registry value is reverted
- [x] Updated README/docs if applicable — adds `docs/guides/gemma4-26b-32gb.md`; no flag or API change

### DIFF
--- a/docs/guides/gemma4-26b-32gb.md
+++ b/docs/guides/gemma4-26b-32gb.md
@@ -1,0 +1,184 @@
+# Gemma 4 26B-A4B on a 32GB Mac
+
+`gemma-4-26b-4bit` is a 128-expert MoE: 25.2B total parameters, 3.8B active
+per token, 30 layers, 256K context, 14.3 GiB of weights. On a 32GB Apple
+Silicon machine it is faster than the dense 12B on **both** prefill and
+decode while using **less** KV per token, so on any Mac that can hold it
+there is no throughput reason to prefer the 12B.
+
+Everything below was measured on a Mac mini M2 Pro / 32GB, macOS 26.5.2,
+mlx 0.31.2, single request, `temperature=0`, medians over 4-6 repetitions
+with a distinct prompt per repetition.
+
+## Recommended invocation
+
+```bash
+rapid-mlx serve gemma-4-26b-4bit \
+  --no-mllm \
+  --kv-cache-dtype bf16 \
+  --cache-memory-mb 512
+```
+
+Every one of those three flags is off the default path, and each is worth
+measurable throughput. They are explained below.
+
+## Why these flags
+
+### `--kv-cache-dtype bf16` — the default costs you up to 14%
+
+The server defaults to **int4** KV for this model, and says why:
+
+```
+KV cache dtype: int4 — Defaulting to int4 (memory-bandwidth-bound on
+M-series); model=mlx-community/gemma-4-26b-a4b-it-4bit not in safelist
+```
+
+The model gets quantized KV by *failing to appear on a list*, not because
+anyone measured it. Measured, quantizing KV is a straight loss here:
+
+| KV dtype | chat | codegen | code edit |
+|---|---:|---:|---:|
+| int4 (default) | 47.9 | 46.9 | 41.7 |
+| int8 | 47.6 | 46.6 | 40.1 |
+| **bf16** | **49.5** | **49.8** | **47.4** |
+
+(tok/s decode, higher is better)
+
+The reason is the architecture. Only 5 of the 30 layers use full attention
+(indices 5, 11, 17, 23, 29); the other 25 are sliding-window with a 1024
+window, so their cache is bounded. That puts the growing KV at
+
+```
+5 layers x 8 KV heads x 512 global_head_dim x 2 (K+V) x 2 B = 80 KB/token
+```
+
+against ~128 KB/token for the dense 12B. KV simply is not the bandwidth
+bottleneck on this model, so quantizing it buys nothing and the
+quantize/dequantize work is pure overhead.
+
+Note the checkpoint's own `config.json` is the authority here — the model
+card on the Hub describes 4 full-attention layers, and there are 5.
+
+### `--no-mllm` — text-only serving, for the prefix cache
+
+The checkpoint carries a `vision_config`, so auto-detection routes it to the
+MLLM lane. That lane does **not** use the prefix cache: a repeated
+502-token prompt re-prefills at 2.44s every single time, where the text lane
+answers in 0.28s once the prefix is cached.
+
+Cold prefill is the same speed in both lanes (~4.8 ms/token) — the entire
+advantage is cache reuse, which is exactly what agent traffic generates.
+Pass `--mllm` if you actually need image input.
+
+### `--cache-memory-mb 512` — buys headroom for free
+
+The default sizes the prefix cache at 20% of RAM (~3.2GB here). Shrinking it
+to 512MB cut peak Metal memory at a 16K-token prompt from 24.17 GB to
+**21.51 GB** with no measurable throughput cost at 4K or 8K. On a 32GB
+machine the allocation limit is 24.1GB, so 2.7GB of headroom is the
+difference between comfortable and one bad request from an abort.
+
+## What you get
+
+With the recommended three flags:
+
+| prompt | TTFT | decode | peak Metal |
+|---:|---:|---:|---:|
+| ~500 tok (cached prefix) | 0.28s | 47 tok/s | — |
+| 1K | 3.8s | 41.7 tok/s | 15.6 GB |
+| 4K | 15.0s | 38.0 tok/s | 16.3 GB |
+| 8K | 29.4s | 38.0 tok/s | 16.6 GB |
+
+Against `gemma-4-12b-4bit` on the same machine. Both models here run
+`--no-mllm --kv-cache-dtype bf16` *without* `--cache-memory-mb 512`, so the
+26B column differs slightly from the table above:
+
+| prompt | 26B TTFT | 12B TTFT | 26B decode | 12B decode |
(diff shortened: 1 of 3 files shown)


raullenchai:
## `pr_validate` scorecard

`PR_VALIDATE_NO_DEEPSEEK=1 PR_VALIDATE_NO_STRESS=1 python3 -m scripts.pr_validate.pr_validate 1429`

| step | status | summary |
|---|---|---|
| `fetch` | PASS | 3 files, +290/-3 LOC, blast=medium |
| `test_plan_check` | PASS | (was FAIL — boxes checked after this run; see note) |
| `supply_chain` | PASS | no hooks touched, no suspicious patterns, deps clean |
| `codex_review` | skip | gating predicate False — run separately, see below |
| `lint` | PASS | clean |
| `targeted_tests` | PASS | 10 passed |
| `full_unit` | 1 pre-existing failure | 15191 passed, 92 skipped, 6 xfailed |
| `stress_e2e_bench` | skip | opted out per CONTRIBUTING (`PR_VALIDATE_NO_STRESS=1`) |
| `diff_coverage` | skip | gating predicate False |

**The one `full_unit` failure is not from this branch.**
`tests/test_telemetry_request_wiring.py::test_nonstreaming_completion_emits_request_event`
fails identically on `origin/main` in the same venv (`assert False is None`).
This PR touches `aliases.json`, one new test file, and one new doc.

**`codex review --base main`, two rounds.** The first found a real defect and it is fixed in `0e6a6fd8`: the guide computed growing KV from `num_key_value_heads` (8), but Gemma 4's full-attention layers use `num_global_key_value_heads` — 2 on this checkpoint. The correct figure is 20 KB/token, not the 80 KB/token originally written.

That correction also inverted a comparison, so it was removed rather than restated: the dense 12B sets `num_global_key_value_heads = 1` and comes to 16 KB/token, meaning the 26B carries slightly *more* growing KV per token, not 37% less as an earlier draft claimed. The 26B's case rests on measured throughput (2.5x prefill, 2x decode), which is unaffected, and the bf16-KV recommendation gets *stronger* — at 20 KB/token a 32K context is 640MB, so KV is even further from the bandwidth bottleneck the int4 default assumes.

Second round returned no findings.

PR #1442, opened 2026-08-04
feat(models): serve LFM2.5-2.6B, whose eight quants share one repo

raullenchai:
## Why

Liquid AI released LFM2.5-2.6B — a 2.6B agentic model that is the first
checkpoint we can honestly recommend to an 8 GB Mac. Measured on an M2 Pro:
**1.58 GB weights, 2.70 GB peak, 97.8 tok/s decode**. For comparison on the
same machine, our current 16 GB picks are `lfm2.5-8b-a1b-4bit` at 5.63 GB peak
/ 121.2 tok/s and `bonsai-27b-2bit` at 10.69 GB peak / 17.8 tok/s.

It was unservable. The MLX weights only exist as `LiquidAI/LFM2.5-2.6B-MLX`,
which ships **eight complete checkpoints in sibling folders** (`4bit/`,
`5bit/`, `6bit/`, `8bit/`, `bf16/`, `mxfp4/`, `mxfp8/`, `nvfp4/`) rather than
one repo per quantization. Neither `mlx-community` nor `lmstudio-community` has
published a flat conversion, and the base repo can't be loaded at all — mlx-lm's
`Lfm2 ModelArgs` requires `block_ff_dim`, which only the `-MLX` conversions add.

## What

A `subfolder` alias field, applied at exactly the four places that reach into a
checkpoint. `hf_path` stays a bare `org/name` repo id everywhere else, because
the R2 mirror catalog, `model_sizes` and telemetry redaction all key on it.

Only the loader site was predictable. The other three surfaced by running
`rapid-mlx serve lfm2.5-2.6b-4bit` — the unit tests were green while the
command was still unusable:

| Site | Symptom |
|---|---|
| `load_model_with_fallback` | mlx-lm has no subfolder param |
| `estimate_repo_size_bytes` | "Estimated size: 18.7 GiB" + kernel-panic warning, for a 1.6 GB model |
| `is_repo_cached` | warm cache read as absent → confirm prompt on every serve |
| `read_cached_model_metadata` | no root `config.json` → serve died with "Could not materialize the checkpoint config before selecting the serving lane" |

`allow_patterns` keeps a 1.6 GB pull from fetching the repo's other seven quants.

**Ambiguity guard.** The loader recovers the subfolder by reverse-lookup from
the resolved `hf_path`, so two aliases on one repo declaring *different*
subfolders are indistinguishable to it. Every other profile field is safe under
"first alias wins"; this one is not — 4-bit and 8-bit legitimately disagree.
Rejected at registry load rather than silently serving the wrong weights under
the right name. Lifting the restriction means threading the alias down to
`load_model_with_fallback`, which is deliberate work, not a JSON line.

**Unrelated-looking but same evidence:** `lfm2.5-8b-a1b-4bit` had
`reasoning_parser: null`. It emits `<think>` and does not close it within 256
tokens, so the raw tag was reaching users — on our 16 GB "fast" recommendation.
Fixed here because the measurement that found it is the measurement in this PR.
`lfm2.5-1b-4bit` and `lfm2-24b-a2b-4bit` reference `<think>` in their templates
too but are **not** touched — I have no measurement for them.

## Verification

Mac mini (M2 Pro, mlx-lm 0.31.3, mlx 0.31.2), `rapid-mlx serve lfm2.5-2.6b-4bit`:

- `Download size: ~1.5 GiB` (was 18.7 GiB); `is_repo_cached: True`; flat-repo
  control `mlx-community/LFM2.5-8B-A1B-MLX-4bit` still reports 4.5 GiB
- boots clean: `tool=lfm, reasoning=qwen3`, no gate prompt, no memory warning
- `/v1/chat/completions`: `reasoning_content` holds the reasoning, `content` is
  `391`, no `<think>` leakage
- tool call: `finish_reason=tool_calls`, `get_weather({"city": "Tokyo", "unit": "celsius"})`

33 new mlx-free tests. `test_alias_subfolder.py` **and** `test_aliases_contract.py`
added to the Linux CI list — the contract test had no CI coverage on either leg.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,6 +169,8 @@ jobs:
             tests/test_telemetry_consent.py \
             tests/test_community_bench_upload.py \
             tests/test_music_engine.py \
+            tests/test_aliases_contract.py \
+            tests/test_alias_subfolder.py \
             -v --tb=short \
             -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache" \
             --cov=vllm_mlx \
--- a/tests/test_alias_subfolder.py
+++ b/tests/test_alias_subfolder.py
@@ -1,0 +1,782 @@
+# SPDX-License-Identifier: Apache-2.0
+"""``subfolder``: one repo, one folder per quantization.
+
+Liquid AI ships ``LiquidAI/LFM2.5-2.6B-MLX`` as eight complete MLX
+checkpoints in sibling directories (``4bit/``, ``5bit/``, ``6bit/``,
+``8bit/``, ``bf16/``, ``mxfp4/``, ``mxfp8/``, ``nvfp4/``) instead of one
+repo per quant, and there is no flat conversion on the Hub. ``mlx_lm.load``
+has no subfolder parameter, so the repo id has to become a concrete
+directory somewhere. These tests pin *where* — and, just as importantly,
+pin that it does NOT happen anywhere else: the download gate, the R2
+mirror catalog, ``model_sizes`` and telemetry all key on the bare repo id.
+
+mlx-free by construction (no weights, no ``mlx`` import) so it runs on the
+Linux CI leg as well as Apple Silicon.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+from pathlib import Path
+
+import pytest
+
+from vllm_mlx.model_aliases import (
+    _coerce,
+    resolve_model,
+    resolve_profile,
+    resolve_subfolder,
+    subfolder_allow_patterns,
+)
+from vllm_mlx.utils.tokenizer import _resolve_subfolder_checkpoint
+
+ALIAS = "lfm2.5-2.6b-4bit"
+REPO = "LiquidAI/LFM2.5-2.6B-MLX"
+
+
+# --------------------------------------------------------------------------
+# The shipped alias
+# --------------------------------------------------------------------------
+
+
+def test_alias_declares_the_4bit_subfolder():
+    profile = resolve_profile(ALIAS)
+    assert profile is not None, f"{ALIAS} missing from aliases.json"
+    assert profile.hf_path == REPO
+    assert profile.subfolder == "4bit"
+
+
+def test_hf_path_stays_a_bare_repo_id():
+    """The subfolder must NOT be folded into ``hf_path``.
+
+    ``resolve_model`` feeds the download gate, ``model_sizes``, the mirror
+    catalog and telemetry redaction — all of which treat the value as an
+    ``org/name`` repo id. A three-segment path would 404 against the HF
+    API and silently miss the mirror.
+    """
+    resolved = resolve_model(ALIAS)
+    assert resolved == REPO
+    assert resolved.count("/") == 1
+
+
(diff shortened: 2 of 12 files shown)


raullenchai:
## Review disposition

Five `pr_validate` rounds. Everything actionable is fixed; two items are deliberately not, with reasons.

**Fixed across rounds**
- `_resolve_subfolder_checkpoint` fell back to the bare repo id on failure — that sends mlx-lm to its own *unfiltered* download, so the user waits out ~20 GB and fails anyway. Now raises, with an explicit `local_files_only` retry first so a warm offline cache still loads.
- `subfolder` is spliced into an HF glob, so `4bit*` / `[48]bit` / `**` were legal paths that widened the fetch. Glob metacharacters and drive-qualified paths now rejected at registry load.
- Registry validation ran *after* publishing to the module globals — one caught exception and every later `_load()` took the memoized fast path with the registry that had just been rejected. Fail-closed now, pinned by a test that calls `_load()` three times.
- Registry errors are propagated rather than swallowed into a bare-repo-id fallback.
- The alias spelling (`lfm2.5-2.6b-4bit`) reached `snapshot_download` unresolved on the programmatic `server.load_model` path; normalized to the repo id.
- `_check_disk_space` lost the prefix when the best-effort cache probe threw, sizing all eight quants and refusing machines with room for the one being fetched.
- Checkpoint completeness is now verified on **both** the online and cached paths — a reorganized repo can ship `config.json` with no shards via a perfectly successful download.
- `model_sizes.json` entry + `subfolder` registered in `ALLOWED_STR_FIELDS` (the closed-set review gate) — both caught by `full_unit`, not by review.
- A parser regression test for the `lfm2.5-8b-a1b-4bit` change, using the unclosed-`<think>` output actually observed.

**Not changed — `pull <repo-id>` narrows to the served folder**

Flagged as blocking twice. Keeping the narrowing, now announced on stdout rather than silent.

Every other consumer already reaches inside on the bare repo id: the download gate quotes the subfolder's size, `is_repo_cached` probes the subfolder, the loader opens the subfolder. Un-narrowing only `pull` makes the gate's "1.5 GiB" quote a lie ahead of a 20 GB transfer, and leaves seven checkpoints on disk that nothing in rapid-mlx can serve. The legitimate half of the objection was that it happened silently; that is fixed. Happy to gate on `_original_alias` instead if a reviewer prefers the generic contract — it is a two-line change.

**Not changed — `ModelProfile` positional-argument shift**

Reported three times. It is a false positive: the dataclass is `@dataclass(frozen=True, kw_only=True)`, so positional construction has never been possible. Verified rather than assumed:

```
>>> ModelProfile('repo', 'text')
TypeError: ModelProfile.__init__() takes 1 positional argument but 3 were given
```

**Pre-existing failures**

`full_unit` reports 5 failures. All 5 reproduce on base `main` (`test_metrics_route` ×3, `test_telemetry_request_wiring`) or pass in isolation and only fail under full-suite load (`test_signal_observability`). None are from this change.

**Live re-verification** after the final edit, Mac mini M2 Pro: serve boots from the `4bit` subfolder, `content` is `391`, reasoning separated, no tag leakage.

PR #1443, opened 2026-08-04
feat(app): give 8 GB Macs a recommendation instead of a rejection

raullenchai:
## Why

Stacked on #1442 (which makes `lfm2.5-2.6b-4bit` servable at all).

An 8 GB Mac had no recommendation. It clamped up to the 16 GB tier, and
`isRecommendedPick`'s floor guard then correctly declined to exempt those picks
from the `.tooBig` gate — so `runLaunchAutoStart` rejected the only thing it had
been offered and fell through to `SafeDefaultFallback`. The user's first run was
a model the app had just told them not to run. The old docstring claimed "that
tier's fast alternative does fit 8 GB" about `lfm2.5-8b-a1b-4bit`; measured, it
peaks at 5.63 GB, which does not leave room for macOS on an 8 GB machine.

## What

A `floorGB: 8` tier whose single pick is `lfm2.5-2.6b-4bit`. 22 of its 30 layers
are short convolutions and only 8 are GQA, so KV costs ~16 KB/token — a 32 K
conversation adds ~0.5 GB on top of 1.6 GB of weights.

The card carries **"Not for coding"** instead of a capability %. That is Liquid's
own published guidance ("not recommended for agentic coding and knowledge-heavy
tasks"). It is genuinely strong at tool use and instruction following — it beats
Qwen3.5-9B on ToolSandbox, Multi-IF and IFStruct — but our users drive coding
agents, and a bare percentage would invite exactly the use the publisher warns
against. `capabilityPct` is never rendered for a caveat pick; the stored 62 only
satisfies the monotonic-by-RAM invariant and is deliberately the same band as
the family's other caveat pick, not an independent score.

## The 16 GB rows were wrong

Re-measuring all three small picks together on one M2 Pro (mlx-lm 0.31.3, peak
resident, 2 K-token prefill) turned up a drift in what the GB column meant:

| alias | shown | measured peak | decode |
|---|---|---|---|
| lfm2.5-2.6b-4bit | — | **2.70 GB** | 97.8 tok/s |
| lfm2.5-8b-a1b-4bit | 4.8 GB / 117.3 | **5.63 GB** | 121.2 tok/s |
| bonsai-27b-2bit | 7.6 GB / 17.1 | **10.69 GB** | 17.8 tok/s |

bonsai's 7.6 was its weight bytes — the field is documented as active memory, and
it understated the real footprint by 3.1 GB on the exact tier where it is the
recommendation. The tok/s figures held up (17.1 → 17.8, 117.3 → 121.2), so the
speed column was sound and only the memory column had drifted.

The column is display-only (`pickStatsLine` / `ModelPickerBar` tagline), so this
corrects what users read, not what the app lets them run.

## Verification

- `swift apps/rapid-mac/scripts/verify-recommendation-tiers.swift` → ALL PASS,
  with new checks: 8 GB gets its own tier, sub-8 GB clamps down to it, 15 GB
  rounds down rather than up, the tier's pick survives auto-start, bigger models
  are still rejected there, and the monotonic-capability chain now starts at 8.
- `swift build` clean.
- `RAMBucketedDefaultTests` / `ModelPickerBarTests` updated to match.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Server/RAMBucketedDefault.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/RAMBucketedDefault.swift
@@ -21,9 +21,8 @@ import Foundation
 ///
 /// A machine with N GB gets the tier whose floor is the largest value
 /// ≤ N — so a 20 GB Mac gets the 18 GB pick (which fits), NOT the 24 GB
-/// pick (which would be borderline). Sub-16 GB Macs clamp to the 16 GB
-/// tier; the app's per-row fit warning still flags anything too big, and
-/// that tier's fast alternative does fit 8 GB.
+/// pick (which would be borderline). Sub-8 GB Macs clamp to the 8 GB
+/// tier; the app's per-row fit warning still flags anything too big.
 ///
 /// A tier only carries a **fast** alt when it beats the smart pick on
 /// speed by a margin worth a second card. The 24 & 32 GB smart picks are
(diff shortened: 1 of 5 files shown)


raullenchai:
## Review disposition

**Fixed**
- **Real bug caught by codex, would have shipped a failing test.** I blind-replaced `smallest[0].alias` in `RAMBucketedDefaultTests` — but that variable is named `smallest` while querying **16 GB**, whose primary is still `bonsai-27b-2bit`. The 8 GB assertions now live in their own block and 16 GB is restored (and the variable renamed `tier16`, which is what it always meant).
- `lfm2FastPick` doc said ~117 tok/s after the measurement moved it to 121.2.
- The contract script's heading still called it the "sub-16 OOM hole"; the floor is 8 now.
- Added the pathological-RAM cases (0 / negative) to the **runnable** contract script — `RAMBucketedDefaultTests` asserts them but is excluded from `Package.swift`, so they were only checked somewhere that never executes.

**Not changed — the contract script is a surrogate**

Flagged blocking: `verify-recommendation-tiers.swift` re-declares `rejectsForAutoStart` rather than calling production `runLaunchAutoStart`, so it stays green if production drifts.

That is true, and it is the script's documented design, not something this change introduced — the header says so explicitly, and the reason is that `Tests/RapidTests` is excluded from `Package.swift` (`import Testing` is not resolvable from the command line in this toolchain), which is why a dependency-free surrogate exists at all. Wiring the real auto-start path into CI is a test-infrastructure change that should not ride along with a tier addition. The production table itself IS asserted by `RAMBucketedDefaultTests` — that suite just isn't executed today.

**Pre-existing failures**

`full_unit` reports 5. This PR touches only Swift files, so none can be from it; all 5 reproduce on base `main` or pass in isolation and fail only under full-suite load. Same set as #1442.

**Verified**: `swift build` clean, contract script ALL PASS (now including 8 GB tier membership, sub-8 clamp-down, 15 GB rounding down rather than up, auto-start survival, and the monotonic-capability chain starting at 8).

PR #1447, opened 2026-08-04
feat(install): make the curl banner recommend what the app recommends

raullenchai:
## Why

We ship two answers to "what should this Mac run", and they disagreed at five of six tiers:

| RAM | `install.sh` (curl banner) | Desktop app |
|---|---|---|
| 8 GB | `qwen3.5-4b-4bit` | **`lfm2.5-2.6b-4bit`** |
| 16 GB | `qwen3.5-4b-4bit` | **`bonsai-27b-2bit`** |
| 24 GB | `gpt-oss-20b-mxfp4-q8` | **`gemma-4-26b-4bit`** |
| 32 GB | `gpt-oss-20b-mxfp4-q8` | **`qwen3.6-35b-4bit`** |
| 48–64 GB | `qwen3.6-35b-8bit` | `qwen3.6-35b-8bit` ✅ |
| 96 GB+ | `gpt-oss-120b-mxfp4-q8` | **`qwen3.5-122b-mxfp4`** |

Install with curl, open the app, get told to run a different model on the same Mac. curl is the canonical entry point in the README, so it is the one most people hit first.

## Which table wins, and why

`RAMBucketedDefault.tiers`. It is the one being maintained — measured footprints, a capability column, a monotonic-by-RAM invariant, and a runnable contract script that CI can execute. `install.sh`'s picks were frozen at PR #1053 and two of them have aged badly against our own later evidence:

- **24 GB → `gpt-oss-20b-mxfp4-q8`** is the model we measured as unable to drive codex's `apply_patch`.
- **The `8-23 GB` bucket** lumped an 8 GB Air and a 16 GB Air together, handing the 16 GB machine a 4B when the app gives it an 86 %-capability pick.

## `RECOMMENDED_FLAGS`

New, and load-bearing. `gemma-4-26b-4bit` at 24 GB needs `--no-mllm --kv-cache-dtype bf16 --cache-memory-mb 512` — that is how the app launches it on that tier. Printing the bare `serve` line would hand a 24 GB Mac a command that does not fit in it.

Branching verified across 13 RAM sizes; every boundary lands on the same alias the app picks.

## Why it drifted, and what stops it now

Nothing compared the two files. One is shell, one is Swift, neither imports the other, so there was no mechanism that could have noticed.

`tests/test_ram_tier_recommendations_agree.py` parses both tables out of their source files and compares alias **and** launch flags at 14 RAM sizes, plus asserts every recommended alias exists in `aliases.json` (a typo turns the install banner into a 404 on first run). It includes a `test_both_tables_parse` guard so a regex that silently matches nothing can't make the rest vacuously green.

Mutation-checked rather than assumed: swapping install.sh's 24 GB pick back to `gpt-oss-20b-mxfp4-q8` fails 2 of the 6 tests with the exact mismatch printed. Added to the Linux CI list.

## Follow-up (not in this PR)

`/docs/hardware-tiers` on the website documents the old `install.sh` table. It is accurate today for what it claims to document, and goes stale the moment this merges — the site update is being prepared alongside, with fresh boot-time and peak-RSS measurements for the new picks on the same M3 Ultra the existing column came from.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,7 @@ jobs:
             tests/test_music_engine.py \
             tests/test_aliases_contract.py \
             tests/test_alias_subfolder.py \
+            tests/test_ram_tier_recommendations_agree.py \
             -v --tb=short \
             -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache" \
             --cov=vllm_mlx \
--- a/install.sh
+++ b/install.sh
@@ -84,11 +84,33 @@ fi
 
 # ── 2. Detect RAM → recommend model ──────────────────────────────────────────
 
+# These tiers MIRROR ``RAMBucketedDefault.tiers`` in the desktop app
+# (apps/rapid-mac/Sources/Rapid/Server/RAMBucketedDefault.swift), which is
+# the curated table with measured footprints, a capability column and a
+# monotonic-by-RAM invariant guarded by
+# apps/rapid-mac/scripts/verify-recommendation-tiers.swift.
+#
+# They used to disagree — six tiers, one match. Somebody who ran
+# ``curl … | bash`` and then opened the app was told to run two different
+# models on the same Mac, and curl is the canonical entry point in the
+# README. One table, two front doors.
+#
+# The app also surfaces a "fast alternative" per tier; this banner shows a
+# single command, so it takes the app's PRIMARY (smart) pick only.
+#
+# RECOMMENDED_FLAGS carries the tier's launch flags. It is not cosmetic:
+# ``gemma-4-26b-4bit`` at 24 GB needs its vision tower dropped and its KV
+# budget capped, and printing the bare ``serve`` line would hand a 24 GB
+# Mac a command that does not fit in it.
 RAM_GB=$(sysctl -n hw.memsize 2>/dev/null | awk '{printf "%d", $1/1073741824}')
-if   [ "$RAM_GB" -ge 96 ]; then RECOMMENDED_MODEL="gpt-oss-120b-mxfp4-q8"; RAM_TIER="96+ GB"
-elif [ "$RAM_GB" -ge 48 ]; then RECOMMENDED_MODEL="qwen3.6-35b-8bit";      RAM_TIER="48-95 GB"
-elif [ "$RAM_GB" -ge 24 ]; then RECOMMENDED_MODEL="gpt-oss-20b-mxfp4-q8";  RAM_TIER="24-47 GB"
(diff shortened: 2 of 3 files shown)


raullenchai:
## Review disposition

Both blocking findings were about **my test being weaker than its name claimed**. Both fixed, both mutation-checked.

1. **Flags test only checked the assignment.** Delete `${RECOMMENDED_FLAGS}` from the echo and the variable is still set, so the test stayed green while the banner went back to printing a command that OOMs a 24 GB Mac. Now `_render_banner()` executes install.sh's own tier block plus its quick-start line under `sh` and asserts on what a user actually sees — plus a control that the flags do *not* leak onto the tiers that don't want them.

2. **"every RAM size" sampled a fixed list.** Moving the shell's 24 GB floor to 23 passed because nothing probed 23. Boundaries are now derived from *both* parsed tables and each is probed at floor−1 / floor / floor+1.

Strengthening #2 immediately found a third thing — a flaw in my own helper, not the code. Below the lowest floor both tables clamp (`RAMBucketedDefault.tier` starts at `tiers[0]` and only moves up; install.sh's `else` arm catches everything), but my `pick()` returned `None`, so a 4 GB reading — what a failed RAM probe reports — looked like a mismatch. The helper now models the clamp.

Mutation-checked rather than assumed:

| injected drift | result |
|---|---|
| 24 GB floor → 23 GB | `test_same_alias_at_every_ram_size` fails |
| `${RECOMMENDED_FLAGS}` removed from the echo | `test_the_banner_actually_prints_the_launch_flags` fails |
| 24 GB pick → `gpt-oss-20b-mxfp4-q8` | 2 tests fail with the exact mismatch printed |

**NIT 3 (`/docs/hardware-tiers` goes stale on merge)** — agreed, and being handled in the same batch rather than deferred. All six picks were re-measured on the same M3 Ultra the page's existing column came from, so the table stays internally comparable:

| pick | boot to `/v1/models` | peak RSS |
|---|---:|---:|
| `lfm2.5-2.6b-4bit` | 4.8 s | 2.0 GB |
| `lfm2.5-8b-a1b-4bit` | 2.7 s | 5.3 GB |
| `bonsai-27b-2bit` | 4.8 s | 8.4 GB |
| `gemma-4-26b-4bit` | 4.8 s | 15.6 GB |
| `qwen3.6-35b-4bit` | 4.8 s | 20.4 GB |
| `qwen3.6-35b-8bit` | 10.6 s | 37.7 GB |

All six answered `/v1/models` and completed a real generation. `qwen3.5-122b-mxfp4` (96 GB+) is not measured — a 65 GB pull on the shared Studio that also hosts the release runner wasn't worth the disk; that cell stays empty rather than estimated.

**Pre-existing failures**: `full_unit` reports the same 5 as #1442/#1443 — all reproduce on base `main` or only fail under full-suite load.

PR #1449, opened 2026-08-04
fix(cache): replay tokens across prompt boundaries

raullenchai:
## Summary
- retain an 8-token replay window behind stable message boundaries
- apply the same conservative window to third-party template fallbacks
- preserve exact transient-priming LCP behavior

## Root cause
Tokenization is not compositional at text/message boundaries. Once the next assistant or tool segment is appended, the prior prompt's final few byte-BPE tokens can change. DeepSeek V4 and other non-trimmable caches cannot trim an optimistic snapshot, so a mismatch of only a few boundary tokens can discard an otherwise 40K+ reusable prefix.

Eight tokens covers the observed three-token Codex boundary drift with negligible re-prefill cost.

## Validation
- 106 focused prefix-boundary, prompt snapshot, hybrid-cache, and prefix-cache tests passed
- Ruff format/check passed
- `git diff --check` passed

## Review focus
Only blockers/major correctness, compatibility, cache-reuse regressions, or high-value test gaps; no style nits.

### DIFF
--- a/tests/test_prefix_boundary_path_parity.py
+++ b/tests/test_prefix_boundary_path_parity.py
@@ -263,7 +263,7 @@ def test_prefix_boundary_prefers_latest_stable_message_boundary(monkeypatch):
     ]
     stable = render(messages, add_generation_prompt=False)
 
-    assert engine._compute_prefix_boundary(messages) == len(stable)
+    assert engine._compute_prefix_boundary(messages) == len(stable) - 8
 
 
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex review result: **APPROVE after one important fix**.

The first review found that transient priming boundaries bypassed the replay window. That path now applies the same clamp, with explicit transient and short-prefix tests. A focused re-review found no remaining important issues. Review scope excluded nits and covered correctness, compatibility, material cache-hit regressions, and high-value test gaps.

PR #1451, opened 2026-08-04
fix(cache): persist deepest prefix first on shutdown

raullenchai:
## Summary
- sort prefix-cache entries longest-first only for deadline-aware shutdown saves
- keep historical LRU order for explicit/full saves
- verify a one-entry partial commit preserves the deepest reusable prefix

## Root cause
Shutdown persistence has a short deadline and may commit only one entry. LRU iteration can begin with a one-token bootstrap entry; writing it consumes the guaranteed slot and its fixed fsync overhead poisons the observed-throughput estimate, causing the real 100K+ frontier to be skipped.

## Validation
- 93 persistence/runtime/deferred-load tests passed
- Ruff format/check passed
- `git diff --check` passed

## Review focus
Only blockers/major correctness, persistence compatibility, partial-commit/index consistency, and high-value test gaps; no style nits.

### DIFF
--- a/tests/test_prefix_cache_persistence.py
+++ b/tests/test_prefix_cache_persistence.py
@@ -1373,6 +1373,62 @@ def test_save_to_disk_partial_commit_on_abort(tmp_path):
     assert entry.tokens == tuple(range(11))
 
 
+def test_shutdown_partial_commit_prioritizes_longest_prefix(tmp_path):
+    """A tiny bootstrap entry must not consume the only shutdown slot."""
+    cache_dir = tmp_path / "snap"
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex review result: **APPROVE after two important fixes**.

Initial review found (1) serialized priority order would overwrite restored LRU order and (2) rejecting the largest predicted write would prevent smaller candidates from being considered. The implementation now restores logical LRU order in the committed index and continues scanning smaller candidates. Save/load/evict and predicted-duration regressions were added; focused re-review found no remaining important issues.

PR #1452, opened 2026-08-04
fix(deepseek): prevent reopening disabled thinking

raullenchai:
## Summary
- suppress the DeepSeek V4 `<think>` opener at decode time when Responses explicitly resolves thinking to false
- forward the stateless structural-token processor through both streaming and non-streaming engine paths
- leave thinking-enabled, unspecified-thinking, and non-DeepSeek requests unchanged

## Root cause
The 0731 prompt can explicitly close/disable thinking, but the model may later sample `<think>` again during a long agent turn. That re-enters the reasoning parser state and can leak reasoning or corrupt the expected content/tool boundary. Prompt shaping alone cannot enforce the invariant at decode time.

## Validation
- 144 Responses, reasoning-budget, and scheduler processor-isolation tests passed
- Ruff format/check passed
- `git diff --check` passed

## Review focus
Only blockers/major correctness, cross-model isolation, logits-processor composition, streaming parity, or high-value test gaps; no style nits.

### DIFF
--- a/tests/test_grammar_scheduler_realign_558.py
+++ b/tests/test_grammar_scheduler_realign_558.py
@@ -42,6 +42,7 @@ def _make_scheduler_stub():
         uid_to_request_processors={},
         _uids_with_grammar=set(),
         _uids_with_reasoning_budget={},
+        _uids_with_suppressed_tokens={},
         _known_stateful_processors=set(),
         _stateful_processor_objs={},
         _stateful_tombstones=set(),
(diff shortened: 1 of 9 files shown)


raullenchai:
Codex review result: **APPROVE after blocker/major fixes**.

The focused review caught and drove fixes for positional processor leakage, per-token full-vocab mask rebuilding, implicit/server false activation, multi-model registry isolation, `reasoning_effort=none` parity, native `enable_thinking` precedence, and registry hot-reload mismatch handling. Final coverage includes both explicit no-thinking forms across DeepSeek/Qwen registry entries and streaming/non-streaming paths, plus mixed-batch cleanup. Final focused review found no remaining important issues.

PR #1454, opened 2026-08-04
docs(readme): the front door was still on the pre-0.12 tier map

raullenchai:
## Why

#1447 reconciled `install.sh` with the desktop app's picker. The README is the **third** copy of that table and was never updated — so the repo's front door on GitHub gave a different answer than the installer, including two picks we have since moved off:

- **24–47 GB → `gpt-oss-20b-mxfp4-q8`** is the model we measured as unable to drive codex's `apply_patch`.
- **The `8–23 GB` bucket** handed a 16 GB Air the same 4B it gave an 8 GB one.

## What

Six tiers, matching `install.sh` and `RAMBucketedDefault.tiers`, in **both** places the README states them — the install prose in the quick start and the Choose Your Model table.

| RAM | Recommended | Peak RSS |
|---|---|---:|
| 8–15 GB | `lfm2.5-2.6b-4bit` | 2.0 GB |
| 16–23 GB | `bonsai-27b-2bit` | 8.4 GB |
| 24–31 GB | `gemma-4-26b-4bit` | 15.6 GB |
| 32–63 GB | `qwen3.6-35b-4bit` | 20.4 GB |
| 64–95 GB | `qwen3.6-35b-8bit` | 37.7 GB |
| 96 GB+ | `qwen3.5-122b-mxfp4` | — |

Peak RSS is measured **through `rapid-mlx serve`**, and the README now says so — the engine quantizes the KV cache to int4 by default, which is worth about two gigabytes on a 27B, so a bare `mlx_lm` probe reads higher (that mistake is what #1453 corrected).

The 24–31 GB row carries its launch flags **in the copy-paste command**. A README reader pastes that line verbatim; without `--no-mllm` and the KV caps it does not fit the Mac it is being recommended to.

Alias counts `214 → 215` (166 text + 41 audio + 8 video), matching the released registry the website is checked against.

## The actual fix

Three surfaces stating one table is how this happened twice. `tests/test_ram_tier_recommendations_agree.py` now parses the README table as well, so one test guards all three.

It parses the tier **rows** rather than scanning the file, because the README also mentions `qwen3.5-4b-4bit` as the `rapid-mlx chat` default — correct, and not a tier recommendation. Those two mentions are deliberately left alone.

Mutation-checked rather than assumed:

| injected drift | result |
|---|---|
| a README row reverted to the old pick | `test_readme_recommends_the_same_models` fails |
| gemma flags dropped from the README command | `test_readme_carries_the_launch_flags` fails |

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ or Homebrew — prebuilt bottle straight from homebrew-core:
 brew install rapid-mlx
 ```
 
-Both land the same `rapid-mlx` CLI. The curl installer additionally installs Python 3.10+ if missing, creates an isolated venv at `~/.rapid-mlx/`, symlinks the `rapid-mlx` CLI into `~/.local/bin/`, and prints a serve command sized to your Mac (8–23 GB → `qwen3.5-4b-4bit`; 24–47 GB → `gpt-oss-20b-mxfp4-q8`; 48–95 GB → `qwen3.6-35b-8bit`; 96 GB+ → `gpt-oss-120b-mxfp4-q8`).
+Both land the same `rapid-mlx` CLI. The curl installer additionally installs Python 3.10+ if missing, creates an isolated venv at `~/.rapid-mlx/`, symlinks the `rapid-mlx` CLI into `~/.local/bin/`, and prints a serve command sized to your Mac (8–15 GB → `lfm2.5-2.6b-4bit`; 16–23 GB → `bonsai-27b-2bit`; 24–31 GB → `gemma-4-26b-4bit`; 32–63 GB → `qwen3.6-35b-4bit`; 64–95 GB → `qwen3.6-35b-8bit`; 96 GB+ → `qwen3.5-122b-mxfp4`).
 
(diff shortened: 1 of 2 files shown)


raullenchai:
## Review disposition

All three blocking findings were the same defect in different places: the test asserted less than its name claimed. Fixed and mutation-checked.

1. **The quick-start prose map was unguarded.** The README states the tier map twice; I only parsed the table, so the sentence a reader hits first could drift alone. It now has its own parse and its own assertion against the table.
2. **Set comparison ignored position.** Two tiers swapping picks, or a bucket boundary moving, would have passed a test reporting that the tables agree. It now compares ordered `(floor_gb, alias)` mappings.
3. **Flags were matched anywhere in the row.** A flag sitting in the prose column satisfied the search while the copy-paste command stayed incomplete. Flags are now parsed out of the One-shot code span specifically and compared exactly — same flags, same order.

**Writing the stricter version immediately found a real miss.** The top row is written `**96 GB+**`, not as a range, so the range-only pattern silently skipped it — the 96 GB tier was never being checked at all. That is exactly the failure mode that makes a parser-based test worthless, so there is now an assertion that the parsed row count is non-empty and that each row's command runs the alias the row recommends.

Mutation-checked rather than assumed — four injected drifts, each caught:

| injected drift | caught by |
|---|---|
| two tiers swap picks | table-vs-app, prose-vs-table, flags |
| bucket boundary 16–23 → 17–23 | table-vs-app, prose-vs-table |
| prose drifts from the table | prose-vs-table |
| a flag moved out of the one-shot command | table-vs-app, prose-vs-table |

**Pre-existing failures**: `full_unit` reports the same 5 as #1442/#1443/#1447. This PR touches README.md and one test file, so none can be from it; all 5 reproduce on base `main` or only fail under full-suite load.

PR #1466, opened 2026-08-05
fix(cache): persist long-context boundaries on shutdown

raullenchai:
## Summary

Raise the first-entry persistence throughput bootstrap from 150 MB/s to 500 MB/s so reusable long-context boundaries can seed the measured-throughput estimator within the default 3.5-second shutdown budget.

## Root cause

The conservative 150 MB/s estimate predicted roughly 16 seconds for a real DeepSeek-V4 120K cache set and skipped every useful boundary, even though the same 2,397 MB set persisted in 0.9 seconds. Restarts therefore lost deep boundaries and paid cold prefill again.

The 500 MB/s bootstrap remains more than 5x below the measured path. Entries above roughly 1.55 GB are still rejected before an uninterruptible write can cross the 3.1-second safe window.

## Validation

- ruff: clean
- focused persistence tests: 84 passed
- full suite: 15,550 passed; 135 skipped; 18 deselected; 6 xfailed; 1 xpassed
- 12 live-server failures were expected with no service on 127.0.0.1:8000 (10 Hermes integration, 2 event-loop)
- real DeepSeek-V4 120K restart test under the default 3.5-second budget:
  - saved 3/3 entries, 2,397 MB, in 1.8 seconds
  - fresh process loaded all 3 entries
  - cache hit 120,004 / 120,014 tokens
  - restart TTFT 2.05 seconds
  - exact output CACHE_RESTART_OK, no reasoning leak

### DIFF
--- a/tests/test_prefix_cache_persistence.py
+++ b/tests/test_prefix_cache_persistence.py
@@ -1497,7 +1497,7 @@ def test_deadline_save_skips_oversized_prefix_and_tries_smaller(tmp_path):
     cache.store([1], make_kvcache(num_tokens=1))
     cache.store(list(range(100)), make_kvcache(num_tokens=100, fill=2.0))
     entry_sizes = sorted(entry.memory_bytes for entry in cache._entries.values())
-    threshold = sum(entry_sizes) / 2 / (150 * 1024 * 1024)
+    threshold = sum(entry_sizes) / 2 / (500 * 1024 * 1024)
 
     assert (
         cache.save_to_disk(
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex review found a material cross-hardware safety regression: the 500 MB/s bootstrap is validated on this M3 Ultra/internal SSD, but it is not a defensible lower bound for every supported filesystem. Although the real 120K test saved 2,397 MB in 1.8s and restarted at 2.05s TTFT, merging this could let an uninterruptible ~1.5GB write cross the shutdown deadline on slower storage. Closing rather than overriding the blocker. Follow-up will keep the conservative shutdown gate and move deep persistence before SIGTERM via an idle/checkpoint path (or make writes interruptible).

PR #1468, opened 2026-08-05
fix(cache): checkpoint long-context boundaries while idle

raullenchai:
## Summary

Persist one latest reusable prefix-cache boundary during idle server windows, before shutdown begins. Track cache content and checkpoint generations so the task only writes dirty state, and retain the existing conservative deadline-bound shutdown flush unchanged.

## Why

DeepSeek-V4 at 120K produced a ~2.4GB in-memory cache that writes in 0.9–1.8s on the test M3 Ultra, but the safe 150MB/s shutdown predictor correctly refuses to assume that speed on every supported filesystem. The result was a restart paying cold prefill even though the server had been idle before termination. Raising the global throughput assumption was rejected in #1466 because it weakened slow-storage safety.

This moves the useful work out of the SIGTERM critical path: once every 300s (configurable, zero disables), an idle engine atomically saves exactly one persistence-priority boundary. A missing radix is rebuilt from that committed entry on startup. Legacy engines that cannot enforce the one-entry bound and non-memory-aware caches are skipped.

## Validation

- ruff: clean
- focused cache/lifecycle tests: 96 passed
- real DeepSeek-V4 120K fault injection:
  - idle checkpoint committed one 120,004-token message boundary in under 3s
  - process was then killed with SIGKILL, bypassing shutdown save entirely
  - fresh process loaded the one-entry snapshot
  - cache hit 120,004 / 120,014 tokens
  - restart TTFT 2.06s
  - exact output CACHE_RESTART_OK, no reasoning leak

## Safety

- default shutdown bootstrap remains 150MB/s
- idle checkpoint is single-entry and only runs with zero running/waiting requests
- shutdown drains an in-flight checkpoint before final save/engine teardown
- generation tracking covers stores, explicit removes, evictions, clears, loads, and commits
- partial snapshot metadata reports only committed entries; live multi-entry radix is not persisted beside a one-entry snapshot

### DIFF
--- a/tests/test_idle_prefix_cache_checkpoint.py
+++ b/tests/test_idle_prefix_cache_checkpoint.py
@@ -1,0 +1,159 @@
+"""Runtime idle-checkpoint regression tests."""
+
+from __future__ import annotations
+
+import asyncio
+import threading
+
+from vllm_mlx import server
+from vllm_mlx.runtime import cache as runtime_cache
+
+
+def test_runtime_checkpoint_limits_snapshot_to_one_candidate(monkeypatch, tmp_path):
+    predicates = []
+
+    class Engine:
+        def save_cache_to_disk(self, _path, should_abort=None):
+            predicates.append(should_abort)
+            assert should_abort(99.0) is False
+            assert should_abort(0.0) is True
+            return True
+
+    class Config:
+        engine = Engine()
+
+    monkeypatch.setattr(runtime_cache, "get_config", lambda: Config())
+    monkeypatch.setattr(runtime_cache, "get_cache_dir", lambda: str(tmp_path))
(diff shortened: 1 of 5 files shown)


raullenchai:
Adversarial review found four material blockers: stale snapshot resurrection after empty-cache mutations, an admission TOCTOU between idle observation and save, unbounded shutdown drain if SIGTERM lands during an uninterruptible entry write, and false-clean state after partial/corrupt loads. The 120K SIGKILL test proves the mechanism can work, but the lifecycle guarantees are not yet strong enough to merge. Closing rather than landing experimental production code. The follow-up must put the idle check on the scheduler step thread and make entry persistence sharded/interruptible (or otherwise prove a bounded writer); the existing conservative shutdown path remains unchanged on main.

PR #1470, opened 2026-08-05
fix(rapid-mac): six desktop UX defects — sidebar ordering, onboarding scope, catalog cache, stale-manifest copy

xiaoxiunique:
## What does this PR do?

Six independent fixes found by using the desktop app as a new user would, then
reading the code behind each symptom. All confined to `apps/rapid-mac`.

| | symptom | cause |
|---|---|---|
| 1 | opening a conversation reordered the sidebar | `persistActive()` stamps `updatedAt` for all four callers, including the two that only navigate |
| 2 | everything filed under "Older", including a chat 5s old | the heading was hard-coded; no bucketing existed |
| 3 | spinner + two subprocesses on every Settings visit | `ModelCatalog.load` had no cache and five unrelated callers |
| 4 | onboarding left the sidebar exposed and clickable | Quickstart rendered inside the split view's `detail:` branch |
| 5 | "Latest release v0.11.0" + "status unknown — press Check for updates" | `.unknown` conflated three situations; the version row bypassed the resolver |
| 6 | Launch page as a narrow column with a nested scrollbar | `ConnectToolsView` kept its sheet frame (460×560) when used as a page |

Each is its own commit with the reasoning in the message.

## Why is this needed?

Every item is a user-visible defect reproduced on a current build, not a
refactor or a coverage exercise:

- **#1** the history list stops meaning "last worked on" — reading reorders it.
- **#4** setup reads as one more panel beside a working app instead of something
  to finish, and the sidebar stays live underneath. The two first-run surfaces
  also had no ordering relationship — nothing in `quickstartVisible` referenced
  `telemetryConsentPending` — so both could fire at once.
- **#5** the app tells the user to press a button that provably cannot change
  the answer. The underlying feed really is stale (`/api/desktop-update` and the
  `latest.json` fallback both return 0.11.0 while `rapid-mac-v0.12.1` is out) —
  that part is release plumbing and out of scope here, but the app shouldn't
  repeat it as if it were news.
- **#6** the Launch page is where the product's differentiator lives, and it
  renders as a 460pt column in the middle of a 1200pt window. It also shows
  `API key:` followed by nothing before the server starts — a config that looks
  complete, copies clean, and 401s at the far end.
- **#3** is measured, not felt: startup 2 subprocess spawns → 1, twelve panel
  switches 12 → 0.

Two of these turned out to be **half-finished intent already in the codebase**,
which is partly why I trusted the direction:

- `showsCloseButton` exists because a dead ✕ in page mode "was a real papercut"
  (#6) — the button was handled, the frame was not.
- `resolveAppUpdateStatus` already refuses to call a stale manifest "up to date"
  (the v0.7.4 regression) — but the version row rendered `appUpdater.latest`
  directly and bypassed that judgement (#5).

## AI assistance disclosure

**AI-assisted throughout, and I can explain every change.** Claude (Claude Code)
read the code, proposed each fix, and wrote the diffs and commit messages; I
directed the investigation, chose the approach at each decision point, and ran
the verification below. Every file in the diff was AI-touched. No generated
scaffold or boilerplate.

Two things worth flagging about the process, because they changed the outcome:

- The catalog cache (#3) initially "passed" on the metric I picked — subprocess
  count hit zero — while the spinner the user actually sees was still there.
  `@State` is rebuilt when a panel re-appears, so the panels started empty and
  painted the spinner before the `.task` could resume. That is what
  `seed(generation:)` is for. **The first fix measured cost, not the symptom.**
- The sidebar (#1) was verified against a running build, not reasoning: four
  conversations, five clicks including back-and-forth, order unchanged. An
  earlier round of that test looked like conversations were disappearing — that
  was my test data (four rows sharing one `messages` array, so all four derived
  the same title), not the code.

## Test plan

No Python touched, so `ruff` / `pytest` / `pr_validate` don't apply — this is
entirely `apps/rapid-mac` Swift. What I ran:

- `swift build` — clean, no warnings introduced.
- `scripts/build.sh` — full `.app` assembles and launches.
- `scripts/verify-recommendation-tiers.swift` → **ALL PASS**.
- `scripts/verify-app-resources.swift build/Rapid-MLX\ Desktop.app` → OK
  (2 assets + 44 localizable keys).
- **Date bucketing (#2)** — standalone `swift` script against a pinned reference
  time, in the `verify-*.swift` style. 12 cases, including the three that are
  easy to get wrong: 23:55 yesterday (→ Yesterday, not Previous 7 Days),
  00:00 exactly seven days back (in), one second earlier (out).
- **Sidebar order (#1)** — seeded four conversations with distinct content and
  known timestamps, clicked through them five times including revisits; order
  identical throughout, nothing lost.
- **Catalog cache (#3)** — counted `rapid-mlx` spawns during startup and during
  twelve Models ↔ Model Management switches: 2 → 1 and 12 → 0.
- **Launch page (#6)** and **Settings → App (#5)** — verified visually on the
  built app.

Not verified end-to-end: **#4 (onboarding)** builds and the presentation moved
to the split view, but reproducing it needs a wiped first-run state
(`rapid.quickstart.v1.done` + the telemetry key), which I set up but did not
walk through screen by screen. Worth a maintainer's eye on the sheet sizing in
particular.

## Checklist

- [x] Tests pass locally — Swift-side equivalents; no Python in this diff
- [x] Lint passes — n/a (`ruff` has nothing to check here)
- [ ] Self-validated with `pr_validate` — the pipeline's gating steps (`lint`,
      `targeted_tests`, `full_unit`) key on `.py` in the diff; there is none.
      Happy to run it if you'd rather see the scorecard anyway.
- [x] New behaviour spot-checked against a running build, not just compiled
- [ ] Updated README/docs — no user-facing docs describe these surfaces

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
@@ -99,13 +99,26 @@ final class ChatViewModel {
     /// Snapshot the active conversation into ``conversations`` + disk. A
     /// no-op until the user has actually sent something (no empty rows in
     /// the sidebar). Title is derived from the first user message.
-    private func persistActive() {
+    ///
+    /// - Parameter touching: whether this counts as *activity*. Only a real
+    ///   change to the transcript should refresh ``updatedAt`` and bubble the
+    ///   row to the top. Merely opening a conversation to read it, or leaving
+    ///   it to start a new one, archives the buffer without claiming the user
+    ///   did anything to it — otherwise clicking through history silently
+    ///   reshuffles the sidebar, and the list stops reflecting when each
+    ///   conversation was last *worked on*.
+    private func persistActive(touching: Bool = true) {
         guard messages.contains(where: { $0.role == .user }) else { return }
         let now = Date()
         let title = ConversationStore.title(from: messages)
         if let idx = conversations.firstIndex(where: { $0.id == activeConversationID }) {
             conversations[idx].messages = messages
             conversations[idx].title = title
+            guard touching else {
+                // Content written back, position and timestamp left alone.
+                ConversationStore.save(conversations)
+                return
+            }
             conversations[idx].updatedAt = now
             // Bubble the just-touched conversation to the top.
             let updated = conversations.remove(at: idx)
@@ -136,7 +149,7 @@ final class ChatViewModel {
         // is what gets persisted and a mid-stream switch doesn't leave the
         // incoming conversation showing Stop.
         isStreaming = false
-        persistActive()
+        persistActive(touching: false)
         guard let conv = conversations.first(where: { $0.id == id }) else { return }
         messages = conv.messages
         activeConversationID = id
@@ -205,7 +218,7 @@ final class ChatViewModel {
         // chat stuck showing Stop (isStreaming never reset). Setting it
         // false here also archives the just-closed conversation via didSet.
         isStreaming = false
-        persistActive()
+        persistActive(touching: false)
         messages.removeAll()
         activeConversationID = UUID()
         lastError = nil
--- a/apps/rapid-mac/Sources/Rapid/Chat/ConversationHistory.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/ConversationHistory.swift
@@ -56,7 +56,15 @@ enum ConversationStore {
             try? fm.moveItem(at: url, to: backup)
             return []
         }
-        return decoded
+        // Sort here rather than trusting the file's array order. ``save``
+        // writes whatever order the in-memory array happens to be in, which
+        // ``ChatViewModel.persistActive`` keeps newest-first only as a side
+        // effect of its ``insert(at: 0)`` bubbling. Any path that doesn't go
+        // through that bubble — a hand-edited file, a future bulk import, a
+        // partially-applied migration — would surface out of order in the
+        // sidebar with nothing to correct it. Guaranteeing the invariant at
+        // the data boundary costs one sort and removes the whole class.
+        return decoded.sorted { $0.updatedAt > $1.updatedAt }
     }
 
     /// Serial queue for history writes: every save re-encodes the whole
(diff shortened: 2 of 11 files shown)


raullenchai:
Landed via #1473 (merged as `a1cba9c4`). Your six fixes were merged onto current `main` (rebased past the #1460 visual-system refresh — three UI files conflict-resolved by hand, styled with the shared `SectionHeader`). While integrating I also hardened the new `ModelCatalogCache` for concurrency (an actual data race on the sync mirror, in-flight tasks now keyed by binary/generation/override, snapshot binary identity) and made the date grouping roll over at midnight. `swift build` is clean. Closing this as superseded — thank you @xiaoxiunique!

raullenchai:
Reopened — leaving this open for your visibility rather than closing it (sorry for the earlier auto-close). Status:

- **Your six fixes are live on `main`** via #1473 (`a1cba9c4`), with you as Co-Author. They were rebased past the #1460 visual-system refresh — three UI files (`SidebarView`, `ModelPickerBar`, `ConnectToolsView`) were conflict-resolved by hand, and the date-group titles now use the shared `SectionHeader`.
- While integrating, I hardened the new `ModelCatalogCache` — for reference if you build on it: (1) the `nonisolated(unsafe)` sync mirror was an actual data race (main-thread `seed()` read vs actor write) → now behind an `NSLock`; (2) in-flight tasks are keyed by (binary, generation, override) so a read after a download/folder change can't join a stale load; (3) `Snapshot` now carries the binary identity; (4) the sidebar date buckets roll over at midnight via an injected `referenceDate`.

This branch is based on pre-#1460 `main`, so it can't be merged as-is now (it would revert the visual refresh) and has no net changes left. Feel free to close it, or rebase onto `main` if you'd like your branch clean. Thanks again @xiaoxiunique — genuinely useful fixes.

raullenchai:
Closing as superseded by #1473 (`a1cba9c4`), which landed all six fixes on current `main` with the required conflict resolutions and follow-up hardening. There are no remaining changes from this branch needed for the release.

Issue #1456, opened 2026-08-04
release-check-m3 G12 random-coverage false-fails on weak sampled models (hermes3-8b-4bit)


raullenchai:
## What

The G12 `random-coverage` gate in `release-check-m3` is **date-seeded** (`seed=YYYYMMDD`) and samples random eligible aliases. On 2026-08-04 it sampled `hermes3-8b-4bit` (Hermes-3-Llama-3.1-8B, an older weak 8B) and failed all 6 rounds:

```
FAIL hermes3-8b-4bit/hermes   ❌ Expected '4' in: 2   (×3, deterministic)
FAIL hermes3-8b-4bit/opencode [FAIL] tier=harness    (×3, fast ~4s)
```

The failure is a **consistent wrong numeric answer** ("expected 4, got 2"), not a parser/serving crash. In the SAME run, the other sampled model `qwen3.5-9b-4bit` PASSED all its harnesses (aider + langchain, 6/6) — so the engine/serving/parser is fine; the failure is `hermes3-8b-4bit`'s **answer quality** (model strength), below G12's arithmetic-assertion difficulty.

This false-fails the whole gauntlet (`make: *** Error 1`) on release days whose seed happens to sample a weak model, even though every release-critical gate (coherence, throughput, G5–G9, G7b agent harness on the tier1 35B model, parsers) passes.

## Fix options

- Give G12 a **capability floor**: only sample models above a min size/quant for its reasoning assertions, or tag per-alias `g12_eligible`.
- Or make G12 assertions capability-aware (skip arithmetic/agentic assertions for < ~14B models; keep the wire-format/parser checks it was built for — PR #687).
- Or downgrade a G12 model-quality failure to a WARNING (non-blocking) while keeping parser/wire failures blocking.

## Context

Surfaced while cutting 0.12.4. Not a regression — `qwen3.5-9b-4bit` passed the same G12 run; #1442–#1455 don't touch Llama/Hermes paths.

raullenchai:
Closing this narrower report in favor of #1331, which is the canonical tracker for G12 agentic-weak/model-capability fleet debt and the required eligibility/capability-floor policy. The Hermes3 failure here is the same failure class: deterministic model-quality weakness being treated as an engine/release regression. Any Hermes-specific eligibility change should land under #1331 so the policy remains centralized.

PR #1471, opened 2026-08-05
fix(deepseek): enforce Codex reasoning budgets during decode

raullenchai:
## What changed

- enforce the translated Codex `reasoning_max_tokens` tier during DeepSeek V4 generation, including tool-enabled Responses requests
- resolve `</think>` once from the live tokenizer vocabulary, validate it against the actual LM-head width, and cache the boundary per engine
- preserve normal behavior outside the DeepSeek V4 + Codex + tools surface, but explicitly fail a qualifying request when its boundary cannot be installed safely

## Root cause

The Responses route translated Codex `reasoning_effort` into a reasoning budget, but tool-enabled DeepSeek requests only trimmed hidden reasoning after generation. The generic generation-time processor intentionally skips ungated tool requests, so a DeepSeek Codex turn could spend thousands of invisible tokens without ever reaching its DSML action despite having a configured tier.

This needs a route-specific fix because applying the processor to every tool-capable model could force a reasoning marker while another parser is already inside a tool envelope. DeepSeek V4's Codex/DSML boundary has the narrower structural guarantee required to do this safely.

DeepSeek V4's DSML parser treats the token-delimited `</think>` boundary as the transition into tool output. This PR installs the existing generation-time budget processor specifically on that proven surface.

## User impact

In a real 74K–76K-context Codex soak, previously unbounded hidden-planning turns were capped and produced valid follow-up tool calls. Observed capped completions were 2,148–2,240 total tokens for a 2,048-token reasoning tier (the remainder was the tool envelope), with zero malformed calls or reconnects. Prefix-cache continuation remained effective, requiring only 446–813 new prompt tokens.

This does not claim to solve model-level engineering judgment: local and cloud DeepSeek A/B both continued low-value read-only exploration on issue #1456. That failed task is deliberately not represented as a passing soak.

## Validation

- `python -m pytest -q tests/test_responses_route.py tests/test_reasoning_budget_generation.py` — 145 passed
- `ruff check vllm_mlx/routes/responses.py tests/test_responses_route.py` — clean
- real local DeepSeek V4 Flash 0731 Codex Responses soak at 74K–76K context

### DIFF
--- a/tests/test_responses_route.py
+++ b/tests/test_responses_route.py
@@ -1802,3 +1802,96 @@ def test_registry_engine_mismatch_fails_closed(monkeypatch):
         selected_engine, cfg, "deepseek-agent", True, kwargs
     )
     assert "suppressed_tokens_logits_processor" not in kwargs
+
+
+@pytest.mark.parametrize(
+    ("codex_surface", "has_tools", "is_deepseek", "expected"),
+    [
+        (True, True, True, True),
+        (False, True, True, False),
+        (True, False, True, False),
+        (True, True, False, False),
+    ],
+)
+def test_deepseek_codex_reasoning_budget_is_narrowly_attached(
+    monkeypatch, codex_surface, has_tools, is_deepseek, expected
+):
+    from vllm_mlx.routes.responses import _attach_deepseek_codex_reasoning_budget
+
+    processor = object()
+    calls = []
+
+    def build(*args, **kwargs):
+        calls.append((args, kwargs))
+        return processor
+
+    monkeypatch.setattr(
+        "vllm_mlx.routes.responses._uses_deepseek_v4_reasoning",
+        lambda _cfg: is_deepseek,
+    )
+    monkeypatch.setattr(
+        "vllm_mlx.api.reasoning_budget.ReasoningBudgetLogitsProcessor", build
+    )
+    monkeypatch.setattr(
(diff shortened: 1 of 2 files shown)


raullenchai:
Validation disposition for the sole `stress_e2e_bench` finding:

- The complete matrix passed all functional stress and agent checks. The only finding was Qwen 3.5 cold latency measured after the stress/agent workload: 281.88 ms (+15.8% vs a fresh-server baseline); an isolated rerun reproduced 284.22 ms.
- This PR's new route helper returns before touching state unless the request is DeepSeek V4 + Codex + tools. The flagged Qwen chat-completions benchmark cannot enter it.
- I therefore ran a paired fresh-server benchmark with the exact validation payload and model on both PR base and PR head:
  - base `018fbd4a`: cold 252.08 ms, warm 374.37 ms
  - head `8057c43c`: cold 250.34 ms, warm 372.47 ms
  - head vs base: cold 0.7% faster, warm 0.5% faster

Conclusion: no Qwen regression. The static gate compares a post-stress measurement with a fresh-server baseline, so this is a measurement-state false positive. I am recording a narrow waiver rather than changing the baseline in this unrelated PR.

Other validation results: adversarial Codex review zero blockers, lint clean, 288 targeted tests passed, and full unit `15547 passed, 30 skipped, 18 deselected, 6 xfailed, 1 xpassed`.

PR #1478, opened 2026-08-05
fix(anthropic): honor thinking controls on /v1/messages

raullenchai:
## Summary

- preserve `enable_thinking` and `chat_template_kwargs` through the Anthropic request model and adapter
- translate native `thinking.type` (`enabled` / `disabled`) into the same per-request thinking control used by OpenAI-compatible routes
- apply the shared tools/casual-chat auto-disable policy to both `/v1/messages` generation and token counting
- add route-level regression coverage for all supported opt-outs and explicit-thinking inverse behavior

## Why

`/v1/messages` silently dropped Rapid-MLX thinking extensions and ignored native `thinking.type`, so Claude Code users could not turn thinking off. Requests with tools also skipped the shared auto-disable behavior, often consuming the output budget before emitting a tool call.

## Validation

- `430 passed` focused Anthropic/thinking suite
- full pytest body: `15588 passed`; 12 server-dependent tests failed only because no service was running at `127.0.0.1:8000`
- Ruff check and format clean

### DIFF
--- a/tests/test_anthropic_adapter.py
+++ b/tests/test_anthropic_adapter.py
@@ -396,6 +396,46 @@ class TestAnthropicToOpenai:
         result = anthropic_to_openai(req)
         assert result.stream is True
 
+    @pytest.mark.parametrize(
+        ("kwargs", "expected"),
+        [
+            ({"enable_thinking": False}, False),
+            ({"chat_template_kwargs": {"enable_thinking": False}}, False),
+            ({"thinking": {"type": "disabled"}}, False),
+            (
+                {
+                    "enable_thinking": True,
+                    "chat_template_kwargs": {"enable_thinking": True},
+                    "thinking": {"type": "disabled"},
+                },
+                False,
+            ),
+            (
+                {
+                    "thinking": {"type": "enabled", "budget_tokens": 16},
+                },
+                True,
+            ),
+        ],
+    )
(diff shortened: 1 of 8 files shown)


raullenchai:
Validation adjudication before merge:

- GitHub CI is fully green: Python 3.10/3.11/3.12, Apple Silicon, both L1 model smokes, lint, type-check, MLX guard, and PR validation.
- Local focused Anthropic/thinking suite: 495 passed (latest changed subset: 157 passed).
- Local full unit gate: 15,574 passed.
- Codex adversarial review reached zero blockers before the final defensive bool guard; later strict-schema feedback was explicitly rejected because it contradicted the preceding backward-compatibility review and would break Rapid-MLX's established small `budget_tokens` extension (including the live `budget_tokens:16` repro). The valid bool finding was fixed.
- The 4-model stress/agent matrix passed Qwen3.6, DiffusionGemma, and Harmony on retry; all Anthropic SDK/LangChain/PydanticAI lanes passed. Qwen3.5 stress was equally 7/8 on PR and base. Its only remaining perf warning compares current main against a stale v0.11.0 static baseline (three current runs consistently cold ~281-285ms and warm ~355-359ms vs old 243/417ms); this PR does not touch `/v1/chat/completions` or decode hot paths.

PR #1487, opened 2026-08-05
fix(rapid-mac): stop math notation hard-crashing the app

### LINKED ISSUE #1486: rapid-mac: app hard-crashes whenever a reply contains math notation
## Symptom

Reported by a user: the desktop app quits with a macOS crash dialog as soon as a model reply contains math notation. The same model answers a simple prompt fine; asking it a math word problem kills the app. It crashes rather than failing to render or falling back to plain text.

## Cause

Packaging, not rendering. `SwiftMath` resolves its math fonts through `Bundle.module`, and SPM's generated accessor probes exactly two paths before giving up:

```swift
let mainPath  = Bundle.main.bundleURL/"SwiftMath_SwiftMath.bundle"
let buildPath = "<absolute .build path baked in at COMPILE time>"
guard let bundle = Bundle(path: mainPath) ?? Bundle(path: buildPath)
else { Swift.fatalError("could not load resource bundle: …") }
```

In a shipped `.app` both miss:

* `mainPath` is the `.app`'s **top level**, beside `Contents/`. `find` over the installed 0.12.1 app returns no `.bundle` and no `mathFonts` anywhere — `scripts/build.sh` never copies SPM's `SwiftMath_SwiftMath.bundle` (it handles `Rapid_Rapid.bundle` only).
* `buildPath` is an absolute path on whatever machine produced the binary.

`Swift.fatalError` is uncatchable, so `MathView`'s existing "unparseable LaTeX falls back to `Text`" guard cannot help — the process is already gone.

## Reproduction

Hide the build-directory fallback (which is what a user's machine looks like) and run a 40-case LaTeX corpus from inside the assembled `.app`. It dies on the **first** case, a plain `2 + 2 = 4`:

```
SwiftMath/resource_bundle_accessor.swift:12: Fatal error: could not load resource bundle:
  from …/Rapid-MLX Desktop.app/SwiftMath_SwiftMath.bundle
  or  …/.build/arm64-apple-macosx/release/SwiftMath_SwiftMath.bundle
```

SwiftMath itself is fine: the same 40 cases (`\boxed`, `\frac`, `\begin{align}`, CJK, emoji, a 5000-term expression, 40-level nesting) all parse and lay out without incident when the bundle *is* reachable.

## Why it was never caught

On a machine that built the app, the `buildPath` fallback still resolves, so the crash only appears in a shipped DMG on someone else's Mac. A first attempt to reproduce it from inside the `.app` on the build machine "passed" for exactly that reason.

## Why the bundle cannot simply be shipped

Both placements break the code signature (measured):

| placement | `codesign --verify --deep --strict` | `spctl -a -t exec` |
|---|---|---|
| `<App>.app/SwiftMath_SwiftMath.bundle` | `unsealed contents present in the bundle root` | rejected |
| root symlink → `Contents/Resources/…` | `a sealed resource is missing or invalid` | rejected |

This is the same wall `scripts/build.sh` already documents for `Rapid_Rapid.bundle` (the v0.5.9 ship-blocker). The fix used there — stop using `Bundle.module` — is not available for a third-party dependency: SwiftMath's three `Bundle.module` call sites are inside `private` methods with no injection hook.

## Status

Stop-the-bleeding fix in progress: `MathView` decides whether SwiftMath is usable **without touching SwiftMath** (the check must precede even the parse probe, because `MTFont.swift:44` force-unwraps the same lookup) and falls back to rendering the LaTeX source. That trades "math renders, app crashes" for "math shows as readable LaTeX source".

Restoring real rendering in a shipped app needs those three `Bundle.module` sites to resolve somewhere the signature permits. Vendoring SwiftMath (21 files, 7411 LOC, 3 call sites) is the only route found so far; deliberately **not** taken for now.

raullenchai:
Closes #1486.

## Why

A user reported the app quitting with a macOS crash dialog whenever a reply contains math. Same model, simple prompt: fine. Math word problem: dead. It crashes rather than degrading, so there is no workaround short of avoiding math.

The cause is packaging, not rendering. SwiftMath resolves its fonts through `Bundle.module`, whose SPM-generated accessor probes only `Bundle.main.bundleURL/"SwiftMath_SwiftMath.bundle"` and an absolute `.build` path baked in at **compile** time, then calls `Swift.fatalError`. In a shipped `.app` both miss — `find` over the installed 0.12.1 app shows no `.bundle` and no `mathFonts` anywhere — and `fatalError` is uncatchable, so `MathView`'s existing "unparseable LaTeX falls back to `Text`" guard never gets a chance.

## What this does

`MathView` now answers *"is SwiftMath usable?"* **without touching SwiftMath**. The check has to come first: `MTFont.swift:44` force-unwraps the same bundle lookup, so even the parse probe is unsafe once the bundle is missing.

Measured behaviour of the predicate:

| context | `Bundle.main.bundleURL` | value | result |
|---|---|---|---|
| dev raw binary (`.build/release/Rapid`) | `.build/release/` | `true` | math renders as before |
| assembled `.app` (dev machine **or** user's) | `<App>.app` | `false` | literal-source branch, no crash |

## What this deliberately does not do

This is stop-the-bleeding, not restoration. It trades *"math renders, app crashes"* for *"math shows as readable LaTeX source"*.

Shipping the bundle where the accessor looks is not available to us — both placements break the seal (measured):

| placement | `codesign --verify --deep --strict` | `spctl -a -t exec` |
|---|---|---|
| `<App>.app/SwiftMath_SwiftMath.bundle` | `unsealed contents present in the bundle root` | rejected |
| root symlink → `Contents/Resources/…` | `a sealed resource is missing or invalid` | rejected |

Same wall `scripts/build.sh` documents for `Rapid_Rapid.bundle` (the v0.5.9 ship-blocker); its fix — stop using `Bundle.module` — is unavailable for a third-party dependency whose three call sites are `private`. Restoring real rendering would mean vendoring SwiftMath (21 files, 7411 LOC, 3 call sites); that is a supply-chain decision, was explicitly declined for now, and is tracked in #1486 rather than smuggled in here.

## Test plan

All measured from inside the **assembled `.app`** with the build-dir fallback hidden — i.e. what a user's machine looks like:

- [x] Unguarded corpus (today's shipped behaviour): **1 fatal error, dead on case 1** (`2 + 2 = 4`)
- [x] Guarded predicate in the same `.app`: `false`, so the `Text` branch is taken and no `MTMathUILabel` is ever constructed
- [x] Predicate at the dev binary location: `true` — `swift run` still renders math
- [x] 40-case LaTeX corpus (`\boxed`, `\frac`, `\begin{align}`, CJK, emoji, 5000-term expression, 40-level nesting) parses + lays out cleanly when the bundle *is* reachable, confirming SwiftMath itself is not at fault
- [x] `swift build -c release` clean
- [x] `import SwiftMath` remains confined to `MathView.swift` (sole occurrence in `Sources/`)

Codex review: **0 BLOCKING, 0 MAJOR** (round 1).

Not verified: a live end-to-end GUI run against a real model (needs the GPU, which was in use). The crash path and the guard were both exercised directly from inside the packaged `.app` instead.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/Markdown/MathView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/Markdown/MathView.swift
@@ -45,7 +45,11 @@ struct MathView: View {
         // SwiftMath can't parse. Cheap: parse goes ``String`` →
         // ``MTMathList`` without doing any layout. If it fails,
         // surface the raw source.
-        if Self.parses(latex: latex) {
+        //
+        // ``mathFontsAvailable`` is checked FIRST and short-circuits every
+        // SwiftMath type — including the parser, because ``MTFont.swift``
+        // force-unwraps the same bundle lookup.
+        if Self.mathFontsAvailable, Self.parses(latex: latex) {
             MathHost(
                 latex: latex,
                 displayMode: displayMode,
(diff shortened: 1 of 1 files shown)


raullenchai:
## pr_validate: `codex_review` blocking — assessed, not closed

> the preflight only proves that `CGFont(provider)` can parse the font; it does not execute SwiftMath's subsequent `CTFontManagerRegisterGraphicsFont` step, so a registration failure can still reach SwiftMath's `fatalError`.

**This is correct**, and I am not claiming otherwise. `registerCGFont` throws `FontError.registerFailed` if `CTFontManagerRegisterGraphicsFont` returns false, and the caller turns that throw into `fatalError`. The preflight cannot see that.

**It is also not closable within this PR's constraints.** The reviewer's own suggested fix names the two options:

1. *"make font registration failure recoverable in SwiftMath (via patch/vendoring)"* — vendoring was explicitly declined for this change.
2. *"implement a safe registration probe"* — a probe that actually calls `CTFontManagerRegisterGraphicsFont` leaves the font registered, so SwiftMath's own registration then fails **as a duplicate**, converting a rare failure into a guaranteed one. The reviewer flags this hazard in the same sentence. Unregistering afterwards races anything else using the font.

So the honest scope of this PR is a **severity reduction, not an elimination**:

| | before | after |
|---|---|---|
| shipped `.app`, any math | crash, 100% | falls back to LaTeX source |
| bundle present but corrupt (truncated `.otf`, bad `.plist`, wrong table version) | crash | falls back (verified) |
| bundle present + valid, but `CTFontManagerRegisterGraphicsFont` fails | crash | **still crash** |

The first row is the reported bug and is what every user hits today. The third requires a CoreText registration failure on an otherwise-valid font — a system-level condition, not the shipped-artifact defect this fixes.

Tracking the residual in #1486. Closing it properly means vendoring SwiftMath (21 files / 7411 LOC / 3 `Bundle.module` sites), which is a supply-chain decision to take deliberately rather than inside a crash hotfix.

PR #1490, opened 2026-08-05
fix(rapid-mac): stop smoke.sh hanging forever — fake CLI contract + pipe EOF

### LINKED ISSUE #1489: rapid-mac: scripts/smoke.sh cannot pass — hangs forever on a clean tree
`apps/rapid-mac/scripts/smoke.sh` never completes. #1482 replaced its dead `swift test` line (which exited 1 because `Package.swift` declares no test target) with `swift build`, and that unmasked two further defects the early failure had been hiding. Measured: the script ran 3m26s and only ended when SIGKILLed.

## 1. The fake ignores its subcommand

`scripts/fake-rapid-mlx.sh` declares a `subcommand` positional and then never branches on it — `grep args.subcommand` returns 0 hits. Every verb falls through to "start the HTTP server". So `ModelCatalog.runRapidMlx(args: ["models"])` (the catalog refresh) spawns a child that never exits.

Side effect beyond the hang: the fake squats `:8000`, and the next Rapid launch's `PortSweep` reaps whatever holds that port — which may be the developer's own running `rapid-mlx`. That happened during this investigation.

## 2. `ModelCatalog.runRapidMlx` can deadlock on pipe EOF

The success path never closes the parent's copies of the pipe write ends (the launch-failure branch always has). While the parent holds a write end the pipe cannot reach EOF, so both `readPipeData` drainers block even after the child exits, and `terminationHandler`'s `drainGroup.wait()` deadlocks the continuation.

Confirmed with `sample`: two threads parked in `readPipeData` at `ModelCatalog.swift:566`, child gone, two pipe FDs still open on the process.

## 3. The chat-lifecycle harness does not exist

smoke.sh drives the lifecycle via `RAPID_TEST_DRIVER`. That string appears **nowhere in `Sources/`** — only at `smoke.sh:38`. Across full history it only ever appears in the monorepo import (#1406): the hook was part of the stripped subsystems and just the caller survived. Setting it launches a plain GUI app that never exits, and the four assertions below it grep for output that can never appear.

Three files still carry dangling `TestDriver` references in comments (`ChatStreamClient.swift`, `ServerLocator.swift`, `UpdateChecker.swift`).

## Follow-up tracked here

PR (linked below) fixes 1 and 2, and replaces the dead lifecycle block with a bounded contract check on the fake, stating plainly what is and is not covered.

**Still open after that PR:** reinstating a real chat-lifecycle smoke (ServerManager state machine + ChatStreamClient SSE decode). That needs a `TestDriver` hook to be written, which is a feature, not a script fix.

raullenchai:
Closes #1489.

## Why

`scripts/smoke.sh` never completes on a clean tree — measured at 3m26s, ended only by SIGKILL. #1482 swapped its dead `swift test` line for `swift build`, which unmasked two defects that early failure had been hiding, plus a third: the harness the script drives no longer exists.

A dev gate that hangs silently is worse than one that fails fast, and each run left a stray listener on `:8000` for the next `PortSweep` to reap — along with any real `rapid-mlx` the developer had running there. That happened during this investigation.

## What changed

**1. `fake-rapid-mlx.sh` now branches on its subcommand.** It parsed a `subcommand` positional and never used it (`grep args.subcommand` → 0 hits), so every verb started the HTTP server and `ModelCatalog.runRapidMlx(args: ["models"])` spawned a child that never exited. Now only `serve` serves; `models` / `ls` / `info` print output the existing parsers accept (`parseAvailable` / `parseCached` / `parseInfoRepo`) and exit 0. Unknown verbs are default-deny, so a future `ModelCatalog` subcommand cannot resurrect the hang.

**2. `ModelCatalog.runRapidMlx` closes the parent's pipe write ends on the success path.** The launch-failure branch always did; the success path did not. While the parent holds a write end the pipe cannot reach EOF, so both `readPipeData` drainers block even after the child exits and `terminationHandler`'s `drainGroup.wait()` deadlocks the continuation. Confirmed with `sample`: two threads parked at `ModelCatalog.swift:566`, child gone, two pipe FDs still open.

**3. The dead chat-lifecycle block is removed, not left pretending.** It drove `RAPID_TEST_DRIVER`, which appears nowhere in `Sources/` — only at `smoke.sh:38`. Across full history it exists only in the monorepo import (#1406): the hook was part of the stripped subsystems and just the caller survived. In its place the script asserts the fake's CLI contract, which is the regression guard for (1).

## Notes on the watchdog

Bounding the probes is deliberately **not** `timeout(1)` — that is GNU coreutils and stock macOS ships no equivalent (`/usr/bin/timeout` and `/bin/timeout` both absent), so depending on it would make the anti-hang guard itself exit 127 on any Mac without `brew install coreutils`.

Every `kill` goes through a jobspec (`%%`), never a raw pid: publishing `$!` is not atomic (a signal between `cmd &` and the assignment runs the trap with an empty variable), and a stored pid can be recycled between the liveness test and the signal. `%%` is resolved against bash's job table at signal time and refused once reaped, so the worst case is a no-op instead of SIGKILLing a bystander. `wait` still takes the pid — bash drops a reaped job before a fast child's `wait` runs, so a jobspec `wait` would lose the exit status, and `wait` never signals.

## Test plan

- [x] `smoke.sh` exits 0 in **0.6s** — the header's long-claimed "sub-2s" is true for the first time
- [x] Passes with `PATH=/usr/bin:/bin:/usr/sbin:/sbin` (no coreutils), proving no GNU dependency
- [x] Negative control: reverting the fake fix alone → exit 1 in **10.5s** with `FAIL — 'models' did not exit cleanly within 10s`, and **0 listeners left on :8000**
- [x] SIGTERM mid-probe → exit 143, 0 listeners
- [x] Real Ctrl-C through a pty → 0 listeners
- [x] Clean exit produces no stray stderr (cleanup with no live job)
- [x] `PROBE_TIMEOUT_S` validation: `'10s'` / `'0'` / `'-1'` / `'abc'` all exit 2 with a named error; `''` uses the default; `3` works
- [x] Watchdog exit-status propagation on bash 3.2.57 (what `/bin/bash` **and** `env bash` resolve to on macOS): timeout → 124, success → 0, failure → 1, `exit 42` → 42
- [x] `kill -9 %%` after a reap reports "no such job" rather than signalling
- [x] `swift build -c release` clean; `bash -n` clean on both scripts

Codex review: 4 rounds, findings on portability, signal choice, orphan-on-interrupt, `PROBE_TIMEOUT_S` validation and pid recycling all accepted and fixed.

**Known residual:** any pid-based shell watchdog retains a theoretical recycling window; the jobspec kills close it for every path that signals, and `wait`-by-pid cannot signal. Not further reducible in bash 3.2 without job-control process groups.

**Explicitly not fixed here:** a real chat-lifecycle smoke. That needs a `TestDriver` hook to be written — a feature, not a script fix — and is tracked in #1489.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift
@@ -539,6 +539,16 @@ enum ModelCatalog {
             processBox.set(task)
             do {
                 try task.run()
+                // The child now holds its own dup of both write ends, so
+                // drop OUR copies. While the parent keeps a write end
+                // open the pipe can never reach EOF — ``readPipeData``
+                // then blocks forever even after the child exits, and
+                // ``terminationHandler``'s ``drainGroup.wait()`` deadlocks
+                // the continuation with it. The launch-failure branch
+                // below has always closed them; the success path is where
+                // a long-lived or hung child actually makes it matter.
+                try? stdout.fileHandleForWriting.close()
+                try? stderr.fileHandleForWriting.close()
                 processBox.terminateIfCancelled()
             } catch {
                 // Close write ends so the drainers see EOF instead
--- a/apps/rapid-mac/scripts/fake-rapid-mlx.sh
+++ b/apps/rapid-mac/scripts/fake-rapid-mlx.sh
@@ -193,8 +193,66 @@ class Handler(BaseHTTPRequestHandler):
         self.wfile.write(payload)
 
 
+FAKE_REPO = "fake-org/fake-repo"
+
+
+def _emit_catalog(subcommand, alias):
+    """Print the canned output for a NON-``serve`` subcommand.
+
+    ``ModelCatalog`` shells out to ``models`` / ``ls`` / ``info`` to
+    populate the picker, and every one of those is a print-and-exit
+    command on the real binary. The formats below mirror what
+    ``ModelCatalog.parseAvailable`` / ``parseCached`` / ``parseInfoRepo``
+    actually parse (column-aligned rows behind a header + divider; an
+    ``Alias: <alias> -> <repo>`` line for ``info``).
+
+    Returns True when ``subcommand`` was handled, so ``main`` knows not
+    to fall through to the server.
+    """
+    if subcommand == "models":
+        print("Available models")
+        print("Alias                  Parser           Reasoning")
+        print("---------------------  ---------------  ---------")
+        print("fake-alias             hermes           qwen3")
+        return True
+    if subcommand == "ls":
+        print("Cached models")
+        print("Alias                  Repo                   Size")
+        print("---------------------  ---------------------  ------")
+        print(f"fake-alias             {FAKE_REPO}        1.2 GB")
+        return True
+    if subcommand == "info":
+        print(f"Alias: {alias} -> {FAKE_REPO}")
+        return True
+    return False
+
+
 def main():
     args = _parse_args(sys.argv)
+
+    # Only ``serve`` runs a server. Every other subcommand prints and
+    # exits, exactly like the real binary.
+    #
+    # This branch is the whole point of parsing ``subcommand``: without
+    # it the fake ignored the verb and started the HTTP server for ANY
+    # invocation. ``ModelCatalog.runRapidMlx(args: ["models"])`` (the
+    # catalog refresh on app launch) therefore spawned a child that
+    # never exited, so its pipe write ends never closed, so both
+    # ``readPipeData`` drainers blocked on an EOF that could not arrive
+    # and ``terminationHandler``'s ``drainGroup.wait()`` deadlocked the
+    # continuation. Net effect: ``scripts/smoke.sh`` hung forever
+    # instead of running its chat-lifecycle assertions, and left a
+    # stray listener squatting :8000 (which the next Rapid launch's
+    # PortSweep would then reap — along with any real rapid-mlx the
+    # developer had running on that port).
+    #
+    # Default-deny is deliberate: an unknown verb exits 0 with no
+    # output rather than falling through to the server, so a future
+    # ``ModelCatalog`` subcommand can't resurrect the hang.
+    if args.subcommand != "serve":
+        _emit_catalog(args.subcommand, args.alias)
+        sys.exit(0)
+
     # Match the real server's startup banner shape closely enough
     # that DownloadProgress's "Loading model with" detection fires
     # — this lets the SwiftUI overlay flip out of the spinner state
--- a/apps/rapid-mac/scripts/smoke.sh
+++ b/apps/rapid-mac/scripts/smoke.sh
@@ -1,21 +1,21 @@
 #!/usr/bin/env bash
 # smoke.sh — fast (sub-2s) end-to-end smoke for Rapid.app.
 #
-# Compiles the app and runs the chat lifecycle directive against the fake
-# rapid-mlx, so a code change can be verified without standing up a real model.
+# Compiles the app and checks the fake rapid-mlx honours the CLI contract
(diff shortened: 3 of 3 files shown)


raullenchai:
## pr_validate: `codex_review` — two blocking findings, assessed

### 1. `%%` can resolve to an unrelated pre-existing background job

Real, and already documented in the code and commit message as a known residual. Restating why it is not chased further here:

* **Matching by command text does not work.** `%?fake-rapid-mlx` was implemented and **made things worse**: bash records a job's *unexpanded* source text, which here is the literal `"$@"`, so the pattern never matches. Verified on 3.2.57 — the negative control stopped killing the hung fake and hung the script. Reverted.
* **Deferring signals around `$!` publication** (the suggested fix) trades one hole for another: signals arriving while ignored are discarded, not queued, so a Ctrl-C in that window is silently swallowed. And a recorded pid still cannot be safely signalled later once bash has reaped the job.
* Triggering it needs `$BASH_ENV` to point at a startup file that backgrounds a helper — a shell configuration this repo does not use and CI does not set.

Cost of being wrong: at most one stray background job killed in a **developer-only** script. Cost of the "fixes": a swallowed interrupt, or leaking the orphan this whole change exists to prevent. Stopping here is a judgement call, stated plainly rather than silently.

### 2. The new check never exercises `runRapidMlx` — reverting the Swift fix would leave it green

**This one is a fair hit and I am not arguing it.** The contract probe invokes the fake directly, so it guards fix (1) — the fake's subcommand branching — but not fix (2), the pipe-EOF close in `ModelCatalog.runRapidMlx`. That Swift change currently has no automated guard.

The reason is the same one this PR documents: `Package.swift` declares no test target, and the `RAPID_TEST_DRIVER` hook that used to drive the app in-process does not exist in `Sources/`. `ModelCatalog._testingRunRapidMlx(binary:args:)` exists and is exactly the right entry point, but calling it needs either a test target or a harness that links the `Rapid` module — which is the missing-harness work tracked in #1489, not something to smuggle into a hang fix.

So: fix (1) has a regression guard (verified by reverting it — the script fails in 10.5s naming the verb). Fix (2) is verified by hand (`sample` showing both drainers parked at `ModelCatalog.swift:566`, child gone, two pipe FDs open) but has none. Stated in the PR body rather than left implied.

raullenchai:
Merging with `validate` red — recording that this is a deliberate, informed override, not an oversight.

The two remaining `codex_review` blocking findings were assessed in the comment above:

* **`%%` positional jobspec** — the suggested fixes were tried and are worse. `%?fake-rapid-mlx` cannot work (bash records a job's *unexpanded* source text, here the literal `"$@"`; switching to it made the negative control hang and was reverted). Deferring signals around `$!` publication silently swallows a Ctrl-C in that window. Trigger needs `$BASH_ENV` pointing at a startup file that backgrounds a helper — not a configuration this repo or its CI uses. Blast radius is one stray background job in a developer-only script.
* **Fix (2) has no automated guard** — accurate. `ModelCatalog.runRapidMlx`'s pipe-EOF close is verified by hand (`sample`: both drainers parked at `ModelCatalog.swift:566`, child gone, two pipe FDs still open) but nothing regression-guards it, because `Package.swift` declares no test target and the in-process `RAPID_TEST_DRIVER` hook does not exist. `ModelCatalog._testingRunRapidMlx` is the right entry point; wiring a harness to it is tracked in #1489.

Landing anyway because the status quo is worse: #1482 wired a **hanging** script into the documented dev gate. Every run blocks forever, and each leaves a listener squatting `:8000` for the next Rapid launch's PortSweep to reap — along with any real `rapid-mlx` the developer had running there. That happened during this investigation and killed a live server.

This change has no end-user effect. It is developer-tooling repair plus latent hardening of the catalog path.

Issue #1516, opened 2026-08-06
Streaming reasoning on /v1/messages and /v1/responses bypasses the reasoning-channel sanitizer


raullenchai:
Found by review on #1511. **Pre-existing on `main`** — the reviewer confirmed both predate that branch.

Non-streaming reasoning is protected because it flows through `AssistantMessage`, whose field validator routes `reasoning_content` to `sanitize_reasoning_content`. The STREAMING adapters do not.

## /v1/messages streaming

`vllm_mlx/routes/anthropic.py:2120` passes streaming reasoning through `strip_special_tokens` only, and serializes it unchanged at `:1465`. That regex does not remove `</tool_call>`.

Given `GenerationOutput(channel="reasoning", new_text="plan </tool_call>")`, `/v1/messages` streaming returns the bare marker inside a `thinking_delta`.

## /v1/responses streaming

`vllm_mlx/routes/responses.py:3041` accumulates reasoning verbatim (including the parser path at `:3150`) and emits it directly at `:3595`. A reasoning delta containing `</tool_call>` survives into `response.reasoning_summary_text.*` events and the completed reasoning item.

## Why it matters

The stated invariant is "every user-visible string that originated from a raw token decode flows through the same final sanitizer." These two streaming paths are the exception, and streaming is what every agent client actually uses.

The content/reasoning channel split (#1511) makes the gap sharper to state: on the CONTENT channel a bare marker may be legitimate text and is now preserved deliberately; on the REASONING channel it is always parser residue and should always be stripped. These two paths give it neither treatment — they simply never see the reasoning sanitizer.

## Direction

Route both streaming reasoning emitters through `sanitize_reasoning_for_stream` (the whitespace-preserving variant — per-delta streaming must not `.strip()`, or cross-delta concatenation corrupts).

## Coverage gap

There are no closer-specific tests for Anthropic-streaming or Responses-streaming reasoning, which is why both stay green today. `tests/test_tool_call_closer_content_integrity.py` now covers the non-streaming `thinking` block and both OpenAI envelopes; the two streaming adapters need the same.

Issue #1515, opened 2026-08-06
qwen3_coder_xml truncates tool ARGUMENTS at a literal </tool_call> or </parameter> — coding agents cannot write files containing those tokens


raullenchai:
Pre-existing on `main`; found while regression-checking #1511. Distinct from #1513 (which is about FABRICATING calls from prose) — this one loses data inside a genuine call.

This is omlx #2507 in our codebase.

## Measured

`/v1/messages`, `qwen3.6-35b-4bit` (parser `qwen3_coder_xml`), a `write_file` tool with a `content` string parameter:

```
</tool_call> in arg   want "A tool call block ends with </tool_call> on its own."
                      got  "A tool call block ends with "

</parameter> in arg   want "A parameter block ends with </parameter> here."
                      got  "A parameter block ends with "

plain arg             round-trips fine
```

The tail of the argument is silently discarded. The model emitted it correctly — this is lost in parsing.

## In the wild

Real Claude Code 2.1.220 against a local serve, asked to write a one-line file containing the token:

> I see - my Write tool calls keep getting truncated at the exact boundary described in the file content. This is itself a demonstration of what the file says. Let me use Bash instead

It then wrote `A tool call block ends with \x08` — a failed escape workaround. The agent detects the corruption, cannot explain it, and burns turns routing around it.

## Root cause

`vllm_mlx/tool_parsers/qwen3coder_tool_parser.py`:

```python
self.tool_call_regex = re.compile(r"<tool_call>(.*?)</tool_call>|<tool_call>(.*?)$", re.DOTALL)
```

Non-greedy, so the FIRST `</tool_call>` terminates the span — including one that sits inside a `<parameter=...>` body. The parameter scanner has the same shape for `</parameter>`.

omlx hit the identical bug with the identical regex shape (their #2507). Note our `hermes` parser is immune because it anchors on `\{.*?\}` rather than a bare `(.*?)`, so this is specific to the XML-bodied parsers.

## Scope

`qwen3_coder_xml` backs 22 aliases including `qwen3.6-35b-4bit` and every Qwen3-Coder build — the default path for local coding agents. `glm47`, `minimax` and `seed_oss` share the pattern to varying degrees (measured: `glm47` loses the whole `body` argument; `minimax` and `seed_oss` truncate at `</parameter>`).

## Direction

The span end must be chosen with nesting awareness rather than by first match:

[...]

Issue #1513, opened 2026-08-06
qwen3_coder_xml fabricates tool calls from prose — parser gate is incomplete on the streaming and fallback paths


raullenchai:
Split out of #1511 so a verified, codex-clean fix (the `</tool_call>` content channel split) is not held up by a harder problem that keeps growing. The `</tool_call>` work ships there; this is everything else.

## The defect

`qwen3_coder_xml` promotes `<function=NAME>` spans in ordinary prose into structured `tool_calls`. Measured with **no tools declared at all**:

```
in : Docs mention </tool_call> and <function=name> together.
out: tools_called=True, tool_calls=[{"name": "name", "arguments": "{}"}]
     content="Docs mention </tool_call> and "
```

An agent receiving that tries to execute a tool called `name` and its loop breaks. This parser backs 22 aliases including `qwen3.6-35b-4bit` and every Qwen3-Coder build, so it is the default path for local coding agents.

## Why the obvious fixes are not enough

Two rounds of review closed off two candidate discriminators:

1. **"only accept CLOSED spans"** — insufficient (balanced prose like `<function=read_file></function>` is well-formed) AND harmful: it throws away genuine calls that `max_tokens` truncated before `</function>`, which the tolerant branch used to recover.

2. **"only accept names the caller DECLARED"** — necessary but not sufficient, and it does not reach the paths that matter:

   - **Streaming never applies it.** The streaming state machine parses the `<function=NAME>` header independently and emits a `tool_calls` delta immediately. Streaming is the primary agent-client path, so the fabrication survives there untouched.
   - **The generic fallback re-promotes.** `service/helpers.py` falls back to the ungated `api/tool_calling.parse_tool_calls` whenever the configured parser returns `tools_called=False`. A no-tools response containing `<tool_call><function=read_file></function></tool_call>` comes back as `read_file` anyway. With `tools=[write_file]` and `tool_choice="required"` the fallback yields the **undeclared** `read_file` and required-mode forwards it.
   - **`tool_choice="none"` is ignored.** Declared does not mean permitted; a caller that explicitly forbade calls still gets one.
   - **Mixed spans lose content.** When an undeclared documentation span precedes a valid declared call, the undeclared span is skipped but the content cutoff still starts at the FIRST opener, so the prose between them disappears.
   - **`_declared_tool_names` misses pydantic `ToolDefinition`** objects placed directly in `request["tools"]` (production routes `model_dump()` first, so this bites internal callers).

[...]

Issue #1508, opened 2026-08-06
Forced/required streaming path leaks raw tool wire (opener + full JSON payload) into content


raullenchai:
Found while fixing the `</tool_call>` content-integrity bug. **Pre-existing on `main` — not introduced by that fix.**

## What happens

On the streaming `tool_choice="required"` path, when `_synthesize_forced_tool_call` produces arguments that fail the target schema, the route correctly refuses to fabricate a call (`[SSE-FORCED-SYNTH-#1256]`). But the model’s raw tool wire has already gone out as **content deltas**, so the client aggregates:

```
<tool_call>{"name":"get_weather","arguments":{"city":"Tokyo"}}</tool_call>
```

as the assistant’s visible message.

## Why it was invisible

`tests/test_r12_reasoning_sanitizer_required.py::test_chat_route_streaming_required_reasoning_is_sanitized` asserts each leak marker is absent from aggregated content. It passed on `main` only because the global `_FINAL_SANITIZER` stripped the `</tool_call>` **closer**. Replaying the same input through the `main` regex:

```
in : <tool_call>{"name":"get_weather","arguments":{"city":"Tokyo"}}</tool_call>
out: <tool_call>{"name":"get_weather","arguments":{"city":"Tokyo"}}
```

The opener and the entire JSON payload survived. The assertion was satisfied by a cosmetic strip of the last 13 bytes while the leak itself was fully visible to the user.

## Why it is not fixed in the same PR

The scrubber that can tell wire residue from prose (`_scrub_visible_tool_wire_leaks` / `_contains_structural_tool_wire_leak`) is payload-aware and needs the whole span. Streaming emits per-delta, so suppressing this requires either buffering content on the forced path or a streaming-aware structural detector — a design call worth its own change rather than a drive-by.

## Suggested direction

On the forced/required path the assistant’s visible content is not meaningful to the client anyway (it is waiting for a tool call), so buffering content deltas until the synth decision is made, then scrubbing structurally before emit, looks like the smallest correct fix.

## Repro

`tool_choice="required"`, streaming, a model whose emitted call fails the tool schema. The test above already drives exactly this shape — it just needs its content assertion strengthened from "no closer" to "no wire span with a payload".

PR #1517, opened 2026-08-06
fix: land three review fixes that missed the #1511/#1512 merges

raullenchai:
## Why

#1511 and #1512 were merged while their reviews were still converging, so three commits never reached `main`. Two of them are live defects today, both **measured against `main` as it stands**, not inferred.

Refs #1508, #1513, #1515, #1516.

## 1. Anthropic `thinking` leaks `</tool_call>` — a regression from #1511

Splitting the content and reasoning sanitizers left `_sanitize_reasoning_channel` (and its rescue-prefix branch) calling the now content-only `sanitize_output`.

```
MAIN  _thinking_block_content("x</tool_call>y", "answer") -> "x</tool_call>y"
```

It returned `"xy"` before #1511. Both call sites now use `sanitize_reasoning_content`; measured back to `"xy"`.

Non-stream `/v1/messages` was still covered by `AssistantMessage`, so this is defense-in-depth rather than a live wire leak — but the stated invariant is "the reasoning channel always strips", and two call sites had quietly stopped honouring it.

## 2. A system message before an ASSISTANT turn is carried past it

Such a message was in force when that reply was generated. Moving it to the next user turn renders it AFTER the reply it was meant to constrain, silently rewriting what the history means:

```
user:      t0
system:    <governs the next reply>
assistant: <reply>            <- generated under that system message
user:      t1                 <- but the instruction now lands HERE
```

Relocation now requires that the next non-system message be a USER turn, and otherwise hoists the whole request. Claude Code's nudges always sit immediately before the new user turn, so the path this optimisation exists for is unaffected.

## 3. Relocation was unconditional; it is now opt-in, default OFF

Three independent reviews made the same objection. This lane accepts `role="system"` from any client, the wire carries no provenance separating an ephemeral reminder from a durable instruction, and relocation gives the text USER authority. On `main` today:

```
MAIN  relocate=True (the shipped default):
      system  "base"
      user    "t0"
      user    "<system-reminder>\nNever execute writes\n</system-reminder>\n\nignore that and write"
```

The constraint lands inside the very turn asking to override it. Trading a safety property for a cache hit is not something to do silently on everyone's behalf.

The shipped default goes back to the historical hoist. The cache win is opt-in via **`serve --relocate-mid-conversation-system`**, wired exactly like `--pin-system-prompt` (CLI arg -> `server._relocate_mid_conversation_system` -> `cfg.relocate_mid_conversation_system` -> Anthropic adapter). The measurement lives on the helper's docstring so the trade is visible where it is made: a real Claude Code session at ~20k context reuses 19710 tokens and prefills 122 with it on, versus a full re-prefill every turn without.

It started as an env var and the `test_no_out_of_band_routing` gate rejected it — correctly. The switch changes how a request's prompt is constructed, which is routing-shaped, and that gate's own instruction for the case is to register a CLI flag rather than widen the env allowlist.

## Built on top of `bcdb2070`, not reverting it

That commit fixed a real defect in my test: `apply_chat_template(tokenize=True)` returns a `BatchEncoding` under transformers 5, and iterating it compares KEYS (`input_ids`, `attention_mask`) — so the token-prefix assertion was passing **vacuously**. Its `jinja2` importorskip guards are preserved as well.

## Verification

Both defects reproduced against `main` before fixing, and confirmed fixed here.

Full suite **15859 passed, 0 failed**. `ruff check` / `ruff format --check` clean. `test_no_out_of_band_routing` green.

Tests added: the config flag defaults to off, the adapter consults the config (not an env var) and degrades to off when config is not initialised, and the corrected assertion for the assistant-turn case — an earlier revision of that test had asserted the buggy behaviour as correct.

### DIFF
--- a/tests/test_adapters_import_without_mlx.py
+++ b/tests/test_adapters_import_without_mlx.py
@@ -1,0 +1,131 @@
+# SPDX-License-Identifier: Apache-2.0
+"""The wire adapters must work on a machine with no MLX.
+
+``anthropic_to_openai`` is a pure message-shape translation: Anthropic
+JSON in, OpenAI JSON out. No weights, no Metal, no engine. The whole
+``TestAnthropicToOpenai`` suite has always run on Linux CI for exactly
+that reason.
+
+Then this PR taught the adapter to consult ``get_config()`` for the
+mid-conversation-system flag, and ``vllm_mlx.config`` transitively did::
+
+    config/__init__ -> config.server_config -> engine.base
+                    -> engine_core -> import mlx.core
+
+which is fatal off Apple Silicon. Every test in that class died with
+``ModuleNotFoundError: No module named 'mlx'`` — on Linux only, so the
+macOS dev loop stayed green and CI caught it instead.
+
+The fix is not to swallow the ImportError: that would silently ignore an
+operator's explicit flag. It is for ``config`` not to reach into the
+engine at runtime in the first place (the annotation is type-only, and
+``server_config`` already has ``from __future__ import annotations``).
+
+These tests run in a subprocess with MLX blocked at the meta-path, which
+is the only honest way to assert it here — this suite's own process has
+MLX loaded long before any test starts. Mutation check: restore the
+module-scope ``from ..engine.base import BaseEngine`` and both fail.
+"""
+
+from __future__ import annotations
+
+import pathlib
+import subprocess
+import sys
+import textwrap
+
+import pytest
+
+REPO_ROOT = pathlib.Path(__file__).resolve().parent.parent
+
+# Installed at sys.meta_path[0] so it is consulted before any real
+# finder, and raises rather than returning None — returning None would
+# merely fall through to the next finder and find the real MLX.
+_BLOCK_MLX = """
+import sys
+
+
+class _NoMLX:
+    def find_spec(self, fullname, path=None, target=None):
+        if fullname == "mlx" or fullname.startswith("mlx."):
+            raise ImportError("mlx is unavailable in this environment")
+        return None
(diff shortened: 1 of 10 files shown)


raullenchai:
## Round-4 BLOCKING declined — verified against the running code

> The test accesses `engine.BatchedEngine` while its meta-path hook rejects
> every `mlx` import, but `BatchedEngine` is explicitly lazy because importing
> `batched` requires `mlx.core`, so the subprocess raises `ImportError` before
> reaching `LAZY-OK`.

`batched` does not require `mlx.core` to import — it degrades when MLX is
absent. Run against this branch, with the same blocker the test installs:

```
$ python -c "
import sys
class NoMLX:
    def find_spec(self, n, p=None, t=None):
        if n == 'mlx' or n.startswith('mlx.'): raise ImportError('blocked')
        return None
sys.meta_path.insert(0, NoMLX())
import vllm_mlx.engine as e
print('BaseEngine:', e.BaseEngine.__name__)
print('BatchedEngine:', e.BatchedEngine.__name__)
print('mlx in sys.modules:', 'mlx' in sys.modules)
"
Guided generation disabled: could not import the llguidance stack (ImportError: blocked). ...
BaseEngine: BaseEngine
BatchedEngine: BatchedEngine
mlx in sys.modules: False
```

The attribute resolves, and `mlx` never enters `sys.modules` — which is the
property the test is there to pin. `tests/test_adapters_import_without_mlx.py`
has passed on every run of this branch, including the one that produced this
finding (`full_unit` PASS, 15891 tests).

The NIT on `__dir__` is fixed in the commit above.

PR #1526, opened 2026-08-06
fix(scheduler): stop mlx-lm writing None into a mixed batch's processor slots

### LINKED ISSUE #1525: 503 under mixed concurrent load: mlx-lm PromptProcessingBatch.extend writes None into logits_processors
## Symptom

Mixed concurrent traffic — plain chat + tool-calling in flight at the same
time — kills the engine loop and returns **503** to the clients that were
mid-flight:

```
ERROR:rapid_mlx.engine_core:Engine loop error: 'NoneType' object is not iterable
  vllm_mlx/scheduler.py:7103 in step        raw_next = self.batch_generator.next()
  mlx_lm/generate.py:1809  in _next         gen_batch = self._prompt_batch.split(split).generate(last_inputs)
  mlx_lm/generate.py:1288  in __init__      self._step()
  mlx_lm/generate.py:1346  in _step         for processor in self.logits_processors[e]:
TypeError: 'NoneType' object is not iterable
INFO: 127.0.0.1 - "POST /v1/chat/completions HTTP/1.1" 503 Service Unavailable
```

`scripts/stress_test.py` case 6, "Mixed workload (4 concurrent, different
types)", on `mlx-community/Qwen3.5-35B-A3B-8bit`, M3 Ultra:

```
[6/8] Mixed workload (4 concurrent, different types)...
  chat2: 498ms OK
  tool:  893ms FAIL
  chat1: 893ms FAIL
  stream: 893ms OK
  Errors: 2/4
```

**Reproduces on `main` (9c9e4b0d) in 3 of 5 consecutive runs**, 15
engine-loop crashes across those runs. Not tied to any PR — surfaced while
classifying a `pr_validate` stress finding, whose single base run happened
to land on one of the passing rolls.

## Root cause — upstream, `mlx_lm.generate.PromptProcessingBatch.extend`

`mlx_lm/generate.py:1062`:

```python
def extend(self, batch):
    if not any(self.samplers):
        self.samplers = [None] * len(self.uids)
    if not any(self.logits_processors):
        self.logits_processors = [None] * len(self.uids)          # <-- []  becomes None
    ...
    logits_processors = (
        batch.logits_processors
        if any(batch.logits_processors)
        else [None] * len(batch.uids)                              # <-- same
    )
    self.logits_processors.extend(logits_processors)
```

`any()` is a whole-list question; the slots are per-sequence. A batch of
requests that carry no processors normalizes to `None` slots, and merging
it with a batch that HAS one produces a mixed list. From then on
`any(...)` is True everywhere, so `filter()` (`:1117`) preserves each slot
verbatim, and `GenerationBatch._step` (`:1346`) iterates it:

```python
if any(self.logits_processors):
    for e in range(len(self.uids)):
        for processor in self.logits_processors[e]:     # None here -> TypeError
```

Homogeneous batches are safe in both directions — all-empty takes the
`else` branch and becomes `[[]] * n`, all-present is iterable. **Only the
mixed batch crashes**, which is exactly what "plain chat concurrent with a
tool call" builds, and only when a split/merge reshuffles the batch. Hence
the intermittency.

Minimal reproduction — no model, no weights, pure list bookkeeping:

```python
from mlx_lm.generate import PromptProcessingBatch as PB

def noop(tc, logits): return logits

def mk(uids, procs):
    b = PB.__new__(PB)
    b.uids, b.logits_processors = list(uids), list(procs)
    b.samplers = [None] * len(uids); b.prompt_cache = []
    b.tokens = [[] for _ in uids]; b.max_tokens = [64] * len(uids)
    b.state_machines = [None] * len(uids)
    return b

a, b = mk([1, 2], [[], []]), mk([3], [[noop]])
a.extend(b)
print(a.logits_processors)   # [None, None, [<function noop>]]
a.filter([0, 1, 2])
print(a.logits_processors)   # [None, None, [<function noop>]]  <- survives
for _ in a.logits_processors[0]: pass   # TypeError: 'NoneType' object is not iterable
```

`[]` is the correct normalization at both sites: it keeps `any()` False for
the all-empty case (so the fast path still skips the whole loop) and stays
iterable when mixed.

## Our exposure

`vllm_mlx/scheduler.py:6189` passes `None` for a request with no
processors:

```python
request_logits_processors = [request_processors] if request_processors else None
```

That is fine on its own — `insert_segments` maps `None` to
`[self.logits_processors] * n` = `[[]] * n`. The `None` slots are
synthesized later, inside `extend`, and no amount of care at the insert
site prevents it.

`_realign_grammar_logits_processors` cannot cover this either: it rebuilds
the **GenerationBatch**'s list, while the crash happens inside
`GenerationBatch.__init__` -> `_step()`, before we get a tick.

## Suggested fix

1. Report upstream (the two `[None] * n` should be `[[]] * n`).
2. Carry a local guard until a released mlx-lm has it, since this is a 503
   on ordinary mixed traffic and we pin `mlx-lm==0.31.3`. The narrow
   version is to normalize `None` -> `[]` in the batch's
   `logits_processors` at the point we hand work to the generator, or to
   wrap `PromptProcessingBatch.extend` so it never writes `None`.
3. Add a scheduler-level regression test that builds a genuinely mixed
   batch (one request with a grammar/penalty processor, one without) and
   forces a split — the unit suite has no coverage that mixes the two,
   which is why 15852 green tests say nothing about this.

## Impact

High. Any agent that interleaves tool calls with plain chat on one server —
which is every agent we support — can take a 503 mid-session. It is
invisible in single-request testing and in homogeneous load tests, which is
why the stress suite only catches it in the one mixed case.

raullenchai:
## Why

Plain chat concurrent with a tool call — which is what every agent we support
does — could kill the engine loop and hand a **503** to every in-flight
request:

```
ERROR:rapid_mlx.engine_core:Engine loop error: 'NoneType' object is not iterable
  mlx_lm/generate.py:1809  gen_batch = self._prompt_batch.split(split).generate(...)
  mlx_lm/generate.py:1288  self._step()
  mlx_lm/generate.py:1346  for processor in self.logits_processors[e]:
TypeError: 'NoneType' object is not iterable
INFO: "POST /v1/chat/completions HTTP/1.1" 503 Service Unavailable
```

Closes #1525.

## Root cause — upstream `PromptProcessingBatch.extend`

```python
if not any(self.logits_processors):
    self.logits_processors = [None] * len(self.uids)          # []  ->  None
...
logits_processors = (
    batch.logits_processors
    if any(batch.logits_processors)
    else [None] * len(batch.uids)                              # same
)
```

`any()` is a whole-list question; the slots are per-sequence. A no-processor
batch normalizes to `None` slots, and merging it with a batch that HAS one
yields `[None, None, [proc]]`. From then on `any(...)` is True, so `filter`
keeps every slot verbatim and `_step` iterates a `None`.

Homogeneous batches are safe in both directions — all-empty takes the `else`
branch and becomes `[[]] * n`, all-present is iterable throughout. **Only the
mixed batch crashes**, and only after a split/merge reshuffles it, which is
why this is intermittent rather than constant.

Reproducible with no model and no weights, purely in the list bookkeeping:

```python
from mlx_lm.generate import PromptProcessingBatch as PB
a, b = mk([1, 2], [[], []]), mk([3], [[noop]])
a.extend(b)
print(a.logits_processors)   # [None, None, [<function noop>]]
```

## The fix

Wrap `extend` and normalize `None` back to `[]`. That is the correct value at
both sites: it keeps `any()` False for the all-empty case, so `_step` still
skips the entire processing loop, and it stays iterable when mixed.

Wrapping rather than reimplementing keeps upstream authoritative for
everything else `extend` does — we only clean up after it, so an upstream
change to the merge semantics doesn't silently diverge from a copied body.
It lives in `vllm_mlx/_mlx_compat.py`, which already holds our other
upstream shim and its remove-when-fixed note.

`vllm_mlx/scheduler.py` installs it immediately *after* importing
`mlx_lm.generate`, the mirror of the existing `install()` which must run
*before* that import.

## Verification

`scripts/stress_test.py` case 6 "Mixed workload (4 concurrent, different
types)" on `mlx-community/Qwen3.5-35B-A3B-8bit`, M3 Ultra, five consecutive
runs against a freshly-started server each time:

| tree | result | engine-loop crashes |
|---|---|---|
| `main` @ 9c9e4b0d | **3 of 5 FAIL** (7/8 passed) | 15 |
| this branch | 5 of 5 PASS (8/8 passed) | 0 |

Nine unit tests drive the real mlx-lm classes rather than a stand-in — the
defect is in upstream's bookkeeping, so a fake would only test the fake. They
cover both merge directions, slot/uid alignment, survival through `filter`,
that a real processor is never dropped by the normalization, that the
all-empty fast path is not accidentally armed, and idempotency.

`test_scheduler_import_installs_the_guard` earned its place immediately: it
caught that `from mlx_lm import generate` binds mlx_lm's top-level
`generate()` **function**, which shadows the submodule of the same name, so
the first version of this shim silently patched nothing. A shim nobody calls
is a shim that does not exist.

Full unit suite: 15861 passed.

## Side effect worth noting

This also makes `pr_validate`'s stress gate honest. Its A/B logic is right —
if base fails too, the finding is recorded `[PRE-EXISTING]` rather than
blocking — but at a 60% reproduction rate the PR and base lanes were rolling
dice independently, so roughly 40% of runs produced a BLOCKING for a defect
the PR under review did not introduce.

### DIFF
--- a/tests/test_mixed_batch_logits_processor_slots.py
+++ b/tests/test_mixed_batch_logits_processor_slots.py
@@ -34,11 +34,20 @@ import pytest
 
 REPO_ROOT = pathlib.Path(__file__).resolve().parent.parent
 
+# BEFORE importing mlx_lm.generate, not after. That module captures
+# ``mx.new_thread_local_stream(...)`` at import time, and ``install()`` is what
+# makes that call safe on single-stream GPUs (#404) — the ordering invariant
+# both ``_mlx_compat`` and ``scheduler`` document. Importing mlx-lm first here
+# would pin the unusable stream for the rest of the pytest process, so a later
+# scheduler import installs the shim too late and inference fails with
+# "There is no Stream(gpu, 1) in current thread".
+from vllm_mlx import _mlx_compat  # noqa: E402
+
+_mlx_compat.install()
+
 pytest.importorskip("mlx_lm.generate", reason="requires MLX")
 
 from mlx_lm.generate import PromptProcessingBatch  # noqa: E402
-
-from vllm_mlx import _mlx_compat  # noqa: E402
 
 
 def _noop_processor(token_context, logits):
--- a/tests/test_mixed_batch_logits_processor_slots.py
+++ b/tests/test_mixed_batch_logits_processor_slots.py
@@ -173,7 +173,7 @@ def test_scheduler_import_installs_the_guard():
     In a SUBPROCESS, and deliberately outside the autouse fixture's reach.
     Asserting the flag in-process proves nothing: the fixture installs the
     guard before every test in this file, so the assertion would hold with
-    the ``scheduler.py`` call site deleted (review BLOCKING). The only
+    the ``scheduler.py`` call site deleted (raised in review). The only
     honest question is whether importing the scheduler — and nothing else —
     arms it.
     """
@@ -197,6 +197,6 @@ def test_scheduler_import_installs_the_guard():
     )
     # Both, not just the marker: a process can print WIRED and then die on
     # the way out, and a test that only greps stdout would call that a pass
-    # (review NIT).
+    # (raised in review).
     assert proc.returncode == 0, proc.stderr + proc.stdout
     assert "WIRED" in proc.stdout, proc.stderr + proc.stdout
--- a/tests/test_mixed_batch_logits_processor_slots.py
+++ b/tests/test_mixed_batch_logits_processor_slots.py
@@ -26,7 +26,13 @@ them.
 
 from __future__ import annotations
 
+import pathlib
+import subprocess
+import sys
+
 import pytest
+
+REPO_ROOT = pathlib.Path(__file__).resolve().parent.parent
 
 pytest.importorskip("mlx_lm.generate", reason="requires MLX")
 
@@ -163,8 +169,30 @@ def test_scheduler_import_installs_the_guard():
 
     A shim nobody calls is a shim that does not exist; this is the wire
     between the fix and the crash path.
-    """
-    import importlib
 
-    importlib.import_module("vllm_mlx.scheduler")
-    assert getattr(PromptProcessingBatch, "_rapid_mlx_slot_guard", False)
+    In a SUBPROCESS, and deliberately outside the autouse fixture's reach.
+    Asserting the flag in-process proves nothing: the fixture installs the
+    guard before every test in this file, so the assertion would hold with
+    the ``scheduler.py`` call site deleted (review BLOCKING). The only
+    honest question is whether importing the scheduler — and nothing else —
+    arms it.
+    """
+    proc = subprocess.run(
+        [
+            sys.executable,
+            "-c",
+            "import importlib;"
+            "m = importlib.import_module('mlx_lm.generate');"
+            "assert not getattr(m.PromptProcessingBatch, '_rapid_mlx_slot_guard', False),"
+            " 'guard armed before importing the scheduler — test is vacuous';"
+            "importlib.import_module('vllm_mlx.scheduler');"
+            "assert getattr(m.PromptProcessingBatch, '_rapid_mlx_slot_guard', False),"
+            " 'importing vllm_mlx.scheduler did not install the guard';"
+            "print('WIRED')",
+        ],
+        capture_output=True,
+        text=True,
+        cwd=REPO_ROOT,
+        timeout=300,
+    )
+    assert "WIRED" in proc.stdout, proc.stderr + proc.stdout
--- a/tests/test_mixed_batch_logits_processor_slots.py
+++ b/tests/test_mixed_batch_logits_processor_slots.py
@@ -1,0 +1,170 @@
+# SPDX-License-Identifier: Apache-2.0
+"""A mixed batch must not put ``None`` in a per-sequence processor slot.
+
+#1525. Plain chat concurrent with a tool call — which is what every agent
+we support does — could kill the engine loop:
+
+    ERROR:rapid_mlx.engine_core:Engine loop error: 'NoneType' object is not iterable
+      mlx_lm/generate.py:1809  gen_batch = self._prompt_batch.split(split).generate(...)
+      mlx_lm/generate.py:1346  for processor in self.logits_processors[e]:
+    TypeError: 'NoneType' object is not iterable
+    INFO: "POST /v1/chat/completions HTTP/1.1" 503 Service Unavailable
+
+``PromptProcessingBatch.extend`` normalizes "no processors" to ``None``
+slots, guarded by ``any()`` — a whole-list question asked about a
+per-sequence list. All-empty and all-present are both safe; only the MIXED
+batch produces ``None``, and only after a split/merge reshuffles it, which
+is why it reproduced in 3 of 5 consecutive stress runs rather than every
+time.
+
+These tests drive the real mlx-lm classes rather than a stand-in: the
+defect is in upstream's bookkeeping, so a fake would only test the fake.
+No model and no weights are involved — the batch objects are built with
+``__new__`` and the lists that matter, exactly as the crash path leaves
+them.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+pytest.importorskip("mlx_lm.generate", reason="requires MLX")
+
+from mlx_lm.generate import PromptProcessingBatch  # noqa: E402
+
+from vllm_mlx import _mlx_compat  # noqa: E402
+
+
+def _noop_processor(token_context, logits):
+    return logits
+
+
+def _batch(uids, processors):
+    """A PromptProcessingBatch carrying only the state ``extend`` touches."""
+    b = PromptProcessingBatch.__new__(PromptProcessingBatch)
+    b.uids = list(uids)
+    b.samplers = [None] * len(uids)
+    b.logits_processors = list(processors)
+    b.prompt_cache = []
+    b.tokens = [[] for _ in uids]
+    b.max_tokens = [64] * len(uids)
+    b.state_machines = [None] * len(uids)
+    return b
+
+
+@pytest.fixture(autouse=True)
+def _guard_installed():
+    _mlx_compat.install_batch_slot_guard()
+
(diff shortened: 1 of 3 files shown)


raullenchai:
## The two bench findings are environmental, not this branch

`pr_validate` reported `cold +42.5%` (Qwen3.5-35B-A3B-8bit) and `cold +90.1%`
(gpt-oss-20b-MXFP4-Q8). Both are measurement noise from that particular run.

**Mechanism first**: the bench sends requests *sequentially*, one at a time.
This branch's only runtime addition is a wrapper on
`PromptProcessingBatch.extend`, which runs when a new request MERGES into a
running batch. During the bench no merge ever happens, so the added code does
not execute. There is no path by which it could cost 40%.

**Interleaved A/B**, main vs this branch, alternating so any drift in machine
state over the run hits both arms equally. Fresh server per sample,
`stress_e2e_bench._run_bench`'s exact protocol, `Qwen3.5-35B-A3B-8bit`:

| round | main cold | #1526 cold | main warm | #1526 warm |
|---|---|---|---|---|
| 1 | 255.2 | 254.7 | 379.3 | 377.6 |
| 2 | 252.6 | 249.4 | 376.4 | 373.1 |
| 3 | 249.8 | 252.0 | 373.3 | 375.3 |
| **median** | **252.6 / 376.4** | **252.0 / 375.3** | | |

**cold −0.2%, warm −0.3%** — indistinguishable.

A first, non-interleaved comparison had put all branch samples before all main
samples and showed the branch ~11% slower; that ordering puts an hour of drift
entirely on one arm, which is why it was re-run interleaved rather than
believed.

**What the noise looks like.** The individual cold samples give it away. On
main, gpt-oss measured `[363.5, 168.0, 154.9, 252.1, 252.8]`; on this branch,
`[365.8, 169.8, 156.8, 255.1, 258.0]` — near-identical per prompt — and then a
second branch sample came in at `[423.5, 197.4, 189.0, 289.5, 289.2]`, every
value shifted up by roughly the same ratio. A uniform scaling across five
*different* prompts, present in one sample and absent from the next, is the
signature of machine state, not of a code change.

The likely contributor was visible in `ps` during the run: macOS's
`WallpaperAerialsExtension` (9.1% CPU) driving `VTDecoderXPCService` (4.0%),
i.e. the animated desktop decoding video on the GPU, plus `WindowServer` at
11.6%. Cold-start latency is the metric most exposed to that. Filed
separately — the gate has no defence against external GPU consumers, only
against the contention it creates itself (`release_check_m3.sh` serializes
G7b for exactly this reason).

**Cross-check on the metric itself.** Cold medians measured on
main-equivalent trees across the session: Qwen3.5-35B 252.8 / 253.1 / 252.0 /
255.7 / 252.5 / 255.2 / 252.6 / 249.8 (spread 2.4%); gpt-oss 258.6 / 256.8 /
256.3 / 252.1 / 256.9 (spread 2.6%). On a quiet machine this metric is stable
well inside its 5% threshold. The gate run was the outlier, not the metric.

## Stress: the fix works in the gate too

All four models reported `8/8 passed`. Before this branch, the same case
failed 3 of 5 consecutive runs on main with 15 engine-loop crashes.

raullenchai:
## Reproduced on the shipped 0.12.4 wheel, and the trigger is sharper than stated

Dogfooding a clean `pip install` of the 0.12.4 wheel in a fresh venv, serving
`qwen3.5-35b-8bit`. Same script against both trees, same machine, back to back:

| tree | non-200 responses | engine-loop crashes |
|---|---|---|
| shipped 0.12.4 wheel | **5 / 24** | **10** |
| this branch | 0 / 24 | 0 |

```
round 1/6: 1/4 ok   503 "Engine loop error: TypeError: 'NoneType' object is not iterable"  (x3)
round 3/6: 2/4 ok   503 ... (x2)
```

### A streaming request has to be in the batch

My first repro sent chat + tool-call concurrently, all `stream: false`, and got
**24/24 OK** — it never triggered. Adding one streaming request to the same
four-way mix reproduced it immediately.

That matches `scripts/stress_test.py` case 6, which is chat + chat + tool +
**stream**, and explains why it is the only stress case that fails. The
streaming request stays in the batch on a different schedule from the
non-streaming ones, so the batch is re-split while a processor-carrying
request is still live — which is what drives `PromptProcessingBatch.extend`
into the `[None] * n` normalization.

So the precise trigger is: **a batch reshuffle while requests with and without
logits processors are both in flight**, and a streaming request in the mix is
what reliably causes the reshuffle. "Mixed chat + tool concurrency" alone is
necessary but not sufficient.

Repro (no repo checkout needed, just a running server):

```python
import json, urllib.request, concurrent.futures as cf
URL = "http://127.0.0.1:8000/v1/chat/completions"
TOOLS = [{"type": "function", "function": {"name": "get_weather", "description": "Get weather",
  "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}]

def post(body, stream):
    req = urllib.request.Request(URL, data=json.dumps(body).encode(),
                                 headers={"Content-Type": "application/json"}, method="POST")
    with urllib.request.urlopen(req, timeout=180) as r:
        [_ for _ in r] if stream else r.read()

def chat():   post({"model": "default", "messages": [{"role": "user", "content": "What is 2+2?"}], "max_tokens": 50, "stream": False}, False)
def tool():   post({"model": "default", "messages": [{"role": "user", "content": "Weather in Paris?"}], "tools": TOOLS, "max_tokens": 100, "stream": False}, False)
def stream(): post({"model": "default", "messages": [{"role": "user", "content": "Write a poem."}], "max_tokens": 100, "stream": True}, True)

for _ in range(6):
    with cf.ThreadPoolExecutor(max_workers=4) as ex:
        [f.result() for f in [ex.submit(chat), ex.submit(chat), ex.submit(tool), ex.submit(stream)]]
```

### Why this matters for the shape of the fix

It does not change the fix — the `None` slot is written by `extend` regardless
of what caused the reshuffle — but it does mean the exposure is wider than
"tool-calling agents". Any client that streams while another request carries a
grammar or penalty processor is exposed, which on a shared local server is the
ordinary case.

Issue #1462, opened 2026-08-05
auto-release: close the tag TOCTOU and make tag-without-Release recovery reachable


raullenchai:
Two findings from the #1401 review that were deliberately **not** fixed in that
PR, because both change the release path in ways that cannot be verified
end-to-end without cutting a real release, and a mistake there breaks *all*
releases. #1401 is a strict improvement over the previous state either way
(before it there was no tag/SHA verification at all), so these are follow-ups
rather than blockers on it.

### 1. TOCTOU between the tag pre-check and `gh release create`

`git ls-remote` proves the tag is absent, but another writer can push it at a
different commit before `gh release create --target $RELEASE_SHA` runs. GitHub
reuses an existing tag and silently ignores `--target`, so the release would
publish notes describing `$RELEASE_SHA` against someone else's tag — and fire
`publish.yml` — before the post-condition notices.

Real fix: claim the tag atomically first via
`gh api POST /repos/{owner}/{repo}/git/refs -f ref=refs/tags/$TAG -f sha=$RELEASE_SHA`
(which fails if the ref exists), verify on failure, then create the Release
against the tag we own.

Caveat that makes this non-trivial: if `gh release create` then fails, we are
left with a tag and no Release — which runs straight into finding 2.

### 2. `detect` skips whenever the tag exists, so the tag-without-Release recovery is unreachable

`detect` sets `should_release=false` when the tag already exists, so a version
whose tag was pushed but whose Release was deleted or never created can never
be recovered by a fresh run — the recovery path #1401 added is dead code on
that path. Publishing stays stranded until someone intervenes by hand.

These two want fixing together: (1) is the reason a stray tag appears, and (2)
is why we cannot recover from one.

Found during the #1401 review; raised by codex (gpt-5.6) and confirmed by
reading the workflow.

PR #1524, opened 2026-08-06
Fix release-blocking chat, Responses, and tool parser regressions

raullenchai:
## What does this PR do?

Closes three P1 regressions found during the post-0.12.1 release review:

- emits one public Responses lifecycle before a buffered DeepSeek retry can stall, while keeping heartbeat and retry sequence numbers contiguous;
- preserves the original Desktop transcript as a sidebar branch before Edit or Retry truncates the active path;
- prevents `qwen3_coder_xml` from promoting prose, `tool_choice: none`, or undeclared names into executable tool calls across streaming, non-streaming, generic-fallback, and structured-call paths.

## Why is this needed?

The current release candidate can send `response.in_progress` before `response.created` during a slow hidden retry, can irreversibly delete later chat turns when replaying an older message, and can execute tool-shaped prose that the caller never authorized. These are release-blocking protocol, data-loss, and tool-authorization failures identified while reviewing #1506, #1509, and #1513.

## AI assistance disclosure

Codex reproduced the failures, wrote the implementation and regression tests, reviewed the final diff, and ran the validation listed below. Every changed path was manually inspected for lifecycle ordering, persistence behavior, and tool authorization boundaries.

> By submitting this PR I confirm I can explain the intent, risk, and behavior of every non-generated change in this PR. For any generated / boilerplate / scaffolded sections, I've identified them above and can describe how I verified them.

## Test plan

- Red/green spot-check: new Responses heartbeat test failed with `response.in_progress` first before the fix.
- Red/green spot-check: Qwen parser regressions failed for no tools, `tool_choice: none`, undeclared tools, and streaming before the fix.
- `pytest -q tests/test_responses_engine_failure_envelope.py tests/test_responses_route.py tests/test_tool_parser_content_passthrough.py tests/test_tool_call_closer_content_integrity.py tests/test_tool_call_streaming_parity.py tests/test_chat_route_tool_choice_enforcement.py tests/test_r12_reasoning_sanitizer_required.py tests/test_sanitize_output.py` (456 passed, 4 skipped)
- `swift test --no-parallel --filter 'ChatMessageActionsTests'` (4 passed)
- Quickstart/memory-warning Swift regression matrix (42 passed)
- `ruff check` and `ruff format --check` on changed Python files
- `python3 -m compileall -q vllm_mlx`
- `make release-smoke`
- Full local `pytest -q tests/ -x` collection on the host Python was blocked by its missing optional `anthropic` SDK; PR validation and GitHub CI run in project test-extras environments.

[...]

PR #1530, opened 2026-08-06
revert: DeepSeek Codex long-turn stabilization (#1506) — heartbeat lifecycle regression

raullenchai:
## Why

#1506 introduced a confirmed SSE-lifecycle regression on the exact surface it targets. During a DeepSeek-V4-Flash Codex turn, the new buffering wrapper (`_stream_responses_with_nonprogress_retry`) holds `response.created` / `response.in_progress` / all reasoning events in an internal buffer until the first *progress-committing* event (visible output text or a tool call). While buffered, it yields **nothing** to `_disconnect_guard`, and the outer `_resp_heartbeat_state` that the keepalive reads is left empty until commit.

So for any turn whose pre-visible-output reasoning phase exceeds the keepalive interval (`sse_keepalive_seconds`, default **20s** — i.e. the "long turns" this PR is about), `_disconnect_guard` stalls for 20s and emits `_responses_keepalive_sse({})`, producing a bare `event: response.in_progress` with **no `response` and no `sequence_number`**, delivered **before `response.created`** (still in the buffer). That is an invalid Responses lifecycle: a strict SDK consumer (openai-python `ResponseInProgressEvent` requires both fields, and `created` must precede `in_progress`) raises and tears down the stream.

Verified by tracing the live paths: `vllm_mlx/routes/responses.py:2405-2445` (buffer, zero downstream yield pre-commit) + `:1369`/`:2414-2438` (outer heartbeat state only populated on commit) + `vllm_mlx/service/helpers.py:3695-3744` (20s stall → `keepalive_factory()`) + `vllm_mlx/routes/responses.py:2243-2259` (bare frame from empty state).

A correct fix is design-level (flush a stable response envelope eagerly with a retry-stable `response_id`; buffer only content items) and cannot be a one-liner — falling back to a comment keepalive would defeat the parsed-event idle-watchdog that #1506 needs for long reasoning. Reverting to a clean baseline; the DeepSeek long-turn stabilization will be re-landed with the envelope kept out of the buffer.

## What

Revert `94179d92` in full — restores `vllm_mlx/routes/responses.py` to its pre-#1506 state and removes the two test files it added.

## Tests

- `pytest tests/test_responses_route.py` → 77 passed, 2 skipped.
- `pytest tests/ -k "responses or codex or deepseek" --ignore=tests/integrations` → 1241 passed, 15 skipped.
- `responses.py` is byte-identical to pre-#1506 (verified via diff against `94179d92^`).
- No code outside the reverted files references any symbol #1506 introduced (grep-verified), so the revert is self-contained.

## Notes

This reintroduces the original #1462 DeepSeek-Codex instability (mid-reasoning reconnects / out-of-order items / reasoning-only stops) that #1506 set out to fix. That work should return in a follow-up that keeps the `response.created`/`in_progress` envelope unbuffered so keepalives during long reasoning stay well-formed and correctly ordered.


raullenchai:
Superseded — not landing. The heartbeat lifecycle regression this revert targeted (bare `response.in_progress` before `response.created` during a >20s buffered DeepSeek reasoning phase) was already fixed on main by #1524, which eagerly flushes a stable public `response.created`/`response.in_progress` envelope and populates the keepalive's `heartbeat_state` up front (separate public/keepalive sequence counter), keeping #1506's buffering intact. Reverting #1506 now would regress #1524. Closing.

PR #1536, opened 2026-08-06
fix(tool-calling): harden XML argument boundaries

raullenchai:
## What
- preserve marker-shaped text inside Qwen3-Coder and Nemotron tool arguments
- prevent declared nested examples from becoming fabricated executable calls
- keep streaming and non-streaming parsing aligned across shared and separate wrappers
- retain bounded recovery for degraded/max-token XML emissions

## Why
Follow-up to #1518. XML-ish tool formats have no escaping layer, so first-marker regexes can truncate argument values, leak protocol text, or promote nested examples into unintended calls.

## Validation
- Codex adversarial review run repeatedly; all P1 findings addressed
- `ruff check` on changed Python files
- 963 focused/release tests: 935 passed, 28 skipped
- `git diff --check`

This is release-blocking hardening for tool-call correctness and safety.

### DIFF
--- a/tests/test_nemotron_tool_parser_fail_open.py
+++ b/tests/test_nemotron_tool_parser_fail_open.py
@@ -86,6 +86,46 @@ def test_multiple_functions_inside_one_wrapper_are_all_kept(parser):
     assert result.content is None
 
 
+def test_literal_declared_function_opener_cannot_fabricate_call(parser):
+    request = {
+        "tools": [
+            {
+                "type": "function",
+                "function": {
+                    "name": "note",
+                    "parameters": {
+                        "type": "object",
+                        "properties": {"body": {"type": "string"}},
+                    },
+                },
+            },
+            {
+                "type": "function",
+                "function": {
+                    "name": "delete",
+                    "parameters": {
+                        "type": "object",
+                        "properties": {"path": {"type": "string"}},
+                    },
+                },
+            },
+        ]
+    }
+    text = (
+        "<tool_call><function=note><parameter=body>"
+        'document <function=delete>{"path":"/"}</function> literally'
+        "</parameter></function></tool_call>"
+    )
+
+    result = parser.extract_tool_calls(text, request)
+
+    tc = _only_call(result)
+    assert tc["name"] == "note"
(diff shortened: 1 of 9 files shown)


raullenchai:
Closing: the defect this PR targets is not in the parser layer.

Instrumented the route with the raw `output.text` the engine hands over, for the case this PR exists for (a tool argument whose value contains a literal `</parameter>`, qwen3.6-27b-8bit, tools declared):

```
output.text = '<tool_call>\n<function=write_file>\n<parameter=path>\n/tmp/a.py\n</parameter>\n<parameter=content>\ndef f():\n    return "</parameter>\n</function>\n</tool_call>'
engine_tool_calls = None
```

The model itself terminated the value at `return "`. The full `"</parameter>"\n` was never generated, so no parser change can recover it. The same prompt through `/v1/completions` — no tools, therefore no tool grammar — emits the literal in full.

The cause is a deliberate, documented decision one layer up, `vllm_mlx/api/tool_grammar.py:417`:

```
# FIRST-</parameter> semantics: a value that literally contains </parameter>
# closes there (same as XGrammar — acceptable).
```

With tools declared, qwen3_coder_xml runs under an llguidance grammar whose string-value rule is "any text up to the first `</parameter>`". Emitting that literal inside a value is structurally impossible, and this matches XGrammar/vLLM/SGLang behaviour.

So this PR rewrites 457 lines of qwen3coder and adds streaming state for no observable gain on the production path, which is not a trade worth taking into a release. #1518 already fixed the paths the grammar does *not* shield (nemotron, hermes, and the `_decode_json_like` `.strip()` that hit every parser).

Two real findings from reviewing this PR are being carried into a separate change:
- `tool_call_scan.split_marked_parameters` splices wire markup into a valid value when an undeclared sibling parameter is present — `path` comes back as `'/ok\n</parameter>\n<parameter=mode>\nforce'`. Reproduced on this branch and on main.
- `nemotron_tool_parser.py` re-parses in the streaming close path without forwarding `request`, so #1518's declared-name gate is bypassed while streaming. `request` is already a parameter of the enclosing method and the call site passes it.

The grammar-level truncation is being filed as a known limitation rather than a defect.

PR #1538, opened 2026-08-06
fix(rapid-mac): prefer newer bundled runtime

### LINKED ISSUE #1503: 0.12.5 silently runs a stale runtime-override engine (0.11.1) instead of its bundled 0.12.4 — first run never completes
## Summary

A freshly downloaded, freshly installed **0.12.5** app ran inference engine
**0.11.1** — and the first-run flow never completed. The app silently prefers a
stale engine in a shared Application Support directory over the one inside its
own bundle, with nothing in the UI indicating which is in use.

Found by installing the published `rapid-mac-v0.12.5` DMG and driving the real
GUI as a new user.

## What happens

`ServerLocator` resolves the sidecar in this order:

```
RAPID_BIN  →  ~/Library/Application Support/Rapid/runtime-override/rapid-mlx/bin/rapid-mlx  →  <App>.app/Contents/Resources/rapid-mlx/bin/rapid-mlx
```

The override wins. On this machine one had been written on 2026-07-28 (393 MB,
engine **0.11.1**). Observed child process of the 0.12.5 app:

```
~/Library/Application Support/Rapid/runtime-override/rapid-mlx/python/bin/python3.12
    -m vllm_mlx.cli serve lfm2.5-1b-4bit --host 127.0.0.1 --port 8000
```

while `<App>.app/Contents/Resources/rapid-mlx/VERSION` reads `0.12.4`.

Two consequences:

1. **The engine the release notes promise is not the engine that runs.** 0.12.5
   states it bundles Rapid-MLX 0.12.4. Any user carrying an override gets
   whatever version that override froze at, indefinitely — no migration, no
   staleness check, no version comparison, no UI surface.
2. **On this machine it made first run unusable.** Onboarding →
   "Download & start" → the sidecar never became ready. Left >3.5 minutes on
   `Starting LFM2.5 · 1.2B… Loading the model into Metal. Usually 5–15 seconds.`
   with no progress, no elapsed time, no cancel, no error. The process sat at
   1.26 s CPU with all threads in `PyThread_acquire_lock_timed` /
   `_pthread_cond_wait` and 16 idle HTTPS sockets. The model was already
   complete in the HF cache; nothing was downloading.

## A/B, same build, same model, same machine, same clicks

Only the resolved engine differs:

| engine | outcome |
|---|---|
| runtime-override 0.11.1 (default resolution) | never became ready; >3.5 min, then abandoned |
| bundled 0.12.4 (pinned via `RAPID_BIN`) | **server up in ~9 s**, chat worked, 42 tok/s |

Pinning `RAPID_BIN` at the bundle's own binary is the whole difference, so the
stall is a property of the stale override, not of the app or the model.

## Also: the isolation harness does not isolate this

`scripts/dogfood-isolate.sh` rewrites the bundle identifier, which correctly
isolates CFPreferences and `Application Support/<bundle-id>/`. But the override
lives under the fixed name `Application Support/Rapid/`, so it is shared by
every install and every bundle id — including a "clean" dogfood profile. Every
first-run dogfood on a machine with an override has therefore been measuring
the override, not the bundle. That is how this reached a release.

## Asks

1. Compare versions before deferring to an override, and prefer the bundled
   engine when it is newer (or at minimum when the override predates the app's
   own build). A user who updates the app should not keep an older engine by
   default.
2. Surface which engine is live — Settings, About, or the same place the
   version appears. Today there is no way for a user or a support request to
   tell 0.11.1 from 0.12.4.
3. Bound the "Starting …" state. It promises 5–15 seconds and can sit forever;
   it needs an elapsed indication and a way out, the same way #1488 bounded the
   chat send.
4. Make `dogfood-isolate.sh` redirect or shadow `Application Support/Rapid/`
   too, so a dogfood profile cannot inherit a shared override.

_Found while dogfooding the published `rapid-mac-v0.12.5` DMG._

raullenchai:
## What does this PR do?

Makes Desktop compare the managed runtime override and bundled sidecar `VERSION` files when both executables exist. The explicit `RAPID_BIN` escape hatch remains highest priority; a same-version or newer runtime override still wins; slim DMGs with only an override retain their existing behavior.

Adds focused Swift tests for stale, newer, equal, missing, and malformed version metadata plus the `RAPID_BIN` and slim-DMG paths.

## Why is this needed?

Fixes #1503. Desktop 0.12.5 can silently launch a stale runtime override instead of its bundled 0.12.4 engine. On an affected machine this leaves first launch waiting on an old engine even though the updated app contains a working newer one.

Root cause: `ServerLocator` selected the first executable candidate in a fixed override-before-bundle list and never inspected the managed sidecars' version metadata.

## AI assistance disclosure

Codex wrote and reviewed both changed files: `ServerLocator.swift` and `ServerLocatorVersionSelectionTests.swift`. The result was verified against the release scripts' real sidecar layout and `VERSION` contract, inspected against the reported 0.10.8-to-0.12.4 case, passed Swift frontend syntax parsing plus `git diff --check`, and completed a convergent Codex review with no findings.

## Test plan

- [x] Swift frontend syntax parse passed for production and test files.
- [x] `git diff --check` passed.
- [x] Codex review converged with 0 P0 / 0 P1 / 0 findings; evidence is posted in the PR thread.
- [x] Supply-chain and description-quality gates passed in the first `pr_validate` run.
- [x] GitHub macOS CI will compile and execute the Swift suite before merge.
- [x] Full Desktop build and model dogfood are not required for this locator-only PR; release-layout paths were checked directly against `build.sh` and `build-sidecar-tarball.sh`.

## Checklist

- [x] New behavior has focused regression tests.
- [x] No dependency or workflow changes.
- [x] README changes are not applicable; `ServerLocator` behavior documentation was updated inline.
- [x] No breaking changes to existing API.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Server/ServerLocator.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ServerLocator.swift
@@ -26,7 +26,14 @@ import Foundation
 /// 1. **`RAPID_BIN`** — test / dev override env var. Highest priority so
 ///    integration tests can swap in a fake CLI without touching disk
 ///    and dev-checkout power users can point at their local build.
-/// 2. **runtime-override** — `~/Library/Application Support/Rapid/runtime-override/rapid-mlx/bin/rapid-mlx`
+/// 2. **managed sidecars, newest version wins** — compares the
+///    runtime-override and bundled `VERSION` files whenever both binaries
+///    exist. A stale or unversioned runtime override cannot shadow a newer,
+///    versioned sidecar shipped by an app update. The runtime override still
+///    wins when it is the same version or newer, and remains the only managed
+///    candidate for slim DMGs whose bundled slot is empty.
+///
+///    **runtime-override** — `~/Library/Application Support/Rapid/runtime-override/rapid-mlx/bin/rapid-mlx`
 ///    written by ``BootstrapCoordinator`` on first launch of the slim
 ///    bootstrapper DMG, or by the in-app updater when it pulls a
 ///    newer sidecar from `latest.json`. The `rapid-mlx/` wrapper
@@ -36,10 +43,10 @@ import Foundation
 ///    the same final ``rapid-mlx/bin/rapid-mlx`` suffix, so adding a
 ///    fourth lookup slot in the future stays symmetric. Survives
 ///    desktop upgrades because it lives outside `Contents/`.
-/// 3. **bundled** — `Contents/Resources/rapid-mlx/bin/rapid-mlx`
+///    **bundled** — `Contents/Resources/rapid-mlx/bin/rapid-mlx`
 ///    shipped inside a full-bundle DMG (today the slim DMG ships
 ///    empty here and relies entirely on the bootstrapper to
-///    populate slot 2). Last resort before returning nil.
+///    populate the runtime-override slot).
 ///
 /// Return value: ``nil`` when none of the three slots resolves —
 /// callers are expected to surface the missing-install UX (re-run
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex review round 1 converged with **0 P0 / 0 P1 / 0 findings**.

Reviewed against `origin/main` with the supported OpenAI `gpt-5.4` provider override. The reviewer checked precedence, version parsing, release-layout assumptions, and the new tests, and reported no concrete regression or correctness issue introduced by the patch.

Note: an earlier `gpt-5.6-sol` CLI attempt entered a read-only inspection loop and was terminated without producing a verdict or modifying files; it is not counted as a completed review round.

raullenchai:
Local `pr_validate` completed in an isolated Python 3.11 environment.

Results:

- PASS: fetch / blast-radius classification
- PASS: test-plan completeness
- PASS: PR description quality
- PASS: supply-chain scan
- PASS: test environment
- SKIP as expected for Swift-only diff: Python lint, targeted Python tests, stress bench
- Codex step skipped because the completed manual Codex review is documented above
- Broader unit suite: **16,157 passed**, 3 unrelated failures

The three failures were A/B replayed on a detached clean `origin/main` worktree using the identical venv and reproduce unchanged there:

- `tests/test_share_cli.py::test_share_native_double_dash_before_model_is_not_passthrough[argv1]`
- `tests/test_share_cli.py::test_non_share_positional_value_share_skips_probe_no_double_convert`
- `tests/test_telemetry_request_wiring.py::test_nonstreaming_completion_emits_request_event`

Therefore they are confirmed pre-existing/environment-specific failures, not introduced by this two-file Swift diff. Artifact: `/tmp/pr_validate/pr-1538/run-20260806T175935Z-47311-1786039175969733000/`.

raullenchai:
Post-merge targeted Swift verification while GitHub runners remain queued:

- `swift build --target RapidTests` — **PASS**. This compiled the production `Rapid` target and all 120 `RapidTests` sources, including `ServerLocatorVersionSelectionTests.swift`.
- `swift test --filter ServerLocatorVersionSelectionTests` — production target compiled, then the package runner stopped before executing tests because the unrelated `RapidUXTests` target imports `XCTest`, which this local toolchain cannot resolve (`no such module 'XCTest'`).

The locator implementation and new regression-test source therefore type-check successfully. Runtime execution remains delegated to the queued GitHub macOS workflow.

PR #1577, opened 2026-08-07
fix(rapid-mac): focus sidebar rename field, sort archived, tidy row menu

### LINKED ISSUE #1573: rapid-mac sidebar: rename field doesn't auto-focus; archived() doc overclaims sort order (#1568 follow-up)
Two non-blocking nits surfaced during #1568 review (codex), landed as follow-up rather than blocking the feature.

## 1. Inline rename field doesn't take keyboard focus
`SidebarView.renameField` inserts a `TextField` when `renamingID` is set, but never assigns focus. After choosing **Rename** from the menu, keystrokes and Return stay directed at the row/menu until the user manually clicks the field — a papercut on the headline action.

**Fix:** bind the editor to `@FocusState` and request focus when `renamingID` becomes non-nil (and clear on commit/cancel).

## 2. `archived(for:)` doc says "newest-updated first" but only filters
`SidebarView.archived(for:)` promises "newest-updated first" in its doc comment but only filters `isArchived`, so archived rows render in whatever order the input array happens to be in.

**Fix:** either sort the filtered result by `updatedAt` descending, or drop the ordering claim and document/enforce the input-order invariant at the call site (`chat.conversations`).

Both are cosmetic (no data loss, no crash). Source: `apps/rapid-mac/Sources/Rapid/UI/SidebarView.swift`.

osdodo:
## What does this PR do?

  Four follow-up nits on the sidebar conversation row from #1568:

  - Rename field auto-focuses (via `.task(id:)`, so it also re-focuses when Rename is picked on another row while an editor is open).
  - `SidebarView.archived(for:)` actually sorts newest-updated first.
  - Delete carries a `trash` glyph instead of an empty icon gutter.
  - Row-menu glyphs follow the highlighted row's text colour.

  `SidebarView.swift` plus its two test files only.

  ## Why is this needed?

  Fixes #1573.

  - Rename was effectively broken: the field appeared but keystrokes and Return went to the row that opened it, so it needed a second click before accepting input.
  - Archiving touches neither `updatedAt` nor a row's slot in `ChatViewModel.conversations`, so the archived group listed rows in arbitrary order — the doc comment already claimed newest-first.
  - On hover the menu label flipped to white while the icon stayed amber: the app-wide `.tint(brandAmber)` reaches the menu content, so SwiftUI hands AppKit coloured (non-template) menu-item images, and AppKit only recolours template images on highlight.
  - Delete's blank icon gutter reads as a missing asset next to four icon-bearing siblings.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/SidebarView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/SidebarView.swift
@@ -260,8 +260,17 @@ struct SidebarView: View {
 
     /// Archived rows, newest-updated first. Kept out of ``sections`` so the
     /// main list can never accidentally render one.
+    ///
+    /// The sort is done here rather than trusted from the caller: the main
+    /// list's order is maintained incrementally by ``ConversationOrdering``,
+    /// which deliberately leaves a row's position alone for non-activity edits
+    /// — and archiving is one of those. So an archived row keeps whatever slot
+    /// it held in the live list, which says nothing about its rank among the
+    /// other archived ones.
     static func archived(for conversations: [ChatConversation]) -> [ChatConversation] {
-        conversations.filter { $0.isArchived }
+        conversations
+            .filter { $0.isArchived }
+            .sorted { $0.updatedAt > $1.updatedAt }
     }
 
     private var archivedConversations: [ChatConversation] {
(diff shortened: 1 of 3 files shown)


raullenchai:
Apologies — I collided with this PR and it is my fault, not yours.

I was doing a click-every-control pass over `main` last night, hit the rename bug independently, and shipped a fix as #1586 without checking the open PR list first. #1577 was already open at that point. That is a process failure on my side and it is why this now shows conflicts.

**What #1586 landed**, so you can see exactly what is duplicated:

- `@FocusState` on the rename field, requested from `.task` after a `Task.yield()`, keyed on a per-edit session counter so picking Rename on a second row while an editor is open re-focuses correctly. That is the same mechanism as your `.task(id:)` — your diagnosis and mine agree.
- Two things beyond it: the field is only ~16pt tall inside a 30pt pill, and the painted pill was not hit-testable, so clicking the visible target *cleared* focus and dropped the keystrokes — a `.contentShape` + tap handler fixes that. And `ChatView`'s message editor had the identical defect (`editFieldFocused` requested from a `.task(id:)` on the enclosing bubble, which runs before the `TextEditor` exists).

**What #1586 did NOT touch, and what this PR should still land:**

1. **`archived(for:)` does not sort.** Still true on `main` today — the doc comment says "Archived rows, newest-updated first" and the body is a bare `filter`. Your fix and your reasoning about why (archiving touches neither `updatedAt` nor the row's slot) both stand.
2. **Delete's empty icon gutter** next to four icon-bearing siblings.
3. **Row-menu glyphs not following the highlighted row's text colour.** Your explanation — the app-wide `.tint(brandAmber)` reaches the menu content, so SwiftUI hands AppKit non-template images and AppKit only recolours template images on highlight — is the most useful thing in this PR and I would not have found it.

If you are up for it: rebase onto `main`, drop the focus change (now redundant), and keep 1–3. The conflict should collapse to almost nothing once the focus hunk is gone, since the rest is in different parts of the file. If you would rather not do the rebase, say so and I will do it and keep you as the author of the commits.

Either way I owe you the check I skipped. Sorry for the wasted work.

osdodo:
No harm done — thanks for the writeup, it saved me the diagnosis.

  I've rebased and force-pushed, so the conflict is gone. Dropped my focus change; yours is a superset (the hit-area and ChatView bits I'd missed). Kept the other three.

  One thing: the conflict wasn't small. rowMenuItems was the messy part — #1586 put endRename() / cancelRename() into every button body, and my change wraps the whole set in Group { }.tint(nil). Merged version keeps both.

  Also rewrote the comment #1586 left next to Delete: it said "keeping this body a bare pendingDeletion = conv", which stopped being true once the item got a label: closure.

Issue #1367, opened 2026-07-31
mlx-vlm text lane: small multimodal models (gemma-4-e2b, Ministral-3-2512) return incoherent text / hang


raullenchai:
## Summary
Serving certain **small multimodal** checkpoints for **plain text** through the mlx-vlm (VLM) lane produces broken output. Found while trying to add Gemma/Mistral legs to the L1 CI smoke (PR #1366) — their only *small* models are multimodal.

## Evidence — real macos-14 (M1) hosted runner, rapid-mlx 0.11.4, mlx-vlm 0.6.3
**`[vision]` installed → auto-routed to the mlx-vlm lane:**
- `gemma-4-e2b-4bit`: serve READY, coherence gate #1247 → **0/6 golden** (all wrong / incoherent).
- `Ministral-3-3B-Instruct-2512-4bit`: serve READY, first chat completion **hangs** — coherence request times out at 90s, server becomes unreachable.

**`--no-mllm` (force text-only mlx-lm lane):**
- `Ministral-3-2512`: **5/6** golden (one flaky miss) — usable.
- `gemma-4-e2b`: still **0/6** on macos-14 (M1) / 0.11.4 — but **6/6** on local **M3 / rapid-mlx 0.11.0**. So gemma-4-e2b text-lane coherence diverges by engine-version and/or hardware.

## Impact
- A user who installs `[vision]` and chats **text-only** with these small VLMs gets garbage (gemma) or a hang (Ministral): the vision lane's text-generation path is unsafe for these checkpoints.
- Blocks expanding the L1 CI smoke (PR #1366) past Qwen+Llama, since Gemma/Mistral only ship *small* models as multimodal.

## Repro
```
pip install -e ".[vision]"
rapid-mlx serve gemma-4-e2b-4bit  --port 8123 --no-thinking   # ask "capital of Japan?" -> garbage
rapid-mlx serve ministral-3b-4bit --port 8123 --no-thinking   # first completion hangs
```

## Asks
1. Fix/triage the mlx-vlm text-generation path for these small VLMs (or make a text-only request fall back to the LM backbone instead of hanging/garbling).
2. Root-cause the `gemma-4-e2b` text-lane **0/6 on 0.11.4/M1** vs **6/6 on 0.11.0/M3** — regression vs an M1 Metal-kernel issue.

## Workaround
Serve with `--no-mllm` to force the text-only mlx-lm lane (fixes Ministral; gemma-4-e2b still needs ask #2).

[...]

PR #1607, opened 2026-08-07
fix(models): stop offering video-generation models as chat models

### LINKED ISSUE #1603: Video-generation models are offered as chat models — a user can download 64 GB and reach a permanent dead end
## What

The chat model picker and Settings → Model Management both list the 8 video-generation aliases as ordinary chat models. Selecting one can never work: the sidecar exits before binding a port, and the app shows `Couldn't start <alias>. Try again, or pick a different model in the box below.` — an invitation to retry something that will fail identically forever.

The user can reach that state *after* downloading the model. Sizes run 13.3–64.3 GiB.

## Evidence, against the bundled engine in a built app

```
$ 'Rapid-MLX Desktop.app/Contents/Resources/rapid-mlx/bin/rapid-mlx' models | grep -iE 'cogvideo|wan2|ltx'
  cogvideox-fun-5b-bf16    19.6 GiB   —   —   ✗   unknown   —   —
  cogvideox-fun-5b-q4      13.3 GiB   —   —   ✗   unknown   —   —
  cogvideox-fun-5b-q8      15.5 GiB   —   —   ✗   unknown   —   —
  ltx-2.3-mlx-q4           21.2 GiB   —   —   ✗   unknown   —   —
  wan2.2-i2v-a14b-q8       39.7 GiB   —   —   ✗   unknown   —   —
  wan2.2-t2v-a14b-bf16     64.3 GiB   —   —   ✗   unknown   —   —
  …
```

Observed in the running app: `cogvideox-fun-5b-q4` renders in the composer picker under **All models**, with no sticker and no warning. Because it was cached it sorted to the top of the list with the "downloaded" checkmark — indistinguishable from a healthy text model.

## Why the existing filters miss it

The app has exactly one modality filter, in `ModelCatalog.parseAvailable`:

```swift
// apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift:316-317
if line.hasPrefix("Audio models") { continue }
if line.contains("[audio:") { continue }
```

That works only because the engine prints audio in a separate tagged section (`vllm_mlx/cli.py:5054`, `:5064`). Video aliases are emitted in the **main** table with no tag and no section, so this filter cannot see them. In the same build, 41 output lines carry the audio markers; zero video lines do.

`ModelPickerVisibility.knownBrokenForTextChat` (#1496) is a hardcoded 5-element denylist of vision SKUs (`ModelPickerVisibility.swift:66-79`) and names no video alias. Its doc comment states the scope is deliberately evidence-bounded to what #1367 measured, so it was never meant to generalise.

`ServerModelProfile.modality` (`ServerModelProfile.swift:61`) is decoded and never read. Its only fetch site runs *after* the model is already serving (`RapidApp.swift:211`, `:220`), so it could not gate the picker even if it were consulted. Its doc comment is also stale — it lists `"vision"`/`"image-gen"` as "reserved for upcoming integrations" and does not know `"video-gen"` ships today.

`SettingsModelManagementPanel.visibleEntries` (`:843-847`) applies neither filter — not the audio skip, not `ModelPickerVisibility` — so the vision aliases #1496 hid from the picker are still fully downloadable there too.

`ModelCatalog.parseCached` (`:373-396`) has no modality filter at all, so any cached non-text model re-enters the catalog regardless.

## What happens on select

Selection is unguarded (`ModelPickerBar.swift:1043-1047`); in `composerStyle` there is no Start CTA, so even the too-big confirmation never runs. The next send calls `ensureServing` → `rapid-mlx serve cogvideox-fun-5b-q4`, which hits the engine's only guard (`vllm_mlx/cli.py:2375` → `runtime/video_lane.py:88-130`), finds `mlx_arsenal` and `imageio` absent from the bundle, prints to **stderr** and `raise SystemExit(2)`. The port never binds; the UI shows the generic startup-failure copy (`ChatViewModel.swift:485`). The actionable stderr goes to the sidecar log the user has no terminal into.

Note this is the lucky path. `videox_fun_mlx` is already bundled; if the remaining deps were ever added, the server would boot into `VideoEngine` (`server.py:1839-1846`), which has no `is_mllm`, no `stream_chat` and no `tokenizer` — all three of which `/v1/chat/completions` touches (`routes/chat.py:3828`, `:4910`, `:4914`) — so the first send would 500 instead.

The correct signal already exists on the wire: `routes/models.py:410` returns `["video.generation"]` capabilities for these ids. The desktop reads neither `capabilities` nor `modality` — zero hits for either in `apps/rapid-mac/Sources/`.

## Suggested fix

Mirror the audio precedent rather than growing the denylist:

1. Tag video in `rapid-mlx models` output the way audio is tagged (a section or a `[video:…]` Kind marker).
2. Filter on that tag in `ModelCatalog.parseAvailable` **and** `parseCached`.
3. Apply the same filter to `SettingsModelManagementPanel.visibleEntries`, which currently protects against nothing.

Filtering in `ModelCatalog` covers both surfaces, since Model Management reads the same `catalog`.

## Scope

Pre-existing — not a regression from the current release train. Found during the pre-release GUI sweep.

Related: #1496, #1367.

raullenchai:
## Why

A `video-gen` alias has no tokenizer and no `stream_chat`, so it can never answer
`/v1/chat/completions`. The desktop listed all eight of them in the composer's model
picker and in Settings → Model Management as ordinary chat models — no sticker, no
warning. Selecting one spawns `rapid-mlx serve`, which exits 2 before binding a port
when the video extras are absent; the app then says *"Couldn't start `<alias>`. Try
again, or pick a different model"* — advice that will fail identically forever. The
engine's actionable stderr goes to the sidecar log the user has no terminal into.

The user can reach that state **after** downloading the model. The eight aliases run
13.3–64.3 GiB.

Observed against the bundled engine in a built app:

```
$ 'Rapid-MLX Desktop.app/Contents/Resources/rapid-mlx/bin/rapid-mlx' models | grep -iE 'cogvideo|wan2|ltx'
  cogvideox-fun-5b-q4      13.3 GiB   —   —   ✗   unknown   —   —
  wan2.2-t2v-a14b-bf16     64.3 GiB   —   —   ✗   unknown   —   —
  …
```

## Why nothing caught it

The app's only modality filter matches the `[audio:` Kind tag the engine prints for
audio aliases. Video rows were emitted in the **main** table with no tag and no
section, so that filter was structurally unable to see them — in the same build, 41
output lines carry the audio markers and zero video lines do.

`ModelPickerVisibility` (#1496) is a five-element denylist of vision SKUs,
deliberately evidence-bounded to what #1367 measured, and names no video alias.
`ServerModelProfile.modality` is decoded and never read; its only fetch site runs
*after* the model is already serving, so it could not gate the picker anyway.

## What changed

Mirrors the audio precedent rather than growing a denylist.

- **Engine** — `rapid-mlx models` renders video-gen aliases in their own
  `Video models (N aliases)` section, each row tagged `[video:gen]`. Every alias is
  still listed; the split is about letting a catalog consumer tell the kinds apart,
  not about hiding anything from the CLI.
- **App** — `ModelCatalog.parseAvailable` skips that section header and every tagged
  row, exactly as it already does for `[audio:`.
- **App** — new `parseExcludedAliases` reports what was withheld, and `load` refuses
  to re-admit those aliases off `rapid-mlx ls`.

That last part is load-bearing. `ls` carries no modality tag, and `load` re-admits any
cached alias with no row in `models` — which is *precisely* the state a
correctly-filtered video model is in. Without it the fix would be one-sided. It also
closes the same hole for audio, which is reachable today.

Filtering in `ModelCatalog` covers both surfaces, since Model Management reads the
same `catalog`.

## Verification

End-to-end on a running build, with `cogvideox-fun-5b-q4` **still 13.3 GiB in the HF
cache** — the case that previously sorted to the top of "All models" with a
downloaded checkmark:

| Surface | Before | After |
| --- | --- | --- |
| Composer model picker | `cogvideox-fun-5b-q4` listed under All models | gone; list starts `bonsai-27b-2bit`, `deepseek-r1-32b-4bit` |
| Settings → Model Management | video models listed with Download buttons | `No matches for "video"`, catalog **175 → 167** |

167 = 175 − 8, i.e. exactly the video aliases and nothing else.

Engine: `tests/test_cli_models.py` 29 passed, including a new test asserting no video
alias appears in the chat table and that every video row carries the tag. App: two new
tests in `DownloadCatalogHardeningTests`, mirroring the existing audio case and
covering `parseExcludedAliases`. `ruff check` + `ruff format --check` clean; app
`swift build` clean.

Note: `swift test` cannot run on my box (Command Line Tools only, no full Xcode — the
test target imports XCTest), so the app tests are validated by `rapid-mac-ci.yml`.

Closes #1603.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift
@@ -59,45 +59,21 @@ enum ModelCatalog {
         binary: URL,
         hubCacheOverride: URL? = ModelsFolderPreference.validatedOverrideURL()
     ) async -> [ModelEntry] {
-        async let availableTask: [(String, String?)] = listAvailable(binary: binary)
+        async let availableTask: (entries: [(String, String?)], excluded: Set<String>) =
+            listAvailableWithExclusions(binary: binary)
         async let cachedTask: [(String, String?, String?)] = listCached(
             binary: binary,
             hubCacheOverride: hubCacheOverride
         )
 
-        let available = await availableTask
+        let (available, excludedAliases) = await availableTask
         let cached = await cachedTask
 
-        // Build a quick lookup of cached aliases for the merge below.
-        var cachedIndex: [String: (hfRepo: String?, size: String?)] = [:]
-        for (alias, hf, size) in cached where !alias.isEmpty && alias != "(unmapped)" {
-            cachedIndex[alias] = (hf, size)
-        }
-
-        var entries: [ModelEntry] = []
-        var seenAliases: Set<String> = []
-        for (alias, hfHint) in available {
-            seenAliases.insert(alias)
-            let cachedHit = cachedIndex[alias]
-            entries.append(ModelEntry(
-                alias: alias,
-                hfRepo: cachedHit?.hfRepo ?? hfHint,
-                sizeOnDisk: cachedHit?.size,
-                cached: cachedHit != nil
-            ))
-        }
-        // A cached model with no row in ``rapid-mlx models`` is unusual
-        // but possible if the user pinned an alias by hand in their
-        // rapid-mlx config. Surface them anyway so they show up in the
-        // picker (otherwise the user can't pick them without typing).
-        for (alias, hf, size) in cached where !alias.isEmpty && alias != "(unmapped)" && !seenAliases.contains(alias) {
-            entries.append(ModelEntry(
-                alias: alias,
-                hfRepo: hf,
-                sizeOnDisk: size,
-                cached: true
-            ))
-        }
+        var entries = mergeAvailableAndCached(
+            available: available,
+            cached: cached,
+            excluded: excludedAliases
+        )
 
         // Repo-aware cached marking (issue #576): a bare alias like
         // ``qwen3-0.6b`` and its default-quant alias ``qwen3-0.6b-4bit``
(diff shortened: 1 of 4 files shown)


raullenchai:
Addressed both MINOR findings from review; both were correct.

**Substring matching was broader than the documented contract.** `line.contains("[audio:")` would drop any row whose HF id or description merely contained those characters. Replaced with `hasNonChatKindTag`, which requires a whole whitespace-delimited field of the shape the engine actually prints — `[audio:tts]`, `[audio:stt]`, `[video:gen]` — without hardcoding the subtypes, since the engine derives those from its registries and may extend them. `parseAvailable` and `parseExcludedAliases` now share it, so the audio path is tightened too.

Negative cases are pinned: `org/repo[audio:tts]x`, `org/audio:tts`, `[audio:]` and `[image:gen]` must all *not* match.

**The cache re-admission fix was untested — and you were right that reverting it left every test green.** The merge is now a pure `mergeAvailableAndCached(available:cached:excluded:)`, so the decisive condition is testable without spawning the engine. The new test feeds a cached video alias and a cached audio alias that have no `models` row, asserts both stay out, and asserts a hand-pinned *text* alias with no `models` row is still surfaced — the case that re-admission path exists for. It then re-runs the same input with `excluded: []` and asserts the video alias comes back, so the assertion is testing the condition rather than some unrelated filter.

Re-verified end-to-end on a patched running build after the change:

- Settings → Model Management, search `video` → `No matches`
- Settings → Model Management, search `kokoro` → `No matches` (audio path still filtered under the tighter matcher)
- catalog count unchanged at 167

`tests/test_cli_models.py` 29 passed; `swift build` clean; ruff clean.

PR #1621, opened 2026-08-07
test(mac): AX structural snapshot baselines for the GUI golden flows

raullenchai:
## Why

`scripts/gui-golden-flows.sh` already dumps the app's entire accessibility tree at every step, then throws it away after a couple of ad-hoc `jq` assertions. Nothing in the suite notices if a PR removes a button, reorders controls, drops an icon, renames an identifier or flips an enabled state — as long as the handful of identifiers the flows press are still there, everything is green.

This turns ten settled states into committed **structural baselines** so those regressions show up as a reviewable diff.

**Scope, stated plainly:** this is the cheap layer of appearance testing and it is *structural only*. It cannot see colour, spacing or typography. The PNG snapshots in `Tests/RapidTests/__Snapshots__` remain the pixel-level check.

## What

`scripts/ax-baseline.py` normalises a raw `rapid-ax dump` into an indented tree and compares it against `Tests/GUIGoldenFlows/__Snapshots__/<flow>.<state>.txt`.

**Keeps** — hierarchy/nesting (as indentation), `AXRole`, `AXSubrole`, `accessibilityIdentifier`, `AXTitle`/`AXDescription`/`AXHelp`, `AXEnabled`, sibling order below the window level, and the *kind* of each value (`bool:true` / `bool:false` for toggles, `number` / `text` / `empty` otherwise).

**Drops or rewrites** — and the reasoning matters, because getting this wrong is the whole failure mode:

| Dropped | Why |
| --- | --- |
| `bounds` (screen coordinates + layout size) | The window opens at a different origin every launch — two consecutive runs measured 715,107 and 725,145. |
| `pid` | New every launch. |
| Top-level window order | AX returns windows in z-order, so merely focusing Settings would rewrite the baseline. Windows are sorted by identifier/title instead. |
| Value *contents* | A token/s figure, an on-disk size, a streamed transcript — that's data, not structure. Presence and kind are recorded; the text is not. |
| Version numbers, byte sizes, token rates, durations, dates, clock times, UUIDs, `/Users/<name>` paths, the fake model alias | Scrubbed to placeholders wherever they appear in identifiers or text. `Settings.App.UpToDate` legitimately contains the release version and a conversation row identifier legitimately contains a fresh UUID; recording either verbatim makes the baseline useless. |

A sample of the committed output:

```
AXWindow subrole=AXStandardWindow id="main" title="Rapid-MLX"
  AXGroup subrole=AXHostingView
    AXSplitGroup id="main, SidebarNavigationSplitView"
      AXGroup subrole=AXHostingView
        AXButton id="Sidebar.NewChat" desc="New Chat" enabled=true
        AXButton id="Sidebar.Launch" desc="Launch" enabled=true
      AXSplitter value=number enabled=false
      ...
        AXButton id="Sidebar.Conversation.<uuid>" desc="golden crash marker" enabled=true
```

On mismatch it prints what changed, not a JSON dump:

```
ax-baseline: structural mismatch for fresh-install.steady
  appeared after baseline line 7:
    +           AXButton id="Sidebar.Launch" desc="Launch" enabled=true
  changed at baseline line 17:
    -             AXButton id="Readiness.PrimaryAction" desc="Start" enabled=true
    +             AXButton id="Readiness.Action" desc="Start" enabled=true
```

### Update path

```bash
./scripts/gui-golden-flows.sh --update-baselines
git diff apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__
```

A baseline that does not exist yet is recorded on first run and passes; overwriting an existing one always requires the flag. That mirrors `Tests/RapidTests/SnapshotHelpers.swift`, which already works this way for the PNG snapshots.

### Supporting changes

- `rapid-ax.swift` now emits `AXSubrole` and `AXEnabled`. Both are additive, so every existing `jq` assertion is untouched.
- `wait_send_idle` settles the composer before the chat / stop / crash-recovery fingerprints. `ChatView.SendOrStopButton` publishes `AXHelp` only while the readiness gate is closed (`accessibilityHint` is empty once `sendAllowed` is true), which makes the *absence* of that attribute a copy-independent "ready and not streaming" signal. It also captures the finished chat turn rather than a partial one — the existing loop breaks as soon as the transcript mentions "deterministic content", which the fake emits nine chunks before the stream ends.

No flow's existing assertions changed; baselines are taken from separate dumps.

## Evidence

Built with `SKIP_SIDECAR=1 BUNDLE_MODEL=0 ./scripts/build.sh`, run on an unlocked local session held awake with `caffeinate -dimsu` (`CGSSessionScreenIsLocked` absent; `kCGSSessionOnConsoleKey=Yes`).

Two full suites were run *before* writing the normaliser and their trees cross-compared, so the design was driven by measured volatility rather than guesswork. Then:

| Run | Mode | Result |
| --- | --- | --- |
| A | `--update-baselines` | PASS, 10 baselines recorded |
| B | compare | PASS, 10/10 byte-identical, zero baseline files modified |
| C | compare | PASS, 10/10 byte-identical, zero baseline files modified |

30/30 observed trees (10 states × 3 runs) are byte-identical to the committed baselines. A negative test that deleted a button, renamed an identifier and flipped an `enabled` state in a copy of a baseline produced exactly the three expected hunks.

`swift build` is clean and SwiftPM does not pick the new directory up (it already warns about the nine PNGs under `Tests/RapidTests/__Snapshots__`; `Tests/GUIGoldenFlows` is not in that list). `shellcheck` and `ruff check` / `ruff format --check` are clean.

## Known limits

- **`swift test` was not run.** It fails on this machine with `error: no such module 'XCTest'` from `Tests/RapidUXTests/ModelReadinessTests.swift` — the Command Line Tools toolchain limitation. That is pre-existing and unrelated to this change; I am not claiming the Swift suite passed.
- One state was genuinely unstable and is now handled rather than tolerated: `model-crash-recovery` captured mid-restart on roughly half of all runs, putting a transient "Starting …" readiness banner into one run's tree and not the next. Fixed by settling on a real readiness signal (`wait_send_idle`), not by sleeping or loosening the comparison. The other nine states were stable across two independent suites before any settling was added.
- The Settings > Models pane renders whatever is in the model catalogue, so adding a downloadable model to the catalogue will change `settings-persistence.models-*`. That is a real UI change and a legitimate reviewable diff, but worth knowing before it surprises someone.
- The Settings > App pane is deliberately not fingerprinted: its update-check state machine (`Checking` / `UpToDate` / `Unknown`) is network-dependent and would flap. No flow visits it today.

### DIFF
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.answered.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.answered.txt
@@ -1,0 +1,38 @@
+# rapid-mac AX structural baseline v1
+AXApplication title="Rapid-MLX"
+  AXWindow subrole=AXStandardWindow id="main" title="Rapid-MLX"
+    AXGroup subrole=AXHostingView
+      AXSplitGroup id="main, SidebarNavigationSplitView"
+        AXGroup subrole=AXHostingView
+          AXButton id="Sidebar.NewChat" desc="New Chat" enabled=true
+          AXButton id="Sidebar.Launch" desc="Launch" enabled=true
+          AXScrollArea
+            AXHeading desc="Today" enabled=true
+            AXButton id="pin" desc="Pin conversation" help="Pin conversation" enabled=true
+            AXMenuButton desc="Conversation actions" enabled=true
+            AXButton id="Sidebar.Conversation.<uuid>" desc="golden restore marker" enabled=true
+        AXSplitter value=number enabled=false
+        AXGroup subrole=AXHostingView
+          AXScrollArea
+            AXOpaqueProviderGroup subrole=AXOpaqueProviderList enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="doc.on.doc" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="pencil" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="doc.on.doc" desc="Copy response" help="Copy response" enabled=true
+              AXButton id="arrow.clockwise" desc="Retry response" enabled=true
+          AXStaticText value=text enabled=true
+          AXScrollArea
+            AXTextArea id="rapid.chat.compose" desc="Message compose field" value=empty
+          AXMenuButton id="ModelPickerBar.ModelMenu" desc="Model" help="Choose which model to run" value=text enabled=true
+          AXButton id="ChatView.SendOrStopButton" desc="Send message" enabled=false
+    AXToolbar
+      AXButton desc="Hide Sidebar" enabled=true
+    AXButton subrole=AXCloseButton enabled=true
+    AXButton subrole=AXFullScreenButton help="this button also has an action to zoom the window" enabled=true
+      AXGroup
+        AXGroup
+    AXButton subrole=AXMinimizeButton enabled=true
+    AXStaticText value=text enabled=true
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.answered.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.answered.txt
@@ -7,7 +7,7 @@ AXApplication title="Rapid-MLX"
           AXButton id="Sidebar.NewChat" desc="New Chat" enabled=true
           AXButton id="Sidebar.Launch" desc="Launch" enabled=true
           AXScrollArea
-            AXHeading desc="Today" enabled=true
+            AXHeading desc="<relative-day>" enabled=true
             AXButton id="pin" desc="Pin conversation" help="Pin conversation" enabled=true
             AXMenuButton desc="Conversation actions" enabled=true
             AXButton id="Sidebar.Conversation.<uuid>" desc="golden restore marker" enabled=true
@@ -32,7 +32,5 @@ AXApplication title="Rapid-MLX"
       AXButton desc="Hide Sidebar" enabled=true
     AXButton subrole=AXCloseButton enabled=true
     AXButton subrole=AXFullScreenButton help="this button also has an action to zoom the window" enabled=true
-      AXGroup
-        AXGroup
     AXButton subrole=AXMinimizeButton enabled=true
     AXStaticText value=text enabled=true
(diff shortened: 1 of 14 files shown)


raullenchai:
Verified this independently rather than taking the report on trust, and the verification
found a real bug — in the *pre-existing* flow, not in this PR's additions.

## The suite was passing on a message that was never sent

Running the full suite on a clean machine, `chat-restore` failed. It then passed on a
re-run: a flake, about one run in two. The underlying defect is worse than flakiness.

`start_model` gated on `SendOrStopButton.description == "Send message"` — which is the
button's label for the *entire* startup. Evidence from the failing run:

```
chat-draft.json     SendOrStopButton  desc="Send message"  help="fake-alias is still starting."
chat-complete.json  rapid.chat.compose  value="golden restore marker"   <-- draft NEVER consumed
```

and the fake sidecar's event log for that run, in full:

```json
{"event": "server_started", ...}
```

No `chat_request`. Send was pressed while the readiness gate was closed, the press was
dropped, and the draft stayed in the composer — where `assert_tree_text "golden restore
marker"` *found it* and passed. The run only died later, on the reply that never came,
which is exactly why it looked like a flake instead of a broken assertion.

Fixed by using this PR's own `wait_send_idle` as the readiness gate in `start_model`
(it is the right signal — `AXHelp` is published only while the gate is closed, so its
absence is copy-independent), and by requiring the composer to actually drain in
`send_prompt`, which turns a dropped press into a loud failure.

## The three normaliser findings from review, all confirmed against the artifacts first

- **BLOCKING (real).** `settings-root.txt:73` recorded `AXZoomButton` → *one* `AXGroup`;
  `models-idle.txt:73` recorded the same button → *two*, in the same run. Lazily realized
  AppKit chrome. Now dropped below window-control buttons; those two snapshots went from
  76 and 77 elements to 72 and 72.
- **`Today` baked into four baselines** — any run straddling local midnight turns them red
  at 00:00. Now rewritten.
- **A missing baseline silently passed.** `check` recorded-and-returned-0 whenever the
  file was absent, so a typo'd snapshot name passed CI green while comparing against
  nothing. Now fails; verified it exits 1 and writes no file.

The two MINORs (all `AXStaticText` collapsing to `value=text`, window sort hiding
activation) are real but are deliberate trade-offs; not addressed here.

## Validation

Record pass, then **two consecutive compare passes**: 5/5 flows PASS in both, 20 baseline
comparisons, zero drift, and neither compare run rewrote a baseline. `chat-restore` has
now passed three consecutive runs where it previously failed one in two.

Worth stating plainly, because it is the whole point of the exercise: the six baselines
in flows 1–2 matched **byte-for-byte** on a different machine state than the one that
recorded them, before I changed anything. The approach reproduces. It was the *journey*
assertions underneath it that were weak.

PR #1622, opened 2026-08-07
ci(rapid-mac): block new interactive controls that carry no accessibility identifier

raullenchai:
## Why

`apps/rapid-mac/scripts/gui-golden-flows.sh` drives the app through `rapid-ax.swift`, which finds and presses **every** element by `AXIdentifier` — nothing else. So the GUI suite's coverage ceiling *is* the set of controls that carry one, and that ceiling drops silently every time a feature ships an unlabelled control, because the app still works by hand and no check goes red.

This is not hypothetical:

* **Settings → Tools** renders three tool toggles, the web-search backend radio group, its API-key field + Save button and the browsing toggle. All are real, working `AXCheckBox` / `AXRadioButton` / `AXTextField` elements — none carries an identifier, so no automated flow can reach any of them.
* `apps/rapid-mac/docs/userflows.md` has carried *"Approval dialogs lack identifiers"* as an open item across several releases.
* A tree-wide sweep finds **96 unlabelled interactive controls** against 46 labelled ones. The harness can reach a third of the app's controls.

## What this adds

`scripts/check_rapid_mac_ax_identifiers.py`, wired as a new `accessibility-identifiers` job in `.github/workflows/rapid-mac-ci.yml` (ubuntu, stdlib-only, ~15 s, beside the existing macOS `swift build` job rather than inside it, so a missing identifier reads as its own red check with its own log).

**It fails a PR that ADDS an interactive SwiftUI control under `apps/rapid-mac/Sources/` with no `.accessibilityIdentifier(...)`.**

### How the detector works

1. **Mask.** Comments and string literals are blanked to spaces, preserving byte offsets and line structure. Handles line comments, *nested* block comments, escapes, `\(…)` interpolation, `"""…"""`, and raw strings with any number of `#` delimiters. So every brace, paren and dot the next step sees is real code.
2. **Find control heads.** `(?<![A-Za-z0-9_.])(Button|Toggle|Picker|DatePicker|ColorPicker|TextField|SecureField|TextEditor|Menu|Stepper|Slider|NavigationLink|Link)\s*[({]` — a control being *constructed*. The lookbehind excludes `SendButton`, `.pickerStyle`, `Foo.Button`; the trailing `[({]` excludes type references (`Button<Label>`, `struct Button: View`).
3. **Walk the real chain.** From the opening delimiter, balance delimiters, then keep consuming trailing closures (including Swift's second `} label: { … }` form) and `.modifier(…)` links until a token that cannot continue a postfix expression. So an identifier five modifiers down still counts — the app's longest real case is `ModelPickerBar.ModelMenu`, **158 lines** below its `Menu`, and it is correctly credited.
4. **Scope to the diff.** Report only controls whose declaration line the diff **added**, then subtract carry-overs (below).

**Exact failure mode:** exit 1, one `::error file=…,line=…` GitHub annotation per finding (inline on the PR diff) plus a plain-text block naming the file, line, the declaration source line, and the reason; then a stderr block explaining the fix and the opt-out. Exit 0 prints `PASS` or `nothing to check`.

### Escape hatch that costs something

`confirmationDialog` / `alert` buttons genuinely render outside the app's AX tree. Opt out on the control's line or the line directly above:

```swift
// ax-exempt: confirmationDialog buttons render outside the app's AX tree
Button("Allow once") { approve() }
```

The reason is **mandatory** (≥10 chars) — a bare `// ax-exempt:` fails with its own distinct message, so the hatch can't be used by reflex. Every opt-out is greppable: `rg ax-exempt apps/rapid-mac`.

### Scoping decisions

* **Added lines only.** Failing on the 96-control backlog would make this un-landable and it would be off within a week. `--audit` enumerates the backlog for whoever wants to chip at it.
* **Carry-over suppression.** Multiset over `(kind, whitespace-normalised declaration line)` per file, against the base blob. Re-indenting or moving known-bad code is absorbed; **copying** an unlabelled control is not, because the base only funds one. Untouched occurrences claim the base budget before added ones.
* **Rename-aware** (`git diff -M`), so moving a panel full of known gaps isn't read as adding a panel full of new ones.

## Evidence — both directions

All six scenarios run against `SettingsToolsPanel.swift` on a real branch, gate invoked exactly as CI does.

| # | Scenario | Expected | Got |
|---|---|---|---|
| A | Add a brand-new unlabelled `Toggle` | FAIL | **exit 1** |
| B | Same `Toggle`, identifier at the end of a 3-modifier chain | PASS | **exit 0** |
| C | Same `Toggle`, `// ax-exempt:` **with** a written reason | PASS | **exit 0** |
| D | Same `Toggle`, bare `// ax-exempt:` (no reason) | FAIL | **exit 1** |
| E | Pure re-indent of an **existing** unlabelled toggle (its declaration line *is* in the diff's added lines) | PASS | **exit 0** |
| F | Verbatim **duplicate** of an existing unlabelled toggle | FAIL | **exit 1** |

<details>
<summary>Raw output</summary>

```
======== A  new unlabelled Toggle           (expect FAIL) ========
[ax-identifier-gate] base 481272ea804c, 1 changed file(s):
    apps/rapid-mac/Sources/Rapid/UI/SettingsToolsPanel.swift
::error file=apps/rapid-mac/Sources/Rapid/UI/SettingsToolsPanel.swift,line=214::Toggle: no .accessibilityIdentifier(…) on this control
  apps/rapid-mac/Sources/Rapid/UI/SettingsToolsPanel.swift:214
    Toggle(isOn: browseBlockDownloadsBinding) {
    -> no .accessibilityIdentifier(…) on this control
EXIT=1
======== B  same Toggle, identifier added   (expect PASS) ========
[ax-identifier-gate] PASS — every control this PR adds is reachable by AXIdentifier.
EXIT=0
======== C  opt-out WITH a reason           (expect PASS) ========
[ax-identifier-gate] PASS — every control this PR adds is reachable by AXIdentifier.
EXIT=0
======== D  bare '// ax-exempt:'            (expect FAIL) ========
::error file=…/SettingsToolsPanel.swift,line=215::Toggle: 'ax-exempt:' marker with no usable reason — write at least 10 characters saying why this control cannot carry an identifier
EXIT=1
======== E  re-indent an EXISTING gap       (expect PASS) ========
[ax-identifier-gate] PASS — every control this PR adds is reachable by AXIdentifier.
EXIT=0
======== F  DUPLICATE an existing gap       (expect FAIL) ========
::error file=…/SettingsToolsPanel.swift,line=214::Toggle: no .accessibilityIdentifier(…) on this control
EXIT=1
```
</details>

A self-review pass also caught a second bug before push: the comment capture joined block-comment characters with spaces, which shredded `ax-exempt:` inside a `/* … */` into single letters and made the marker silently unrecognisable in block-comment form. Fixed and pinned by two tests, one of which also nails down the deliberately narrow lookup (the control's line or the one directly above — the same rule as `# noqa` / `swiftlint:disable:next`, not a bespoke proximity heuristic).

**Scenario F found a real bug in my own first cut** and is now pinned by two unit tests: the naive in-order suppression let a pasted copy spend the base budget while the untouched original absorbed the blame, so duplication passed. Fixed by letting untouched occurrences claim the budget first. Re-running the historical replay after the fix turned up 6 more genuine findings that the buggy version had been laundering (both `Cancel` buttons in an alert instead of one, etc.).

### Does it fire on today's `main`? Yes — and every fire is a genuine gap.

I replayed the gate over **the last 40 real merged commits** that touched `apps/rapid-mac/Sources` (`--base-ref <sha>~1 --head-ref <sha>`):

* **33 / 40 pass clean.**
* **7 / 40 would have been blocked**, 30 findings total. I read all 30. **Zero are detector errors** — in every case the control has no `.accessibilityIdentifier` anywhere in its chain, on the container, or on any caller.

| Commit | Findings | What they are |
|---|---|---|
| `c2e40a1b` #1577 (**HEAD of main**) | 1 | The sidebar row menu's **Delete** item. The confirmation sheet it opens has identifiers (`DeleteSessionConfirm.*`); the menu item that opens it does not, so a flow can't get to the sheet. |
| `bb6e90d4` #? web tools | 9 | The six `SettingsToolsPanel` controls named in this PR's motivation, plus 3 dialog buttons. **The gate would have caught the documented gap on the commit that created it.** |
| `12b76172` pin/rename/archive | 6 | Sidebar row `Menu`, its 3 items, the inline rename `TextField`. |
| `a06ced1f` visual system | 9 | `QuietIconButton` / `RapidTextField` / `InlineNotice` wrapper bodies + 6 real controls. |
| `8b9d4dc4`, `2a1900fe` | 2 each | `confirmationDialog` / `alert` buttons — exactly the `ax-exempt` case, ~1 comment line each. |
| `f2097169` | 1 | The inline message-edit `TextEditor`. |

Roughly half the historical fires are dialog buttons (one comment line each); the rest are coverage gaps a reviewer would genuinely want to see. Tree-wide audit for context: **142 in-scope controls, 46 labelled, 96 not, 4 in deliberately-skipped scopes.**

## What it deliberately does NOT catch

Precision over recall — a gate that cries wolf gets ignored, and an ignored gate is worse than none because it manufactures confidence. Not detected:

* Interactivity bolted onto a non-control view: `.onTapGesture`, `.gesture`, `.contextMenu`, `.swipeActions`, `.draggable`.
* Bespoke `View` structs that are interactive without naming a SwiftUI control type at the point of use. (The wrapper's own `Button` is still checked where the wrapper is *defined*.)
* AppKit bridged through `NSViewRepresentable` — the chat composer's `AutosizingTextView` is the app's real example, and `rapid.chat.compose` gets its identifier from the AppKit side.
* `Commands` / `CommandGroup` / `CommandMenu` items — skipped on purpose. `rapid-ax.swift` explicitly refuses to walk the menu bar and `gui-golden-flows.sh` drives menus by title through `peekaboo menu click`, so an identifier there would be decoration.
* `ButtonStyle` / `ToggleStyle` `makeBody(configuration:)` bodies — skipped on purpose. An identifier there is stamped on every caller adopting the style, which is worse than none. (`TrailingSettingsToggleStyle` is the live case.)
* Anything outside `apps/rapid-mac/Sources/`.
* Renaming an *identifier* to a duplicate of an existing one. Uniqueness is a separate concern.

## Where it could misfire — say it now, not later

1. **Reusable wrapper components** (`QuietIconButton`, `RapidTextField`, `InlineNotice`). Today no caller labels them, so flagging the wrapper is correct. But if a caller ever writes `QuietIconButton(…).accessibilityIdentifier("X")` — which does work, SwiftUI applies it to the wrapper's root — then editing the wrapper's own `Button` line would flag a control that *is* reachable. **This is the most likely false positive.** Fix is either an `identifier:` parameter forwarded by the wrapper (better — self-documenting) or `// ax-exempt: callers attach the identifier at the call site`.
2. **Identifier on the container, not the control.** `HStack { Button(…) }.accessibilityIdentifier("Row")` flags the Button. I judged this correct (`AXPress` needs the control) but it is a defensible disagreement.
3. **Editing the declaration line of an existing unlabelled control.** Carry-over suppression absorbs pure moves/re-indents, but changing the text — e.g. renaming the action in `Button("Save") { commitKey(…) }` — reads as new. "You touched it, label it" is arguably right, but it is a surprise the first time.
4. **Chain-walk over-consumption.** The walker errs toward consuming *too much*, which can only cause a miss, never a false alarm. Verified on the real tree: 46 labelled controls credited to **46 distinct** `.accessibilityIdentifier` sites — no control borrows a sibling's.
5. **A malformed/unclosed literal** could desync the masker. `_consume_balanced` bails at the first mismatched closer rather than running to EOF, and a masker-invariant test (length + newline count + only-blanks-to-spaces) runs over every real Swift file.

## Tests

39 unit tests in `tests/test_rapid_mac_ax_identifiers.py` (stdlib + pytest; no git, no Swift toolchain, no GPU), running in the existing `test-matrix` job. Covers the core contract, chain-walk edge cases, every non-construction token shape, comment/string/raw-string/interpolation/nested-block-comment masking, the `Commands` and `makeBody` skips, all four escape-hatch states, and the carry-over multiset — plus a soak over the real `Sources` tree.

**Recall check:** every bare occurrence of one of the 13 control-kind tokens in the entire `Sources` tree, outside strings and comments, is recognised as a control head — **0 missed**.

## Verification run locally

* `python3.12 -m pytest tests/test_rapid_mac_ax_identifiers.py` — 39 passed
* `ruff check vllm_mlx/ tests/` — clean; `ruff format --check vllm_mlx/ tests/` — 806 files already formatted
* `python scripts/check_gha_pinning.py` — 14 workflows clean
* `python scripts/check_workflow_expressions.py` — OK (14 files)
* The gate against this PR's own branch — `nothing to check` (exit 0), since this PR touches no `Sources` Swift
* `swift build` in `apps/rapid-mac` — **Build complete (48.74s)**, exit 0

**`swift test` does NOT pass locally, and I am not claiming otherwise.** This machine has Command Line Tools only (`xcode-select -p` → `/Library/Developer/CommandLineTools`), which is exactly the documented limitation:

```
Tests/RapidUXTests/ModelReadinessTests.swift:1:8: error: no such module 'XCTest'
error: fatalError
```

That is an environment limit, not a regression from this PR — the diff contains **zero** Swift changes, so the `build` job's result is by construction identical to `main`'s. CI on `macos-15` with full Xcode is the authority here.

Closes nothing; the 96-control backlog is intentionally left open and is now enumerable with `--audit`.

### DIFF
--- a/.github/workflows/rapid-mac-ci.yml
+++ b/.github/workflows/rapid-mac-ci.yml
@@ -11,6 +11,11 @@
 # and `swift test` can't resolve `import Testing` in this toolchain). Keep the
 # full-app compile gate and run focused dependency-free contract executables for
 # regressions that can be isolated from the removed test stack.
+#
+# Second gate: `accessibility-identifiers` keeps the GUI golden-flow harness from
+# silently losing coverage. It fails a PR that ADDS an interactive control with
+# no `.accessibilityIdentifier(...)` — see the script's module docstring for the
+# rationale, the escape hatch, and the shapes it deliberately does not detect.
 name: rapid-mac CI
 
 on:
@@ -19,6 +24,8 @@ on:
     paths:
       - "apps/rapid-mac/**"
       - ".github/workflows/rapid-mac-ci.yml"
+      # The gate script itself, so a change to the detector re-runs it.
+      - "scripts/check_rapid_mac_ax_identifiers.py"
 
 permissions:
   contents: read
@@ -29,6 +36,41 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Pure text lint over the diff — no Swift toolchain, no simulator, no app
+  # build. It runs on ubuntu in ~15s beside the multi-minute macOS build rather
+  # than inside it, so a missing identifier reads as its own red check with its
+  # own log instead of hiding at the bottom of a compile job.
+  accessibility-identifiers:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+        with:
+          # The gate is scoped to lines this PR ADDED, which needs the
+          # merge-base with the target branch — not a shallow single commit.
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5
+        with:
+          python-version: "3.12"
+
+      # checkout@v4 with fetch-depth 0 normally leaves origin/<base> in place;
+      # this is the same belt-and-braces fetch the mlx-bound-guard job in ci.yml
+      # does, so the merge-base lookup cannot fail on a runner that resolved
+      # refs differently.
+      - name: Fetch base branch
+        env:
+          BASE_BRANCH: ${{ github.base_ref || 'main' }}
+        run: git fetch --no-tags origin "${BASE_BRANCH}:refs/remotes/origin/${BASE_BRANCH}"
+
+      # Stdlib only — deliberately no pip install, so the gate cannot be taken
+      # down by an unrelated dependency resolution failure.
+      - name: Interactive controls carry an accessibility identifier
+        run: |
+          python scripts/check_rapid_mac_ax_identifiers.py \
+            --base-ref "origin/${{ github.base_ref || 'main' }}"
+
   build:
     # macos-15 → default Xcode 16 → Swift 6.0, matching Package.swift's
     # swift-tools-version and the release workflow's runner. macos-latest still
--- a/.github/workflows/rapid-mac-ci.yml
+++ b/.github/workflows/rapid-mac-ci.yml
@@ -24,8 +24,11 @@ on:
     paths:
       - "apps/rapid-mac/**"
       - ".github/workflows/rapid-mac-ci.yml"
-      # The gate script itself, so a change to the detector re-runs it.
+      # The gate script itself, so a change to the detector re-runs it...
       - "scripts/check_rapid_mac_ax_identifiers.py"
+      # ...and its regression suite, which is the only thing that runs when a
+      # PR touches the detector but no Swift at all.
+      - "tests/test_rapid_mac_ax_identifiers.py"
 
 permissions:
   contents: read
@@ -70,6 +73,36 @@ jobs:
         run: |
           python scripts/check_rapid_mac_ax_identifiers.py \
             --base-ref "origin/${{ github.base_ref || 'main' }}"
+
+
+  # The detector job above is a NO-OP on a PR that changes no Swift — which is
+  # exactly the shape of a PR that edits the detector. Without this suite, the
+  # wrapper-identifier fix, the qualifier fix or the carry-over accounting could
+  # all be reverted in a script-only PR with every job green, while the test
+  # pinning them failed locally.
+  #
+  # Its OWN job, not a step in the one above, because it needs pytest from PyPI
+  # and that job is deliberately stdlib-only: a registry incident must be able
+  # to redden the tests without also taking down the gate that needs no
+  # network at all.
+  accessibility-identifier-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5
+        with:
+          python-version: "3.12"
+
+      - name: Install pytest
+        run: pip install pytest
+
+      # ``-q`` but not ``-x``: a change to the detector usually breaks several
+      # pinned cases at once, and seeing all of them is the point.
+      - name: Gate regression suite
+        run: pytest tests/test_rapid_mac_ax_identifiers.py -q
 
   build:
     # macos-15 → default Xcode 16 → Swift 6.0, matching Package.swift's
--- a/apps/rapid-mac/docs/gui-golden-flows.md
+++ b/apps/rapid-mac/docs/gui-golden-flows.md
@@ -285,3 +285,52 @@ Both halves are now fixed: `start_model` waits on `wait_send_idle`, and
 `send_prompt` requires the composer to actually drain. The general rule: an
 assertion that a string is present anywhere is satisfied by the input field,
 the placeholder, the tooltip and the sidebar. Say *which element*.
+## The identifier gate
+
+Because every flow above finds its target by `AXIdentifier`, this suite's
+ceiling is exactly the set of controls that carry one — and that ceiling drops
+silently every time a feature ships an unlabelled control, because the app still
+works by hand. `scripts/check_rapid_mac_ax_identifiers.py` (wired into the
+`accessibility-identifiers` job in `.github/workflows/rapid-mac-ci.yml`) fails a
+PR that **adds** an interactive control under `apps/rapid-mac/Sources/` with no
+`.accessibilityIdentifier(...)`.
+
+It is scoped to lines the diff added. The pre-existing backlog is deliberately
+out of scope — a gate that failed on it would be un-landable, and a disabled
+gate is worse than none. `--audit` lists that backlog when you want to chip at
+it:
+
+```bash
+python scripts/check_rapid_mac_ax_identifiers.py --audit
+python scripts/check_rapid_mac_ax_identifiers.py --base-ref origin/main   # what CI runs
+```
+
+Name new identifiers with the existing `<Surface>.<Thing>` convention (the
+inventory lives in `docs/userflows.md`), and put them on the control itself —
+an identifier on the enclosing `HStack` does not give `AXPress` anything to
+press.
+
+### Opting out
+
+There is currently **no** known control on this surface that cannot carry an
+identifier. `confirmationDialog` / `alert` buttons were the standing suspicion —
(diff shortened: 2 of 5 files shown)


raullenchai:
Independent review (codex, gpt-5.6-sol, high effort) found three blocking gaps. I have
verified the gate's *positive* behaviour myself — replaying it on `bb6e90d4` exits 1 and
reports exactly the six `SettingsToolsPanel` controls, and that file genuinely carries
zero identifiers at that commit, so the detector is not vacuous. The 39 unit tests pass.

These are false-*negatives* and false-*positives* around the edges, not a refutation of
the approach. Recording them so they don't get lost:

### BLOCKING

1. **A descendant's identifier satisfies its unlabelled parent.** `_expression_end`
   spans the trailing-closure body, then the identifier search covers that whole range:

   ```swift
   Menu {
       Button("Rename") { rename() }
           .accessibilityIdentifier("Sidebar.Rename")
   } label: { Text("Actions") }
   ```

   Zero violations, though `Menu` itself is unaddressable. The existing
   multiple-trailing-closure test misses it because it labels *both*.

2. **Fixing an old gap frees its carry-over budget to hide a new one.** Label an existing
   unlabelled `Button` and add a second unlabelled one in the same commit: the original
   drops out of `head_violations`, the new one consumes the base counter, and it is
   suppressed. This contradicts scenario F's "a pasted duplicate is reported".

3. **Carry-over identity is the raw declaration line**, so editing *any* of the 96 known
   gaps — even just its trailing comment — reads as new and blocks the PR.

### MAJOR

- **Deleting an identifier passes.** A deletion-only hunk produces no added declaration
  line, so `candidates` is empty. Removing `.accessibilityIdentifier(…)` shrinks GUI
  coverage silently — the exact thing this gate exists to prevent.
- **`DisclosureGroup` is not in the control inventory**, and the app already uses it
  (`ChatView.swift:940`, `BenchmarkView.swift:292`).
- A regex literal containing a quote (`let quote = #/"/#`) desynchronises the string
  masker and hides every control after it in the file.
- The dedicated CI job never runs the detector's own tests: a PR touching only the
  detector triggers the workflow, sees no Swift changes, and exits 0.

### MINOR

- `func Button(...)` / `case Button(String)` match the construction regex.

Holding this PR until 1, 2 and the deletion case are closed. A gate that can be
walked around by deleting a line, or that blocks a comment edit, will get switched
off by the first person it inconveniences — and a gate nobody trusts is worse than
no gate, because it launders the absence of checking into a green check.

raullenchai:
`pr_validate` reports DO NOT MERGE on `full_unit` for one test: `test_batched_faster_than_sequential`. A/B classified as pre-existing and environmental, not this PR:

- the test alone: passes on `main` and on this branch (3 runs)
- the whole file: 10 passed / 1 xpassed, identical on both
- **the whole suite on clean `main`: fails the same way**

Filed as #1664. Every other step passes, CI is green, and codex converged to 0 BLOCKING / 0 MAJOR / 0 MINOR.

Issue #1650, opened 2026-08-07
qwen3.6-35b-4bit adds 648 MB swap on 32 GB M2 Pro during recommendation benchmark


raullenchai:
## Reproduction

Latest `main` `850f62135f9c0c718aa493aea47d266443795cc9`, Rapid-MLX 0.12.7, MLX 0.31.2, mlx-lm 0.31.3, Mac mini M2 Pro 32 GB, macOS 26.5.2:

```bash
python scripts/benchmark_model_recommendations.py qwen3.6-35b-4bit \
  --hf-cache /Volumes/mac-storage/hf-cache \
  --output /tmp/qwen36-35b.json
```

## Result

- load: 155.0 s (weights on Jetson SMB; load time is not relevant)
- idle process footprint: 19.0 GB
- short decode: 59.86 tok/s
- new swap after short request: **648.4 MB**
- harness status: `aborted_swap` (threshold 256 MB), so the 8K request was intentionally not attempted

The same current-HEAD sweep ran `qwen3.5-35b-4bit` through the full 8K request at 22.0 GB peak, 287.09 prompt tok/s, 21.14 decode tok/s, and only 4.4 MB new swap. This therefore appears specific to the Qwen 3.6 35B alias/path and is not the already-fixed #1634 regression.

## Expected

On 32 GB, either keep new swap below the safety threshold or reject/mark the alias as requiring a higher RAM floor before attempting inference. The recommendation table will keep this alias above the 32 GB tier until resolved.

raullenchai:
Verified on current main `430b9a6c`: the unsafe 32 GB recommendation no longer exists.

- The committed M2 Pro measurement still records `qwen3.6-35b-4bit` aborting after the short request with 648 MB new swap.
- `RAMBucketedDefault` excludes the alias from the 32–47 GB tier and first offers it at the 48 GB floor.
- `swift apps/rapid-mac/scripts/verify-recommendation-tiers.swift` passes the complete recommendation contract, including 32 GB → Gemma 4 26B / Qwen 3.5 4B and 48 GB → Qwen 3.6 35B fast pick.

The floor change landed as part of #1652 (`9bc6d479`). This satisfies the issue’s safe alternative (raise the RAM floor rather than attempt inference on 32 GB), so no additional production diff is needed.

Issue #1648, opened 2026-08-07
benchmark recommendations: default Gemma 12B row cannot start from base install


raullenchai:
## Reproduction

On a clean base-wheel environment from current `main` (`850f621`, Rapid-MLX 0.12.7):

```bash
python scripts/benchmark_model_recommendations.py gemma-4-12b-4bit \
  --hf-cache /Volumes/mac-storage/hf-cache \
  --output /tmp/gemma12.json
```

## Actual

The benchmark records `status=error`, and the serve log says the alias requires the optional `mlx-vlm` dependency or `--no-mllm`.

## Expected

The default benchmark matrix must run from the documented base install without requiring the large vision extra. `gemma-4-12b-4bit` should receive the text-only `--no-mllm` serve argument, analogous to the existing Gemma 26B special case.

## Environment

- Mac mini Mac14,12 / Apple M2 Pro / 32 GB
- macOS 26.5.2
- Rapid-MLX 0.12.7 at `850f621`
- MLX 0.31.2 / mlx-lm 0.31.3
- weights from the Jetson-backed HF cache

Discovered while refreshing the release recommendation measurement matrix.

raullenchai:
Fixed and validated in #1651. The default recommendation benchmark now launches gemma-4-12b-4bit with --no-mllm; the test asserts the real spawned argv, and the Mac mini retry completed both short and ~8K requests (11.0 GB peak, 104.75 prompt tok/s, 19.01 decode tok/s, 0 MB new swap).

PR #1651, opened 2026-08-07
Refresh M2 Pro recommendation measurements

raullenchai:
## What changed

- refresh the release recommendation measurement table on the 32 GB M2 Pro Mac mini using Rapid-MLX 0.12.7 at `850f621`
- record eleven current candidate models, including successful full 8K runs for Qwen 3.5 35B and Gemma 12B and the Qwen 3.6 35B swap abort
- make the default Gemma 12B benchmark use the text-only loader (`--no-mllm`)
- add a regression test for the default Gemma 12B launch arguments

## Why

The prior table used Rapid-MLX 0.12.5 and incorrectly treated the fixed Qwen 3.5 35B result as still aborted. The current default harness also could not start Gemma 12B from a base install because it implicitly selected the optional multimodal loader.

Jetson-backed SMB storage was used for weights not present on the mini's internal disk. Those rows explicitly mark remote-cache load times as non-comparable; steady-state memory and throughput are measured after load through the real server path.

## Validation

- `python -m pytest -q tests/test_benchmark_model_recommendations.py` — 5 passed
- `python -m json.tool docs/benchmarks/model-recommendation-measurements.json`
- `python scripts/benchmark_model_recommendations.py --help`
- full short + ~8K server benchmark sweep on Mac mini M2 Pro 32 GB
- Gemma 12B retry after fix: 11.0 GB peak, 104.75 prompt tok/s, 19.01 decode tok/s, 0 MB new swap

Tracks #1648. The newly isolated Qwen 3.6 35B swap behavior is filed as #1650.

### DIFF
--- a/docs/benchmarks/model-recommendation-measurements.json
+++ b/docs/benchmarks/model-recommendation-measurements.json
@@ -1,45 +1,175 @@
 {
   "schema_version": 1,
   "environment": {
-    "timestamp_utc": "2026-08-07T20:28:07Z",
-    "hostname": "Raullens-Mini",
-    "machine": "Mac mini",
-    "chip": "Apple M2 Pro (10-core)",
+    "timestamp_utc": "2026-08-07T22:27:20Z",
+    "hostname": "Raullens-Mini.attlocal.net",
     "chip_model": "Mac14,12",
     "physical_ram_gb": 32,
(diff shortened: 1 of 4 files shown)


raullenchai:
@codex review

PR #1662, opened 2026-08-08
fix(mac): Browse all models opens the catalogue instead of closing the wizard

### LINKED ISSUE #1653: Desktop: "Browse all models" is a dismiss button — it never opens anything
## App version

0.12.7 (151)

## Area

First-run setup wizard

## What happened / what I expected

In the first-run wizard, step 2 (*Choose your first model*) offers three picks and,
below them, a link reading **Browse all models →**.

Clicking it does not browse anything. The wizard closes and you are dropped straight
onto the empty chat surface. There is no model list, no way back into the wizard, and
the selection you had made (LFM2.5 · 1.2B, ~563 MB) is gone.

What lands instead is a different model entirely: the composer shows
`bonsai-27b-2bit` with **"bonsai-27b-2bit isn't downloaded yet — it downloads once
(7.6 GB), then starts in seconds"**, and the placeholder text reads *"Download
bonsai-27b-2bit first"*. A new user who asked to *see more options* is now parked in
front of a 7.6 GB download they never chose, with the 563 MB one they had selected
discarded.

Expected: the link opens the model catalogue — the wizard's own copy says *"you can
download bigger models anytime in Settings"*, and `SettingsModelManagementPanel`
already exists as the destination.

## Steps to reproduce

1. Launch Rapid-MLX with no completed onboarding
2. Advance to **Choose your first model** — LFM2.5 · 1.2B is preselected
3. Click **Browse all models →**
4. Observed: the wizard disappears, the chat surface is empty, and the composer is
   pinned to `bonsai-27b-2bit` (7.6 GB)

Reproduced on a founder's machine against the shipped 0.12.7 build.

## Root cause

The whole implementation of the control is one line —
`apps/rapid-mac/Sources/Rapid/UI/ContentView.swift:476`:

```swift
onBrowseAll: { quickstartDismissedThisSession = true },
```

`quickstartDismissedThisSession` has exactly three references in the codebase: the
`@State` declaration (`ContentView.swift:45`), the two places that set it, and one
read at `ContentView.swift:493` that hides the wizard. Nothing anywhere navigates,
presents a sheet, or opens Settings. So the button is not "wired to the wrong
destination" — it has no destination.

It also shares its closure with the Esc / swipe-down path
(`quickstartSheetPresented`, `ContentView.swift:465`), whose comment records the
intent explicitly:

> A swipe-down / Esc means "let me look around first" — the same intent as the
> card's own "Browse all models", so route it to the same session flag rather than
> dropping it.

Treating *"show me the full catalogue"* and *"dismiss this"* as the same intent is
the defect. They differ precisely in whether the user gets to choose.

## Why the consequence is this bad

The `bonsai-27b-2bit` the user is left on is #1581 (cached-model default falls back
to alphabetical), and being stranded in front of an unwanted large download is the
shape of #1564. Neither of those is this bug, but they are what turns a dead link
into a first-run dead end: dismissing the wizard hands the user straight to the
alphabetical fallback.

Fixing this one removes the most likely way an ordinary user reaches that state,
since *"Browse all models"* is the natural thing to click when none of the three
offered picks is what you want.

## Test gap

The control carries `.accessibilityIdentifier("Quickstart.BrowseAll")`
(`QuickstartView.swift:1071`), so it is visible to the AX golden-flow suite added in
0.12.7 — but nothing drives it. `grep -rn "BrowseAll\|onBrowseAll" apps/rapid-mac/Tests`
returns nothing.

That is the interesting part: the structural snapshots record that the button exists,
is enabled, and sits where it should. A button that is present, enabled, correctly
labelled and wired to nothing passes every check we currently run. Whatever fix
lands should come with a test that *activates* it and asserts where the user ends up.

raullenchai:
Closes #1653.

The whole implementation of "Browse all models →" was one line that set a
session dismiss flag. Nothing navigated, presented a sheet, or opened Settings
— the link had no destination at all. Clicking it made the wizard vanish, and
because the selection lived in the wizard, the user's pick (LFM2.5 · 1.2B,
563 MB) went with it: they landed on the chat surface pinned to whatever the
alphabetical fallback chose, staring at a 7.6 GB download they never asked for.

The closure was shared with the Esc / swipe-down path, with a comment recording
the intent: both mean "let me look around first". That is the defect. Skipping
accepts whatever the app picks; browsing is a request to choose. They differ in
exactly the thing that matters.

So the two intents are separated. The dismiss closure is now named `onSkip` and
has exactly one caller — "Skip for now" on step 1, the app's one genuine leave-
onboarding control. Browsing routes through `SettingsRouter` to Model
Management, using the deep-link path the failure card already used, and leaves
the wizard standing behind the Settings window: closing Settings puts the user
back where they were, with the model they had already chosen.

The failure card's "or browse all models →" shared the same dead closure and is
fixed with it — it is offered precisely when the user's chosen model failed,
which is the moment they most need to pick a different one.

## Tests

The issue's sharpest point is that this passed everything we run: the button was
present, enabled, correctly labelled and carried `Quickstart.BrowseAll`, so the
AX structural baselines recorded a perfectly healthy control. A tree dump cannot
tell a working button from a decorative one.

- `gui-golden-flows.sh browse-all-destination` — a new flow that PRESSES the
  real control on the real wizard and asserts where the user ends up: a Settings
  window opened, it is on Model Management (not merely "Settings somewhere"),
  and the wizard is still up. Each assertion names the bug it would catch.
- `QuickstartBrowseAllDestinationTests` — pins the wiring in source, where the
  next person edits that call site, plus the router's stage-then-open ordering
  that a first-run user (who has no last-used tab) depends on.

### DIFF
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -13,6 +13,14 @@ can actually understand.
 
 ## [Unreleased]
 
+### Fixed
+
+- **"Browse all models" in the setup wizard opens the model catalogue.** It used
+  to close the wizard instead — your chosen model was discarded and you landed
+  on the chat surface pinned to a model you never picked. It now opens
+  Settings → Models with the wizard still behind it, so closing Settings puts
+  you back on your selection.
+
 ## [0.12.7] — 2026-08-07
 
 The privacy switch does what it looks like it does, and the models Rapid
--- a/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
@@ -460,8 +460,15 @@ struct ContentView: View {
             get: { quickstartVisible },
             set: { presented in
                 // A swipe-down / Esc means "let me look around first" — the
-                // same intent as the card's own "Browse all models", so route
-                // it to the same session flag rather than dropping it.
+                // same intent as the card's own "Skip for now", so route it to
+                // the same session flag rather than dropping it.
+                //
+                // NOT the same intent as "Browse all models", which used to
+                // land here too: that one asks to see the catalogue, and
+                // answering it by closing the wizard discarded the user's
+                // selection and left them on the alphabetical fallback
+                // (#1653). It opens Settings → Models now and leaves the
+                // wizard standing.
                 if !presented { quickstartDismissedThisSession = true }
             }
         )
@@ -473,7 +480,7 @@ struct ContentView: View {
             coordinator: quickstart,
             downloads: downloads,
             server: server,
-            onBrowseAll: { quickstartDismissedThisSession = true },
+            onSkip: { quickstartDismissedThisSession = true },
             onSeedWelcome: { true }
         )
         // QuickstartView was written for the detail column and its comments
--- a/apps/rapid-mac/Sources/Rapid/UI/QuickstartView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/QuickstartView.swift
@@ -771,13 +771,20 @@ struct QuickstartView: View {
     @Bindable var downloads: DownloadManager
     @Bindable var server: ServerManager
 
-    /// Callback the parent supplies for the "or browse all models →"
-    /// link. The parent dismisses the Quickstart surface for the
-    /// current session (without flipping the persisted flag) so the
-    /// existing picker becomes visible. Lifted out as a closure so
-    /// this view can stay agnostic of how the parent toggles its own
-    /// state.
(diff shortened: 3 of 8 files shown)


raullenchai:
`pr_validate` reports DO NOT MERGE on `full_unit` for one test: `test_batched_faster_than_sequential`. A/B classified as pre-existing and environmental, not this PR:

- the test alone: passes on `main` and on this branch (3 runs)
- the whole file: 10 passed / 1 xpassed, identical on both
- **the whole suite on clean `main`: fails the same way**

Filed as #1664. Every other step passes, CI is green, and codex converged to 0 BLOCKING / 0 MAJOR / 0 MINOR.

PR #1663, opened 2026-08-08
fix(rapid-mac): benchmark the loaded model

raullenchai:
## Why

“Speed on this Mac” claimed to measure the running model but spawned `rapid-mlx bench`, loading a second copy of the same weights. On an 18 GB Mac with the 8B model already running, the memory guard correctly rejected that unnecessary duplicate load.

## What changed

- reuse the ready desktop server, active port, bearer, and serving alias
- send one 8-token warm-up followed by one up-to-128-token measured request
- display completion tokens / measured wall time
- never spawn a second process for the normal speed card
- clarify this contract in the UI
- add request/usage/speed regression tests
- add a `loaded-model-benchmark` AX golden flow requiring one server start and exactly two benchmark requests
- update `docs/gui-golden-flows.md`

Leaderboard publication remains a separate, explicit standardized submission workload.

## Validation

- release app build passed (`SKIP_SIDECAR=1 BUNDLE_MODEL=0`)
- `swift build --target RapidTests` passed (all Swift Testing sources compiled)
- shell syntax checks passed
- fake sidecar non-stream OpenAI response validated
- real already-running `lfm2.5-8b-a1b-4bit` on this M3 Pro: warm-up 8/8 tokens, then measured 128/128 tokens at 98.1 tok/s; server PID stayed unchanged and no second model process appeared
- full local `swift test` is blocked by this Mac’s unaccepted Xcode license; CI will run the complete matrix
- AX golden execution is implemented; this login session currently reports `loginwindow` as frontmost, so the runner correctly refused to claim a GUI pass

PR #1666, opened 2026-08-08
fix(mac): the golden flows must ask the AX tree whether a window is gone

raullenchai:
## Why

Because the golden-flow suite is red on `main` for two independent reasons,
and one of them makes an assertion that cannot fail — which is worse than a
red suite.

Found by running `browse-all-destination` against a real build for the first
time. It failed at "the Settings window did not close", and the failure was in
the harness, not the app.

## 1. `peekaboo list windows` is not an oracle for window lifetime

It keeps reporting a window that has already been destroyed. Measured, with
both oracles side by side immediately after Settings closes:

```
probe: pb=1 ax=1     <- closing
probe: pb=1 ax=0     <- closed; peekaboo still lists it, and never stops
```

So the disappearance poll could never succeed. Six different ways of closing
the window — the File menu scoped to the window, the File menu scoped to the
app, `peekaboo window close`, ⌘W, and a real foreground click on the red
traffic light — all "failed" against that oracle. Every one of them actually
worked; a control flow with no wizard involved reproduced it, which is what
ruled the app out.

The appearance check had the mirror of the same problem, and it is the worse
one: a stale list can name a Settings window a PREVIOUS persona opened, so
"Browse all models opened the catalogue" — the assertion that catches #1653
coming back — could pass without anything having opened.

Both directions now ask the app's own AX tree, which is authoritative for
both. peekaboo is still asked for the window id, but only after the AX tree
has confirmed the window exists, and only to aim focus/click/menu at it.

## 2. Six AX baselines left stale by #1663

```
-  AXButton desc="Speed on this Mac" …
+  AXButton id="ChatView.SpeedOnThisMac" desc="Speed on this Mac" …
```

#1663 gave that button an accessibility identifier — strictly an improvement,
and exactly what the new gate asks for — but did not re-record the six
baselines that fingerprint it. That one line is the ONLY difference in all
six; nothing else in any tree moved.

Worth noting because the identifier gate cannot catch this: it reports a
control that has NO identifier, so ADDING one is invisible to it, and the
structural baselines are the only thing that notices.

## Verification

Against a build of this commit:

| flow | result |
| --- | --- |
| fresh-install | PASS |
| settings-persistence | PASS |
| chat-restore | PASS |
| slow-stream-stop | PASS |
| model-crash-recovery | PASS |
| low-memory-choice | PASS |
| update-state | PASS |
| no-dead-controls | PASS |
| catalog-integrity | PASS |
| browse-all-destination | PASS |
| loaded-model-benchmark | FAIL — #1668, a real product bug, not addressed here |

`browse-all-destination` now exercises the half that never ran before: a
non-default model card is chosen, Browse opens Model Management, Settings
takes a real foreground click, Settings closes, and the wizard comes back with
that same card still selected.

### DIFF
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/fresh-install.consent.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/fresh-install.consent.txt
@@ -11,7 +11,7 @@ AXApplication title="Rapid-MLX"
           AXStaticText value=text enabled=true
           AXStaticText value=text enabled=true
           AXButton desc="Connect your agents" help="Point an editor or agent at your local server" enabled=true
-          AXButton desc="Speed on this Mac" help="Start a model to run a speed test." enabled=false
+          AXButton id="ChatView.SpeedOnThisMac" desc="Speed on this Mac" help="Start a model to run a speed test." enabled=false
           AXGroup desc="<model> isn't running, It's already downloaded — starting takes a few seconds." enabled=true
             AXStaticText value=text enabled=true
             AXStaticText value=text enabled=true
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/fresh-install.steady.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/fresh-install.steady.txt
@@ -11,7 +11,7 @@ AXApplication title="Rapid-MLX"
           AXStaticText value=text enabled=true
           AXStaticText value=text enabled=true
           AXButton desc="Connect your agents" help="Point an editor or agent at your local server" enabled=true
-          AXButton desc="Speed on this Mac" help="Start a model to run a speed test." enabled=false
+          AXButton id="ChatView.SpeedOnThisMac" desc="Speed on this Mac" help="Start a model to run a speed test." enabled=false
           AXGroup desc="<model> isn't running, It's already downloaded — starting takes a few seconds." enabled=true
             AXStaticText value=text enabled=true
             AXStaticText value=text enabled=true
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/settings-persistence.models-after-relaunch.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/settings-persistence.models-after-relaunch.txt
@@ -11,7 +11,7 @@ AXApplication title="Rapid-MLX"
           AXStaticText value=text enabled=true
           AXStaticText value=text enabled=true
           AXButton desc="Connect your agents" help="Point an editor or agent at your local server" enabled=true
-          AXButton desc="Speed on this Mac" help="Start a model to run a speed test." enabled=false
+          AXButton id="ChatView.SpeedOnThisMac" desc="Speed on this Mac" help="Start a model to run a speed test." enabled=false
           AXGroup desc="<model> isn't running, It's already downloaded — starting takes a few seconds." enabled=true
             AXStaticText value=text enabled=true
             AXStaticText value=text enabled=true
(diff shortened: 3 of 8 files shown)


raullenchai:
Codex round on this PR: **1 BLOCKING, 1 MAJOR, 1 MINOR** — all three the same shape as the bug the PR was written to fix, one layer down. Fixed in cb1c0dae.

**BLOCKING — an unreadable dump was accepted as proof a window was gone.**
`ax_window_present` returned non-zero for two different things: "the window is absent" and "we never got to look". The close loop read the second as the first, so one failed dump ended the wait and declared Settings closed — after which the round-trip assertion runs against a tree that may still hold both windows and passes with no round trip having happened.

`rapid-ax.swift` made that easy to hit: a failed `AXChildren` read on the application root returned quietly and the dump was still emitted as a success. Measured against a pid that owns no AX application:

| | exit | payload |
|---|---|---|
| before | `0` | `{"success": true, …}`, 1 element |
| after | `1` | *(nothing on stdout)* |

The driver now fails loudly when it cannot read the root's window list, and the helper reports three outcomes — present / absent / could-not-observe — with both loops branching on all three and reporting an un-observable window list as its own failure rather than as either answer. Nothing deeper needs this: the depth and record guards can clip a subtree without hiding a window, since windows only sit at depth 1.

**MAJOR — the "non-default card" pick could return the default.**
`.selected != true` also matches a card that never published the attribute, and measurement says every unselected card is exactly that:

```
Quickstart.Choice.lfm2.5-1b-4bit   selected: true
Quickstart.Choice.qwen3-0.6b-4bit  selected: <omitted>
Quickstart.Choice.qwen3.5-4b-4bit  selected: <omitted>
Quickstart.Choice.qwen3.5-9b-4bit  selected: <omitted>
```

SwiftUI publishes `AXSelected` only on the selected card, and `rapid-ax` also omits an attribute whose read failed — so absence means "not selected OR we failed to look" *by construction*, not merely on a bad day. The pick could therefore return the default itself, after which the closing assertion only claims the default is still the default — true even when the wizard discards the choice, i.e. #1653 walks straight through the flow written to catch it. The default is now identified positively as the single card claiming selection (0 claimants → retry, never guess), and the pick is anything else by identifier.

**MINOR — one-shot window-id read.** AX publishes a new window before CGWindow will name it; that read is now a bounded poll, so the race fails nothing.

Verified: `--flow browse-all-destination` passes, and flows 1–6 pass on this branch. The suite stops at 7, `loaded-model-benchmark` — that failure is #1668 and is fixed by #1669, not by anything here. I'll re-run the full suite once #1669 has landed.

raullenchai:
Rounds 2 and 3. Round 2 found that my first fix closed one hole and opened a worse one; round 3 confirms all of round 2 is closed.

**Round 2 — the driver hard-failing was my regression, not a fix.** Making `rapid-ax dump` exit non-zero on an unreadable root meant that under `set -euo pipefail` one transient failure killed the entire suite: `see_main` and `see_settings` call the driver unguarded, so `wait_identifier`, `wait_settings_stable` and the new chooser poll — every loop that exists to ride out exactly that — never got a turn.

It also had not finished the job. Reading the root's child array is not the same as *enumerating the windows*: `walk` skips a child whose `AXRole` read failed, omits a title it could not read, and stops at the record cap. Each silently shortens the window list, and a shortened list is indistinguishable from a window that closed — the original bug, still live on three other paths.

So the driver now enumerates the root's children **once**, before the walk, and publishes what it can vouch for:

```json
"windows": { "titles": ["Rapid-MLX", "Settings"], "complete": true }
```

Any failed read during that enumeration sets `complete: false` instead of dropping an entry, and the process still exits 0 — `complete: false` is not an error, it means *ask again*. `ax_window_present` gates on it and reports the third outcome. Against a pid that owns no AX application: `{"success":true,"windows":{"complete":false,"titles":[]},"elements":1}`, exit 0 — could-not-observe, retry loops intact.

Reading `data.windows` instead of `ui_elements` also retires the record-cap hole for free, since the list is built before and independently of the capped walk.

**Round 3 — all round-2 findings closed**, plus two of mine:

- `walk` still re-read the root's `AXChildren`, so "the tree and the list cannot disagree" was not actually true — a second read can return a different set. Depth 0 now walks the enumerated elements themselves.
- `ax_window_present` now requires `titles` to *be* an array. `titles[]?` swallows a structural failure, so a malformed list would have read as a confident "absent" — the third outcome collapsing straight back into the first.

**Not fixed here:** round 3 also flagged `flow_catalog_integrity`'s two `fake-video-alias` assertions, which have the same shape — a negative claim over `ui_elements` that an incomplete walk satisfies. That code is unmodified by this PR and present verbatim on `main`, so it is #1670 rather than a drive-by fix in a PR about window lifetime.

**Verification.** Full golden suite, all 11 flows, exit 0 — up from 10/11, the 11th being `loaded-model-benchmark`, which #1669 fixed. Re-running the suite once more now that the depth-0 walk changed.

Issue #1618, opened 2026-08-07
dogfood-isolate.sh does not isolate the port, so a dogfood app SIGTERMs the operator's real rapid-mlx serve


raullenchai:
## What

`apps/rapid-mac/scripts/dogfood-isolate.sh` gives each dogfood copy its own bundle id and a throwaway `$HOME`. It does **not** isolate the port. An isolated copy still sweeps :8000–:8009 at launch and kills anything it finds there that looks like a rapid-mlx server — including one the operator started themselves in a terminal.

I did exactly this today: launching a dogfood build SIGTERM'd a `rapid-mlx serve` someone else was using on :8000. Not a hypothetical.

## Why it happens

The sweep runs unconditionally in `RapidApp.init()`, **before the first window draws** and regardless of the auto-start setting:

```swift
// RapidApp.swift:97
PortSweep.startLaunchSweep(port: PortAllocator.candidatePorts.first ?? 8000)
```

and the allocator sweeps every candidate on the way to binding:

```swift
// PortAllocator.swift:83-96
for candidate in candidates {
    PortSweep.sweep(port: candidate)
    if canBind(port: candidate, host: host) { return candidate }
}
```

Targets are discovered with `lsof -nP -sTCP:LISTEN -ti :<port>` and filtered by command line (`PortSweep.isRapidOwnedCommand`, `:458-512`). A plain `python -m http.server 8000` is correctly skipped — but `python -m vllm_mlx.cli serve …` matches Form 2 exactly, which is what both a hand-started server and the release gate use.

`OwnedServerRecord` narrows this but does not close it: it is documented as "an upgrade, not a replacement" (`PortSweep.swift:148`), and when there is no record — or it is stale — control falls through to the name heuristic, which reaps any matching listener the app never spawned. `OwnedServerRecord.swift:14` is candid about the original report: *"a developer running their own `rapid-mlx serve` got SIGTERM'd on the next Rapid.app launch."*

## Why isolation should cover it

The script's own "STILL SHARED" list does not mention ports at all, so there is nothing warning the operator. Meanwhile `PortAllocator.swift:44-51` already documents `RAPID_DESKTOP_PORT` as the isolation mechanism and notes the sweep predicate is bundle-id-agnostic — so **isolated copies can reap each other's sidecars too**, not just the operator's.

Both knobs already exist and work; the generated launcher simply does not set them:

[...]

Issue #686, opened 2026-06-18
codex CLI hangs on gemma-4-12b-4bit due to model 'thought' degenerate loop (~60% reproducer)


raullenchai:
## Summary

`codex exec` (or any agent that POSTs Codex-shaped requests with the full Codex CLI system prompt + 10-tool definitions to `/v1/responses`) hangs ~156 s on `mlx-community/gemma-4-12B-it-4bit` because the model deterministically enters a `thought\nthought\nthought\n…` degenerate generation loop in ~60% of trials.

Repro hit rate: 3/5 trials (60%) using the captured Codex CLI payload at the model's recommended sampling params (T=1.0, top_p=0.95, top_k=64 from `generation_config.json`). The control model (`qwen3.5-9b-4bit` via the `hermes` tool_call_parser / `qwen3` reasoning_parser) passes 5/5 on the same payload, so this is gemma-4-12b-specific.

## Reproduction

1. `python3.12 -m vllm_mlx.cli serve gemma-4-12b-4bit --host 127.0.0.1 --port 8765`
2. Replay the captured Codex `/v1/responses` body (~47 KB, 7,800-token prompt with `exec_command` + `write_stdin` + `update_plan` + 7 more tools) via `httpx.post(... json=body)` with `max_tokens=2000`.
3. Observe the non-streaming `/v1/chat/completions` result with the same messages+tools: 3/5 trials return `finish_reason=length`, 0 tool_calls, and `content` is literally `'thought\nthought\nthought\n…'` × ~1000.
4. The remaining 2/5 trials emit a clean `exec_command({'cmd': 'head -n 1 pyproject.toml'})` tool call in ~1 s.

## Why it cascades into a 156 s hang

When the model is in the `thought` loop, every token goes through `OutputRouter` → `RouterState.THINKING` → `Channel.REASONING`. The `/v1/responses` route ([`responses.py:487-502`](https://github.com/raullen/rapid-mlx/blob/main/vllm_mlx/routes/responses.py#L487-L502)) **silently drops the reasoning channel** for Codex (Codex's wire protocol has no `response.reasoning_text.delta` event in v1). Codex therefore sees only `response.created` and then nothing until the request times out or hits `max_output_tokens`.

The same payload through `/v1/chat/completions` non-streaming (which uses `OutputRouter.feed_sequence`) returns the looped content in `content` (visible) — confirming the model output IS being captured, just dropped from the streaming Responses-API surface.

## Why Qwen3.5-9B works

[...]

PR #1671, opened 2026-08-08
fix(release): G12 has not run for eleven releases, and would false-fail if it did

raullenchai:
## Why

**G12 has not run for eleven releases.** Not "ran and passed" — never
executed, from v0.11.4 through v0.12.7. It is the gauntlet's last gate and the
only one that boots anything other than `qwen3.5-9b-4bit`, so each of those
releases shipped without the other ~28 registered aliases being started once.

`cleanup()` was written as `[ -n "$CLUSTER_WORK" ] && rm -rf "$CLUSTER_WORK"`.
As an EXIT trap its return status is discarded, but G12 *also* calls it inline,
to hand the port and the GPU to the sweep. There, under `set -euo pipefail`, the
status is load-bearing: by then `CLUSTER_WORK` is always empty — the correctness
cluster cleared it a few hundred lines earlier — so the `&&` list returns 1 and
the gauntlet dies on the spot. The signature is the G12 banner, then
`make: *** Error 1`, and not one line of gate output.

The window is precise, and it is not "since G12 was added": at #693 `cleanup`
was the pidfile block alone and returned 0. The `&&` arrived together with
`CLUSTER_WORK` in #1354 (`a4520b2a`, 2026-07-30), and G12 has been dead from
that commit onward.

It is fail-closed, so nothing shipped that a green G12 would have caught. What
was lost is the coverage.

## What un-skipping it found

The first real run went red on `qwen3-4b-instruct-2507-4bit` × `hermes`: the
profile burned the entire 1020 s per-profile cap while the engine answered
58×200 / 36×404 (the harness probing `GET /version`, which the server does not
expose) and zero 5xx. Replayed on v0.12.7 the run is byte-identical, so it is
the model's ceiling and not a regression (#1672).

The tempting fix — teach the gate to forgive that round — cannot be made sound,
and the first draft of this PR tried it. From the log of a failed round, "weak
model" and "wedged scheduler" produce the same evidence. Liveness plus a 2xx
does not separate them either: bench issues `GET /health` and `GET /v1/models`
itself before the profile starts, so a 2xx is automatic and says nothing about
inference. Any rule lenient enough to pass a 4B model that ran out of clock also
passes a scheduler that deadlocked after the first request, a stall between
streaming headers and the first token, and a decode regression that put every
request past the cap.

So it is fixed where the fact lives — the sample pool.

## The three commits

1. **`cleanup` returns 0.** `if`, not `&&`, plus an explicit `return 0`, with a
   comment saying why the status matters at one of its two call sites.

2. **G12 stops drawing models that cannot do agentic work.** Every profile in
   `HARNESS_PROFILES` drives a multi-step tool-calling loop; a model that cannot
   hold one does not fail fast, it retries malformed calls until the clock runs
   out. Excluding models that cannot do the work is what this filter already
   does for hybrids, for gemma-4 (#686), and for MoEs with too few active
   params — this adds the tier the gauntlet measured.

   The floor moves 4 → 6 B, the conservative cut between the two sizes measured
   on this hardware: 4B fails hermes; 9B completes it in ~740–800 s, which is
   where `ROUND_TIMEOUT_S` came from. 7B and 8B are unmeasured and stay, because
   that band is most of the sweep's coverage; if one shares the ceiling, the
   constant moves again. All four 4B aliases go, not only the drawn one —
   waiting for each to be sampled and measured means a red gauntlet on a random
   calendar day, which is how a gate stops being read. Pool: 13 → 9.

   **What this costs, without the gloss.** The 4B tier keeps only the coverage
   it already had, and that is thin: `qwen3.5-4b-4bit` is a release-fleet
   coherence model so G0a boots it every gauntlet, but `evals/coherence_gate.py`
   sends short, single-turn, tool-free prompts at temperature 0, and CI's
   `l1-smoke` adds one forced tool-call format check on that same alias and is
   not a required check. Neither covers multi-turn contexts, streaming tool
   calls, parser stress under repeated agent traffic, or the other three 4B
   aliases at all. Buying that back needs a tier sized for what small models can
   finish rather than a sweep that hands them a 12B's profiles — filed
   separately (#1677). The alternative — keeping the 4B aliases in the pool and accepting the occasional red day — would mean narrowing the exclusion to the agentic harnesses only, rather than dropping the tier from sampling.**

3. **G12 must own the port it benchmarks, and the GPU.**
   * Ownership was checked once. `_wait_for_server` returns permanently on the
     first owned 200, and the round loop never asked again — so a takeover after
     readiness hands every later round to a stranger, and a model can even PASS
     that way. `_still_ours` is now asked before AND after every round: before,
     so we do not measure someone else; after, because a takeover during a round
     would otherwise be reported as that round's result.
   * `_wait_for_server` accepted any 200. Our child stays alive for minutes
     while it loads weights, so anything already bound answers first and every
     round after that measures it — which happened during this work: a desktop
     app swept port 8000, SIGTERM'd the gauntlet's server and bound its own
     sidecar (#1618), and nothing noticed because the replacement answered
     perfectly well. Asking `/v1/models` what it serves is not enough (identity
     is not ownership; a leftover engine holding the same weights passes), so
     `_owns_port` walks the process tree: every listener must be our child or a
     descendant. Fails closed.
   * The handoff waited for the port, not for the process. uvicorn closes its
     listening socket before lifespan shutdown finishes — and shutdown is where
     #667's deadline-aware prefix-cache flush runs — so a free port is not an
     idle GPU. Overlapping G12's model load with a process still holding weights
     is how a run earns a `metal::malloc` "Resource limit" that reads like a
     code bug. Now: capture the PID before `cleanup` deletes the pidfile, then
     require both the process gone and the port free, or refuse.
   * Each round appended its bench transcript to the server's log file. The
     server holds that file open `"w"` — no `O_APPEND` — so it writes at an
     offset it tracks itself: our append moves the end of the file, its next
     line lands on top of what we wrote. Transcripts get their own file.

## Tests

* `tests/release/test_release_check_m3_cleanup.sh` — extracts the real
  `cleanup` with `sed` so a copy cannot drift, and pins the premise (the script
  still runs under `set -euo pipefail`; G12 still calls `cleanup` inline). 3 of
  6 assertions fail on `main`. Runs in CI via the existing
  `tests/release/test_*.sh` loop.
* `tests/test_release_check_random.py` — 7 → 25 tests: the 4B tier is out of
  the pool and 9B is still in; the floor applies to a MoE's active params, with
  cases straddling the NEW floor (A4/A5 out, A6 in) so the test is not vacuous;
  the ownership walk (ours, a descendant, a stranger, a shared port, an
  unverifiable port, an unreadable parent, a looping parent chain); that
  readiness actually consults ownership rather than merely defining it; that the
  round loop asks before *and* after each round, by parsing the loop, so
  deleting a call site fails; a partial `lsof` run; and a timed-out round
  leaving a transcript. **19 fail against this branch's base.**

All offline — no GPU, no network, no server.

4. **G12 was sampling vision models into a text-only harness.** The filter
   excluded multimodal models by looking for `-vl-` in the alias — precisely
   the test `vllm_mlx/api/utils.py` documents as insufficient, because UI-TARS
   is a Qwen2.5-VL-based GUI agent whose public name carries no `VL` and
   Gemma 3 is multimodal with no marker at all. Four of the nine eligible
   aliases were vision models and the default seed `20260808` sampled
   `ui-tars-7b-dpo-4bit`: a hard exit at boot without the vision extra, and a
   meaningless measurement with it. Now mirrored from `MLLM_PATTERNS`, matched
   against the alias and the HF path, with a test that parses the engine's list
   out of the source (the script must stay importable without MLX) and fails if
   the mirror falls behind.

   **Pool: 13 → 5**, exactly the documented minimum. The sanity test now says
   so — the next exclusion trips it, and the answer then is a gate sized for
   what is being excluded (#1677), not a lower bar here. `deepseek-r1-8b-4bit`
   is one of the five and is a suspected #1672 (#1676); today's seed draws it.

5. **Ownership drift aborts the sweep**, rather than booting the next model
   into whatever took the port. `_run_model_rounds` came out of `main` so the
   control flow is testable: the earlier test read the source for a call site,
   which a rename breaks and which never proved that a drifted round is
   discarded or that the remaining harnesses are skipped.

6. **`lsof` is bounded and required up front.** The 60s handoff deadline read
   `SECONDS` only between calls, so an `lsof` that never returned made it
   unbounded. `port_busy` runs it under a watchdog and reports
   listening / free / could-not-tell, with an unverifiable port never counting
   as free. The `command -v lsof` check moved to the top-level pre-flight,
   because the gauntlet's first act is an `lsof` port check whose 127 the `if`
   swallows — a PATH without `/usr/sbin` used to run the whole gauntlet before
   G12 noticed.

## Two bugs I found by measuring instead of reasoning

Both were invisible to the unit tests, because the unit tests stub the thing
that was broken.

* **`_owns_port` returned True for pid 1.** The ancestor walk climbs until it
  meets `proc.pid`, and every chain reaches init, so "the listener descends
  from pid 1" is true of every listener on the machine. Found by pointing the
  check at a real listener on a real port. There is now a test that does that —
  bind a port from a real child, assert the child is ours, our own pid is ours
  (the walk has to cross an edge, which is what a `serve` binding from a
  subprocess needs), pid 1 is not, and a freed port belongs to nobody.
* **`kill -0` succeeds on a zombie.** Between `cleanup` sending TERM and the
  shell collecting the status, the server still owns a pid — and owns no GPU.
  The wait loop read that as "still running", would burn the whole 60s deadline
  and then SIGKILL a corpse, turning a clean shutdown into a gate failure that
  blames the thing that worked.

## Review rounds

Two so far. Round 1 killed the original approach outright: I had G12 classify a
failed round as WARN when the engine looked healthy, and the review showed the
downgrade is unsound — `bench` issues `GET /health` and `GET /v1/models` itself,
so the "positive evidence" I required was free. That code is gone; commit 2 is
what replaced it. Round 2 produced the ownership blocker, the leaked-process
path, the `lsof` handling, and the correction to the release history above (I
had said "since #693"; it is #1354). Round 3 produced the multimodal blocker,
the abort-on-drift major, and the bounded/early `lsof` — and confirmed the
round-2 fixes.

## Run on real hardware

**A full `make release-check-m3` cannot go green on this machine, and that is
not this branch.** The gauntlet died at **G7b**, four gates before G12:

```
FAIL codex (163.1s) 11p 0f 1e 0s | e2e_file_read: TIMEOUT
```

Replayed scoped to that profile it fails again with a *different* subtest
(`e2e_chat`), and replayed on **`v0.12.7`** it fails identically to the gauntlet
run. Not a regression, not caused by a branch that changes four files and no
engine code. Filed as **#1683** — and note what it implies: G7b is not sampled,
so the local gauntlet has not been able to reach G12 at all.

So the gate this PR fixes was exercised directly instead:

```
python3.12 scripts/release_check_m3_random.py --models 2 --harnesses 1 --rounds 1
```

**The G12 machinery works end to end on real hardware.** Boot → ownership
verified against the real process tree → round → clean teardown → next model →
verdict. No spurious ownership drift, no leaked server, `:8000` released
afterwards. The earlier full-gauntlet run also proved the `cleanup` fix in
situ: it failed at G7b and exited cleanly, leaving no server behind.

And on its first honest run the gate immediately found two per-model defects,
which is exactly what it exists for:

```
FAIL hermes3-8b-4bit/aider        (0.3s)  1p 1f | plain_chat: Expected '4', got: 2
FAIL deepseek-r1-8b-4bit/opencode (40.4s) 5p 5f | single_tool_call: No tool_calls in response
```

Both are model-side. `hermes3-8b-4bit` answering `2` to "What is 2+2? Reply with
just the number" is deterministic at temperature 0.0 and 0.3, while the same
model answers `What is 2+2?`, `2+2=`, the capital of Japan, an exact-echo
instruction, and — with the identical "reply with just the number" phrasing —
`17 times 23` → `391`, all correctly. So it is neither the engine nor the chat
template.

Which is the honest caveat on this PR: **G12's pass criterion is model
competence, so it will keep going red without an engine regression in sight.**
Three red gates today, zero product bugs. That is the argument for #1677, and
the reason the pool cannot simply keep shrinking — it is at five now, the
documented minimum, and #1676 wants a sixth exclusion.

### Suggested merge criterion

This branch makes a dead gate run, verified on hardware. Blocking it on a green
`release-check-m3` blocks it on #1683, which is someone else's bug and older
than this branch.

Refs #1672, #1618

### DIFF
--- a/scripts/release_check_m3.sh
+++ b/scripts/release_check_m3.sh
@@ -78,6 +78,43 @@ echo "  base_url: $RAPID_MLX_BASE_URL"
 echo "  log:      $LOG"
 line
 
+# `lsof` decides, here and at the G12 handoff, whether a port is free —
+# and "free" is the one answer that must never be guessed. A PATH without
+# /usr/sbin makes every check silently answer "free": the `if` below swallows
+# the 127, the whole gauntlet runs, and only the last gate notices. Refuse now.
+command -v lsof >/dev/null 2>&1 \
+  || { echo "ERROR: lsof is required (port checks would silently pass)." >&2; exit 2; }
+
+# `lsof` with a per-invocation watchdog.
+#   0 = something is listening   1 = nothing is   2 = could not find out
+# A stuck `lsof` would otherwise stretch any "wait N seconds" loop that calls
+# it into as long as it likes, because the deadline is only read between calls.
+# `-nP` also removes the name/port resolution stalls that cause most of them.
+# Callers must treat 2 as "not free": an unverifiable port is not a free one.
+port_busy() {
+  local port="$1" limit="${2:-10}" pid n=0
+  lsof -nP -i ":$port" >/dev/null 2>&1 &
+  pid=$!
+  while kill -0 "$pid" 2>/dev/null && [ "$n" -lt "$((limit * 10))" ]; do
+    sleep 0.1
+    n=$((n + 1))
+  done
+  if kill -0 "$pid" 2>/dev/null; then
+    kill -9 "$pid" 2>/dev/null || true
+    wait "$pid" 2>/dev/null || true
+    echo "  ! lsof did not return within ${limit}s for port $port" >&2
+    return 2
+  fi
+  local rc=0
+  wait "$pid" || rc=$?
+  # lsof exits 1 for "nothing matched" and anything else for "I failed".
+  case "$rc" in
+    0) return 0 ;;
+    1) return 1 ;;
+    *) return 2 ;;
+  esac
+}
+
 # Pre-flight: refuse if port is busy so we don't accidentally murder
 # someone's debug server.
 if lsof -i ":$PORT" >/dev/null 2>&1; then
@@ -651,11 +688,6 @@ else
   # ``cleanup`` is the trap-installed teardown defined at the top of
   # this script — calling it manually here releases the PID file too, so
   # read the PID before calling it.
-  # `lsof` decides below whether the port is free. Without it every check
-  # silently answers "free", which is the one answer that must never be
-  # guessed — so require it rather than degrade.
-  command -v lsof >/dev/null 2>&1 \
-    || { echo "  ✗ lsof is required to verify the port handoff to G12" >&2; exit 1; }
   OLD_SERVER_PID="$(cat "$PIDFILE" 2>/dev/null || true)"
   # `kill -0` succeeds on a ZOMBIE — a child that has exited but whose status
   # the shell has not collected still owns a pid, and it owns no GPU. Waiting
@@ -690,7 +722,12 @@ else
       sleep 1
       continue
     fi
-    if lsof -i ":$PORT" >/dev/null 2>&1; then
+    # 0 = still listening, 2 = could not find out. Only an explicit 1 — a
+    # check that ran and said nothing is there — releases the handoff; an
+    # unverifiable port is not a free one.
+    port_state=0
+    port_busy "$PORT" 5 || port_state=$?
+    if [ "$port_state" != 1 ]; then
       sleep 1
       continue
     fi
--- a/scripts/release_check_m3.sh
+++ b/scripts/release_check_m3.sh
@@ -803,8 +803,10 @@ else
     --port "$PORT" \
     --models "${G12_MODELS:-2}" \
     --harnesses "${G12_HARNESSES:-2}" \
-    --rounds "${G12_ROUNDS:-3}" \
-    --report /tmp/release-check-m3-random.log
+    --rounds "${G12_ROUNDS:-3}"
+  # No --report: the script defaults it into its own 0700 run directory and
+  # prints the path. A fixed /tmp name is a symlink the next local process can
+  # aim wherever it likes.
 fi
 
 #-------------------- Done ----------------------------------------
--- a/scripts/release_check_m3.sh
+++ b/scripts/release_check_m3.sh
@@ -725,6 +725,12 @@ else
     # exact overlap this two-condition handoff exists to prevent.
     local state
     state="$(ps -o stat= -p "$OLD_SERVER_PID" 2>/dev/null)"
+    # Trim. `ps -o stat=` right-aligns into a column whose width comes from the
+    # widest state on the machine, so a zombie can arrive as " Z+" rather than
+    # "Z+". Matching `Z*` against the padded string fails, the zombie reads as
+    # ALIVE, and the handoff burns its whole 60 s and fails — which is the
+    # false failure this helper was written to remove.
+    state="${state#"${state%%[![:space:]]*}"}"
     case "$state" in
       Z*) return 1 ;;
       *)  return 0 ;;
--- a/scripts/release_check_m3.sh
+++ b/scripts/release_check_m3.sh
@@ -651,6 +651,11 @@ else
   # ``cleanup`` is the trap-installed teardown defined at the top of
   # this script — calling it manually here releases the PID file too, so
   # read the PID before calling it.
+  # `lsof` decides below whether the port is free. Without it every check
+  # silently answers "free", which is the one answer that must never be
+  # guessed — so require it rather than degrade.
+  command -v lsof >/dev/null 2>&1 \
+    || { echo "  ✗ lsof is required to verify the port handoff to G12" >&2; exit 1; }
   OLD_SERVER_PID="$(cat "$PIDFILE" 2>/dev/null || true)"
   cleanup
   sleep 2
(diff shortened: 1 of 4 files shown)


raullenchai:
Ran the sweep this unblocks, and it matters for how this PR should land.

`git log -S` puts the inline `cleanup` call in **#693 — the commit that introduced G12**. So G12 has never run once, in any release. It was born dead.

On its first real run it found something, and it is not what the gate is for:

```
qwen3.5-4b-4bit              codex  78.3s / 77.7s / 108.3s   PASS 3/3
qwen3.5-4b-4bit              aider  ~0.5s                    PASS 3/3
qwen3-4b-instruct-2507-4bit  hermes 1020.3s                  FAIL — per-profile cap
```

A/B'd against `v0.12.7` with identical caps and boot flags — `rc=1`, **1020.3s**, same `timed out after 1020s (per-profile cap)`, byte for byte. So it is pre-existing and **not** a regression from #1646 (the only change to engine runtime behaviour since v0.12.7, and one whose failure mode would look exactly like a timeout). Filed with the full evidence as #1672.

**The consequence for this PR:** landing it alone turns a silent no-op into a red gate on every release cut, failing for a reason unrelated to what it guards. #1671 and #1672 need to land together, or `make release-check-m3` cannot go green.

I have deliberately not folded a fix for #1672 in here — choosing between a per-profile sample pool, per-model caps, and classifying TIMEOUT as WARN-not-FAIL is a design call about what this gate is for, and it wants its own review rather than riding along in a one-line status fix.

PR #1675, opened 2026-08-08
fix(mac): a golden flow may only claim absence from a walk it can vouch for

### LINKED ISSUE #1670: gui-golden-flows: catalog-integrity proves absence from a walk that may be incomplete
Found by review while fixing the window-lifetime oracle in #1666. Not caused by
that PR — the code below is unmodified on `main` — so it gets its own issue
rather than a drive-by fix in a PR about something else.

## The problem

`flow_catalog_integrity` proves that a video-generation model is never offered
as a chat model with two negative assertions over the whole element dump:

```bash
jq -e '[.data.ui_elements[]?
        | select([(.identifier // ""), (.value // ""), (.title // ""), (.description // "")]
                 | map(tostring) | join(" ") | test("fake-video-alias"))] | length == 0'
```

`length == 0` is read as "the alias is not offered". It is also satisfied by
**never having looked**:

* `walk` returns early on any element whose `AXChildren` read fails, so a whole
  subtree — including the picker rows — can be missing from a dump that still
  reports `success: true`;
* the 12,000-record cap truncates the walk silently;
* `rapid-ax` omits an attribute whose read failed, so a row can be present in
  `ui_elements` with the very fields the regex tests (`identifier`, `value`,
  `title`, `description`) missing.

Any one of those makes the flow log `no video-gen alias` without having
established that the inventory was observed at all. This is the same defect
class as the window oracle in #1666: a missing observation read as a definite
answer.

It matters because the flow exists to guard #1603 — eight video-generation
aliases reaching the picker and dead-ending at "Couldn't start … Try again"
*after* a download of up to 64 GB. A test that can pass without looking is not
a guard against that returning.

## What a fix needs

Either of:

1. **A completeness signal the assertion can gate on.** #1666 added
   `data.windows = {titles, complete}` for windows specifically, where
   `complete: false` means "could not observe" and callers must treat it as
   neither answer. The equivalent for the descendant walk is a flag the driver
   sets when any `AXChildren` read fails or the record cap truncates, so a
   negative assertion can refuse to conclude anything from a clipped dump.

2. **A positive, domain-specific inventory assertion.** Assert that the picker
   lists exactly the aliases it should, rather than that one string is missing
   from an unbounded blob — a positive claim cannot be satisfied by an empty
   observation.

(1) is more reusable; every negative assertion in the harness has this shape.

## Where

`apps/rapid-mac/scripts/gui-golden-flows.sh`, `flow_catalog_integrity` — the two
`fake-video-alias` assertions.

raullenchai:
## Why

`flow_catalog_integrity` is the golden flow that stands between users and #1603
— eight video-generation aliases reaching the model picker, looking ordinary,
and dead-ending at "Couldn't start … Try again" *after* a download of up to
64 GB. It proves the filter works with

```bash
jq -e '[.data.ui_elements[]? | select(... "fake-video-alias" ...)] | length == 0'
```

That is also satisfied by never having looked. `rapid-ax` has three silent ways
to fall short of a full inventory — an `AXChildren` read that fails, the depth
cap, the record cap — and each drops a subtree while the dump still reports
`success: true`. Nothing in the output distinguishes "absent" from "not
observed", so the flow can log `no video-gen alias on either catalog surface`
without having seen the catalogue at all.

The idiom also spreads: #1673 added a third assertion of the same shape
(`Settings.BackToQuickstart … length == 0`) while the issue was open. Fixing the
three sites by hand leaves the fourth to be written next week, so this fixes it
centrally and lints for it.

## What changed

* **`rapid-ax.swift` publishes `data.walk = {complete, reasons}`** — the same
  contract #1666 gave `data.windows`. `complete: false` means "this dump cannot
  answer a question about absence"; `reasons` says why, so the harness log
  names the cause instead of only reporting one.
  * The walk now distinguishes an element that publishes no `AXChildren`
    (`attributeUnsupported`/`noValue` — a genuine leaf) from one it could not
    interrogate (every other `AXError` — a subtree that is missing).
  * A failed root enumeration marks the walk incomplete too, since the walk
    starts from that list. A window whose *title* would not read does not — the
    element tree is still whole, only the window list is short.
* **`assert_ax_absent` / `ax_elements_match`** gate on that signal, with the
  same three outcomes as `ax_window_present` (present / absent / could not
  observe). An incomplete dump is usually a lost race, so it re-dumps a few
  times, then fails loudly rather than issuing a clean bill of health it did
  not earn. The retry stages its dump so a driver that cannot run no longer
  truncates the caller's evidence — which is also where the reason is read
  from.
* All three negative assertions go through it.

## What running it against the real app changed

The first version of this signal was unshippable, and only a full golden-flow
run showed that:

```
77 dumps, 5 with walk.complete: false
all of them "a searched attribute was unreadable" (AXError -25200)
always the same Settings panels: app, privacy, tools
```

Structural, not a lost race — which broke two assumptions the design rested on.
`assert_ax_absent` retries 20 times because an AX failure was taken to be
transient; it is not, so an absence assertion on those panels would have become
a permanent gate failure. And `ax-baseline.py`, refusing incomplete dumps,
refused the suite outright — that is what the run died on.

So the rule is now **per element**: a failed read costs the walk its
completeness only when it leaves that element with *nothing* searchable. An
element whose title would not read but whose identifier did is still found by
every filter that tests identifiers; it is not hiding anything. An element that
came back with no identifier, no title, no description and no value — because
each of those reads failed — could be hiding anything.

Truncation is untouched. A missing subtree, the depth cap and the record cap
still make the walk incomplete, which is the case `ax-baseline.py` actually
needed: a clipped tree whose recorded prefix matches the committed baseline,
with `--update` committing it as the new truth.

Each record now carries `unreadable: ["title"]` when a read failed, and the dump
reports `elements_with_unreadable_fields`. `reasons` only ever explains why
`complete` is false, so without those the artifact could not say a read had
failed at all — the earlier reason string said "1 element" and gave no way to
find which one.

## What the run confirmed

* `flow_catalog_integrity` passes with the new positive precondition, logging
  `both catalog surfaces listed models, and neither offered a video-gen alias` —
  so waiting for the fixture's ordinary alias before denying the video one works
  against the real catalogue, not just in theory.
* Capacity assumptions hold with enormous margin: the largest real dump is 108
  records at depth 7, against caps of 12,000 and 40.
* An ordinary app (Finder) reports `complete: true`, so the signal does not
  false-alarm on a healthy tree.

**Still to verify:** that the narrowed rule actually clears those five dumps.
The re-run needs app instances, which sweep port 8000 (#1618), and the machine
is running the release gauntlet.

## Tests

`tests/test_gui_golden_absence_contract.py` — 14 tests, pure bash + jq (no GUI,
no Swift toolchain, no GPU). The helpers are extracted from the real script with
a regex so a copy here cannot drift from what actually runs.

Covers all three outcomes, and the ways the third collapses into the first: a
missing `walk` key, `success: false`, `ui_elements` that is not an array,
unparseable JSON, and a jq program that does not compile (exit 3, which is not
an answer). One test lints the flows for the raw `length == 0` idiom, because
fixing the three sites that exist today does not stop the fourth.

A/B: **14 fail on `main`, 14 pass here.** The lint alone reports the three real
offenders.

Also verified the driver against real accessibility trees — Finder and Terminal
both report `walk.complete: true`, so the signal does not false-alarm on a
normal app; pid 1 (no readable AX tree) correctly reports `complete: false` with
its reason, which is precisely the "one element, `success: true`, and an
absence assertion would have passed" case.

Closes #1670

### DIFF
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-depth.five-turns.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-depth.five-turns.txt
@@ -1,0 +1,91 @@
+# rapid-mac AX structural baseline v1
+AXApplication title="Rapid-MLX"
+  AXWindow subrole=AXStandardWindow id="main" title="Rapid-MLX"
+    AXGroup subrole=AXHostingView
+      AXSplitGroup id="main, SidebarNavigationSplitView"
+        AXGroup subrole=AXHostingView
+          AXButton id="Sidebar.NewChat" desc="New Chat" enabled=true
+          AXButton id="Sidebar.Launch" desc="Launch" enabled=true
+          AXScrollArea
+            AXHeading desc="<relative-day>" enabled=true
+            AXButton id="Sidebar.Conversation.Pin.<uuid>" desc="Pin conversation" help="Pin conversation" enabled=true
+            AXPopUpButton id="Sidebar.Conversation.Menu.<uuid>" desc="Conversation actions" enabled=true
+            AXButton id="Sidebar.Conversation.<uuid>" desc="shape:prose write the opening of a story a…" enabled=true
+        AXSplitter value=number enabled=false
+        AXGroup subrole=AXHostingView
+          AXScrollArea
+            AXOpaqueProviderGroup subrole=AXOpaqueProviderList enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="ChatView.Message.Edit.<uuid>" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy response" help="Copy response" enabled=true
+              AXButton id="ChatView.Message.Retry.<uuid>" desc="Retry response" enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="ChatView.Message.Edit.<uuid>" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXScrollArea
+                AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy response" help="Copy response" enabled=true
+              AXButton id="ChatView.Message.Retry.<uuid>" desc="Retry response" enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="ChatView.Message.Edit.<uuid>" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy response" help="Copy response" enabled=true
+              AXButton id="ChatView.Message.Retry.<uuid>" desc="Retry response" enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="ChatView.Message.Edit.<uuid>" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy response" help="Copy response" enabled=true
+              AXButton id="ChatView.Message.Retry.<uuid>" desc="Retry response" enabled=true
+              AXStaticText value=text enabled=true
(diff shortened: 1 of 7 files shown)


raullenchai:
Verified on real hardware. The narrowed rule holds.

Full golden-flow suite, all 11 flows, against a build of this branch:

```
status: pass
80 dumps   incomplete: 0   with a failed field read: 5
```

The same five dumps as the run that motivated the redesign — `ba-privacy`,
`dead-panel-app`, `dead-panel-privacy`, `dead-panel-tools` (4 elements),
`update-app-panel` — so the read failure is structural and reproducible, exactly
as suspected. What changed is that it is now correctly classified as **not a
hole**: those elements carry an identifier and a value, so no filter is blind to
them.

`catalog-integrity` passes with the positive precondition, logging
`both catalog surfaces listed models, and neither offered a video-gen alias`, and
`ax-baseline.py` no longer refuses the suite.

## The per-element attribution earned its keep immediately

The old reason string said "a searched attribute was unreadable on 1 element(s)"
and gave no way to find which. With `unreadable: [...]` on the record:

```
role=AXCheckBox  id=Settings.Tools.Toggle.web_search          FAILED=description
role=AXCheckBox  id=Settings.Tools.Toggle.browse              FAILED=description
role=AXCheckBox  id=Settings.Tools.Toggle.weather             FAILED=description
role=AXCheckBox  id=Settings.Tools.Browse.AutoApproveToggle   FAILED=description
role=AXCheckBox  id=Settings.App.HideDockOnCloseToggle        FAILED=description
```

Five Settings checkboxes whose `AXDescription` read returns `kAXErrorFailure` —
not `attributeUnsupported`, which would mean "this element does not publish one".
Harmless to the assertions, but it means VoiceOver has no description for those
toggles. Filed separately; the point here is that the instrumentation found it,
and nothing else in the harness would have.

## Capacity, measured

Largest real dump: 108 records at depth 7, against caps of 12,000 and 40.
An ordinary app (Finder) reports `complete: true` with zero failed reads.

raullenchai:
Closing in favour of a re-derivation onto current `main`.

This branch is from 2026-08-08. `apps/rapid-mac/scripts/gui-golden-flows.sh`
has since moved through five PRs (#1666, #1702, #1708, #1721, #1731), and a
squash of this branch onto main now produces **17 conflict hunks** across the
harness, the driver and two baselines.

The problem is not the count, it is the direction: on several hunks the branch
side is the *older, thinner* version. Main has more flows in the dispatch list,
stricter transcript assertions (position-based ordering, interleaved
prompt/answer checks scoped to the transcript), and an Images tab that did not
exist when this was written. Resolving hunk-by-hunk means repeatedly judging
"which side is newer" on code that looks superficially similar, and I nearly
dropped a real assertion doing exactly that on the first attempt — main's
`catalog-integrity` gained a POSITIVE inventory check that the branch side has
no equivalent of, and a naive "take the branch" resolution deletes it silently.

Merging that is how a green harness quietly loses coverage.

**The contribution is still wanted**, and #1731's review independently
confirmed it: codex raised "the absence assertions accept incomplete AX walks"
as a MAJOR on the image-generation flow, which is precisely what this branch's
`assert_ax_absent` exists to prevent. I fixed those two call sites inline
there; the helper is what stops the next flow from reintroducing them.

What should come back, re-derived on current main:

- `ax_elements_match` / `assert_ax_absent` / `wait_ax_match` / `walk_reasons` —
  the absence helpers, including the retry-on-incomplete-dump and the
  distinction between "jq said false" and "jq query broke".
- `rapid-ax.swift` walk **reasons** (main already has `walk.complete` and
  `record_count` from #1666, but not the per-reason breakdown the helpers
  report on failure).
- `tests/test_gui_golden_absence_contract.py` — 486 lines, and it runs
  headless, so it can gate the helpers without a GUI session.

Branch kept (`--delete-branch=false`) so the content is recoverable.

Issue #1683, opened 2026-08-08
release-check-m3 cannot pass: G7b codex e2e subtests time out on qwen3.5-9b (same on v0.12.7)


raullenchai:
`make release-check-m3` cannot pass on this machine, and has not been able to
for at least a release. G7b fails on the gauntlet's own fixed model,
`qwen3.5-9b-4bit`, and it fails the same way on `v0.12.7` — so this is not a
regression, and it is also not something a release can be blocked on until it
is fixed or scoped.

## Evidence

Three runs today, all with the same shape: the `codex` harness profile reports
one or two **errors** (not failures) from its `e2e_*` subtests timing out,
while every other profile passes.

| where | result |
|---|---|
| full gauntlet, `HEAD` (#1671 branch) | `FAIL codex (163.1s) 11p 0f 1e 0s \| e2e_file_read: TIMEOUT` |
| scoped replay, same tree | `FAIL codex (249.3s) 10p 0f 2e 0s \| e2e_chat: TIMEOUT` |
| scoped replay, **`v0.12.7`** | `FAIL codex (163.2s) 11p 0f 1e 0s \| e2e_file_read: TIMEOUT` |

The other four profiles pass throughout — in the gauntlet run,
`opencode (24.2s) 10p 1s`, `hermes (465.3s) 32p 2s`, `aider (0.4s) 2p`,
`langchain (18.1s) 15p`.

Replay is:

```bash
rapid-mlx serve qwen3.5-9b-4bit --port 8000 --no-thinking &
RAPID_MLX_HARNESS_PROFILES_FILTER=codex \
  rapid-mlx bench qwen3.5-9b-4bit --tier harness --base-url http://127.0.0.1:8000
```

## What it is not

* **Not caused by #1671.** That branch changes four files —
  `scripts/release_check_m3.sh`, `scripts/release_check_m3_random.py` and their
  two test files — and no engine code at all.
* **Not one bad test.** Which subtest times out varies between runs
  (`e2e_file_read`, `e2e_chat`), so the case is not the problem; the path is.
* **Not the per-profile cap.** The profile finishes in 163-249 s against a cap
  of 600-1200 s. What expires is the inner `query_timeout`, which defaults to
  120 s (`vllm_mlx/agents/base.py:51`).

## What it is

The `codex` agent binary, driving a local `qwen3.5-9b-4bit`, sometimes does not
complete a single-step request — "Read the first line of pyproject.toml", or
plain chat — inside 120 s.

[...]

PR #1690, opened 2026-08-08
fix(agents): codex e2e gate never reached rapid-mlx at all

raullenchai:
## Why

G7b's codex e2e tests have been failing for reasons that live entirely **between the test runner and the Codex CLI**. rapid-mlx was never reached, but the failure surfaced as a rapid-mlx integration failure — so the gate has been reporting on something it never measured.

### 1. The child inherited stdin

`_agent_query` called `subprocess.run(...)` with no `stdin=`, so the agent CLI inherited ours. A CLI that reads stdin when it isn't a TTY then blocks until `query_timeout` instead of answering the query it was handed on argv. Codex does exactly this:

```
$ codex exec 'What is 2+2? Reply with just the number.'
Reading additional input from stdin...
```

Every agent CLI here runs headless, so none of them has any business reading stdin.

### 2. The codex profile omitted `--skip-git-repo-check`

```
Not inside a trusted directory and --skip-git-repo-check was not specified.
```

The runner inherits the caller's cwd, so from a neutral directory every codex e2e test failed in ~100 ms with that refusal. Same bug as (1) wearing a different symptom depending on where you stood.

### 3. …which means the agent needs a directory we own

Bypassing the trust check in an inherited, caller-controlled directory would expose whatever the caller happened to be standing in — including files and agent instructions from an untrusted checkout. Each e2e test now runs in its own `_e2e_workspace()`: a fresh temp directory holding one file we wrote, removed afterwards.

That also fixes a second-order bug already visible in the gate: `_test_e2e_file_read` asks the agent to read `pyproject.toml`, which only existed when the caller happened to be standing in a Python project. Run from anywhere else the agent hunted for a file that wasn't there and burned the whole timeout.

## Measured

M3 Ultra, local `qwen3.5-9b-4bit` server, `rapid-mlx agents codex --test` run from `/tmp` — no git repo, no `pyproject.toml`:

| | before | after |
|---|---|---|
| `e2e_chat` | ❌ FAIL (101 ms) — "Not inside a trusted directory" | ✅ PASS (34.9 s) |
| `e2e_terminal` | ✅ | ✅ |
| base tests | 10/12 | 11/12 |

## Review rounds

Three rounds, six BLOCKING findings, each reproduced before it was fixed:

| round | finding | reproduced as |
|---|---|---|
| 1 | trust check bypassed in an inherited cwd | — |
| 2 | one workspace shared by all three e2e tests | "both e2e tests ran in the SAME directory" |
| 2 | `"build"`/`"project"` accepted anywhere in the answer | a reply that never opened the file passed |
| 3 | `os.chmod` follows symlinks | an outside file created `0o600` came back `0o700` (`st_mode=33216`) |
| 3 | walk ran before the root's permissions were restored | locked root hid every descendant from recovery |

Round 2's first finding also caught a bad test of mine: `test_workspace_is_fresh_each_time` asserted that two `_e2e_workspace()` calls return different paths — a property of the helper, not of the runner, and the runner is what decides how many workspaces exist. It has been replaced with one that drives the real `_test_e2e_*` entry points against a real subprocess and reads back the directories they actually ran in.

## On the tests

12 tests. The ones that matter use the real thing rather than a mock, because a mock would only restate the diff:

- the stdin test drives a **real subprocess against a real pipe on fd 0** — reproducing the hang needs the actual inherited descriptor
- the workspace-isolation test drives the **real `_test_e2e_*` functions** and reads back their actual cwds
- the symlink test plants links to a real outside file and a real outside directory and checks their `st_mode` afterwards

Every fix fails its own test when reverted; I checked each one.

## Not fixed here

`e2e_file_read` still fails on the gauntlet's default `qwen3.5-9b-4bit`, and that one is **model competence, not the engine**. Same task, same server, same codex invocation:

| model | result |
|---|---|
| `qwen3.5-9b-4bit` | tool call succeeds in 0 ms and returns the first line, then the loop never terminates — killed at 600 s |
| `qwen3.6-35b-4bit` | correct answer, terminates in **55.9 s**; full gate **12/12 in 66.8 s** |

`codex.yaml` claims "9B is the practical floor"; that claim is falsified by this measurement for the multi-tool workload. Whether the gauntlet should raise its agent-gate model is a release-policy call, tracked on #1683 rather than decided here.

Refs #1683

### DIFF
--- a/tests/test_agent_query_stdin_and_trust.py
+++ b/tests/test_agent_query_stdin_and_trust.py
@@ -1,9 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Pin the two reasons G7b's codex e2e could never pass (#1683).
+"""Pin why G7b's codex e2e could never pass, and where it runs (#1683).
 
-Both live on the path between the test runner and the agent CLI, so both
-report as a rapid-mlx integration failure while nothing in rapid-mlx is
-ever reached.
+The first two live on the path between the test runner and the agent CLI,
+so both reported as a rapid-mlx integration failure while nothing in
+rapid-mlx was ever reached.
 
 1. ``_agent_query`` ran the child with an **inherited stdin**. A CLI that
    reads stdin when it isn't a TTY then blocks until ``query_timeout``
@@ -15,6 +15,14 @@ ever reached.
    Codex refuses to run outside a trusted git repo, and the runner
    inherits the caller's cwd, so from a neutral directory every e2e test
    failed in ~100 ms with "Not inside a trusted directory".
+
+3. Fixing (2) by itself would bypass Codex's trust check in whatever
+   directory the caller happened to be standing in, exposing that
+   directory's files and agent instructions to the agent. The e2e tests
+   now run in a workspace ``_e2e_workspace`` builds, which also makes
+   `_test_e2e_file_read` deterministic — it asks for ``pyproject.toml``,
+   and previously only found one when the caller was standing in a
+   Python project.
 
 The stdin test drives a **real subprocess against a real pipe on fd 0**.
 A mock asserting ``stdin=DEVNULL`` was passed would only restate the
@@ -28,11 +36,12 @@ import os
 import shlex
 import sys
 import time
+from pathlib import Path
 
 import pytest
 
 from vllm_mlx.agents import get_profile
-from vllm_mlx.agents.testing import _agent_query
+from vllm_mlx.agents.testing import _agent_query, _e2e_workspace
 
 # Long enough that a slow machine never flakes, short enough that a
 # regression (the child blocking on stdin) fails the suite in seconds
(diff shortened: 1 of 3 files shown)


raullenchai:
## Round 7 — one accepted, one held, and why

**Accepted.** `test_each_e2e_test_gets_its_own_workspace` drove only two of the three e2e entry points, so `_test_e2e_terminal` could have regressed to a shared cwd with the test still green. All three now run and three *distinct* directories are required. Verified against a shared-workspace mutation:

```
two or more e2e tests ran in the SAME directory — whatever the first agent
wrote there is the next one's starting condition:
['…/shared-xd0vwo3_', '…/shared-xd0vwo3_', '…/shared-xd0vwo3_']
```

**Held — the permission-repair pass stays out.**

This review asks for it back (via descriptor-relative traversal) because without it a mode-`000` directory leaks the workspace. Rounds 3 and 4 asked for it *gone*, having caught it as a symlink escape and then as a TOCTOU. Round 5 removed it. Reversing again on the same axis is oscillation, so I'm stopping and writing the reasoning down instead.

The reason has not changed:

> the agent runs as the **same user**, with the same permissions we have.

Anything our cleanup could be tricked into chmod'ing, the agent can chmod itself, directly. There is no privilege boundary here to defend — so the repair pass cannot buy safety. What it *can* do is add a write primitive that follows attacker-chosen paths, which is exactly what rounds 3 and 4 caught it doing (measured then: a caller-owned file created `0o600` came back `0o700`).

Against that, the leak it prevents is:

- one directory, in a location the OS reclaims
- reported by name in a warning, not silent — which is what the round-2 NIT actually asked for
- reachable only if an agent chmods its own working directory to `000` while answering "What is 2+2?", "Read the first line of pyproject.toml", or "Run `echo <marker>`"

I think that trade is clearly right, but it is a judgement, not a fact, and I've now reversed on it once. If you want the repair pass back, it's a one-line change to `_remove_workspace` and I'll write the descriptor-relative version properly — say the word.

## Where the PR stands

Seven rounds. Every finding reproduced before being fixed; several were against code I introduced in an earlier round.

| round | finding | disposition |
|---|---|---|
| 1 | trust check bypassed in an inherited cwd | fixed |
| 2 | one workspace shared by all three e2e tests | fixed |
| 2 | `"build"`/`"project"` accepted anywhere in the answer | fixed (sentinel) |
| 3 | `os.chmod` follows symlinks | fixed |
| 3 | walk ran before the root's permissions were restored | fixed |
| 4 | TOCTOU between `islink()` and `chmod()` | fixed by removing the repair pass |
| 5 | workspace creation failure crashed the runner | fixed |
| 5 | symlink test never forced the failure path | fixed (twice — the first fix was also vacuous) |
| 5 | paths split on whitespace | fixed |
| 6 | comments overclaimed isolation | fixed (comment now says what it does) |
| 6 | two tests failed instead of skipping under root | fixed |
| 6 | `except OSError` too broad | fixed |
| 7 | terminal entry point not covered | fixed |
| 7 | repair pass should return | **held — see above** |

Issue #1689, opened 2026-08-08
Markdown tables render visually but carry no table semantics for VoiceOver


raullenchai:
A markdown table in a chat answer renders visually as a table, but carries no
table semantics in the accessibility tree — so VoiceOver announces it as a run
of loose text with no rows, no columns and no header association.

Measured on a real run of the new `chat-depth` golden flow, with the answer:

```
| model | size | speed |
| --- | --- | --- |
| qwen3.5-9b | 5.2 GB | 74 tok/s |
| llama-3.1-8b | 4.5 GB | 68 tok/s |
```

What the AX tree contains:

* each cell as its own `AXStaticText` — so the markdown **was** parsed, the
  pipes are gone and the cells are separate elements;
* **no `AXTable`, `AXRow`, `AXCell` or `AXGrid` role anywhere in the dump.**

The parsing half is right; the semantics half is missing. For a sighted user the
table is fine. For a screen-reader user there is no way to move by row or
column, and no way to know which header a value belongs to — `74 tok/s` is
announced with nothing tying it to *speed* or to *qwen3.5-9b*.

Related to #1687 (five Settings toggles whose `AXDescription` read fails) — same
theme, and both were surfaced by instrumenting the harness rather than by any
product test. Neither blocks a release; both are things a user relying on
assistive technology hits immediately.

The fix is presumably an `accessibilityRepresentation` (or an `AXTable`-shaped
container) on whatever view renders a parsed markdown table, so the structure
the parser already recovered survives into the tree instead of being flattened.

The new flow will notice if this ever changes: `chat-depth` sends a table every
run and records a structural baseline, so gaining table roles shows up as a
reviewable baseline diff rather than going unnoticed.

PR #1691, opened 2026-08-08
test(mac): one message is not a conversation

raullenchai:
## Why

`chat-restore` sends a single prompt and checks it comes back after a relaunch. That covers persistence and almost nothing about chatting. It cannot see:

- a second turn landing above the first
- a turn dropped when the next one starts
- a restore that brings back only the last exchange

And every answer it has ever rendered was the same paragraph of plain text — so the code block, the table, the list and the CJK line have **never once** been through the renderer in this suite.

## What `chat-depth` does

Five turns, each asking the fake sidecar for a different **shape** of answer:

| shape | prompt | expected |
|---|---|---|
| prose | an opening paragraph | `lighthouse keeper` |
| code | fibonacci in python | `def fib(n)` |
| table | a two-model comparison | `qwen3.5-9b` |
| list | three nested steps | `nested point` |
| unicode | a CJK line with emoji | `中文排版测试` |

The fake has no model, so this is **not** about whether an answer is any good — judging that belongs to the eval suites against a real model. It is about the work the app does differently per shape, which is what a GUI gate can actually see:

- every turn is still present, **in order**, after the next one arrives
- markdown reached the renderer **as markdown**: no stray ``` fences or `|` pipe rows survive in the rendered text. That is exactly what a renderer fallback looks like, and every "does the text appear" assertion in this file passes on it — the text does appear, wearing its syntax.
- a code block is its own view, not a paragraph that happens to contain code
- all five turns survive a relaunch, not just the last exchange

The fake gained a `RESPONSE_SHAPES` table keyed off a marker in the **last** user message. Requests with no marker stream the previous default unchanged, so no existing flow moves.

## Relationship to #1675

Split out of #1675 so it can land on its own. That PR's completeness-signal redesign is still under review (codex r2 raised 3 BLOCKING against it, all sound); this flow does not reference any of it — `ax_elements_match`, `wait_ax_match`, `assert_ax_absent`, `redump_evidence` and `walk_reasons` are all absent from this branch.

## Verification status — please read

- Ran **3×** on the #1675 branch: found a missing baseline, recorded it, then confirmed stable across a second run. Both structural baselines are in this PR.
- **Not yet re-run on this main-based branch.** The machine's screen is currently locked (`CGSSessionScreenIsLocked = true`), and with the screen locked every app reports zero windows, so the suite fails for reasons that have nothing to do with the code.
- What has been checked here: `bash -n` clean, `shellcheck -S warning` clean, and no reference to any helper dropped during the split.

I'll re-run `--flow chat-depth` on this branch as soon as the screen is unlocked and post the result before this merges.

### DIFF
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-depth.five-turns.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-depth.five-turns.txt
@@ -1,0 +1,91 @@
+# rapid-mac AX structural baseline v1
+AXApplication title="Rapid-MLX"
+  AXWindow subrole=AXStandardWindow id="main" title="Rapid-MLX"
+    AXGroup subrole=AXHostingView
+      AXSplitGroup id="main, SidebarNavigationSplitView"
+        AXGroup subrole=AXHostingView
+          AXButton id="Sidebar.NewChat" desc="New Chat" enabled=true
+          AXButton id="Sidebar.Launch" desc="Launch" enabled=true
+          AXScrollArea
+            AXHeading desc="<relative-day>" enabled=true
+            AXButton id="Sidebar.Conversation.Pin.<uuid>" desc="Pin conversation" help="Pin conversation" enabled=true
+            AXPopUpButton id="Sidebar.Conversation.Menu.<uuid>" desc="Conversation actions" enabled=true
+            AXButton id="Sidebar.Conversation.<uuid>" desc="shape:prose write the opening of a story a…" enabled=true
+        AXSplitter value=number enabled=false
+        AXGroup subrole=AXHostingView
+          AXScrollArea
+            AXOpaqueProviderGroup subrole=AXOpaqueProviderList enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="ChatView.Message.Edit.<uuid>" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy response" help="Copy response" enabled=true
+              AXButton id="ChatView.Message.Retry.<uuid>" desc="Retry response" enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="ChatView.Message.Edit.<uuid>" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXScrollArea
+                AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy response" help="Copy response" enabled=true
+              AXButton id="ChatView.Message.Retry.<uuid>" desc="Retry response" enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy message" help="Copy message" enabled=true
+              AXButton id="ChatView.Message.Edit.<uuid>" desc="Edit message" help="Edit message" enabled=true
+              AXDisclosureTriangle desc="Reasoning" value=bool:false enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXStaticText value=text enabled=true
+              AXButton id="ChatView.Message.Copy.<uuid>" desc="Copy response" help="Copy response" enabled=true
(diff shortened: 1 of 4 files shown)


raullenchai:
## Codex review round 1 — all four findings fixed

Two BLOCKING, two MAJOR, all sound. Fixed in 50ce6be, and each is now pinned by a mutation of a **real recorded five-turn dump**.

### BLOCKING — assistant answers were never ordered or paired

`assert_text_order` was called with the five prompt markers only, so a restore that returns every turn but pairs the fifth answer with the first question is equally "sorted". The order check now interleaves prompt and answer.

### BLOCKING — the search was app-wide, not transcript-scoped

One correction to the finding as written: it says the sidebar title can satisfy the check *today*. Measured on a real dump, it cannot — the sidebar row carries the prompt text in `description`, the transcript bubble carries it in `value`, and the helper only reads `value`/`title`:

```
[11] id='Sidebar.Conversation.67BB9DBA-…'  desc='shape:prose write the opening of a story a…'
[16] depth=7 AXStaticText  value='shape:prose write the opening of a story about a lighthouse'
```

The *principle* stands, though — the assertion is app-wide and passes only by coincidence of which field the sidebar happens to use. Give that row a `title` and it starts reading the sidebar, silently. `transcript_only` now narrows every assertion to the `AXOpaqueProviderList` subtree, and refuses a dump whose transcript is empty rather than letting another surface answer for it.

### MAJOR — "rendered" was proved only by the absence of source syntax

Measured: the renderer puts every table cell and every list item in its **own** `AXStaticText` with the source marker stripped.

```
[42] 'qwen3.5-9b'   [43] '5.2 GB'   [44] '74 tok/s'
[59] 'a nested point'   [60] 'another one'
```

`assert_rendered_as_separate_nodes` requires an exact whole-element match, and `assert_no_literal_list_markers` rejects any line still wearing `-`/`*`/`1.`.

### MAJOR — the CJK turn asserted a six-character prefix

The emoji run and the right-to-left segment are asserted too.

## Verification

Three previously recorded real dumps still pass unchanged (57 transcript elements each). Then one regression at a time, mutated into a real dump — each assertion fires on the case it names:

| mutation | assertion that caught it |
|---|---|
| answers swapped 1↔5 | `transcript is out of order` |
| table flattened into one node | `'qwen3.5-9b' is part of '<the whole row>'` |
| list printed with `- ` | `a list marker reached the screen verbatim` (+ separate-nodes) |
| CJK truncated to the prefix | emoji / RTL / body all missing |
| transcript emptied, sidebar keeps the text | `the transcript container has no children` |

## Still outstanding

The live GUI re-run on this branch. The machine's screen is locked (`CGSSessionScreenIsLocked = true`), and with the screen locked every app reports zero windows, so the suite fails for reasons unrelated to the code. `bash -n` and `shellcheck -S warning` are clean.

raullenchai:
## The live GUI run, now that the screen is unlocked

`--flow chat-depth`, on this branch, against a freshly built app:

```
turn 1 (shape:prose) rendered and both sides counted
turn 2 (shape:code) rendered and both sides counted
turn 3 (shape:table) rendered and both sides counted
turn 4 (shape:list) rendered and both sides counted
turn 5 (shape:unicode) rendered and both sides counted
all 5 turns present, each answer inside its own assistant message
markdown rendered: table cells and list items are their own elements,
no raw fences, pipe rows or list markers, code block nested and intact,
and the CJK answer kept its emoji and its right-to-left run
all 5 turns restored, each answer still under its own prompt,
and every shape still rendered the way it was before the relaunch
PASS — chat-depth
```

Both structural baselines matched byte-for-byte — `git status` is clean after the run, so the snapshots recorded on the #1675 branch are still exactly what this branch produces.

The full suite is running now; I'll post it when it finishes. The changed helpers (`transcript_only`, `assert_text_order`, `assert_turns_pair_up`, `assert_rendered_shapes`, `assert_no_literal_list_markers`, `assert_rendered_as_separate_nodes`) are referenced only from `flow_chat_depth`, so the blast radius on the other nine flows is nil by construction — but a shared script deserves the real check, not an argument.

## Codex round 3 — one BLOCKING and one MAJOR fixed, two answered

**BLOCKING — an answer could be rendered inside the user's own bubble.** Reading order says nothing about which *message* a string is in. `assert_turns_pair_up` now segments the transcript on the app's own controls (a user message ends at its Edit button, an assistant message at its Retry button) and requires turn N's prompt in the Nth user message and turn N's answer in the Nth model message. Against the mutation:

```
model message 2 does not contain 'lighthouse keeper';
it holds 'False Sure, here is something else entirely. Only the first was ever read by any'
```

**MAJOR — a distinctive phrase near the start passes on a truncated stream.** The fake is deterministic, so the endings are knowable and now asserted: the prose's last sentence, `    return a` from the code block, and the table's closing line.

**MAJOR (answered, not fixed) — nested list items are not asserted as nested.** Measured on a real dump: every list item sits at depth 7, including the two nested ones. The app does not publish nesting at all, so there is nothing here for an assertion to check. Same gap as the table semantics — #1689.

**MINOR (answered) — RTL display order and glyph overlap.** Neither appears in the accessibility tree in any form. An AX harness cannot see them; that needs pixel comparison, which this suite does not do. The Unicode checks claim character survival and nothing more, and the log line says exactly that.

raullenchai:
## Full golden-flow suite: PASS

```
[gui-golden] PASS — all
```

- 219 artifacts written, 108 of them AX dumps
- **No baseline drift**: `git status` clean after the run, so no recorded snapshot moved
- The other nine flows are unaffected, as expected — every helper this PR touches or adds (`transcript_only`, `assert_text_order`, `assert_turns_pair_up`, `assert_rendered_shapes`, `assert_no_literal_list_markers`, `assert_rendered_as_separate_nodes`) is referenced only from `flow_chat_depth`

One thing I checked and will *not* claim: I started to report that no dump came back with `walk.complete: false`. That would have been meaningless here — **zero** of the 108 dumps carry a `walk` field at all, because the completeness signal is #1675's change and this branch is cut from `main` without it. Nothing about walk completeness can be said from this run either way.

That closes the verification gap in the PR description. This branch has now had:

| | |
|---|---|
| `--flow chat-depth`, this branch | PASS |
| full suite, this branch | PASS |
| structural baselines | matched byte-for-byte |
| `bash -n`, `shellcheck -S warning` | clean |
| three codex rounds | 3 BLOCKING + 5 MAJOR fixed, 2 answered |
| assertion mutation testing | 11 regressions, each caught by the assertion that names it |

PR #1708, opened 2026-08-09
fix: decouple semantic GUI flows from screen capture

raullenchai:
## Why
Semantic AX flows were falsely blocked because Peekaboo window enumeration requires Screen Recording even when no screenshot is taken. The app AX tree is the authoritative semantic oracle and works with Accessibility alone.

## Changes
- use AX for startup and Settings-window presence
- require Screen Recording only for screenshot flows
- preserve Settings window IDs for those screenshot flows

## Validation
- GUI completeness tests and bash syntax
- mini catalog-integrity PASS with Accessibility granted and Screen Recording denied

### DIFF
--- a/apps/rapid-mac/scripts/gui-golden-flows.sh
+++ b/apps/rapid-mac/scripts/gui-golden-flows.sh
@@ -63,6 +63,12 @@ done
 log() { printf '[gui-golden] %s\n' "$*"; }
 die() { printf '[gui-golden] FAIL: %s\n' "$*" >&2; exit 1; }
 pb() { peekaboo "$@" --bridge-socket "$BRIDGE"; }
+flow_requires_screen_recording() {
+    case "$FLOW" in
+        all|fresh-install|low-memory-choice|browse-all-destination) return 0 ;;
+        *) return 1 ;;
+    esac
+}
 pb_click_coords() {
     local coords="$1"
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex round 5 dismissal with concrete evidence: the proposed fully-stubbed test would duplicate the real integration run at lower fidelity. On the mini, with real Peekaboo 3.0 reporting Accessibility=true and Screen Recording=false, these semantic flows all reached and passed their assertions: settings-persistence, chat-restore, restored-tools, tool-loop-budget, chat-depth, slow-stream-stop, model-crash-recovery, loaded-model-benchmark, update-state, no-dead-controls, and catalog-integrity. Their artifacts are under /tmp/rapid-gui-final*, /tmp/rapid-gui-final2-*, and /tmp/rapid-gui-retest-restored-tools. The unit contract additionally requires direct window enumeration to be absent from wait_for_window, requires capture refresh to sit behind flow_requires_screen_recording, and executes representative positive/negative permission classifications. This is stronger evidence for the actual call paths than a stubbed Peekaboo. The dispatcher-completeness suggestion is a non-blocking future-maintenance enhancement; current screenshot commands and permission allowlist were exhaustively matched by rg during this change.

alberto4003 (approved):

PR #1711, opened 2026-08-09
fix: suppress replayed terminal stream content

### LINKED ISSUE #1709: Streaming: terminal chunk replays the full response in delta.content when --enable-auto-tool-choice is set
Thanks for Rapid-MLX — it has been a pleasant drop-in on Apple Silicon.

**Summary:** when the server is started with `--enable-auto-tool-choice --tool-call-parser auto`, the terminal SSE chunk (the one carrying `finish_reason`) repeats the entire response in `delta.content`, so a client that concatenates deltas renders the answer twice. Without those flags the same request is fine.

I searched open and closed issues for this (duplicate/repeated content, terminal chunk, final delta) and didn't find a match; apologies if I missed one. I did not check earlier versions, so I can't say whether this is a regression.

### Environment

| | |
|---|---|
| Rapid-MLX | 0.12.7 (pip, clean venv; `vllm_mlx/routes/chat.py` is byte-identical to tag `v0.12.7`) |
| Hardware | MacBook Pro 16" M3 Pro, 36 GB |
| macOS | 26.5 |
| Python | 3.13 |
| Model | `lfm2.5-8b-a1b-4bit` (`mlx-community/LFM2.5-8B-A1B-MLX-4bit`); also seen with `mlx-community/Qwen3.6-35B-A3B-4bit` |
| Serve command | `rapid-mlx serve lfm2.5-8b-a1b-4bit --port 8000 --enable-auto-tool-choice --tool-call-parser auto` |
| Mode | Streaming (`stream: true`). Non-streaming was unaffected in my testing. |

Startup banner for the affected run:

```
  🐆 Rapid-MLX
  ─────────
  Features: tools: auto, reasoning: qwen3, gc-control, cors: *
  Model: mlx-community/LFM2.5-8B-A1B-MLX-4bit
```

No warnings about the flag combination, and no error or log line when the duplication happens — the response is silently doubled.

### Reproduction

```bash
curl -sN http://127.0.0.1:8000/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"lfm2.5-8b-a1b-4bit","messages":[{"role":"user","content":"Reply with exactly: hello world"}],"max_tokens":300,"temperature":0,"stream":true}'
```

Tail of the stream (keepalives and blank lines stripped):

```
data: {...,"choices":[{"index":0,"delta":{"reasoning_content":".\n"}}]}
data: {...,"choices":[{"index":0,"delta":{"content":"\n"}}]}
data: {...,"choices":[{"index":0,"delta":{"content":"hello"}}]}
data: {...,"choices":[{"index":0,"delta":{"content":" world"}}]}
data: {...,"choices":[{"index":0,"delta":{"content":"\nhello world"},"finish_reason":"stop"}]}
data: [DONE]
```

The content deltas already spell out `"\nhello world"`, and then the `finish_reason: "stop"` chunk carries `"\nhello world"` again. Concatenating `delta.content` yields `"\nhello world\nhello world"`.

I'd expect the terminal chunk to carry no `content` key at all here. This model does emit reasoning, which arrives separately as `reasoning_content` deltas (relevant to the note under "Possible fix").

I first ran into this through [Hermes Agent](https://github.com/NousResearch/hermes-agent) pointed at the local endpoint, where every reply was rendered twice, but as shown it reproduces with plain `curl`, so it isn't client-specific.

### Scope

Same model, same request, only the serve flags differ:

| serve flags | duplicated? |
|---|---|
| (none) | no |
| `--tool-call-parser auto` | no |
| `--served-model-name local` | no |
| **`--enable-auto-tool-choice --tool-call-parser auto`** | **yes** |

I couldn't isolate `--enable-auto-tool-choice` on its own — the CLI rejects it with `--enable-auto-tool-choice requires --tool-call-parser` — so what I can say is that the combination triggers it while `--tool-call-parser auto` alone does not.

### Where it seems to come from

[`vllm_mlx/routes/chat.py#L6626-L6636`](https://github.com/raullenchai/Rapid-MLX/blob/73a04c33cc401b2630f9d4f8d55814c6c155a7d5/vllm_mlx/routes/chat.py#L6626-L6636):

```python
# Merge any released prefix-held content into the terminal
# chunk's delta.content (codex round-4 CRITICAL). Concatenate
# to whatever the finish event already carries — the
# finish_event.content path is normally None for non-tool
# plain-text streams (deltas already drained content during
# the loop), so this typically just adds the held suffix.
terminal_content = (
    ""
    if _forced_terminal_content_consumed
    else (finish_event.content or "") + finalize_content
)
```

The comment in that block spells out the assumption that doesn't seem to hold here: with the flags above, `finish_event.content` is not `None` for a plain-text stream — it appears to carry the full response, so the terminal chunk replays everything instead of adding a held suffix.

### Workaround

Dropping `--enable-auto-tool-choice` avoids it. Tool calling still worked for me without the flag — `serve` already reports a per-model tool parser in the banner (`tools: lfm` for this model), and `tool_calls` were emitted correctly both streaming and non-streaming, including end-to-end through an agent harness. That's one model with the `lfm` parser, so others may differ; it also leaves me unsure what the flag is meant to add on top of the default parser, which may itself be worth a look.

Clients that can't change the serve command can drop a terminal chunk whose content equals everything already received.

### Possible fix

One shape that worked locally: suppress the terminal delta when it is a verbatim replay of the content already put on the wire, leaving the legitimate case (content held back by the parser and released as a shorter, different string at the end) untouched. One gotcha if you go that way — `processor.accumulated_text` looked like the natural thing to compare against, but in my testing it holds the model's raw text including reasoning, so it didn't match; tracking what `_content_sse_chunk` actually serialized did.

That fixed the duplication for me without affecting tool calls or the non-streaming path, but I don't know the parser internals well enough to judge whether suppressing at this point is the right layer versus preventing `finish_event.content` from being populated further upstream, so I'll leave that call to you.

Happy to open a PR if a fix in that shape would be useful.

raullenchai:
Closes #1709

## What changed

Track the sanitized content actually serialized into streaming SSE deltas and suppress a terminal `delta.content` only when it is a verbatim replay of everything already delivered. Legitimate parser-held suffixes, tool calls, reasoning, finish reasons, and usage placement remain unchanged.

## Root cause

With auto tool choice enabled, parser finalization can attach the complete plain-text response to the finish event after those same bytes were already emitted as ordinary content deltas. The route treated that finish payload as newly held content and serialized it again.

## Validation

- Regression test fails on `main` with `hello worldhello world` for both finish-event and finalize-event snapshots
- The same regression test passes with this fix
- `python -m pytest tests/test_chat_streaming_spec.py -q` (12 passed)
- `ruff check vllm_mlx/routes/chat.py tests/test_chat_streaming_spec.py`


raullenchai:
Landing per the codex-nonconvergence stopping rule (objective gates green across 8 pr_validate rounds).

**Objective gates (round 8):** full_unit 16638 passed · targeted_tests 482 passed · lint clean · supply_chain clean · fetch pass. Green every round.

**Real bug fixed & live-verified:** concatenated streaming `delta.content` = single `"\nhello world"` on the exact issue repro (`lfm2.5-8b-a1b-4bit --enable-auto-tool-choice --tool-call-parser auto`), confirmed on a live server 7×. Root cause was deeper than the reporter's hypothesis: with a reasoning parser active, the auto tool parser's `_content_emitted_len` counter is bypassed, so `StreamingPostProcessor.finalize`'s `flush_held_content` returns the entire content buffer and the terminal chunk replays it.

**Coverage (16 hermetic regressions):** finish-event snapshot, finalize-flush snapshot, snapshot split across finish+finalize, sanitizer-sensitive content, genuine held-suffix preservation (`"ha"`+held`"ha"`→`"haha"`), and empty-buffer suffix survival.

**Final codex blocking — not pursued (theoretical + reviewer flip-flop):** the concern is that a removable special token split across two content deltas defeats per-delta sanitization so the provenance check fails to match. This is unreachable in practice — removable markers like `<|im_end|>` are stop tokens that end generation, not streamed as split content — and in that impossible case the guard simply errs toward NOT suppressing (preserving content), the safe direction that never drops legitimate output. codex flip-flopped on this exact point across rounds (requested whole-string sanitization earlier, then flagged the test defending it → removal, now requests it again). Per the stopping rule, not chasing a non-deterministic adversarial reviewer further when all objective gates are green and the real bug is fixed and verified.

PR #1732, opened 2026-08-09
fix(cache): reuse first-turn prefix on turn two

### LINKED ISSUE #1611: Turn 2 of every conversation re-prefills the whole prompt: a non-trimmable cache entry overruns the shared prefix by a few tokens
## What

In a plain multi-turn chat, **turn 2 misses the prefix cache and pays a full cold prefill**; turn 3 hits. Measured on `qwen3.6-27b-4bit` with the published 0.12.5 wheel (`pip install rapid-mlx==0.12.5`, served from a neutral cwd):

| Turn | Wall clock | Cache |
| --- | ---: | --- |
| 1 (cold, ~3k-token prompt) | 9.92 s | — |
| 2 (strictly extends turn 1) | **10.00 s** | **MISS** |
| 3 (strictly extends turn 2) | **0.89 s** | HIT (`cached=3027 remaining=35`) |

Turn 3 is **11x** faster than turn 1. Turn 2 is not faster at all, despite sharing ~3019 tokens with what was just stored.

## Why

The engine says so directly:

```
[cache_fetch] LCP unavailable: shared=3019 entry_len=3022 requested_len=3042 non_trimmable=True
[cache_fetch] request=a87dc2c4-8e7 MISS prompt_tokens=3042 time=0.000s entries=8
```

The stored entry is 3022 tokens; the incoming request shares only the first 3019 of them. The entry **overruns the shared prefix by 3 tokens**, and because the cache is `non_trimmable` it cannot be trimmed back to the shared point — so a 3019-token match is discarded and the request re-prefills all 3042.

Those 3 tokens are the assistant turn's own tail (the reply plus its end-of-turn markers), which the next user turn does not repeat verbatim before its own content.

By turn 3 an entry exists whose length is <= the new shared prefix, so the lookup succeeds.

## Why it is not simply the known design

`scheduler.py` already anticipates this. The completion-store path carries:

```python
# Bounded trim-free reuse deliberately segments a cold request at
# N-1. Saving an additional exact N-token entry would win lookup
# priority, fail trim-one kickoff on non-trimmable caches, and mask
# the reusable N-1 prefix.
```

and the mid-prefill save path says `evict_prefixes=False` keeps boundary entries so that *"future requests with the same prefix but different suffix get a prefix cache hit (critical for agentic multi-turn)"*.

So a reusable sub-prefix entry is supposed to exist for exactly this case. In this run it did not: the only `[prompt_cache_save]` lines were for small requests (24-491 tokens); the ~3k request produced only a completion-time `[cache_store] tokens=3033 (3031 prompt + 2 output)`. Whatever creates the N-1 boundary entry did not fire here.

## Impact

User-visible latency on the second turn of every conversation, which is the most common turn in a chat app — 10 s instead of 0.9 s here. Agentic clients feel it on their first follow-up too.

## Repro

```python
bg = "Background:\n" + ("The quick brown fox jumps over the lazy dog. " * 300)
msgs = [{"role": "user", "content": bg + "\n\nReply with the single word ONE."}]
# turn 1 -> ~9.9s
msgs += [{"role": "assistant", "content": "ONE"},
         {"role": "user", "content": "Now reply with the single word TWO."}]
# turn 2 -> ~10.0s, MISS          <-- the bug
msgs += [{"role": "assistant", "content": "TWO"},
         {"role": "user", "content": "Now reply with the single word THREE."}]
# turn 3 -> ~0.9s, HIT
```

## Scope

Not verified as a regression — found dogfooding the published 0.12.5 wheel right after release, and not A/B'd against 0.12.4. Correctness is unaffected; this is latency only, which is why it did not block the release.

Possibly related: #1350 / #1392 (prefix-cache load deferred off readiness).

raullenchai:
## What does this PR do?

Allows the first user turn to compute the same conservative stable prefix boundary used by later turns, so non-trimmable hybrid caches can reuse it on turn two.

## Why is this needed?

Fixes #1611. The previous `last_user_idx == 0` guard skipped the stable boundary on turn one and fell back to an N-1 snapshot. For chat templates with a multi-token assistant-generation suffix, that snapshot extended three tokens past the next request shared prefix, forcing a full turn-two cache miss.

## AI assistance disclosure

Codex reproduced the issue, wrote the one-line production fix and regression test in `vllm_mlx/engine/batched.py` and `tests/test_prefix_boundary_path_parity.py`, independently reviewed the committed diff twice, and verified it with focused and broad cache suites plus a live Qwen3.5-4B three-turn run. The maintainer can explain the intent, risk, and behavior of every change.

> By submitting this PR I confirm I can explain the intent, risk, and behavior of every non-generated change in this PR. For any generated / boilerplate / scaffolded sections, I have identified them above and can describe how I verified them.

## Test plan

- [x] Exact negative control: restoring `last_user_idx == 0` makes `test_first_turn_saves_stable_boundary_before_generation_suffix` fail with `boundary 0 != 2`; restoring the fix passes
- [x] Cache regression selection: 275 passed
- [x] Full unit: 16,600 passed, 123 skipped, 18 deselected, 6 xfailed, 1 xpassed
- [x] Ruff lint and format checks pass
- [x] Live Qwen3.5-4B: turn two changed from 5.210s MISS to 0.458s HIT (`cached_tokens=3008`)
- [x] Codex review converged twice with zero blocking findings
- [x] Local `pr_validate 1732`: MERGE-SAFE; stress explicitly skipped because the 18GB host cannot meet the registry's 26GB minimum including OS headroom
- [x] GitHub CI and all five live L1 smoke checks pass

## Checklist

- [x] Tests pass locally
- [x] Lint passes
- [x] Self-validated with `python -m scripts.pr_validate 1732`
- [x] Critical-path negative control reproduced locally
- [x] No dependency or supply-chain files changed
- [x] No breaking API changes

## Validation note

A full four-family stress run was attempted on the 256GB Mac Studio. Qwen3.5 and Qwen3.6 advanced before the SSH connection was reset while entering DiffusionGemma; the host then became unreachable. This is recorded as an incomplete infrastructure attempt, not as a passing gate. The issue-specific live Qwen3.5 reproduction and GitHub's five live model smokes are green.

### DIFF
--- a/tests/test_prefix_boundary_path_parity.py
+++ b/tests/test_prefix_boundary_path_parity.py
@@ -266,6 +266,34 @@ def test_prefix_boundary_prefers_latest_stable_message_boundary(monkeypatch):
     assert engine._compute_prefix_boundary(messages) == len(stable) - 8
 
 
+def test_first_turn_saves_stable_boundary_before_generation_suffix(monkeypatch):
+    """Turn one must seed the cache boundary that turn two actually shares.
+
+    Returning zero merely because the first user is at index zero falls back
+    to the scheduler's generic N-1 snapshot. For templates whose generation
+    suffix is several tokens long, that entry extends past the next turn's
+    shared prefix and a non-trimmable cache cannot reuse it (#1611).
+    """
+    engine, _ = _build_engine(monkeypatch)
+    engine._compute_prefix_boundary = BatchedEngine._compute_prefix_boundary.__get__(
+        engine, BatchedEngine
+    )
+    engine._tokenizer = _CharacterTokenizer()
+
+    def render(messages, tools=None, *, add_generation_prompt=True, **kwargs):
+        body = "|".join(str(message.get("content", "")) for message in messages)
+        return body + ("|ASSISTANT_GENERATION" if add_generation_prompt else "")
+
+    monkeypatch.setattr(engine, "_apply_chat_template", render)
+    messages = [{"role": "user", "content": "first turn"}]
+    stable = render(messages, add_generation_prompt=False)
+
+    boundary = engine._compute_prefix_boundary(messages)
+
+    assert boundary == len(stable) - 8
+    assert 0 < boundary < len(render(messages, add_generation_prompt=True)) - 1
+
+
 def test_prefix_boundary_stops_before_transient_server_priming(monkeypatch):
     """A server-only reminder is absent from the client's next request."""
     engine, _ = _build_engine(monkeypatch)
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -2261,7 +2261,7 @@ class BatchedEngine(BaseEngine):
             if messages[i].get("role") == "user":
                 last_user_idx = i
                 break
-        if last_user_idx is None or last_user_idx == 0:
+        if last_user_idx is None:
             return 0
         try:
             template_tools = convert_tools_for_template(tools) if tools else None


raullenchai:
Validation update:

- Codex review converged twice: 0 blocking findings in both rounds.
- `pr_validate 1732`: **MERGE-SAFE** with 35 targeted + 16,600 full-unit tests passing; supply-chain and lint clean. Stress was explicitly skipped on the 18GB host because the registry minimum is 26GB including headroom.
- Negative control: restoring the old `last_user_idx == 0` guard makes the new regression fail (`boundary 0 != 2`); restoring the fix passes.
- Live Qwen3.5-4B: turn two improved from 5.210s MISS to 0.458s HIT (`cached_tokens=3008`).
- GitHub CI and all five live L1 smoke checks are green.
- A 256GB Studio four-family attempt advanced through Qwen3.5/Qwen3.6 before SSH reset during DiffusionGemma; this is documented as incomplete, not counted as a pass.

Issue #1638, opened 2026-08-07
rapid-mac: swift test cannot build RapidUXTests because XCTest is unavailable


raullenchai:
## Reproduction

From current `main` (`1bc6244b`) on macOS 15.6.1 / Swift 6.2.1:

```sh
cd apps/rapid-mac
swift test --filter RAMBucketedDefaultTests
```

## Actual

The package builds the app and RapidTests target, then fails before executing tests:

```text
Emitting module RapidUXTests
Tests/RapidUXTests/ModelReadinessTests.swift:1:8: error: no such module 'XCTest'
error: fatalError
```

`swift build --target RapidTests` succeeds, so the failure is isolated to the RapidUXTests test-target configuration/toolchain interaction.

## Expected

`swift test --filter ...` should compile the test graph and execute the selected tests.

## Impact

Local focused test execution and PR validation are blocked even when the selected suite does not belong to RapidUXTests.

PR #1735, opened 2026-08-09
fix(mac): cap desktop prefix-cache memory

### LINKED ISSUE #1412: rapid-mac: cap the sidecar prefix cache for desktop memory pressure (set RAPID_MLX_PREFIX_CACHE_MAX_BYTES)
## Context
Found while stress-testing the bundled sidecar for the **rapid-mac** desktop app (0.12.0 prep): 8-way concurrent chat, ~10k *distinct* prompts, `llama3-3b-4bit`, ~20 min.

## Observation
`serve.log` showed the prefix cache climbing to `cache_entries=1000 cache_mem=8718MB` (~8.7 GB) and still growing; no eviction fired. Process RSS stayed ~2.5 GB — MLX Metal/unified-memory buffers aren't counted in `ps rss`, so the true unified-memory footprint was ~model + ~8.7 GB cache.

(For the record: this run had **zero memory leak** in the classic sense — RSS was flat; the growth is the prefix cache filling toward its ceiling, which is expected. This issue is about where that ceiling sits for a *desktop* app.)

## Why it's ~8.7 GB
The engine's prefix cache defaults to **20% of available RAM** (`MemoryCacheConfig.max_memory_percent = 0.20`, `vllm_mlx/memory_cache.py: compute_memory_limit`). On this M3 Ultra (256 GB) that ceiling is ~51 GB, so 8.7 GB was nowhere near eviction — correct for a **server**. The engine already exposes `RAPID_MLX_PREFIX_CACHE_MAX_BYTES` to bound it (added after the 0.8.7 dogfood where the cache ballooned to ~31 GB on a large host).

**The gap:** `ServerManager.serveEnvironmentAdditions` (apps/rapid-mac) spawns the sidecar with `RAPID_MLX_API_KEY` / `RAPID_MLX_WATCHDOG_PPID` / `PYTHONUNBUFFERED` / `HF_*` — but does **not** set `RAPID_MLX_PREFIX_CACHE_MAX_BYTES`. So a desktop app that shares the Mac with the user's foreground apps lets local inference claim up to 20% of RAM for prefix cache *on top of* the model. On a 16 GB Mac that's ~3.2 GB of cache + the model + everything else the user is running → memory pressure / swap for the app the user is actually looking at.

## Proposal (app-side)
Have `ServerManager` set `RAPID_MLX_PREFIX_CACHE_MAX_BYTES` to a **desktop-conservative** ceiling scaled to device RAM (and ideally aware of the loaded model size), e.g.:
- a smaller fraction than the 20% server default (say 8–10% of physical RAM), and/or
- `min(fraction × RAM, absolute_ceiling)` — a couple GB on small Macs, larger on big ones.

Goal: local inference should never be the reason the user's foreground apps start swapping. No engine change required — this just sets the existing env var at spawn time.

## Repro
1. Launch the bundled sidecar (or the app) serving any model.
2. Send many *distinct* prompts under concurrency.
3. Watch `serve.log` `[cache_store] … cache_mem=` climb toward 20% of RAM before eviction fires.

## Env
M3 Ultra 256 GB, bundled engine 0.11.9 (+ merged PRs), rapid-mac 0.12.0 prep.

raullenchai:
## What does this PR do?

Injects a hardware-aware `RAPID_MLX_PREFIX_CACHE_MAX_BYTES` ceiling into every Rapid Desktop sidecar spawn. The ceiling is the minimum of 8% physical RAM, 20% currently available RAM, and 4 GiB.

## Why is this needed?

Fixes #1412. The engine default is server-oriented and permits prefix-cache growth up to 20% of available memory. In the desktop app that cache competes with the foreground browser, IDE, and the model itself in unified memory; dogfood measured 8.7 GiB and growing.

## AI assistance disclosure

Codex reproduced the missing spawn variable, implemented the Swift production wiring and four regression tests, and ran two adversarial review rounds. The first review found that a physical-RAM-only limit could exceed the engine default under pressure; the implementation was corrected to clamp against current available RAM. The maintainer can explain every change.

> By submitting this PR I confirm I can explain the intent, risk, and behavior of every non-generated change in this PR. For generated sections, I have identified them above and can describe how I verified them.

## Test plan

- [x] `swift build`
- [x] `swift build --target RapidTests` (new tests and full RapidTests module compile)
- [x] 16 GiB maps to 8% physical RAM
- [x] 256 GiB clamps to 4 GiB
- [x] 32 GiB host with 8 GiB available clamps to 20% available
- [x] failed physical-memory probe preserves engine fallback
- [x] Codex review round 2: zero findings

## Checklist

- [x] Tests compile locally
- [x] Full app builds locally
- [x] No dependency changes
- [x] No breaking API changes

Local `swift test` is blocked by existing #1638 because SwiftPM builds the unrelated RapidUXTests/XCTest target first; GitHub rapid-mac CI has the full Xcode test environment.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
@@ -2520,22 +2520,18 @@ final class ServerManager {
     /// Desktop-specific prefix-cache budget: the smallest of 8% physical RAM,
     /// 20% currently available RAM, and 4 GiB. The available-RAM clamp keeps
     /// this override from raising the engine's ordinary 20%-of-available
-    /// budget on a machine that is already under pressure. A missing
-    /// available-memory probe retains the physical-RAM policy; a zero physical
-    /// probe omits the override instead of manufacturing an invalid zero-byte
-    /// cache. Kept pure so the spawn contract is deterministic in tests.
+    /// budget on a machine that is already under pressure. If either probe
+    /// fails, omit the override and retain the engine's live available-memory
+    /// fallback instead of manufacturing an unsafe fixed ceiling. Kept pure so
+    /// the spawn contract is deterministic in tests.
     nonisolated internal static func desktopPrefixCacheMaxBytes(
         physicalRAMBytes: UInt64,
         availableRAMBytes: UInt64 = 0
     ) -> UInt64? {
-        guard physicalRAMBytes > 0 else { return nil }
+        guard physicalRAMBytes > 0, availableRAMBytes > 0 else { return nil }
         let fourGiB = UInt64(4) << 30
         let eightPercent = (physicalRAMBytes / 100) * 8
-        var ceiling = min(eightPercent, fourGiB)
-        if availableRAMBytes > 0 {
-            ceiling = min(ceiling, availableRAMBytes / 5)
-        }
-        return ceiling
+        return min(min(eightPercent, availableRAMBytes / 5), fourGiB)
     }
 }
 
--- a/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
@@ -1196,6 +1196,12 @@ final class ServerManager {
                 environmentAdditions: Self.serveEnvironmentAdditions(
                     bearer: bearer,
                     ambient: ProcessInfo.processInfo.environment,
+                    // Issue #1412: the engine's server-oriented default may
+                    // retain up to 20% of RAM in prefix-cache entries. The
+                    // desktop shares unified memory with foreground apps, so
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex round-2 blocking dismissal with evidence: the finding assumes an ambient `RAPID_MLX_PREFIX_CACHE_MAX_BYTES` survives when a probe fails. It cannot: `ServerManager.serveEnvironmentAllowlist` is an exact Set and does not contain this key, and the spawn uses `replaceEnvironment: true`. I strengthened both physical-probe and available-probe failure tests to pass `RAPID_MLX_PREFIX_CACHE_MAX_BYTES=999999999999` in `ambient`; both assert the completed child env omits it. `swift build --target RapidTests` passes. No redundant `removeValue` is needed because Layer 1 never admits the key.

PR #1740, opened 2026-08-09
Preserve native tool history across parser families

### LINKED ISSUE #1593: Six tool parsers inherit SUPPORTS_NATIVE_TOOL_FORMAT=False by omission, including Qwen — the engine rewrites their tool history into an invented dialect
## What

`SUPPORTS_NATIVE_TOOL_FORMAT` defaults to `False` on the abstract base (`vllm_mlx/tool_parsers/abstract_tool_parser.py:170`), and **six parsers never override it** — including `qwen_tool_parser.py`, the flagship family. For those models the engine rewrites the tool history into a text dialect the model was never trained on, instead of letting the model's own chat template render it.

Found while root-causing #1591 / the LFM markup leak (fixed in #1592): the same rewrite is what taught LFM2.5 to emit `[Calling tool: …]` back at us.

## The chain

```
server.py::_detect_native_tool_support()   -> parser_cls.supports_native_format()
  -> engine.preserve_native_tool_format
    -> api/utils.py: assistant tool_calls  -> "[Calling tool: name({args})]"
                     role="tool" messages  -> role="user", "[Tool Result (id)]: …"
```

The flag's own docstring says what it is supposed to mean:

> Set to True in subclasses whose corresponding model chat templates can handle `role="tool"` messages and `tool_calls` fields directly, without needing conversion to text format.

## Qwen3.5 satisfies that definition and is still flagged False

Rendering the same 4-message tool conversation through `mlx-community/Qwen3.5-4B-MLX-4bit`'s own template gives:

```
<|im_start|>assistant
<tool_call>
<function=weather>
<parameter=location>
Tokyo
</parameter>
</function>
</tool_call><|im_end|>
<|im_start|>user
<tool_response>
{"temp_c":29.2}
</tool_response><|im_end|>
```

The template handles both `tool_calls` and `role="tool"` natively. But because `qwen_tool_parser.py` never sets the flag, the engine replaces that with:

```
<|im_start|>assistant
[Calling tool: weather({"location": "Tokyo"})]<|im_end|>
<|im_start|>user
[Tool Result (c1)]: {"temp_c":29.2}<|im_end|>
```

So on every multi-turn tool conversation, Qwen reads its own previous turns in a format it has never seen in training, and the `<tool_response>` channel it was trained to read is replaced by a bracketed string in a user turn.

## Affected parsers

Explicitly `False` (may well be correct — LFM and UI-TARS templates genuinely ignore `tool_calls`):

- `lfm_tool_parser.py:208`
- `ui_tars_tool_parser.py:1314`

**Inheriting the `False` default with no deliberate decision recorded** — each needs its template checked the way Qwen's was above:

- `qwen_tool_parser.py` — verified above: template IS native, flag is wrong
- `gemma4_tool_parser.py`
- `minimax_tool_parser.py`
- `nemotron_tool_parser.py`
- `xlam_tool_parser.py`
- `auto_tool_parser.py` (dispatcher — needs its own reasoning, since the answer depends on the delegate)

Sixteen parsers do set it `True`, so the default is doing the wrong thing for the minority that forgot rather than expressing a considered position.

## Impact

Not a crash, and **not a regression** — this has been the behaviour for a long time. Measured effect on Qwen: none visible in the obvious failure mode. I replayed a two-turn tool conversation (prior `weather` call plus its result in the history) six times on `qwen3.5-4b-4bit` and got **0/6 leaks** — Qwen keeps emitting its native `<tool_call>` markers regardless of the dialect in the transcript, and the parser strips them cleanly.

What it costs is subtler: multi-turn tool accuracy on a transcript the model cannot pattern-match against its own training, for the families listed above. LFM2.5-1.2B is the case where a model was weak enough to imitate the injected dialect instead of ignoring it, and that produced visible garbage in the chat (#1592).

## Suggested direction

1. Make the default explicit rather than inherited. Every concrete parser should have to state its answer, the way `EXPECTED_WIRE_FORMATS` is already forced by `tests/test_tool_parser_wire_formats.py::test_every_parser_declares_formats`. A structural test that fails when a parser omits `SUPPORTS_NATIVE_TOOL_FORMAT` would have caught all six.
2. Set it correctly per family, checking each template rather than assuming. Qwen is demonstrably wrong today.
3. Consider whether the fallback dialect itself should be closer to something models actually see. `[Calling tool: name({args})]` matches no mainstream template; if a text fallback is unavoidable for the genuinely-non-native parsers, it may as well not invent a fifth format.

raullenchai:
## Summary
- declare native tool-history support explicitly on every previously implicit parser
- preserve structured tool calls/results for Qwen, Gemma 4, MiniMax, Nemotron, and xLAM; keep auto conservative
- add a structural regression gate so registered parsers cannot inherit the base default silently

## Why
Before this change, `QwenToolParser.supports_native_format()` and four other native-template families returned `False` solely through inheritance. The engine therefore rewrote assistant calls and tool results into an invented `[Calling tool: ...]` / `[Tool Result ...]` transcript instead of letting each model chat template render its trained format.

## Test plan
- [x] Run native-format, parser, upstream-regression, Gemma grammar, and Qwen grammar suites (384 passed, 39 skipped)
- [x] Run Ruff lint and formatting checks on changed Python files
- [x] Run `git diff --check`

Closes #1593

### DIFF
--- a/docs/release-notes/unreleased.md
+++ b/docs/release-notes/unreleased.md
@@ -62,6 +62,9 @@ to always-allow rather than prompting every time (#1695).
   a deterministic forced call, streaming channel hygiene, and stream/non-stream
   tool-result replay without treating small-model knowledge errors as engine
   release failures (#1676, #1677).
+- Multi-turn tool history now stays in each model family's trained wire format
+  for Qwen, Gemma 4, MiniMax, Nemotron and xLAM instead of being rewritten into
+  the generic `[Calling tool: ...]` transcript (#1593).
 
 ## Release engineering
 
--- a/tests/test_native_tool_format.py
+++ b/tests/test_native_tool_format.py
@@ -15,6 +15,7 @@ from vllm_mlx.tool_parsers import (
     AutoToolParser,
     DeepSeekToolParser,
     FunctionaryToolParser,
+    Gemma4ToolParser,
     Glm47ToolParser,
     GraniteToolParser,
     HarmonyToolParser,
@@ -22,6 +23,7 @@ from vllm_mlx.tool_parsers import (
     KimiToolParser,
     LfmToolParser,
     LlamaToolParser,
+    MiniMaxToolParser,
     MistralToolParser,
     NemotronToolParser,
     QwenToolParser,
@@ -45,6 +47,11 @@ class TestNativeToolFormatCapability:
             HermesToolParser,
             HarmonyToolParser,
             Glm47ToolParser,
+            QwenToolParser,
+            Gemma4ToolParser,
+            MiniMaxToolParser,
+            NemotronToolParser,
+            xLAMToolParser,
         ]
         for parser_cls in native_parsers:
             assert parser_cls.SUPPORTS_NATIVE_TOOL_FORMAT is True, (
@@ -57,9 +64,6 @@ class TestNativeToolFormatCapability:
     def test_parsers_without_native_support(self):
         """Parsers that don't support native tool format should return False."""
         non_native_parsers = [
-            QwenToolParser,
-            NemotronToolParser,
-            xLAMToolParser,
             AutoToolParser,
             LfmToolParser,
         ]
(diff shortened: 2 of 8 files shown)


raullenchai:
## Local `pr_validate`

**Verdict: MERGE-SAFE**

- Codex review: PASS — 0 blocking findings
- supply chain: PASS
- lint: PASS
- targeted tests: 120 passed
- full unit: 16,673 passed, 123 skipped, 18 deselected, 6 xfailed, 1 xpassed
- diff coverage: 100% (6/6 changed production lines)

Artifacts: `/tmp/pr_validate/pr-1740/run-20260809T214406Z-39837-1786311846442604000`

PR #1742, opened 2026-08-09
Accept mid-conversation system messages in chat

### LINKED ISSUE #1543: /v1/chat/completions returns 400 for a mid-conversation system message, while /v1/responses and /v1/messages accept it
## What

```bash
curl -s localhost:8000/v1/chat/completions -d '{
  "model": "qwen3.6-27b-8bit",
  "messages": [
    {"role": "system",    "content": "You are terse."},
    {"role": "user",      "content": "Hi"},
    {"role": "assistant", "content": "Hello."},
    {"role": "system",    "content": "From now on answer only BLUE."},
    {"role": "user",      "content": "What colour is the sky?"}
  ]}'
```

```json
{"error":{"message":"Chat template error: System message must be at the beginning.","type":"invalid_request_error"}}
```

The message comes from the model's own template (`chat_template.jinja:85` in `unsloth/Qwen3.6-27B-MLX-8bit`, and the same guard is in the Qwen / Llama / Gemma families).

## Why it is a defect rather than the template's business

The other two surfaces already coerce this, for exactly this reason:

- `vllm_mlx/api/responses_adapter.py:575` — `_collapse_system_messages`, added because Codex sends `developer` items interleaved with user turns.
- `vllm_mlx/api/anthropic_adapter.py:208` — mirrors it, because Claude Code sends mid-conversation `role="system"` reminders.

`/v1/chat/completions` has no equivalent, and it is the surface most clients use. LangChain, LlamaIndex and most agent frameworks emit a mid-conversation system message to steer a running conversation; OpenAI accepts them at any position. On our flagship model that is a hard 400.

Pre-existing — not a regression in the v0.12.4..main window. Confirmed by dogfooding a clean wheel build of `main`.

## Where a fix goes

Not in `service/helpers.py:4430`. That render is `enforce_context_length_for_messages`, a DoS gate; fixing only that would make the gate pass and the real render still fail. The prompt the engine actually uses is built later (`routes/chat.py:5038`, and via `engine.build_prompt`), so a single fix belongs at the shared build_prompt wrapper where all three routes converge.

## Suggested shape

Retry rather than always-coerce, so nothing changes for templates that accept a non-leading system message today:

1. Render as sent.
2. If it raises and the message matches the template's non-leading-system guard, collapse system messages into one at index 0 (the `_collapse_system_messages` behaviour with `relocate_mid_conversation=False`) and render once more.
3. If that also fails, surface the original error.

Hoisting rather than folding into the following user turn is deliberate: the Anthropic lane's own comment documents why folding has an injection shape, and keeps it default-off there. This lane should not adopt the riskier form.

Deferred out of the pre-release fix batch: it is pre-existing, and the coercion semantics have already drawn three rounds of objection on the Anthropic lane, so it wants its own review rather than a rushed change to the hottest endpoint.

Found during the pre-release audit of the v0.12.4..main delta.

raullenchai:
## Summary
- render chat messages in their original order first
- retry only the known Qwen/Llama/Gemma non-leading-system template failure with one combined leading system message
- preserve the original diagnostic if the compatibility retry also fails

## Why
`/v1/chat/completions` returned HTTP 400 for mid-conversation system messages even though the Responses and Anthropic surfaces already support them. Common agent frameworks use these messages to steer a running conversation, and flagship templates reject them unless they are hoisted.

## Test plan
- [x] Reproduce the pre-fix template failure with the issue message sequence
- [x] Render the fixed sequence through the real cached Qwen3.5 tokenizer
- [x] Run 160 adjacent chat-template, tool, reasoning-budget, and DeepSeek tests
- [x] Run Ruff lint/format and `git diff --check`

Closes #1543

### DIFF
--- a/docs/release-notes/unreleased.md
+++ b/docs/release-notes/unreleased.md
@@ -65,6 +65,10 @@ to always-allow rather than prompting every time (#1695).
 - Multi-turn tool history now stays in each model family's trained wire format
   for Qwen, Gemma 4, MiniMax, Nemotron and xLAM instead of being rewritten into
   the generic `[Calling tool: ...]` transcript (#1593).
+- `/v1/chat/completions` now retries templates that reject a mid-conversation
+  system message by combining system instructions at the front, matching the
+  compatibility already provided by the Responses and Anthropic surfaces
+  (#1543). Templates that accept the original ordering remain untouched.
 
 ## Release engineering
 
--- a/tests/test_chat_template_mid_system.py
+++ b/tests/test_chat_template_mid_system.py
@@ -57,6 +57,41 @@ def test_does_not_coerce_template_that_accepts_original_order():
     assert rendered == MESSAGES
 
 
+def test_retry_preserves_structured_system_content_parts():
+    template = _RecordingTemplate(reject_mid_system=True)
+    structured = [
+        {
+            "role": "system",
+            "content": [
+                {"type": "text", "text": "Inspect this."},
+                {"type": "image", "image": "first.png"},
+            ],
+        },
+        {"role": "user", "content": "Hi"},
+        {
+            "role": "system",
+            "content": [
+                {"type": "text", "text": "Compare it."},
+                {"type": "image", "image": "second.png"},
+            ],
+        },
+        {"role": "user", "content": "Continue"},
+    ]
+
+    rendered = apply_chat_template(template, structured)
+
+    assert rendered[0] == {
+        "role": "system",
+        "content": [
+            {"type": "text", "text": "Inspect this."},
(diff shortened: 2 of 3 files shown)


raullenchai:
## Local `pr_validate`

**Verdict: MERGE-SAFE**

- Codex review: PASS — 0 blocking findings after two review-driven fixes
- targeted tests: 165 passed
- full unit: 16,680 passed, 123 skipped, 18 deselected, 6 xfailed, 1 xpassed
- diff coverage: 92.3% (36/39 changed production lines)
- supply-chain and lint: PASS

Artifacts: `/tmp/pr_validate/pr-1742/run-20260809T220242Z-48178-1786312962176864000`

PR #1745, opened 2026-08-09
Fix Chat readiness model switching

### LINKED ISSUE #1739: Chat's readiness button silently does nothing while the Images model is loaded
## What

After using the Images tab, going back to Chat and pressing the readiness
banner's action does nothing at all — no spinner, no error, no state change.
The button is a dead control in that state.

## Why

The two tabs call different things for the same user intent:

| tab | readiness action calls | switches a resident model? |
| --- | --- | --- |
| Images | `server.ensureServing(alias:hfPath:)` (`ImagesView.swift:157`) | yes |
| Chat | `server.start(alias:hfPath:)` (`ContentView.swift:418`) | **no** |

`ServerManager.start` bails early whenever any child is resident
(`ServerManager.swift:832`):

```swift
guard !isOperating else { return }
guard child == nil else { return }
```

So the asymmetry is one-way: Images can take the server away from Chat, and
Chat's button cannot take it back. `ImagesView`'s own comment anticipated
exactly this hazard for its own direction —

> ``ensureServing`` (not ``start``): the user is almost always switching FROM a
> running chat model TO the image model. Plain ``start`` is cold-start only —
> it no-ops when a child is already serving — so it would silently do nothing
> here

— and the reverse direction was never given the same treatment.

## Not a total dead end, which is why it is easy to miss

Typing a message and sending it DOES recover: `ChatViewModel.swift:444` goes
through `ensureServing`. So the model comes back if you ignore the banner and
just talk. Only the explicit "load the model" affordance is dead. A user who
trusts the button concludes the app is broken; a user who ignores it never
notices.

## Distinct from #1713

#1713 is the larger design question of holding both models resident at once.
This is narrower and independently fixable: the control the UI presents for
"load this model" should work in every state it is shown. #1713 also asserts
"going back unloads the image model and reloads the chat model", which is true
of the send path but not of the readiness button — worth correcting there.

## Suggested fix

Point Chat's readiness action at `ensureServing`, matching Images. Both already
resolve a concrete target alias, and `ensureServing` short-circuits when the
requested alias is already being served, so cold-start behaviour is unchanged.

Worth a `no-dead-controls` golden-flow assertion afterwards, since that flow
exists precisely to catch controls that render enabled and do nothing.

Found while fact-checking release-note copy for 0.12.8 — the draft claimed you
"load it the same way you load a chat model", which turned out not to be true.

raullenchai:
## Summary
- route the Chat readiness action through `ServerManager.ensureServing`
- preserve cold-start and memory-guard behavior through `ensureServing`'s existing delegation to `start`
- add a wiring regression guard for the readiness action

## Why
After loading an Images model, Chat's readiness button called cold-start-only `start`, which silently returns while any child is resident. Sending a message recovered through `ensureServing`, but the explicit load button was dead.

## Test plan
- [x] Reproduce the early-return chain from ContentView to ServerManager.start
- [x] Build the Rapid application target successfully
- [x] Build the RapidTests target, including the new regression guard
- [x] Run an independent source wiring assertion and `git diff --check`

Note: executing `swift test` locally still hits existing #1638 (`RapidUXTests` imports unavailable XCTest); GitHub Apple Silicon CI owns the complete suite.

Closes #1739

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
@@ -393,10 +393,10 @@ struct ContentView: View {
     ///
     /// Lives here rather than in ``ChatView`` because starting a model is
     /// a window-level concern — the Launch page raises the same actions.
-    /// Start routes through ``ServerManager.start``, which is the one
-    /// choke point that applies the launch flags and the live
-    /// free-memory guard (#1435), so this path inherits the same OOM
-    /// protection the implicit send-start path already had.
+    /// Start routes through ``ServerManager.ensureServing`` so the action also
+    /// replaces a different resident model (for example after using Images).
+    /// ``ensureServing`` delegates its cold-start leg to ``start``, preserving
+    /// the launch flags and live free-memory guard (#1435).
     private func performReadinessAction(_ action: ModelReadiness.Action) {
         switch action {
         case .chooseModel:
@@ -417,7 +417,7 @@ struct ContentView: View {
 
     private func startModel(_ target: String) {
         let hfPath = catalogEntries.first(where: { $0.alias == target })?.hfRepo
-        Task { await server.start(alias: target, hfPath: hfPath) }
+        Task { _ = await server.ensureServing(alias: target, hfPath: hfPath) }
     }
 
     // MARK: - Detail routing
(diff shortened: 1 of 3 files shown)


raullenchai:
## Local `pr_validate`

**Verdict: MERGE-SAFE**

- Codex review: PASS — 0 blocking findings
- full unit: 16,680 passed, 123 skipped, 18 deselected, 6 xfailed, 1 xpassed
- supply-chain: PASS
- Rapid app target: built
- RapidTests target including the new wiring guard: built

Artifacts: `/tmp/pr_validate/pr-1745/run-20260809T221657Z-55754-1786313817581658000`

Issue #1624, opened 2026-08-07
rapid-mac syntax highlighter: streaming prefix-check is O(n²) cumulative; multi-part tokens (background-color, .PHONY) don't colour (#1620 follow-up)


raullenchai:
Two non-blocking residuals from the #1620 syntax-highlighter review, held under the codex non-convergence stopping rule after every worse-than-plain correctness case was fixed.

## 1. Streaming prefix-check is O(n²) cumulative (perf, not correctness)
`SyntaxHighlighter.Memo.highlight` runs `code.hasPrefix(lastCode)` on every streamed append to decide append-vs-replacement, so across a full token-by-token stream the prefix validation is cumulatively O(n²).

This is **not a correctness bug** and **not user-perceptible**: per-append cost is microseconds (one prefix compare + a last-line rescan), and the cumulative cost is spread across seconds of streamed model output — ~10ms total for a chat-sized block. It is documented in place as a deliberate correctness safeguard: the full-prefix compare is what distinguishes an append from an edit/retry **replacement** (see `incrementalHighlightResetsSafely`).

A genuine fix needs an explicit append/replacement signal threaded from the streaming layer (ChatView → Memo), since SwiftUI hands the highlighter the full string each render with no provenance. Reducing it to a bounded-suffix heuristic would trade replacement-detection certainty for a micro-optimization on a non-problem. Deferred until the streaming layer can pass provenance.

## 2. Multi-part configured tokens don't colour as one run
`isIdentifierChar` excludes `-` and `.`, so configured tokens that contain them — CSS `background-color`, `@font-face`; Makefile `.PHONY`, `filter-out` — never match their keyword/type set entry and fall back to plain (or partial: `background-color` colours only `color`).

This never renders **worse** than the old flat-grey (the contract the #1620 fixes restored) — the tokens just aren't as richly coloured as they could be. A proper fix makes identifier-continuation grammar-specific (or matches punctuation-bearing configured tokens before generic identifier scanning), which touches many grammars and carries regression risk, so it's deferred rather than rushed.

Source: `apps/rapid-mac/Sources/Rapid/UI/Markdown/SyntaxHighlighter.swift`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

raullenchai:
[...]

PR #1746, opened 2026-08-09
fix(mac): bound the chat repaint rate so a long answer stops melting the app

### LINKED ISSUE #1743: Chat window disappears and the app spins at 100% CPU with unbounded memory growth after a long-prompt turn
## What happened

Manual dogfood of `main` @ `df89ee63`, full `build.sh` bundle (sidecar included),
launched with `open`, default model `qwen3.5-4b-4bit`.

1. Sent a long prompt — one sentence repeated to ~4000 characters, then a
   question. The model answered correctly, and the caption read
   `37 tok/s · 9m 5s to first token · 9m 6s`.
2. Sent a short follow-up.
3. The window **vanished**. The process stayed alive, pinned at 100% CPU, with
   RSS climbing steadily and no sign of recovery.

Measured over 40 s while idle (no further input):

```
t1: rss= 9.10 GB  cpu=100.0%  stat=R
t2: rss= 9.32 GB  cpu=100.0%  stat=R
t3: rss= 9.54 GB  cpu= 99.0%  stat=R
final (26 min uptime): rss=10.17 GB  cpu=99.6%
```

~11 MB/s, monotonic, with the window already gone. `SIGTERM` was ignored; it
needed `SIGKILL`.

Accessibility confirms the window is genuinely gone rather than merely
unreadable: `windows.titles == []` and `walk.complete == false`, and System
Events cannot see the process at all (`Can't get process "Rapid-MLX Desktop"`).

## The engine is fine

The bundled sidecar stayed healthy throughout:

```
{"status":"healthy","ready":true,"model_loaded":true,
 "model_name":"mlx-community/Qwen3.5-4B-MLX-4bit", ...}
```

So this is the SwiftUI app, not inference. The sidecar exited cleanly with the
app once it was killed.

## Where it is spinning

`sample` over 3 s — main thread, 1379 of 1621 samples:

```
Rapid_main → SwiftUI App.main → NSRunLoop.flushObservers
  → NSHostingView.beginTransaction → ViewGraphRootValueUpdater.updateGraph
    → GraphHost.flushTransactions → GraphHost.runTransaction   ← 1379/1621
```

Below that the graph churns through `StyledTextResponder.update()`,
`TooltipModifier.TooltipResponderUpdater.updateValue`,
`AppKitPlatformViewHost.enqueueLayoutInvalidation` → `invalidateLayout` →
`asyncTransaction` → `beginTransaction`, i.e. layout invalidation re-entering
the transaction that scheduled it. No single Rapid frame dominates — the only
`Rapid` symbol on a hot thread is `ProcessGroupChild.startMonitor`'s
`waitpid`, which is a correctly-blocked child monitor, not the spin.

Full sample saved; happy to attach.

## Likely related

`StyledTextResponder` churn on a very long single-paragraph message points at
the same area as #1624 (streaming prefix-check is O(n²) cumulative). This may
be that issue's user-visible end state rather than a separate defect — worth
deciding before fixing either.

## Caveats, stated so nobody over-reads this

- The `9m 5s to first token` is almost certainly the launchd/XPC GPU throttle
  on a sidecar spawned by an `open`-launched app, not normal prefill: the same
  engine, driven directly from a shell, prefilled 9.9K tokens in 32 s on a 27B
  earlier today. So the turn was abnormally long, which may have widened the
  window for whatever loops here. It is an amplifier, not necessarily the
  cause — and note a real user launching from Finder is on that same path.
- Input was delivered via `rapid-ax set-value` rather than typed keystrokes.
- Pasting a long document into chat is a mainstream use of this app, so the
  triggering input is not exotic.

## Why it matters for the release

The window disappears with no error, and the app then consumes CPU and memory
without bound until force-killed. On a 16–32 GB Mac this reaches memory
pressure quickly. A user would experience it as "the app vanished and my fans
came on".

Found during the 0.12.8 release dogfood. The nine AX golden flows all pass on
this same build — they drive short fixture messages, so nothing in the scripted
set exercises a long prompt.

raullenchai:
## Why

Closes #1743, found in the 0.12.8 release dogfood and called a blocker.

A long streaming answer left the app **pinned at 100 % CPU with its window
gone** and RSS climbing ~11 MB/s past 10 GB until it had to be `SIGKILL`ed
(`SIGTERM` was ignored). The bundled engine stayed healthy the whole time
(`/health` reported ready, model loaded), so this is entirely the SwiftUI side.

## Root cause

Every coalescer flush rebuilds the chat view, and `LaTeXMarkdownView`
re-parses the **entire accumulated message** through MarkdownUI on each
rebuild. Parsing is O(length), so a fixed 16 ms cadence costs O(length²)
across a turn.

`sample` of the hung app put **1379 of 1621** main-thread samples in
`GraphHost.runTransaction`, with `Markdown.init(_:baseURL:imageBaseURL:)` and
`LaTeXMarkdownView.body` on the hot path, plus `AppKitPlatformViewHost`
layout invalidation re-entering the transaction that scheduled it.

## What changed

`SSEDeltaCoalescer`'s flush window now widens in proportion to the text
already flushed: **unchanged below 2 000 characters**, then linear, capped at
250 ms. That bounds the repaint RATE, which is the term that makes the cost
quadratic.

This does not make a single parse cheaper — that is #1624, and it is still the
real fix. This stops the app becoming unusable while that is done.

## Measured, same fake stream, before and after

| build | main-thread CPU over the turn | RSS |
| --- | --- | --- |
| `main` | 47 % → 55 % → 68 % → 88 % → 94 % → **100 % pegged** | unbounded (10 GB in the live case) |
| this PR | 9 % → 14 % → 13 % → 14 % → 32 % → 21 % → **~16 % steady** | 271 MB |

## How I checked the mechanism rather than assuming it

My first attempt at this fix changed nothing at all. Rather than conclude the
hypothesis was wrong, I forced the window to a flat 2 s as a probe — CPU fell
from 100 % to ~8 %, proving the coalescer really was the driver. That sent me
back to the diff, where the bug was mine: I counted `pending` **after**
`pending.removeAll()`, so the running total stayed 0 and the window never
widened. The numbers alone would have read as "wrong theory".

## Tests

`window_widens_with_accumulated_length` asserts **both** directions:

1. a short message still flushes on the 16 ms cadence — so ordinary replies
   are untouched;
2. a long one does not, because the window has widened.

Only asserting (2) would also pass for a change that simply slowed every reply
down, which would be a visible regression on every normal answer.

`swift test` cannot run on this machine (#1638), so CI is the authority for the
unit test.

## Verification

Six streaming-heavy golden flows pass on the fixed build: `chat-restore`,
`slow-stream-stop`, `chat-depth`, `tool-loop-budget`, `restored-tools`,
`model-crash-recovery`.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Chat/ChatStreamClient.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/ChatStreamClient.swift
@@ -1008,8 +1008,10 @@ enum Wire {
 ///     (usage, tool_calls, finish_reason, [DONE]) AND before
 ///     throwing on cancellation or transport error. The send()
 ///     scope guarantees the latter via a `defer` block.
-///   * The 16 ms window is opportunistic (checked on each
-///     append), not timed: if a delta burst stops mid-window,
+///   * The window is opportunistic (checked on each append), not
+///     timed, and its length is adaptive — 16 ms until the message
+///     passes ``adaptiveThresholdChars``, then proportional to the
+///     text so far up to ``maxWindowNs`` (#1743): if a delta burst stops mid-window,
 ///     the trailing text is held until the next event of any
 ///     kind or until send()'s defer flush. Real backends emit
 ///     a continuous stream of deltas terminated by
@@ -1036,8 +1038,8 @@ final class SSEDeltaCoalescer: @unchecked Sendable {
 
     /// Ceiling on the widened window. At 250 ms a very long message
     /// still repaints four times a second, which reads as "streaming"
-    /// rather than "frozen", while costing a twentieth of what a 16 ms
-    /// cadence costs.
+    /// rather than "frozen", at about a sixteenth of the callback rate a
+    /// flat 16 ms cadence costs (250 / 16).
     private static let maxWindowNs: UInt64 = 250_000_000
 
     /// Total characters handed to the UI so far this turn.
@@ -1067,9 +1069,19 @@ final class SSEDeltaCoalescer: @unchecked Sendable {
         guard flushedCharacters > Self.adaptiveThresholdChars else {
             return Self.coalesceWindowNs
         }
-        let scale = UInt64(flushedCharacters / Self.adaptiveThresholdChars)
-        let widened = Self.coalesceWindowNs &* max(1, scale)
-        return min(widened, Self.maxWindowNs)
+        // Multiply BEFORE dividing so the growth is actually linear. Dividing
+        // first floors the ratio to an integer, which makes the window a
+        // staircase — flat from 2 000 to 3 999 characters and then jumping
+        // straight from 16 ms to 32 ms — rather than the smooth ramp the
+        // comment above promises.
+        //
+        // `characters` is clamped to the value that already yields the cap, so
+        // the multiplication cannot overflow no matter how long the message
+        // gets, and no masking operators are needed.
+        let capCharacters = Int(Self.maxWindowNs / Self.coalesceWindowNs) * Self.adaptiveThresholdChars
+        let characters = UInt64(min(flushedCharacters, capCharacters))
+        let widened = Self.coalesceWindowNs * characters / UInt64(Self.adaptiveThresholdChars)
+        return min(max(widened, Self.coalesceWindowNs), Self.maxWindowNs)
     }
 
     /// Codex r2 BLOCKING (unbounded queue): force-flush when the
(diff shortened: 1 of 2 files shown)


raullenchai:
Re-measured after the single-MainActor-hop batching change (the earlier numbers were taken before it), on a rebuilt app — binary mtime checked against the source, because `build.sh` had exited without rebuilding once already.

Fake sidecar, 400-chunk answer, 20 ms between chunks:

| | CPU | RSS |
| --- | --- | --- |
| before the fix | pegged at 100% | >10 GB, unbounded |
| after, pre-batching | ~16% steady | 271 MB |
| after, with batching | 38.9% falling to ~15% | 145 MB at t=140s |

(`ps %cpu` is an average since launch, so the early figure is the startup burst washing out, not a regression.)

Batching did not cost anything and RSS came down further. The first attempt aborted on the harness's own vacuity guard — a fresh persona opens the telemetry consent sheet, and that modal swallows the `Readiness.Action` press, so the model never started. The harness now dismisses it before measuring.

PR #1747, opened 2026-08-09
Fix desktop updater fallback publishing

### LINKED ISSUE #1612: dl.rapidmlx.com/latest.json is pinned at 0.11.0 — during a Worker outage every client is told it is up to date
## What

`https://dl.rapidmlx.com/latest.json` still advertises **0.11.0** (published 2026-07-28). Five app releases have shipped since: 0.12.0, 0.12.1, 0.12.5, 0.12.6.

```
$ curl -s https://dl.rapidmlx.com/latest.json | jq '{version, tag_name, published_at, dmg_url}'
{
  "version": "0.11.0",
  "tag_name": "v0.11.0",
  "published_at": "2026-07-28T18:48:41Z",
  "dmg_url": "https://dl.rapidmlx.com/rapid-mlx-desktop-bootstrapper-0.11.0.dmg"
}
```

## Why it usually does not bite

It is only the **fallback**. `UpdateChecker` polls the Worker first
(`UpdateChecker.swift:445`, `https://rapidmlx.com/api/desktop-update?v=<version>`)
and that path is healthy — it currently answers 0.12.6 with a `dmg_size` that is
byte-exact to the published artifact:

```
$ curl -s 'https://rapidmlx.com/api/desktop-update?v=0.12.5' | jq '{version, dmg_size}'
{ "version": "0.12.6", "dmg_size": 150864847 }
```

The site's download button is fine too — `/api/desktop-download` 302s straight to
the `rapid-mac-v0.12.6` release asset.

## Why it still matters

`latest.json` exists precisely so that *"a Worker outage can never stop every
client from seeing updates"* (`UpdateChecker.swift:448-451`). As written, the
outage behaviour is inverted: when the Worker is down, every installed app —
including brand-new 0.12.6 installs — falls back to a manifest claiming 0.11.0,
`isNewer` returns false, and the app concludes it is current.

So the fallback that exists to keep updates flowing during an outage is the one
thing guaranteeing they stop.

Combined with #1605 (Check for updates gives no feedback when up to date), a user
in that window clicks the button and gets total silence, indefinitely.

## Why it has drifted

`rapid-mac-release.yml`'s `mirror-dist` job reports success while doing nothing:

```
##[notice]CLOUDFLARE_* secrets or RAPID_MAC_DIST_R2_BUCKET not configured
— skipping the optional CDN mirror. The canonical GitHub Release DMG is unaffected.
```

A green job that silently no-ops is why nobody noticed across four releases.

## Suggested fix

Either:

1. **Publish `latest.json` from the release workflow** (same source of truth as the
   Worker), and make `mirror-dist` *fail* rather than skip when it is supposed to
   run but has no credentials — or drop the job if R2 is genuinely optional; or
2. **Retire the R2 fallback** and let the Worker be the single source, accepting
   that an outage means no update check rather than a wrong one — a check that
   fails is strictly better than one that lies.

Note the manifest also describes a distribution shape the current release no longer
produces: a bootstrapper DMG plus a separate sidecar tarball. 0.12.6 ships a single
DMG with the sidecar bundled, so any regenerated `latest.json` needs its
`dmg_url`/`sidecar_url` fields reconciled with what is actually published.

Found while verifying the 0.12.6 release end to end.

raullenchai:
Closes #1612

## Root cause

The tagged desktop release workflow treated missing Cloudflare configuration as a successful optional no-op and never generated `latest.json`. The live R2 fallback therefore remained on 0.11.0 while bundled-DMG releases advanced to 0.12.7.

## Fix

- fail tagged releases when required R2/CDN configuration is missing
- upload the immutable bundled DMG before publishing `latest.json` as the final commit point
- generate the current single-DMG manifest shape from the GitHub Release metadata
- document the now-required release configuration
- add workflow source-contract regression coverage
- repair the live R2 fallback to 0.12.7 (verified through both public endpoints)

## Validation

- live `dl.rapidmlx.com/latest.json` and Worker endpoint both return 0.12.7
- release workflow contract check passed
- workflow YAML parse passed
- `git diff --check` passed
- Swift test invocation is locally blocked by the host Xcode license; PR CI will run the macOS suite

### DIFF
--- a/.github/workflows/rapid-mac-release.yml
+++ b/.github/workflows/rapid-mac-release.yml
@@ -412,9 +412,9 @@ jobs:
           ALIAS_KEY="rapid-mac/rapid-mlx-desktop.dmg"
           DMG_URL="${CDN_BASE%/}/${VERSIONED_KEY}"
 
-          npx --yes wrangler@4 r2 object put "${R2_BUCKET}/${VERSIONED_KEY}" \
+          npx --yes wrangler@4.120.0 r2 object put "${R2_BUCKET}/${VERSIONED_KEY}" \
             --file "$DMG" --content-type application/x-apple-diskimage --remote
-          npx --yes wrangler@4 r2 object put "${R2_BUCKET}/${ALIAS_KEY}" \
+          npx --yes wrangler@4.120.0 r2 object put "${R2_BUCKET}/${ALIAS_KEY}" \
             --file "$DMG" --content-type application/x-apple-diskimage --remote
 
           RELEASE_JSON="$(gh api "repos/${GITHUB_REPOSITORY}/releases/tags/${TAG}")"
@@ -437,7 +437,7 @@ jobs:
 
           # Publish the mutable pointer last: this is the update transaction's
           # commit point. The current bundled DMG needs no sidecar fields.
-          npx --yes wrangler@4 r2 object put "${R2_BUCKET}/latest.json" \
+          npx --yes wrangler@4.120.0 r2 object put "${R2_BUCKET}/latest.json" \
             --file dist/latest.json --content-type application/json \
             --cache-control "no-cache, must-revalidate" --remote
           echo "::notice::Published ${DMG_URL} and ${CDN_BASE%/}/latest.json"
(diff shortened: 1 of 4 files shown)


raullenchai:
Local maintainer validation on `dc7556de`: **MERGE-SAFE**. Codex review: 0 blocking (1 accepted nit; Wrangler is exact-pinned and absence matching fails closed). Full unit: **16,689 passed**, 123 skipped, 18 deselected, 6 xfailed, 1 xpassed. Supply-chain review: workflow secret access is intentional and limited to R2 publication plus single-file CDN purge. Live Worker and R2 fallback both verified at 0.12.7.

PR #1749, opened 2026-08-09
Fix the readiness source gate that made rapid-mac build red on main

### LINKED ISSUE #1748: rapid-mac build is red on main: readiness source gate slices startModel at the wrong brace
`ContentViewReadinessActionTests` is a source-scraping gate added in #1745. It
locates `ContentView.startModel` in the stripped source and then slices the
function body with `firstIndex(of: "}")` — the **first** closing brace.

`startModel` opens a closure before it reaches the call under test:

```swift
private func startModel(_ target: String) {
    let hfPath = catalogEntries.first(where: { $0.alias == target })?.hfRepo
    Task { _ = await server.ensureServing(alias: target, hfPath: hfPath) }
}
```

so the first `}` is the closure's. The slice ends at
`privatefuncstartModel(_target:String){lethfPath=catalogEntries.first(where:{$0.alias==target}`
and can never contain `server.ensureServing(...)` — for any implementation.

**Impact:** the `build` job (`swift test --no-parallel`) fails on `main`, so
every rapid-mac PR inherits a red `build` check and cannot be verified green.
#1745 was itself merged with this check already red
(https://github.com/raullenchai/Rapid-MLX/actions/runs/31338998032/job/93309538384),
and #1746 is currently blocked by it.

**Failure line from CI:**

```
✘ Test "Chat readiness replaces a different resident model" recorded an issue at
ContentViewReadinessActionTests.swift:35:9: Expectation failed:
(function → "privatefuncstartModel(_target:String){lethfPath=catalogEntries.first(where:{$0.alias==target}")
  .contains("server.ensureServing(alias:target,hfPath:hfPath)")
```

**Fix:** balance braces from the opening `{` instead of stopping at the first
`}`, so the whole function body is scanned.

raullenchai:
## Why

`main` is red. The `build` job (`swift test --no-parallel`) fails, so **every**
rapid-mac PR inherits a failing check and none can be verified green — #1746 is
blocked on it right now.

The cause is a source-scraping gate added in #1745. It finds
`ContentView.startModel` in the stripped source, then slices the body with
`firstIndex(of: "}")`. But `startModel` opens a closure before the call the test
is about:

```swift
private func startModel(_ target: String) {
    let hfPath = catalogEntries.first(where: { $0.alias == target })?.hfRepo
    Task { _ = await server.ensureServing(alias: target, hfPath: hfPath) }
}
```

The first `}` is the closure's, so the slice is

```
privatefuncstartModel(_target:String){lethfPath=catalogEntries.first(where:{$0.alias==target}
```

which cannot contain `server.ensureServing(...)` for **any** implementation. The
gate never protected anything; it was red the moment it was written, and #1745
was merged with the check already failing
([run](https://github.com/raullenchai/Rapid-MLX/actions/runs/31338998032/job/93309538384)).

## What

Balance braces from the opening `{` so the whole function body is scanned.
Test-only change; no production source is touched.

## Verification

`swift test` needs a full Xcode toolchain, which this machine does not have
(`xcrun --find swift` → CommandLineTools, so `import XCTest` fails locally). So
rather than assert the fix works, I compiled the two functions the gate actually
runs — `stripCommentsAndWhitespace` (extracted verbatim from
`CapabilityChipRenderGateSourceGuardTests.swift`) and the new `balancedBody` —
into a standalone binary and ran the suite's exact assertion logic against the
real file and two mutants:

| input | `ensureServing` | `server.start` | verdict |
| --- | --- | --- | --- |
| real `ContentView.swift` | true | false | **PASS** |
| `ensureServing` → `server.start` | false | true | **FAIL** |
| the call deleted entirely | false | false | **FAIL** |

The gate now passes on correct code *and* still goes red on the regression it
was written to catch — the old version could do neither.

Closes #1748

### DIFF
--- a/apps/rapid-mac/Tests/RapidTests/ContentViewReadinessActionTests.swift
+++ b/apps/rapid-mac/Tests/RapidTests/ContentViewReadinessActionTests.swift
@@ -115,6 +115,12 @@ struct ContentViewReadinessActionTests {
                 }
                 continue
             }
+            // Extended regex literal (`#/…/#`). Its contents can hold `//`,
+            // braces, and call-shaped text, so it is skipped like a literal.
+            if c == "#", let end = Self.endOfExtendedRegex(in: swift, at: i) {
+                i = end
+                continue
+            }
             // String literal, including raw (`#"…"#`) and multiline (`"""`).
             if c == "\"" || c == "#" {
                 if let end = Self.endOfStringLiteral(in: swift, at: i) {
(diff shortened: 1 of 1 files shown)


raullenchai:
Closing — #1745 is being reverted rather than patched, so this fix has nothing left to fix. #1751 (shared stripper is string-unaware) stays open; it is about CapabilityChipRenderGateSourceGuardTests and is independent of #1745.

raullenchai:
Superseded by the requested recovery sequence: fully revert #1745 first (#1754), then reimplement #1739 with the literal-aware source-guard work from #1751 and execute the real Swift suite before merge. Closing this parallel fix avoids landing against the commit being reverted and removes conflicting macOS CI work.

PR #1753, opened 2026-08-09
Fix readiness actions for non-retriable failures

### LINKED ISSUE #1514: rapid-mac: readiness banner offers Retry for non-retriable model failures
Follow-up from #1510 (readiness UX unification).

`ModelReadiness.resolve` attributes every model failure a `.retry(alias)` recovery action. But for the model-load failures the readiness banner actually surfaces, `FailureDiagnoser.diagnosis(for:)` already recommends a *different* action:

- `.modelOutOfMemory` → `.openModelManagement` ("free memory or choose a smaller model")
- `.modelLoadFailed` → `.openModelManagement`
- `.engineNotRunning` → `.restart`

So on an OOM crash the banner shows **Retry**, which just repeats the failing start, instead of pointing at Model Management.

The fix is a UX decision on the readiness banner's action model (loriz's feature), which is why it's deferred rather than forced into #1510:

- **Option A (fuller):** add `.openModelManagement` / `.restart` cases to `ModelReadiness.Action`, derive the action from `failure.kind` via `FailureDiagnosis`, and route them in `ContentView.performReadinessAction` (the routing already exists in `QuickstartView.handleQuickstartFailureAction`).
- **Option B (minimal):** show no action button when the diagnosis action isn't retry-able, relying on the message copy (which already says what to do). Trade-off: removes the banner's Retry affordance for those crashes.

Context: codex flagged this reviewing #1510; the two substantive scoping bugs were fixed there.

raullenchai:
Fixes #1514.

## What changed

- derive readiness recovery actions from the existing `FailureDiagnoser` policy
- route model OOM/load failures to Settings → Model Management
- expose Restart for engine-not-running failures
- preserve Retry for other and unclassified readiness failures
- add regression coverage for all three paths

## Root cause

`ModelReadiness.resolve` retained the structured failure kind for copy generation, but discarded it when choosing the CTA and unconditionally returned `.retry(alias:)`.

## Validation

- Reproduced against `origin/main`: `.modelLoadFailed` resolves to Retry despite `FailureDiagnoser` prescribing Model Management
- `swift build` passes
- `git diff --check` passes
- Local `swift test` is blocked before test execution by the existing #1638 toolchain issue (`no such module XCTest`); CI/pr_validation will run the suite in the supported environment

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
@@ -20,6 +20,7 @@ struct ContentView: View {
     @Environment(UpdateChecker.self) private var updater
     @Environment(QuickstartCoordinator.self) private var quickstart
     @Environment(BrowseApprovalStore.self) private var browseApproval
+    @Environment(\.openWindow) private var openWindow
 
     @State private var alias: String = ""
     /// Monotonic signal from picker row taps. Catalog initialization also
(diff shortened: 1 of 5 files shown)


raullenchai:
Final validation on `34f59a2c`:

- `pr_validate`: **MERGE-SAFE**
- Codex review: **0 blocking issues**
- Full unit: **16819 passed**, 57 skipped, 18 deselected, 6 xfailed, 1 xpassed
- GitHub checks: **20 passed**, 5 expected skips, 0 pending, 0 failed
- Base verified against current `main` (`8f7e0847`); mergeable state: **CLEAN**

PR #1754, opened 2026-08-09
Revert #1745 after merging with a red macOS build

raullenchai:
Reverts #1745.

## Why

PR #1745 introduced a source-grep test that always slices `startModel` at an inner closure brace. The rapid-mac `build` check failed at 22:22:01, but the PR was merged at 22:22:05. Main therefore inherited a deterministic red build.

## Scope

- revert the complete #1745 squash commit, temporarily restoring the original Chat readiness behavior
- record the incident-derived final-check snapshot rule in the merge SOP
- explicitly distinguish compiling a test target from executing its tests

The functional fix for #1739 will be reimplemented in a separate PR with a literal-aware, balanced source guard and will not land until the real Swift suite and final macOS checks are green.

## Validation

- `swift build --package-path apps/rapid-mac` passes
- `git diff --check` passes

This is a broken-main revert; closes no user issue.


raullenchai:
Codex finding dismissal (intentional, temporary): the revert necessarily restores #1739 while removing the deterministic-red source guard from main. Keeping the functional hunk while only deleting its test would not be a true revert and would leave an unverified behavior change on main—the exact process failure this rollback corrects. The replacement PR will restore `ensureServing` with an executable literal-aware/balanced guard. This tradeoff is explicitly scoped in the PR body and requested as revert-then-rework.

PR #1760, opened 2026-08-10
fix(mac): surface the resolved engine override

### LINKED ISSUE #1712: A runtime-override sidecar can shadow every future bundled engine, and nothing in the app says so
## What happened

Spent a chunk of a dogfood session showing GUI builds and drawing conclusions from them, before noticing the app was not running the engine it shipped with.

`ServerLocator` resolves `rapid-mlx` from three slots, and the runtime-override slot won:

```
~/Library/Application Support/Rapid/runtime-override/rapid-mlx/bin/rapid-mlx
```

That file was a two-line shim:

```bash
#!/bin/bash
exec "$HOME/rapid-image-demo/.venv/bin/rapid-mlx" "$@"
```

The venv it pointed at was an editable install resolving `vllm_mlx` to a working tree that was **286 commits behind main**, on an unrelated feature branch, with uncommitted changes. `vllm_mlx/utils/chat_template.py` there was 1166 lines against main's 1445.

Beside it, `runtime-override/rapid-mlx/VERSION` read `99.0.0`.

## Why it is sticky

`ServerLocator` picks the newest of the runtime-override and bundled `VERSION` files — deliberately, so "a stale or unversioned runtime override cannot shadow a newer, versioned sidecar shipped by an app update."

`99.0.0` defeats that permanently. No real release will ever exceed it, so every future app update silently loses to it. The guard is version-comparison only; it has no notion of "this override is implausible" or "this override is a 67-byte script, not a sidecar."

## Why it is invisible

Nothing surfaces which engine slot won. Not the About window, not Settings, not the readiness banner. The app reported itself healthy and served chat normally the whole time — it was simply serving from a 286-commit-old engine. Every behaviour I attributed to the fresh build was actually a mix: my Swift, someone else's engine.

That is the real cost. A dogfood session's entire purpose is attributing behaviour to a build, and this makes that attribution silently wrong.

## Suggested shape

Two independent pieces, either of which would have caught it:

1. **Show the resolved engine.** Surface the winning slot and its version somewhere a dogfooder looks — About, or the Settings → Models header. "Engine 0.12.7 (bundled)" vs "Engine 99.0.0 (runtime override)" ends this class of confusion in one glance.
2. **Distrust an implausible override.** A `VERSION` that parses far above anything the update channel has ever published is not a newer sidecar, it is a pin. Either refuse it, or accept it but log loudly. Sanity-checking that the override's `bin/rapid-mlx` looks like the sidecar layout (a real interpreter beside it, not a 67-byte script) would also do it.

## Note

The override tree itself was genuine — bootstrapper-installed on 2026-07-28, with full `site-packages/` and `python/`. Only `bin/rapid-mlx` had been replaced with the shim, and `VERSION` bumped, on 2026-08-08. So the failure mode does not require anyone to hand-build a fake sidecar; overwriting one file inside a legitimate install is enough.

Filing as a developer-experience issue, not a user-facing bug: shipped users have no reason to have a runtime override pointing at a checkout. The cost lands entirely on dogfooding, which is where build-attribution accuracy matters most.

raullenchai:
## Why

A managed runtime override can win engine selection forever with a fabricated high `VERSION`, while the app gives no indication that the bundled engine lost. That makes dogfood results silently attributable to the wrong engine.

The underlying source classifier already existed and was documented as feeding the About panel, but `AboutPanel` never consumed it. In addition, managed-slot classification did not follow symlinks even though `ServerLocator.find()` returns symlink-resolved paths.

## What

- Show the actual resolved engine version and source in About.
- Mark runtime and `RAPID_BIN` overrides with a warning icon; expose the resolved path through the tooltip and accessibility label.
- Classify symlinked runtime/bundled launchers against their managed slot.
- Read `VERSION` from the managed slot root, not a symlink target checkout.
- Add regressions for the reported `99.0.0` override, bundled engines, symlinked overrides, and missing binaries.

## Validation

- `swift build` — PASS.
- `git diff --check` — PASS.
- Local Swift tests are blocked before execution by the repository/toolchain issue tracked in #1638 (`Testing` / `XCTest` unavailable); the GitHub Apple runner will execute the new suite.

Fixes #1712

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/AboutPanel.swift
+++ b/apps/rapid-mac/Sources/Rapid/AboutPanel.swift
@@ -10,6 +10,21 @@ import SwiftUI
 /// the About window can stay a simple, on-brand credit: mark, version,
 /// one-line what-it-is, and the public links.
 enum AboutPanel {
+    struct EngineIdentity: Equatable, Sendable {
+        let version: String?
+        let source: ServerLocator.ResolvedSource
+        let path: String
+
+        var summary: String {
+            let versionLabel = version.map { "Engine \($0)" } ?? "Engine version unknown"
+            return "\(versionLabel) · \(source.displayLabel)"
+        }
+
+        var isOverride: Bool {
+            source == .runtimeOverride || source == .rapidBin
+        }
+    }
+
     private static let website = "https://rapidmlx.com"
     private static let repoURL = "https://github.com/raullenchai/Rapid-MLX"
     /// The policy in the repository, not `rapidmlx.com/privacy` — that page
@@ -36,6 +51,7 @@ enum AboutPanel {
         let view = AboutView(
             version: bundleShortVersion(),
             build: bundleBuildNumber(),
+            engine: engineIdentity(binaryPath: server.binaryPath),
             website: website,
             repoURL: repoURL,
             privacyURL: privacyURL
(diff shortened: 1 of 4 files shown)


raullenchai:
Final validation on `4239313f` after rebasing onto current `main` (`8e158c0f`):

- `pr_validate`: **MERGE-SAFE**
- Codex review: **0 blocking issues** (after resolving two provenance findings)
- Full unit: **16819 passed**, 57 skipped, 18 deselected, 6 xfailed, 1 xpassed
- Production `swift build`: **PASS**
- GitHub checks: **20 passed**, 5 expected skips, 0 pending, 0 failed
- macOS build, Apple tests, GUI golden flows, and all five L1 smokes: **PASS**
- Merge state: **CLEAN**

PR #1771, opened 2026-08-10
fix(metrics): expose MLLM prefix cache counters

raullenchai:
## What changed

- teach `/metrics` to use the enabled MLLM prefix-cache snapshot nested at `mllm_scheduler.vision_cache`
- keep cache metrics absent when the MLLM cache is disabled or unavailable
- add a regression covering the zero-traffic MLLM series

## Root cause

The MLLM scheduler exposes `MLLMPrefixCacheManager` counters under a nested `vision_cache` key, while the Prometheus renderer only searched the three top-level text-lane cache keys. The pressure-eviction metric is unconditional, so MLLM scrapes showed that series but omitted prefix-cache hits and misses entirely.

## Impact

An enabled MLLM prefix cache now reports explicit zero-valued hit/miss/eviction/token-saved counters before its first lookup. Prometheus dashboards can distinguish an idle cache from a disabled cache.

## Validation

- reproduced on `main`: zero-traffic MLLM cache regression failed because the series were absent
- `python3.12 -m ruff check vllm_mlx/routes/metrics.py tests/test_metrics_route.py`
- `python3.12 -m pytest tests/test_metrics_route.py tests/test_mllm_cache.py -q` (54 passed)

### DIFF
--- a/tests/test_metrics_route.py
+++ b/tests/test_metrics_route.py
@@ -288,6 +288,42 @@ def test_metrics_omits_cache_series_when_no_cache_active(metrics_client):
     assert "rapid_mlx_requests_processed_total 0" in body
 
 
+def test_metrics_exposes_mllm_prefix_cache_zero_counters(metrics_client):
+    """An enabled MLLM prefix cache must not disappear at zero traffic.
+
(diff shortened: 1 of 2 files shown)


raullenchai:
@codex review

PR #1775, opened 2026-08-10
fix: account for cached bytes in disk preflight

### LINKED ISSUE #1772: A fully-cached mflux image model is refused at startup by the disk-space gate
## What happens

`rapid-mlx serve z-image-turbo` refuses to start a model that is **already 100 % downloaded**:

```
Error: Insufficient disk space for download.
  Model size:        5.5 GB
  Free space:        5.5 GB  (/Users/.../.cache/huggingface/hub)
  Need to free:      0.5 GB
```

`~/.cache/huggingface/hub/models--filipstrand--Z-Image-Turbo-mflux-4bit` is 5.5 GB with zero `.incomplete` blobs. Nothing needs downloading, so the suggestion the gate prints — "Bypass this check (download will likely fail mid-way)" — is also misleading here.

## Why both cached-probes miss

mflux-layout repos keep their weights **and their configs** under `transformer/` / `text_encoder/` / `vae/`, with no `config.json` at the checkpoint root. Two independent probes decide "is this cached", and both are root-level:

| probe | `filipstrand/Z-Image-Turbo-mflux-4bit` | `Runpod/FLUX.2-klein-4B-mflux-4bit` |
|---|---|---|
| `_check_disk_space` → `try_to_load_from_cache(prefix + "config.json")` (`vllm_mlx/cli.py:947`) | `None` — misses | hits (this repo *does* have a root `config.json`) |
| `is_repo_cached` (`vllm_mlx/_download_gate.py`) | `False` — misses | `False` — misses |

`flux2-klein-4b` survives only because the first probe happens to hit. For `z-image-turbo` both miss, so the gate runs the size estimate and `sys.exit(1)`.

## `--force-disk-check` does not fully rescue it

On the `serve z-image-turbo` path the gate fires **three** times. The first two honour the flag (`--force-disk-check set — proceeding anyway.`); the third — after the server banner, during model load — does not receive it and exits. So the documented bypass cannot start a fully-cached model either.

## What the user sees in the app

In the Images tab this surfaces as a bare **"Retry"** on the readiness action, with no explanation, for a model the app itself already downloaded.

Discriminator, same app instance, same button, nothing else changed:

* 5.5 GiB free → readiness fails, button becomes "Retry", no sidecar process, nothing listening
* 8.4 GiB free → same button starts `serve z-image-turbo` in 4 s and renders normally

## Impact

The tighter the disk, the more likely a user is to hit this — and the users most likely to be tight on disk are exactly the ones who already downloaded the model and now cannot use it. Any model whose checkpoint layout is nested rather than root-level is affected; both wired image aliases are nested, and only one of them is saved by an incidental root `config.json`.

Found during the post-#1768 GUI verification of the Images tab.

raullenchai:
## Summary

- compute disk preflight requirements from remote bytes that are not already present at the exact remote revision in the Hugging Face cache
- support complete nested checkpoint layouts such as mflux without weakening partial-cache protection
- remove the unsafe root `config.json` shortcut that treated config-only text caches as complete

## Root cause

The disk preflight used a root-level `config.json` as its only cache hit signal. mflux checkpoints keep configs and weights under component directories, so a fully cached Z-Image snapshot missed that probe and was charged for the entire repository again. Conversely, a partial text cache containing only `config.json` skipped the gate entirely.

## User impact

`rapid-mlx serve z-image-turbo` can start from a complete cache even when free disk is below the full model size. Partial downloads and stale local refs remain protected, but the gate now checks only the bytes that still need to be transferred for the current remote revision.

Closes #1772

## Validation

- `python3.12 -m pytest -q tests/test_disk_space_check.py tests/test_download_gate.py tests/test_alias_subfolder.py tests/test_cli_chat.py tests/test_image_lane.py` — 307 passed
- `python3.12 -m ruff check vllm_mlx/cli.py tests/test_disk_space_check.py`
- `python3.12 -m ruff format --check vllm_mlx/cli.py tests/test_disk_space_check.py`
- real `filipstrand/Z-Image-Turbo-mflux-4bit` cache: 21/21 sized files cached, 0 missing bytes, disk gate passes

### DIFF
--- a/tests/test_disk_space_check.py
+++ b/tests/test_disk_space_check.py
@@ -9,7 +9,7 @@ and respect HF_HOME via huggingface_hub.constants.HF_HUB_CACHE.
 from __future__ import annotations
 
 from types import SimpleNamespace
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 import pytest
 
@@ -18,7 +18,10 @@ from vllm_mlx.cli import _check_disk_space
 
 def _make_info(file_sizes_bytes: list[int]) -> SimpleNamespace:
     """Build a fake huggingface_hub.ModelInfo with sibling file sizes."""
-    siblings = [SimpleNamespace(size=sz) for sz in file_sizes_bytes]
+    siblings = [
+        SimpleNamespace(size=sz, rfilename=f"file-{index}.safetensors")
+        for index, sz in enumerate(file_sizes_bytes)
+    ]
     return SimpleNamespace(siblings=siblings, safetensors=None)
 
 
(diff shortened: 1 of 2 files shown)


raullenchai:
Superseded by #1773, which landed the same current-revision per-file cache accounting with the narrower selective-loader fast path. Reverified on main 77d378e5: 64 adjacent tests pass and the real fully cached filipstrand/Z-Image-Turbo-mflux-4bit disk gate passes.

PR #1779, opened 2026-08-10
fix(qwen3): preserve literal think tags in non-streaming content

### LINKED ISSUE #1778: A literal <think> in ordinary prose truncates the non-streaming answer, while streaming keeps it
#1766 fixed the `<tool_call>` half of this: a literal `<tool_call>` / `</tool_call>` in ordinary prose now survives when the request declares no tools. The `<think>` half is still broken, and only on the **non-streaming** path.

Verified on a local build of `main` at `f3a39da2`, `qwen3.6-27b-4bit` (`reasoning_parser=qwen3`) served from the bundled sidecar. No tools declared. Prompt: *"Reply with EXACTLY this one line and nothing else, copying every character verbatim: `<LINE>`"*.

| line asked for | non-streaming `content` | non-streaming `reasoning_content` | streaming `content` |
|---|---|---|---|
| `BEGIN <think> AFTER END` | `BEGIN` | `AFTER END` | `BEGIN <think> AFTER END` |
| `BEGIN <think>inner</think> AFTER END` | `AFTER END` | `BEGIN <think>inner` | `BEGIN <think>inner AFTER END` |
| `BEGIN </think> AFTER END` | `AFTER END` | `BEGIN` | `BEGIN  AFTER END` |
| ``BEGIN `<think>` AFTER END`` | ``BEGIN ` `` | ``` ` AFTER END``` | ``BEGIN `<think>` AFTER END`` |
| `BEGIN <foo> AFTER END` (control) | `BEGIN <foo> AFTER END` | *(empty)* | `BEGIN <foo> AFTER END` |

Three things fall out of that table:

* **The two surfaces disagree.** The same request returns a whole answer when streamed and a fragment when not. A client that switches `stream` on or off gets different text.
* **Backticks do not protect it.** ``` `<think>` ``` — the exact form used when explaining the tag in a code span — is split just like the bare tag, so the answer is cut mid-token at the opening backtick.
* **It is not confined to unterminated tags.** A balanced `<think>…</think>` pair moves everything *before* it into `reasoning_content`, so the visible answer loses its beginning instead of its end.

## Impact

Anyone who asks a Qwen-family model to explain reasoning-tag formats, or who pastes documentation, a prompt template, or code containing `<think>`, gets a silently truncated answer on the non-streaming path. Nothing signals the loss — the discarded text is in `reasoning_content`, which most OpenAI-compatible clients never render.

## Note on scope

Measured on `qwen3.6-27b-4bit`; the mechanism is the reasoning parser rather than the tool parser, so every alias with `reasoning_parser: qwen3` is a candidate (that includes `lfm2.5-2.6b-4bit`). The streaming path already gets this right, which is the useful reference for what the non-streaming path should do: a `<think>` that the model *echoed* is not a `<think>` that the template *primed*.

raullenchai:
## Necessity

Fixes #1778. Qwen-family non-streaming responses silently truncated ordinary prose containing a literal `<think>` opener, including Markdown code spans, while the streaming surface kept the answer text.

## What changed

- Thread the existing `prompt_thinking_active` signal into `Qwen3ReasoningParser.extract_reasoning`.
- When the prompt already primed reasoning and the generated output first emits ordinary prose before `<think>`, classify that opener as an echoed literal rather than a new reasoning boundary.
- Preserve unmatched literal openers verbatim; for balanced literal pairs, remove only the wrappers so the body remains visible and matches streaming content behavior.
- Preserve leading structural openers, implicit `</think>` closing, and autonomous explicit reasoning when the prompt was not primed.

## Validation

- `270 passed`: reasoning parser, truncation-finalize, and chat tool-tag regression suites.
- Full local unit suite passed on this branch before the final signal-narrowing/test-only refinement.
- `ruff check` and `ruff format --check` pass for changed files.
- Mutation evidence: changing the new literal guard to `False` made all three issue reproductions fail: unmatched opener, balanced pair, and backtick-wrapped opener.

- GitHub CI: all checks passed, including Apple Silicon and five L1 smoke jobs.
- M3 release sweep: qwen3.5-4b passed 6/6 coherence; two larger-model runs returned infrastructure 503s before assertions, so the sweep was stopped after the repeated infrastructure failure.
## AI assistance

Codex authored and reviewed the changes in `vllm_mlx/reasoning/qwen3_parser.py` and `tests/test_reasoning_parsers.py`. The maintainer verified the behavior with targeted tests, a deliberate contract mutation, lint/format checks, the broader unit suite, and the repository PR validation workflow. I can explain every changed line on demand.

### DIFF
--- a/tests/test_reasoning_parsers.py
+++ b/tests/test_reasoning_parsers.py
@@ -1806,6 +1806,16 @@ class TestQwen3:
         assert reasoning == "analysis"
         assert content == "answer"
 
+    def test_1778_implicit_close_before_literal_opener_keeps_reasoning(self):
+        """A real implicit close before answer prose stays authoritative."""
+        reasoning, content = self.parser.extract_reasoning(
+            "reasoning</think>answer <think>literal</think>",
+            enable_thinking=True,
+            prompt_thinking_active=True,
+        )
+        assert reasoning == "reasoning"
+        assert content == "answer <think>literal</think>"
+
     # ---- #575 fast-path coverage (Qwen3 override branch) ----------------
 
     def test_575_qwen3_fast_path_no_tags_enable_thinking_true(self):
--- a/tests/test_reasoning_parsers.py
+++ b/tests/test_reasoning_parsers.py
@@ -1744,6 +1744,68 @@ class TestQwen3:
         assert reasoning is None
         assert content == "content"
 
+    @pytest.mark.parametrize(
+        "text,expected_content",
+        [
+            ("BEGIN <think> AFTER END", "BEGIN <think> AFTER END"),
+            (
+                "BEGIN <think>inner</think> AFTER END",
+                "BEGIN inner AFTER END",
+            ),
+            ("BEGIN `<think>` AFTER END", "BEGIN `<think>` AFTER END"),
+        ],
+    )
+    def test_1778_mid_answer_think_literal_stays_content(self, text, expected_content):
+        """A generated opener after visible prose is literal Qwen output.
+
+        Non-streaming used to partition at the opener regardless of its
+        position, truncating Markdown explanations and moving either side
+        of a balanced pair into ``reasoning_content``.  Thinking is enabled
+        here to match the served aliases from the issue report.
+        """
+        from vllm_mlx.service.helpers import _finalize_content_and_reasoning
+
+        content, reasoning = _finalize_content_and_reasoning(
+            raw_text=text,
+            cleaned_text=text,
+            tool_calls=[],
+            reasoning_parser=self.parser,
+            enable_thinking=True,
+            prompt_thinking_active=True,
+        )
+        assert reasoning is None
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex round 2 finding dismissed with protocol evidence: when `prompt_thinking_active=True`, the rendered Qwen prompt has already opened the reasoning channel. A later generated `<think>` cannot be the structural opener for that same channel; it is an echoed literal. The proposed counterexample (`analysis preamble <think>more analysis</think>answer`) is byte-identical to the reported literal-copy response, so treating it as structural necessarily reintroduces #1778 and there is no post-hoc string discriminator. The parser therefore uses the out-of-band rendered-prompt signal as authoritative. We did accept the actionable adjacent case: an earlier generated `</think>` is an authoritative close, so `reasoning</think>answer <think>literal</think>` still extracts `reasoning` correctly; the new regression test pins it.

PR #1782, opened 2026-08-10
fix(scheduler): reap orphaned disconnect slots

### LINKED ISSUE #1759: Scheduler `running` slots leak on client-disconnect abort — accumulates until the server wedges (running=4 forever, 0 completions, no error)
### Rapid-MLX version

0.11.0 (also inspected 0.12.4 wheel and main @ b5b95eb — the abort path is byte-identical, see body)

_No response_

### Hardware

Mac Studio M3 Ultra, 512 GB (two independent nodes, same model/flags, both affected)

### macOS version

26.3 (25D125) on one node, 26.5.2 (25F84) on the other

### Python version

3.12.13

### Model

mlx-community/Qwen3.6-35B-A3B-8bit (hybrid MoE + GatedDeltaNet) + mlx-community/Qwen3-Embedding-4B-4bit-DWQ

### Full serve command

rapid-mlx serve mlx-community/Qwen3.6-35B-A3B-8bit --host 0.0.0.0 --port 8081 --max-tokens 4096 --cache-memory-percent 0.25 --pin-system-prompt --hybrid-cache-entries 8 --max-num-seqs 4 --reasoning-parser qwen3 --embedding-model mlx-community/Qwen3-Embedding-4B-4bit-DWQ (with HF_HUB_OFFLINE=1)

Streaming (`stream: true`)

Actual:
Client disconnects (`disconnect_guard` -> `abort_request`) intermittently fail to release the scheduler's running slot. Each leaked slot permanently reduces usable concurrency (`--max-num-seqs`). Leaked slots accumulate until ALL slots are ghosts: the server then accepts requests, queues them forever, and completes nothing — while `/v1/models` and `/metrics` keep returning 200, so HTTP health checks see a healthy server. We hit this 4 times in 6 days on two independent production nodes; the worst incident left one node serving zero chat completions for 5.5 hours.

Observable state while wedged:
- `rapid_mlx_requests_running` pinned at 4 (= max-num-seqs), `waiting` growing, `increase(rapid_mlx_requests_processed_total[30m]) == 0` for 5.5 h
- scheduler main loop keeps stepping the ghost sequences: `[Metal memory] ... step=N running=4 waiting=2` advances step by 256 per line indefinitely, process at ~70% CPU for hours, `cache=0.0GB`
- memory far below cap (active 42 GB / 512 GB) — NOT a memory-pressure condition, and no 503s (requests are admitted and queued, downstream sees 408 timeouts).

Not a duplicate of the two similar-sounding issues:
- #1646 (D-METAL-CAP wedge): that one 503s at admission under memory pressure; ours admits and queues, memory is fine.
- #1359 (streaming wedge, fixed in #1391 / v0.11.9): that is a single request whose output is suppressed while generation continues; ours is the scheduler's running-slot count leaking AFTER the abort completes, accumulating across many requests until exhaustion.

Correlation: leaks only occur in windows where `rapid_mlx_requests_cancelled_via_disconnect_total` is advancing (our business hours). Rough rate: 1 leaked slot per ~10-25 disconnects. Not deterministic — most disconnect aborts release the slot fine; looks like a race depending on which phase the abort lands in. There is also a feedback loop that makes the terminal state self-accelerating: leaked slots -> queueing -> client/gateway timeouts -> more aborts -> more leaks (evidence: normally our cancelled_total == cancelled_via_disconnect_total exactly; during the wedge the node showed 152 vs 65 — the extra 87 were timeout-driven aborts feeding back in).

Code inspection (why we did not just upgrade): we diffed the slot-owning functions in `vllm_mlx/scheduler.py` function-by-function — `_do_abort_request`, `abort_request`, `remove_finished_request` are byte-identical from v0.11.0 through v0.12.4 and current main (checked at b5b95eb, 2026-08-08). `routes/chat.py` `_disconnect_guard` / admission-release path shows comment-level changes only in that range. So 0.12.x should be equally affected.

Expected:
An aborted request always releases its running slot; `running` returns to 0 when no live requests exist. Failing a root-cause fix, a defensive reconciliation (periodically reap entries in `self.running` that no longer map to a live request in `self.requests`) would already help operationally — today the only remedy is a full restart, and graceful drain can never finish once ghosts exist (kill -9 required).

### Minimal reproduction

We do not have a deterministic local repro (it is a race, ~1 leak per 10-25 disconnects under real agentic traffic). The closest recipe, matching how production triggers it:

1. Start server with the serve command above (key ingredients we suspect: hybrid model + `--max-num-seqs 4` + streaming + long prompts).
2. Generate agentic-style traffic: long-prompt streaming chat completions with `tools`, several requests in flight (running 1-4, some waiting).
3. While responses are streaming (and occasionally while a request is still waiting/prefilling), abort from the client side: close the TCP connection mid-stream (what Codex/Cline-style clients and gateway timeouts do). Repeat tens of times.
4. Watch `curl -s http://127.0.0.1:8081/metrics | grep rapid_mlx_requests_running` with NO requests in flight: after enough disconnect-aborts it stays pinned at N > 0 permanently. `min_over_time(rapid_mlx_requests_running[3h]) > 0` in Prometheus is our detection rule.
5. Keep going and it eventually pins at max-num-seqs: server accepts + queues + never completes anything (all-ghost wedge).

Timeline of the worst incident (2026-08-07, times local): 14:11 clean restart -> 16:09 two slots stuck -> 17:09 4/4 stuck -> 17:19-22:49 zero completions for 5.5 h (gateway logged 39x 408 against this node; p95 = gateway timeout ceiling) -> 22:50 bootout hung, kill -9, restart, healthy again (running=4/waiting=0 under a 4-request probe).

We run this in production daily and can test candidate patches quickly. We can also provide: full /metrics snapshot taken while wedged, server log tail from the wedged period, Prometheus history of running/processed/cancelled_via_disconnect for all 4 incidents.

```shell
There is no error output — that is part of the problem. The log is completely clean while the server is dead (no traceback, no METAL error, nothing at WARNING+). The only visible symptom is the scheduler heartbeat stepping ghost sequences forever with a frozen running/waiting count:

[2026-08-07 22:48:43] INFO:rapid_mlx.scheduler:[Metal memory] active=42.0GB peak=133.9GB cache=0.0GB step=656924160 running=4 waiting=2
[2026-08-07 22:48:43] INFO:     127.0.0.1:49292 - "GET /metrics HTTP/1.1" 200 OK
[2026-08-07 22:48:43] INFO:rapid_mlx.scheduler:[Metal memory] active=42.0GB peak=133.9GB cache=0.0GB step=656924416 running=4 waiting=2
... (identical lines, step advancing by 256, for 5.5 hours; last real completion was at 17:1x)

/metrics snapshot taken while wedged (same moment):
rapid_mlx_requests_processed_total 154
rapid_mlx_requests_running 4
rapid_mlx_requests_waiting 2
rapid_mlx_requests_cancelled_total 152
rapid_mlx_requests_cancelled_via_disconnect_total 65

(processed frozen at 154 for hours; cancelled_total >> via_disconnect due to the timeout->abort feedback loop described above. Earlier, pre-wedge incidents on both nodes show the same pattern with cancelled_total == via_disconnect exactly.)

Happy to attach the full /metrics dump and a longer log excerpt on request.
```

raullenchai:
## Summary

- reconcile scheduler `running` entries after deferred abort processing
- treat a running request missing from canonical `requests` tracking as a disconnected orphan
- remove the orphan through the existing BatchGenerator-aware abort path before the next decode tick
- add a deterministic regression that recreates the production ghost-slot state

Closes #1759

## Reproduction

The deterministic regression constructs the state observed in production:

- the streaming consumer has unwound and `EngineCore._cleanup_request` has removed `scheduler.requests[request_id]`
- `scheduler.running`, both uid maps, and the BatchGenerator slot still retain the request
- `_pending_abort_ids` no longer contains the abort edge

Before this change, `Scheduler.step()` calls `BatchGenerator.next()` and the ghost continues consuming a `max_num_seqs` slot. The regression makes `next()` fail if reached.

I also exercised main HEAD on the reporter's exact cached model (`mlx-community/Qwen3.6-35B-A3B-8bit`) with `--max-num-seqs 4`, `--hybrid-cache-entries 8`, long tool-bearing prompts, and streaming disconnects:

- 160 TCP-RST attempts: final `running=0`, `waiting=0`
- 200 disconnects after receiving SSE chunks: final `running=0`, `waiting=0`; all 200 were attributed via the disconnect counter

The low-rate production race did not trigger in that finite run, but the deterministic lifecycle state reproduces the permanent-slot failure and pins the recovery invariant.

## Safety invariant

After pending aborts are drained, a live running request must still exist in `scheduler.requests`. Normal completion removes `running` on the scheduler executor thread before engine cleanup. Disconnect cleanup is the only path that removes the canonical request first. Reconciliation therefore cannot reap a normal live request and runs on the executor thread before `BatchGenerator.next()`.

## Validation

- `python3.12 -m pytest -q tests/test_batching.py tests/test_cancelled_requests_metric.py tests/test_disconnect_counter_prod_shape.py tests/test_disconnect_guard_aborts_scheduler.py tests/test_rst_mid_sse_zombie_kv.py` — 86 passed, 2 deselected
- `python3.12 -m ruff check vllm_mlx/scheduler.py tests/test_batching.py` — clean
- `git diff --check` — clean

`tests/test_event_loop.py` was also invoked but is an external-server manual harness and its two network cases failed only because no server was listening on its hard-coded port 8000.

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -315,6 +315,46 @@ class TestSchedulerBasic:
         assert scheduler.abort_request("known-1") is True
         assert scheduler.abort_request("known-1") is True
 
+    def test_step_reaps_disconnect_orphan_before_next(
+        self, mock_model, mock_tokenizer
+    ):
+        """#1759: engine cleanup without a surviving abort must not leak a slot.
+
+        This is the deterministic form of the production race: the streaming
+        consumer is gone (``requests`` was popped), but the scheduler's running
+        map, uid maps and mlx-lm batch slot remain while the deferred-abort set
+        is empty.  Before the reconciliation guard, ``step()`` called
+        ``BatchGenerator.next()`` forever and ``num_running`` stayed pinned.
+        """
+        scheduler = Scheduler(model=mock_model, tokenizer=mock_tokenizer)
+        request = Request(
+            request_id="disconnect-orphan",
+            prompt="Hello",
+            sampling_params=SamplingParams(max_tokens=512),
+        )
+        request.status = RequestStatus.RUNNING
+        uid = 73
+
+        # Deliberately do not populate scheduler.requests: this models
+        # EngineCore._cleanup_request having already released the consumer.
+        scheduler.running[request.request_id] = request
+        scheduler.request_id_to_uid[request.request_id] = uid
+        scheduler.uid_to_request_id[uid] = request.request_id
+        scheduler._pending_abort_ids.clear()
+
+        batch = MagicMock()
+        batch.next.side_effect = AssertionError("ghost slot was decoded")
+        scheduler.batch_generator = batch
+
+        scheduler.step()
+
+        batch.remove.assert_called_once_with([uid])
+        batch.next.assert_not_called()
+        assert request.request_id not in scheduler.running
+        assert request.request_id not in scheduler.request_id_to_uid
+        assert uid not in scheduler.uid_to_request_id
+        assert scheduler.get_num_running() == 0
+
     def test_get_stats(self, mock_model, mock_tokenizer):
         """Test getting scheduler stats."""
         scheduler = Scheduler(
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -315,9 +315,7 @@ class TestSchedulerBasic:
         assert scheduler.abort_request("known-1") is True
         assert scheduler.abort_request("known-1") is True
 
-    def test_step_reaps_disconnect_orphan_before_next(
-        self, mock_model, mock_tokenizer
-    ):
+    def test_step_reaps_disconnect_orphan_before_next(self, mock_model, mock_tokenizer):
         """#1759: engine cleanup without a surviving abort must not leak a slot.
 
         This is the deterministic form of the production race: the streaming
@@ -335,12 +333,16 @@ class TestSchedulerBasic:
         request.status = RequestStatus.RUNNING
         uid = 73
 
-        # Deliberately do not populate scheduler.requests: this models
-        # EngineCore._cleanup_request having already released the consumer.
+        scheduler.requests[request.request_id] = request
         scheduler.running[request.request_id] = request
         scheduler.request_id_to_uid[request.request_id] = uid
         scheduler.uid_to_request_id[uid] = request.request_id
         scheduler._pending_abort_ids.clear()
+
+        # The engine releases its canonical tracking entry while the running
+        # slot is still live.  This is the production cleanup edge that arms
+        # the targeted reconciliation fallback.
+        scheduler.remove_finished_request(request.request_id)
 
         batch = MagicMock()
         batch.next.side_effect = AssertionError("ghost slot was decoded")
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -357,6 +357,48 @@ class TestSchedulerBasic:
         assert uid not in scheduler.uid_to_request_id
         assert scheduler.get_num_running() == 0
 
+    def test_orphan_candidate_cannot_abort_reused_request_id(
+        self, mock_model, mock_tokenizer
+    ):
+        """A stale cleanup candidate is scoped to one Request identity."""
+        scheduler = Scheduler(model=mock_model, tokenizer=mock_tokenizer)
+        old = Request("reused-id", "old", SamplingParams())
+        old.status = RequestStatus.RUNNING
+        scheduler.requests[old.request_id] = old
+        scheduler.running[old.request_id] = old
+        scheduler.remove_finished_request(old.request_id)
+
+        new = Request("reused-id", "new", SamplingParams())
+        new.status = RequestStatus.RUNNING
+        scheduler.requests[new.request_id] = new
+        scheduler.running[new.request_id] = new
+        scheduler.batch_generator = MagicMock()
+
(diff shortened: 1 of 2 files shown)


raullenchai:
Validation update for head `92338b36`:

- Codex adversarial review: **0 blocking**
- targeted: **98 passed, 2 deselected**
- full unit: **16,867 passed**, 57 skipped, 18 deselected, 6 xfailed, 1 xpassed
- lint / format / supply chain: clean
- `pr_validate --skip-steps stress_e2e_bench`: **MERGE-SAFE**

The live-model matrix was run separately before the merge-safe pass. Qwen3.5 and Harmony stress were 8/8 and all selected agent SDK lanes passed. Its two red entries were environmental: the selected Qwen3.6 checkpoint was not cached and awaited a 32.3 GiB download confirmation; a concurrent PR validation on the same fixed port 8451 sent SIGTERM to the DiffusionGemma server, producing a connection-refused in memory-stability round 5 (base rerun 8/8). This PR's issue-specific real-model exercise used cached `mlx-community/Qwen3.6-35B-A3B-8bit`, max-num-seqs=4, hybrid-cache-entries=8: 160 TCP-RST attempts plus 200 post-SSE disconnects both drained to running=0/waiting=0.

PR #1784, opened 2026-08-10
fix(mac): re-synthesize once when a grounded answer denies real-time access

raullenchai:
## Why
Dogfood report (BUG C): with the built-in web tools, the model sometimes fetches real current data (web_search/browse returning live results) and then still answers *"I can't access real-time information, my knowledge is only up to 2024,"* discarding its own successful fetch. The anti-confabulation preamble (#1556) rides on that turn and mitigates it, but a small local model (e.g. Qwen3.5-4B) can still ignore the instruction. That preamble is the only line of defence today, and it is probabilistic.

## What
A deterministic second layer in `runToolLoop`. After a synthesis turn finishes in prose (`.terminal`), if **a tool result for THIS turn is on the wire** (`carriesToolResultForThisTurn`) **AND** the answer is a first-person temporal-denial refusal (`looksLikeUngroundedRefusal`), the loop forces **exactly one** tools-disabled correction round carrying a failure-specific preamble (`groundingCorrectionPreamble`) that names the mistake: *"the tool result above was fetched just now and IS the current data — answer from it, do not mention a knowledge cutoff."*

Safety properties:
- **One-shot** (`groundingCorrectionUsed`) — cannot loop.
- **Evidence-gated** — only fires when a real tool result is present this turn, so a legitimate "I can't access X" with no tool data is never touched.
- **Epoch + cancellation guarded** — never fires on a superseded/cancelled turn.
- The detector matches only first-person capability denials, so a grounded answer that merely quotes "real-time" or a year *from the tool result* is not flagged.

`num_metal_cap_violations`-style always-on counters and the existing preamble paths are untouched; this is purely additive.

## Tests
`apps/rapid-mac/Tests/RapidTests/GroundingConfabulationRetryTests.swift`:
- A confabulated denial over a live `web_search` result triggers exactly one tools-disabled correction round, and the grounded retry becomes the final answer.
- A clean grounded answer is shipped as-is (no second round).
- Both **fail on the pre-fix loop** — verified by stashing the source change (final answer == the confabulation, `requestBodies.count == 1`).

`UngroundedRefusalDetectionTests`: parameterized positive/negative/empty coverage of `looksLikeUngroundedRefusal`, including grounded answers that mention time words but do not refuse (EN + ZH).

Full `rapid-mac` suite: **1838 tests pass**, no regression.

## Notes
Scoped to the deterministic layer; it does not (and cannot) eliminate all model confabulation, but it converts the most jarring, user-visible failure ("fetched the news, then denied having it") into a single automatic correction. The other dogfood items in the same report (web_search/weather in existing threads, denied-browse re-ask) were verified already fixed on main by #1694/#1696/#1556 and are pinned by existing tests.

🤖 Written by Claude (Opus 4.8, 1M context) at the direction of the repository owner.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
@@ -1687,23 +1687,40 @@ final class ChatViewModel {
         return history[start...].contains { $0.role == .tool }
     }
 
-    /// Like ``carriesToolResultForThisTurn`` but requires a SUCCESSFUL tool
-    /// result — a ``.tool`` row for this turn that did not fail. The grounding
+    /// Like ``carriesToolResultForThisTurn`` but requires usable current data
+    /// from one of the built-in live-data tools. The grounding
     /// correction (BUG C) asserts "the tool result above was fetched just now
     /// and IS the current data", so it must not fire when the only tool result
     /// this turn is a failed/empty one (e.g. a search that errored): there the
     /// model's "I can't get current data" answer is correct, not a
-    /// confabulation. All three built-in tools (web_search, browse, weather)
-    /// are live-data sources, so a non-failed result from any of them is
-    /// real-time evidence.
+    /// confabulation. Tool rows are joined back to their calls by ID so a
+    /// successful calculator or MCP result cannot accidentally arm this
+    /// live-data-specific retry.
     static func carriesSuccessfulToolResultForThisTurn(_ history: [ChatMessage]) -> Bool {
         let start =
             history.lastIndex { $0.role == .user }
             .map { history.index(after: $0) } ?? history.startIndex
-        return history[start...].contains {
-            $0.role == .tool
-                && $0.status == .complete
-                && !$0.content.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        let turn = history[start...]
+        let liveToolNames: Set<String> = ["web_search", "browse", "weather"]
+        var liveCallIDs: Set<String> = []
+        for message in turn where message.role == .assistant {
+            for call in message.toolCalls ?? []
+            where liveToolNames.contains(call.function.name) {
+                liveCallIDs.insert(call.id)
+            }
+        }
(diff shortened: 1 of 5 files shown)


raullenchai:
## Status for review — stopping detector-precision iteration (codex non-convergence)

All **concrete** issues codex raised across 5 rounds are fixed, including two real bugs it caught that I had introduced:
- **content destruction**: the confabulated draft is now stashed and **restored** if the correction round fails, is cancelled, or returns empty — a wrong-but-present answer beats a blank message;
- **cancel-restore**: a cancelled correction also restores the original rather than leaving a blank.

Also landed: first-person-anchored refusal detection (+ Unicode-apostrophe normalization), successful/non-empty tool-result gating, and concrete-only grounding signals.

**Why I'm not chasing the remaining codex blocking item.** The last outstanding note is detector-precision: a refusal that *embeds* a grounding token (a bare URL last round, a Markdown link this round) can slip the `answerReliesOnEvidence` check. This is non-convergent by nature — for **any** finite string signal used to recognize "grounded", an adversarial reviewer can construct a refusal that contains that signal (bare URL → Markdown link → result-phrase → …). The only convergent fix is structured/NLP sentence-level attribution, which is disproportionate for a local-model heuristic mitigation.

**Why that's acceptable here — the design is safe-by-construction:**
- **One-shot** and **non-destructive**: a false positive costs at most one extra correction round and, if that round doesn't produce a usable answer, the original is restored. It can never lose or degrade an answer.
- A false negative (a confabulation the detector misses) is **no worse than today's behavior** — without this change, *every* confabulation ships uncorrected. This strictly improves the common case.

Tests: streamed end-to-end coverage of trigger / no-retry-on-clean / no-retry-on-caveat / restore-on-empty-correction, plus unit coverage of the detectors (EN+ZH, typographic apostrophes, third-party/quoted negatives, bare-link and vague-connector refusals). Full rapid-mac suite green (one unrelated timing-flaky `DownloadCatalogHardening` test passes in isolation).

Leaving this for a human merge call rather than overriding the codex gate autonomously.

Issue #1256, opened 2026-07-27
bug(tools): tool_choice=required can synthesize schema-invalid empty arguments


raullenchai:
## Summary

With a real Qwen3.5 server, forced tool choice can return a synthetic tool call with `{}` arguments even when the selected tool's JSON schema requires fields. The response says `finish_reason: "tool_calls"`, so clients may execute a call that Rapid-MLX already knows is invalid.

## Environment

- Local wheel built from `main` at `93981cf4` (package version `0.11.0`)
- Qwen3.5-4B
- Tested both streaming and non-streaming chat completions

## Reproduction

Provide one tool:

```json
{
  "name": "add",
  "parameters": {
    "type": "object",
    "properties": {
      "a": {"type": "integer"},
      "b": {"type": "integer"}
    },
    "required": ["a", "b"]
  }
}
```

Prompt: `Call add with a=7 and b=8.`

Send the request with `tool_choice: "required"`.

## Actual behavior

Both streaming and non-streaming responses returned:

```json
{
  "finish_reason": "tool_calls",
  "tool_calls": [{"function": {"name": "add", "arguments": "{}"}}]
}
```

The server log shows that parsing produced no tool call and the required-tool fallback synthesized one:

```text
SSE-FORCED-SYNTH ... synthesizing terminal call
tool_choice='required' ... produced no tool_calls; synthesising ...
```

As a control, the same prompt and schema with `tool_choice: "auto"` produced the correct `{"a":7,"b":8}` arguments.

## Expected behavior

Rapid-MLX should never report a successful forced tool call whose arguments fail the supplied schema. It should either recover/generate valid arguments (ideally through the constrained path) or fail explicitly so the client cannot mistake the result for an executable tool call.

PR #1785, opened 2026-08-10
Discover models another MLX runtime already downloaded (#1718)

### LINKED ISSUE #1718: Models already on disk in another MLX runtime's store are invisible, so we re-download them
## The friction

Anyone who has run MLX models on their Mac before installing Rapid already has
tens of gigabytes of weights on disk. We list none of them unless they happen
to sit in the Hugging Face hub cache in the exact `models--org--repo` layout —
that is what the catalog enumerates
(`apps/rapid-mac/Sources/Rapid/Server/QuickstartModel.swift:8`):

```
``rapid-mlx ls`` — which enumerates ``<HF_HUB_CACHE>/models--*/``
```

The folder picker in Settings does not help. Read its own description
(`SettingsModelManagementPanel.swift:372-378`):

```swift
panel.message = "Pick the folder where Rapid should keep downloaded models."
```

It sets **where we download to**, and it expects hub-cache layout. Pointing it
at a model tree that another MLX runtime wrote — those typically use
`<root>/<publisher>/<repo>/` rather than `models--publisher--repo` — surfaces
nothing.

The result: a new user's first experience is downloading a model they already
have, on a machine where disk is usually the scarce resource.

## Explicitly out of scope: GGUF

We cannot serve GGUF, and listing it would offer models that fail on start.
This is recorded in `vllm_mlx/_download_gate.py:72-73`:

```
#   * ``.gguf`` — mlx-lm has *export* support (convert_to_gguf) but
#                 no load path; ``mx.save_gguf`` is one-way.
```

So any scan must identify MLX-servable repos and skip GGUF stores entirely,
rather than listing everything that looks like a model directory.

## Proposal

Identify models **by content, not by directory-name convention**:

* A directory is a candidate if it contains weights (`*.safetensors` / `*.npz`)
  plus a `config.json` or `params.json`.
* Scan a user-nominated root recursively to a sane depth, and additionally
  offer to scan well-known MLX store locations if they exist.
* Present found models as servable in place — **no copying**. Disk is the
  constraint we are trying to respect.
* Show clearly that a model is external (not managed by us), and never delete
  outside our own managed folder.

## Acceptance

* A user with an existing MLX model tree points at its root once and those
  models appear in the picker and serve.
* No GGUF-only directory is ever listed.
* Nothing is copied or moved; external models are served from where they are.
* Deleting a model from Settings can never remove files outside the managed
  folder (guard this with a test — `ModelDeletion.swift` already resolves
  `HF_HUB_CACHE`/`HF_HOME`, so the boundary check belongs next to it).
* A scan of a large tree does not block the UI.

xiaoxiunique:
## What does this PR do?

Makes models another MLX runtime already downloaded visible and usable in Rapid, read-only.

`_scan_external_model_dirs` walks the `<root>/<publisher>/<repo>/` layout other MLX tools write, from roots named by `RAPID_MLX_EXTRA_MODEL_ROOTS`. The desktop app passes the folder the user picked in Settings — which previously only set `HF_HUB_CACHE`, so pointing it at a non-hub tree surfaced nothing.

What counts as a model is decided by `_snapshot_is_complete`, the same check `serve` uses. That mirrors mlx-lm's real loader glob, so we never advertise a directory the loader would then refuse, and it excludes GGUF structurally rather than by blacklist (mlx-lm can *export* GGUF but has no load path, so listing one would offer a model that fails on start).

**External models are read-only.** Both `rm` and the app's delete path rebuild a target as `<hub-root>/models--<repo>`, which is not where these live — a delete would either silently miss or remove an unrelated hub entry of the same name. We did not download them and cannot manage them. Enforced in two places on purpose: the Settings panel omits the delete affordance (mirroring how a serving model already renders), and `ModelCacheActions.runDeletion` refuses outright, because UI omission alone is one missed `if` away from a data-loss bug.

Also fixes an independent CLI bug found on the way — see the separate commit below.

## Why is this needed?

Fixes #1718.

Anyone who ran MLX models before installing Rapid already has tens of GB of weights on disk. We listed none of them unless they sat in the HF hub cache in the exact `models--org--repo` layout, so a new user's first experience was re-downloading a model they already had — on a machine where disk is usually the scarce resource. The engine could already serve from a plain directory (`model_aliases.resolve_model` has `if os.path.exists(name): return name`); only discovery was missing.

The Settings copy already conceded the gap: *"Models downloaded by other apps in other formats won't appear here."*

## AI assistance disclosure

Claude (Opus 4.8) wrote the implementation and tests across all four commits; I directed the work, made the design calls, and verified each claim below myself.

Two design decisions were mine after the AI surfaced the constraint:
- **Read-only rather than deletable** — the AI found that `ModelDeletion` rebuilds the hub path from the repo string and never consults the model's real location.
- **Rejecting parenthesized aliases by shape, not by allow-list** — an allow-list would have admitted `(external)` as deletable, and would silently re-open the hole for whatever status the engine adds next.

The AI's first attempt was wrong in a way worth naming: it implemented "not deletable" as "not present", dropping external rows before they reached `ModelEntry`. That satisfied the safety half by abandoning the goal — the model stayed invisible in the GUI and the user still re-downloaded it. Caught by running the built app rather than trusting the CLI output, and fixed in commit 2.

Verification I actually performed:
- Every new test was checked to **fail** with the corresponding production line broken — depth cap removed, `(external)` marking removed, `_snapshot_is_complete` swapped for a lax "directory is non-empty" probe, status filter narrowed back to `(unmapped)`, delete guard deleted, scan switched back to `_dir_size_bytes`, Size column reverted to 9.
- Ran the built app against a real 633 MB model copied into an external-layout tree, and confirmed in the AX tree that the row appears with its true size and that **no delete element exists** anywhere in the window.

## Test plan

- `pytest tests/test_cli_models.py tests/test_download_gate.py tests/test_models_command_layout.py tests/test_first_run_guide.py` → 153 passed
- `swift test` (apps/rapid-mac) → 1819 of 1821 pass; the 2 failures are pre-existing (see below)
- `ruff check` + `ruff format --check` → clean
- Manual: built the app, pointed the models folder at an external-layout tree containing a real LFM2.5-1.2B-4bit copy, verified the row renders `632.7 MiB` with no delete button, and that the footer reads `Total: 663.4 MB across 1 model`
- Manual: repeated with a symlinked tree (weights linked out of the hub cache) to confirm the size is measured rather than reported as 0 B

**Pre-existing failures on `main`, confirmed by running the same tests on a clean checkout — not touched by this PR:**
- `DownloadCatalogHardeningTests` "terminates catalog subprocesses when the async task is cancelled"
- `WebToolsHardeningTests` "resolve yields empty for an NXDOMAIN reserved TLD" — needs a network that returns NXDOMAIN; behind a fake-IP proxy every name resolves
- Separately, `ls` renders `(incomplete)` for several models that serve fine (looks like `refs/main` pinning in `is_repo_cached`). Out of scope here; happy to file it.

## Commits

1. `feat(desktop): discover models another MLX runtime already downloaded` — the scanner, `RAPID_MLX_EXTRA_MODEL_ROOTS`, Swift env plumbing
2. `fix(desktop): surface external models in the UI, read-only` — the correction described above, plus the two-layer read-only enforcement
3. `fix(cli): widen the Size column so ls rows stay machine-parseable` — **independent bug, reviewable on its own.** A 9-character size like `632.7 MiB` exactly fills its 9-wide field, leaving one literal space; the desktop parser splits on 2+ spaces, so size and modified arrived glued together and the footer degraded to "(+1 unmeasured)" for a model whose size had just been printed. Needs a long repo id to surface, which is why it survived.
4. `fix(desktop): measure symlinked external models, and stop promising a delete` — `_dir_size_bytes` skips symlinks (right for the hub's blob/snapshot split, wrong for an external tree where the link *is* the model); tooltip no longer promises a delete that cannot happen

## Checklist

- [x] Tests pass locally
- [x] Self-validated with `pr_validate` (`PR_VALIDATE_NO_DEEPSEEK=1 PR_VALIDATE_NO_STRESS=1`)
- [x] New tests spot-checked to fail when the corresponding production line is broken
- [x] No breaking changes to existing API — `RAPID_MLX_EXTRA_MODEL_ROOTS` is empty by default, so behaviour is unchanged unless a root is set


xiaoxiunique:
## `pr_validate` results

```
PR_VALIDATE_NO_DEEPSEEK=1 PR_VALIDATE_NO_STRESS=1 python3 -m scripts.pr_validate.pr_validate 1785
```

| step | status | summary |
|---|---|---|
| `fetch` | PASS | 6 files, +622/-22 LOC, blast=high |
| `supply_chain` | PASS | no hooks touched, no suspicious patterns, deps clean |
| `lint` | PASS | clean |
| `full_unit` | FAIL | 15 failed, 16658 passed (see below) |
| `test_plan_check` | FAIL | stale checkbox in my own description — now fixed |
| `stress_e2e_bench` | skip | no Apple Silicon headroom here; opted out |

**Verdict line says DO NOT MERGE. Both failures are mine-to-explain, and neither is this PR's code.**

### `full_unit` — pre-existing on `main`

The 15 failures are concentrated in two files I never touched. Checked out clean `main` and ran them there:

```
$ git checkout main
$ pytest tests/test_agent_setup_config_merge.py tests/test_kv_cache_dtype_cli.py -q
8 failed, 32 passed
```

Same failures, same tests, without a line of this branch applied — `TestTomlMerge::*` (7) and `test_serve_rejects_reasoning_plus_legacy_kv_cache_quantization_bits_4`. The 27 collection errors were missing test deps in my local venv (`hypothesis`, `jsonschema`), fixed before this run.

None of the failures are in the files this PR changes. `targeted_tests` reports "3 fail (neg control unavailable)" — the negative control that would have auto-classified these as pre-existing could not run locally, which is exactly the case CONTRIBUTING describes.

Happy to file the `TestTomlMerge` breakage as its own issue if it isn't already tracked.

### Verified for this PR specifically

```
pytest tests/test_cli_models.py tests/test_download_gate.py \
       tests/test_models_command_layout.py tests/test_first_run_guide.py
→ 153 passed

swift test (apps/rapid-mac) → 1819/1821
```

The 2 Swift failures are also pre-existing on `main` (`DownloadCatalogHardeningTests` subprocess cancellation, and a `WebToolsHardeningTests` NXDOMAIN case that needs a network returning NXDOMAIN — behind a fake-IP proxy every name resolves).

### Manual verification

Built the app, pointed the models folder at an external-layout tree holding a real 633 MB LFM2.5-1.2B-4bit copy, and read the AX tree:

```
Settings.ModelManagement.Row.mlx-community/LFM2.5-1.2B-Instruct-4bit
  mlx-community/LFM2.5-1.2B-Instruct-4bit
  Model · 1.2B · 4-bit
  Help: Measured size on disk. Downloaded by another app — Rapid can't delete it.
        On disk, 632.7 MiB
  Footer: Total: 663.4 MB across 1 model

Delete elements in window: none
```

Repeated with a symlinked tree to confirm the size is measured rather than reported as 0 B.

raullenchai:
## Maintainer validation — ready to land

Reviewed and hardened on exact head `035d32333ecb431f31fce6062803f40b152024e3`.

- Codex adversarial review: **0 blocking** (1 performance nit: external trees are traversed twice; non-blocking and correctness-safe)
- Lint: clean
- Diff-aware targeted tests: **558 passed**
- Full unit: **16,915 passed**, 57 skipped, 18 deselected, 6 xfailed, 1 xpassed
- Swift: `swift build` clean
- GUI GoldenFlow: `catalog-integrity` **PASS**; external model is visible, labelled as owned by another app, and exposes no delete action. Artifacts: `/tmp/rapid-gui-golden-20260810T152023Z`
- Stress/agent matrix: all four model families completed. One stochastic qwen3.5 PydanticAI structured-output sample was 5/6 while same-run base was 6/6; exact-head isolated rerun passed **6/6 twice consecutively**. The rejected sample was a schema-invalid model generation correctly blocked with HTTP 422 by #1256, not an external-model path failure.
- GitHub: no unresolved review threads; head is mergeable.

Key maintainer fixes added during review: serve discovered paths in place; preserve managed-hub precedence; propagate selected roots to the serve child; canonical containment and race-safe scan handling; explicit managed/external provenance; safe CLI identifiers; duplicate/excluded catalog guards; two-layer read-only deletion protection; and GoldenFlow coverage.

PR #1786, opened 2026-08-10
feat(mac): add audio transcription, speech, and voice previews

### LINKED ISSUE #1714: Audio is engine-complete and unreachable from the app: the picker filters audio models out and nothing else offers them
## What's missing

The engine serves a full audio surface — `vllm_mlx/routes/audio.py` registers:

```
POST /v1/audio/speech           text  -> audio
POST /v1/audio/transcriptions   audio -> text
POST /v1/audio/translations     audio -> text (translated)
GET  /v1/audio/voices
POST /v1/audio/music
```

with `serve --enable-audio` behind it. None of it is reachable from the
desktop app. The only place the word "audio" appears in the app source is the
catalog filter that **removes** audio models from the chat picker
(`apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift:376-380`):

```swift
/// Matching the bare substring ``"[audio:"`` would let any row whose
/// … ``[audio:tts]``, ``[audio:stt]``, ``[video:gen]`` …
```

So the app knows audio models exist, correctly keeps them out of the chat
picker, and then offers the user nowhere else to go. A user who pulls a
speech model gets a model that shows up in `rapid-mlx ls` and is invisible in
the GUI.

## Proposal

An Audio surface with the two flows that carry their own weight:

* **Transcription** — drop or pick an audio file, get text, copy/save it.
  Progress while it runs; the file never leaves the machine.
* **Speech** — text in, audio out, with voice selection off
  `GET /v1/audio/voices`, playback and save.

Settings model management already has capability tabs and a reserved slot for
future ones (`SettingsModelManagementPanel.swift:396`):

```swift
/// Capability tabs — Chat / Image (/ Video, once it has aliases).
```

Audio should become one of those tabs so audio models are downloadable and
deletable like every other kind.

## Acceptance

* An audio model can be discovered, downloaded, and deleted from Settings.
* A user can transcribe a local file and copy the result without touching the
  CLI.
* A user can synthesise speech, pick a voice, and play or save it.
* Audio models stay out of the chat picker (do not regress the filter above).
* Failure to load an audio model does not prevent the rest of the app from
  working — the surface degrades to an explanatory empty state.

## Notes

Depends on how model residency is resolved: today, serving an audio model
means not serving a chat model. If multi-model residency lands first, this
gets much better; if not, the Audio surface must be honest that starting it
swaps the loaded model.

osdodo:
## What does this PR do?

  Adds a native Audio surface to the macOS desktop app with two local workflows:

  - Speech synthesis with model selection, voice loading, playback, saving, and per-voice previews.
  - Local audio transcription with file picking or drag-and-drop, progress, copy, and save actions.
  - Voice metadata showing each Qwen speaker's native language, dialect, and gender.
  - Native-language preview phrases for Chinese, English, Japanese, and Korean speakers.
  - Audio model discovery, download, status, and deletion in Settings.
  - Local conversion of M4A, AAC, and MP4 audio to 16 kHz mono WAV before transcription.
  - Clear model replacement messaging for the current single-model residency behavior.
  - Explanatory empty states when the audio runtime or a compatible model is unavailable.

  Audio models remain excluded from the Chat model picker.

  ## Why is this needed?

  Fixes #1714: **Audio is engine-complete and unreachable from the app: the picker filters audio models out and nothing else offers them.**

  The engine already provides speech synthesis, transcription, translation, voice listing, and music routes behind `serve --enable-audio`. The desktop app intentionally filters audio models out of Chat, but
  previously provided no alternative UI for discovering or using them.

  As a result, users could download an audio model that appeared in `rapid-mlx ls` but remained invisible and unusable in the GUI. This PR makes the two primary desktop audio workflows available without
  requiring the CLI, while keeping all files and inference local.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Audio/AudioClient.swift
+++ b/apps/rapid-mac/Sources/Rapid/Audio/AudioClient.swift
@@ -1,0 +1,395 @@
+import Foundation
+import AVFoundation
+
+struct AudioTranscriptionResult: Equatable, Sendable {
+    let text: String
+    let language: String?
+    let duration: Double?
+}
+
+struct SynthesizedAudio: Equatable, Sendable {
+    let data: Data
+    let contentType: String
+
+    var fileExtension: String {
+        switch contentType.lowercased() {
+        case let type where type.contains("mpeg"): return "mp3"
+        case let type where type.contains("flac"): return "flac"
+        case let type where type.contains("ogg"): return "ogg"
+        default: return "wav"
+        }
+    }
+}
+
+enum AudioClientError: Error, LocalizedError, Equatable {
+    case fileTooLarge(maxBytes: Int)
+    case unreadableFile(String)
+    case http(status: Int, message: String?)
+    case invalidResponse
+    case emptyAudio
+    case transport(String)
+
+    var errorDescription: String? {
+        switch self {
+        case .fileTooLarge(let maxBytes):
+            return "That file is larger than \(maxBytes / 1024 / 1024) MB. Choose a shorter recording."
+        case .unreadableFile(let detail):
+            return "The audio file couldn't be read: \(detail)"
+        case .http(let status, let message):
+            return message ?? "Audio request failed (HTTP \(status))."
+        case .invalidResponse:
+            return "The audio server returned an unreadable response."
+        case .emptyAudio:
+            return "The audio server returned no sound."
+        case .transport(let detail):
+            return detail
+        }
+    }
+}
+
+/// Native loopback client for the OpenAI-compatible audio routes. Port and
+/// bearer are supplied per request because both rotate when ServerManager
+/// replaces the single resident model.
+struct AudioClient {
+    static let maxUploadBytes = 25 * 1024 * 1024
+    static let requestTimeout: TimeInterval = 30 * 60
+
+    static let sharedSession: URLSession = {
+        let config = URLSessionConfiguration.ephemeral
+        config.timeoutIntervalForRequest = requestTimeout
+        config.timeoutIntervalForResource = requestTimeout
+        return URLSession(configuration: config)
+    }()
+
+    var session: URLSession = AudioClient.sharedSession
+
+    private struct TranscriptionWire: Decodable {
+        let text: String
+        let language: String?
+        let duration: Double?
+    }
+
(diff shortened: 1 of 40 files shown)


raullenchai:
Rebased onto `main` and landed maintainer fix-ups. Summary of changes on top of the original branch:

**Rebase / hygiene**
- Rebased 8 commits onto current `main`; resolved conflicts in `ModelCatalog.swift` (kept both the `(external)`-model handling from #1785 **and** this PR's token-split parser + `knownRepo` threading) and `ModelCacheActions.swift` (kept the `isExternal` deletion guard + widened the delete test-seam to carry `knownRepo`).
- Dropped the committed `uv.lock`. `main` doesn't track it, nothing in the build consumes it, and `release_fleet.py` treats a committed lock as fleet-selection input ("Developer-only untracked lockfiles must not make the same candidate select different fleets") — committing it would have changed release-fleet selection and tripped the stress-bench environment gate.

**Fix: audio must replace the process, not residency-load** (`2fa677`)
- Codex review caught a P1: switching to an audio model while a chat/image model was already resident failed *every* audio operation. `ensureServing` tries the in-process residency `/v1/models/load` path when a model is `.ready` and only falls back to stop/start on a 404/405 — but the engine's residency loader raises for the `audio`/`video-gen` modality, which the route maps to **HTTP 500**, so the attempt became a hard failure instead of a process swap.
- Audio needs its own `serve <alias>` audio-mode process and can never be an in-process resident. Audio callers now pass `residencyEligible: false`, sending them straight to the process swap. Added `ServerManager.residencyLoadApplies` (pure, `nonisolated static`) with a regression test. Chat/image behavior is unchanged (they keep `residencyEligible: true`, which reduces to the original condition).

**Verification**
- `swift build` + `smoke.sh` green; RapidTests compiles clean (RapidUXTests needs the CI Xcode toolchain for XCTest).
- 20/20 Python audio tests pass; `ruff check` + `ruff format --check` clean.
- Confirmed the diagnosis against the engine: `_load_dynamic_resident_model` raises for `modality in ("video-gen", "audio")` → `/v1/models/load` 500; `serve <audio-alias>` auto-routes to `_serve_audio_mode`.

Excellent, thorough work overall — thanks @osdodo. The audio surface, supply-chain-conscious sidecar trimming, and test coverage are all first-rate.

PR #1790, opened 2026-08-10
Wire MCP through to the desktop app (#1716) — reviewed + hardened

### LINKED ISSUE #1716: MCP is engine-complete but the app only passes --mcp-config through once at startup
## Current state

The engine has a complete MCP implementation:

```
vllm_mlx/mcp/client.py     config.py    executor.py
             manager.py    security.py  tools.py     types.py
```

plus a `--mcp-config` serve flag and `routes/mcp_routes.py`.

The desktop app's entire relationship with it is one flag passed at spawn
(`apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift:102`):

```swift
/// a restart: the child already received its one-time ``--mcp-config``
```

So a user who wants MCP tools in the desktop app must hand-author a JSON file
somewhere the app happens to read, and restart the server to change anything.
There is no way to see which servers connected, which tools they exposed, or
whether a tool call succeeded.

## Why the app is the right place for this

MCP servers are arbitrary local processes that the model can invoke. The
consent decision — *should this model be allowed to call this tool* — is a
user decision that belongs in a UI, not in a JSON file.

We already have the pattern in-app. `Sources/Rapid/Tools/BrowseApprovalStore.swift`
implements per-action approval for the built-in browse tool, and
`BrowseSSRFGuard.swift` shows the shape of "capability with a guard in front of
it". MCP tool approval should reuse that model rather than invent a second one.

## Proposal

1. **Server management in Settings** — add, edit, remove, enable/disable MCP
   servers; show connection state and the error when one fails to start.
2. **Tool inventory** — list the tools each server exposes, with their
   descriptions, and let the user disable individual tools.
3. **Per-tool consent** — first invocation of a tool prompts; the decision is
   remembered per tool (with a way to review and revoke), matching
   `BrowseApprovalStore`'s semantics.
4. **Apply without a restart** — or, if a restart is unavoidable for now, say
   so in the UI instead of silently ignoring the edit.
5. **Visibility in chat** — an MCP tool call should render like the built-in
   tool calls already do (`ChatView.swift:864` renders `message.toolCalls`).

## Acceptance

* A user can add an MCP server, see its tools, and use one in chat without
  editing a file by hand.
* Disabling a tool prevents the model from calling it, verified by a test.
* A tool call that was never approved does not execute.
* A misbehaving or unstartable MCP server surfaces an actionable error and does
  not take the chat surface down with it.

## Related

There is an open hardening question about tool-call arguments being able to
smuggle invocations of *declared* tools past schema filtering. Per-tool consent
narrows the blast radius but does not close that hole; the two should be
designed together.

raullenchai:
Supersedes #1787 (@Jevin-F). Carries **Jevin-F's two original commits verbatim** (`de77abd3`, `5eb42e8e`) and adds one review-follow-up commit on top, so authorship is preserved. Closes #1716.

## Why a new PR

#1787 is from a fork; the review turned up concrete fixes that had to be committed on top, which can't be pushed to the fork branch. This branch = Jevin-F's work + the fixes, landed together with credit intact.

## What the base PR does (unchanged)

Wires the engine's complete-but-unreachable MCP implementation through to the rapid-mac desktop app: a Connectors settings panel (add/edit/remove servers, per-tool switches, per-tool consent reusing `BrowseApprovalStore`'s shape), `--mcp-config` launch gating (off by default), hot-reload without a model restart, and engine-side resilience (non-fatal init, tolerant per-entry config load, stderr capture, `POST /v1/mcp/reload`).

## Review follow-ups (the added commit)

An adversarial review (independent agent + codex) of the fork PR found real defects in the surface it introduced. All fixed here:

- **Server-side sandbox was inert on the desktop path.** `/v1/mcp/execute` calls `manager.execute_tool` directly, never through `ToolExecutor`, so the default-deny on high-risk tools (shell/exec/eval), the argument scrub, and the `allowed_high_risk_tools` allowlist that `_start_mcp` wires up were never consulted — the UI approval click was the *only* gate. The route now validates against the sandbox before dispatch, on the same `(server, tool)` split it runs, via a new `resolve_tool_target`.
- **Connectors master switch is now the point-of-truth gate.** Turning it off clears the catalog, but the running child keeps its connectors loaded, so a later `/healthz` ready transition could repopulate the catalog and re-advertise the tools. `MCPToolRegistry` now gates both advertise and dispatch on the switch, so "off" holds across repopulation.
- **Captured-stderr temp file leaked** one fd per stdio server per reload; `disconnect()` closes it now, and a reconnect closes the prior handle.
- **`MCPCatalog.reload()` returned success even when the tool re-fetch failed**, leaving stale tools advertised; it now reports the refresh outcome.
- **Imported configs weren't name-validated** (write-path only), so `my.server` was forwarded to the engine and its tools became silently uncallable; `load()` surfaces it.
- **`_start_mcp` could orphan child processes** on a mid-start failure; it builds locally and stops the half-started manager before re-raising.
- **`reload_mcp` is serialized** by an asyncio lock against concurrent reloads.
- **Settings rows scrub server-supplied text** through the same `displaySafe` the approval sheet uses.

## Tests

- Python: `239 passed` across the MCP-related suites; new regression coverage for the sandbox gate (block high-risk / allow benign / honor allowlist), the fd close, and the reload-refresh outcome. Fixes a pre-existing test that mocked `load_mcp_config` without the new `tolerant` kwarg.
- Swift: full rapid-mac suite **1869 tests passed**; new coverage for the master-switch gate (advertise + dispatch even with a populated catalog) and the imported-invalid-name path.

## AI disclosure

Reviewed and the follow-up fixes authored with Claude Code (Opus 4.8). The base feature is Jevin-F's.

### DIFF
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -15,6 +15,19 @@ can actually understand.
 
 ### Added
 
+- **Connectors: give the model tools from MCP servers, from Settings.** A new
+  **Settings → Connectors** panel adds, edits and removes MCP servers, shows
+  whether each one connected — and the reason when it didn't — and lists the
+  tools it exposes with an on/off switch each. The first time the model calls
+  one, Rapid asks: **Allow once / Always allow / Don't allow**, showing which
+  connector it came from and exactly what arguments it was given. "Always
+  allow" applies to that one tool, and every remembered answer can be reviewed
+  and reset. Connectors are off by default, edits apply without restarting the
+  model, and a connector that won't start now reports why instead of taking
+  the whole local server down with it. Previously this was command-line only:
+  you had to hand-write `~/.config/rapid-mlx/mcp.json` and had no way to see
+  what connected or what it was doing.
+
 - **A new Images tab generates pictures on your Mac.** Choose an image model,
   follow the readiness prompt to load it, type a prompt and generate. Each
   render joins a filmstrip below the picture; selecting an earlier one brings
--- a/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
@@ -183,6 +183,21 @@ final class ChatViewModel {
         tools.definitions.filter { !disabledTools.contains($0.function.name) }
     }
 
+    /// Just the built-in tools, for Settings → Tools.
+    ///
+    /// Issue #1716: since the registry became a ``CompositeToolRegistry``,
+    /// ``tools/definitions`` also carries connector tools. Those get their own
+    /// switch in Settings → Connectors, backed by a different defaults key
+    /// (``MCPToolRegistry``). Listing them in both panels would give one tool
+    /// two independent off switches — flip the wrong one and the tool stays
+    /// live with no indication why. Each surface owns exactly its own set.
+    ///
+    /// Falls back to the whole registry when it isn't a composite, which is
+    /// what unit tests and the dev-snapshot harness construct.
+    var builtinDefinitions: [ToolDefinition] {
+        (tools as? CompositeToolRegistry)?.builtin.definitions ?? tools.definitions
+    }
+
     // MARK: - Conversation history (M3)
 
     /// Snapshot the active conversation into ``conversations`` + disk. A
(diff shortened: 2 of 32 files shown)


raullenchai:
Superseded: the same review-follow-up commit was pushed directly onto #1787's branch (maintainer edits allowed), so the fixes land on the original PR with @Jevin-F's authorship intact. Closing this in favor of #1787.

PR #1791, opened 2026-08-10
feat(models): Muse Glimmer support — ATEM tool parser, recipient-channel reasoning, aliases

raullenchai:
## Why

Meta released Muse Glimmer 30B today (2026-08-10, Apache 2.0) and pitched it directly at local agent workflows — its benchmark table targets Gemma4-31B and Qwen3.6-27B, and Ollama shipped day-0 support marketed at Claude Code/Codex local agents. Its tool-call wire is a format no existing parser handles: an Anthropic-style `<atem:function_calls>` XML block routed over Harmony-style channels (`<|start|>assistant to=…<|message|>…<|eot|>`), with reasoning on a `to=self` channel. Landing the parsers now means we work the day mlx-lm ships the `muse_glimmer` architecture (not on their `main` yet; Meta says MLX integration lands "in the coming days").

## What

- **`MuseToolParser`** (`--tool-call-parser muse`, wire label `muse_atem`): schema-typed bare-string/JSON-container values per the model's own `chat_template.jinja`. The template warns its output "is not expected to be valid XML", so parsing is a **positional structural scan** — a closer only closes when its lookahead is a legal grammar continuation, and nothing inside a parameter value is structural. Literal openers/closers of every level survive in values; every irreducible ambiguity degrades to visible content, never a wrong or truncated call. In-flight scan distinguishes INCOMPLETE (hold) from MALFORMED (resume; bytes stay content), with canonical-byte openers so the two classifiers stay prefix-stable under streaming.
- **`MuseReasoningParser`**: `to=self` → `reasoning_content`; tool/user segments → content with the ATEM block passing through (harmony division of labor). Stateless per-delta phase detection; holds in-flight headers, partial sentinels, and undecidable trailing terminators; `finalize_streaming` releases holds at end of stream. Literal channel tokens mid-prose survive in both modes.
- `model_auto_config` pattern for `muse[-_]glimmer` HF paths; parity fixture; **54 dedicated tests**, stream + non-stream, mostly per-character.

## Deliberately NOT in this PR

- **Aliases are deferred** (codex r6): publishing `muse-glimmer-30b-*` before mlx-lm can load the arch would list a model that downloads ~20 GB then fails at startup (the #1603/#1729 class). A follow-up re-adds them with an e2e smoke against real weights once upstream lands. Until then, an upgraded mlx-lm + explicit HF path picks the parsers up via auto-config.
- DFlash drafter wiring (`supports_dflash`) and the multimodal perception encoder — follow-ups after e2e.

## Verification

- `tests/test_muse_parsers.py`: **54 passed**; affected suites (wire-formats, streaming parity, `tests/parsers/`, alias contract, postprocessor): **3026 passed, 1 skipped, 6 xfailed** on the mlx-free lane (py3.14).
- `ruff format` + `ruff check` clean.
- **7 rounds of adversarial codex review** (gpt-5.6-sol via pr_validate), all substantive findings fixed with regression tests pinning each one; the only standing decline is per-delta full-rescan cost, which is the deliberate repo-wide stateless-parser pattern (gpt-oss/DSML/harmony) — codex itself downgraded it to NIT.
- **Not verified**: end-to-end generation — blocked on mlx-lm's `muse_glimmer` support. Wire shapes are pinned against the HF repo's `chat_template.jinja`; must be re-verified against real model output before the aliases ship.

### DIFF
--- a/tests/test_muse_parsers.py
+++ b/tests/test_muse_parsers.py
@@ -203,13 +203,52 @@ def test_fake_opener_inside_value_filtered_by_schema():
     assert json.loads(result.tool_calls[0]["arguments"]) == {"text": value}
 
 
-def test_null_for_nullable_non_string():
+def test_null_only_when_schema_allows_it():
+    # Codex r4 #1: null -> None ONLY for nullable schemas; against a
+    # non-nullable type the raw string survives so strict validation
+    # rejects it visibly instead of receiving a forged None.
     parser = _tool_parser()
-    text = _block(_invoke("f", {"limit": "null"}))
+    text = _block(_invoke("f", {"limit": "null", "count": "null"}))
     result = parser.extract_tool_calls(
-        text, request=_request("f", {"limit": {"type": "integer"}})
+        text,
+        request=_request(
+            "f",
+            {
+                "limit": {"type": ["integer", "null"]},
+                "count": {"type": "integer"},
+            },
+        ),
     )
-    assert json.loads(result.tool_calls[0]["arguments"]) == {"limit": None}
+    assert json.loads(result.tool_calls[0]["arguments"]) == {
+        "limit": None,
+        "count": "null",
+    }
+
+
+def test_duplicate_name_fake_opener_cannot_overwrite_value():
+    # Codex r4 #2: a fake opener REUSING the declared name must not
+    # become a second parameter that overwrites the real value.
+    parser = _tool_parser()
+    value = 'real start <atem:parameter name="text">evil</atem:parameter> real end'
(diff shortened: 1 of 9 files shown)


raullenchai:
Scope change after codex r6: the two `muse-glimmer-30b-*` aliases are **deferred** — publishing them before mlx-lm ships the `muse_glimmer` architecture would list a model that downloads ~20 GB and then fails at load (the #1603/#1729 class). The parsers and the `model_auto_config` pattern remain, so the day mlx-lm lands, serving the mlx-community repo by explicit HF path works immediately; a follow-up PR re-adds the aliases with an e2e smoke against real weights.

PR #1803, opened 2026-08-10
Add --disk-stream: stream MoE routed-expert weights from disk

MIt9:
## Summary

Adds `--disk-stream` — an opt-in flag for `rapid-mlx serve`/`bench` that streams MoE routed-expert weights straight off the safetensors checkpoint per forward pass instead of holding them all resident in RAM. Lets an operator run a model whose declared `min_memory_gb` floor (see `_check_alias_min_memory` in `cli.py`, e.g. `hy3-preview-4bit`'s 192GB) exceeds their Mac's actual RAM. Existing behavior is completely unaffected when the flag isn't passed.

## Research: what was verified before writing any production code

Before touching `vllm_mlx/`, the approach was proven end-to-end in a standalone local spike (not part of this repo) against real downloaded checkpoints, in three rounds:

**Round 1 — core mechanism (LFM2.5-8B-A1B, routed-only MoE, no shared experts).** `mlx_lm.load(..., lazy=True)` skips the final `mx.eval(model.parameters())`, so every parameter — including the whole MoE expert stack — stays an unevaluated lazy graph node until something forces materialization. Combined with a class-level monkeypatch of the MoE block's `__call__` (instance-level overrides don't intercept `obj(...)` in Python) that fetches only the selected experts' weights fresh off disk via safetensors byte-offset reads, the resident expert stack is never materialized at all. Verified empirically: `mx.get_peak_memory()` immediately after lazy load, before any forward pass, is ~0 bytes. End-to-end: **4.817GB → 0.955GB peak memory**, generated token ids **exact match** against a normal full-load run (32/32 tokens, greedy decoding, fixed seed).

**Round 2 — throughput.** Streaming re-fetches every expert from disk on every token even if it was just used — 14,520 disk reads / 43.4s for a 32-token generation. Added a byte-budget LRU cache (`ExpertCache`) between the streaming forward and the disk reader: **43.4s → 12.5s** (~3.5x), 87.9% hit rate, peak memory bounded by the configured cache budget and still far below the full-load baseline (1.41GB vs 4.82GB). Also tried batching the per-expert-bundle disk reads into a single `open()` — correctness held, but at that hit rate the miss path was already too small a fraction of total time to produce a measurable end-to-end speedup, so it wasn't worth carrying into this PR.

**Round 3 — generalizing beyond routed-only MoE.** Real-world MoE checkpoints often also have a resident, always-active shared-expert term (LFM2.5 doesn't have one). Verified the mechanism generalizes to that shape too, using Qwen1.5-MoE-A2.7B (`qwen2_moe` architecture — chosen over the original shared-expert target, `hy_v3`/Hunyuan-3, because that model's only available checkpoint declares a 192GB memory floor and simply can't be downloaded/run for verification on ordinary hardware): stream only the routed experts, leave `shared_expert`/`shared_expert_gate` on the normal resident path (they're active every token regardless, so streaming them buys nothing). **8.603GB → 1.959GB peak memory**, 32/32 token-exact.

Full spike write-up, scripts, and raw numbers available on request (kept outside this repo since it was throwaway exploration, not shippable code).

## Approach used to turn the spike into this PR

- **Architecture-registry design**, not a hardcoded single-model patch. `vllm_mlx/registry.py` maps a checkpoint's real `model_type` (as read from its own `config.json`) to a `StreamingAdapter`: which MoE block class to patch, a tensor-naming template describing that architecture's on-disk expert-weight layout, and which streaming-forward function reproduces its exact combination math. Two layouts are implemented in `vllm_mlx/offset_reader.py`: `"stacked"` (LFM2.5-style — every expert stacked on axis 0 of one tensor, sliced by byte offset) and `"direct"` (qwen2_moe-style — each expert is its own separately-named tensor, possibly sharded via `model.safetensors.index.json`). Adding architecture #3 is one new registry entry (plus a new streaming-forward module only if its combination math genuinely differs) — zero changes to the cache, the patch installer, or the CLI.
- `vllm_mlx/expert_cache.py` — the round-2 LRU cache, ported with no behavior changes.
- `vllm_mlx/disk_stream_patch.py` — the glue: looks up the adapter, wires the cache, installs the class-level monkeypatch across every MoE layer. Raises immediately and clearly for an unregistered `model_type` (no silent fallback to resident loading, no confusing crash mid-forward). Also exposes `is_installed()`/`uninstall()`, following this repo's own `deepseek_v32_indexer_gate.py` precedent for class-level monkeypatches.
- CLI wiring (`cli.py`, `server.py`, `engine/batched.py`, `utils/tokenizer.py`) hooks `--disk-stream` into the real `serve`/`bench` boot path (`load_model_with_fallback(..., lazy=True)` → `disk_stream_patch.install()`, before the model reaches `AsyncEngineCore`). The existing `_check_alias_min_memory` RAM-floor warning still fires unchanged — this doesn't suppress it, since resident components (attention, KV cache, the shared-expert term, the cache budget itself) still consume real RAM.

## Testing

Followed this repo's existing conventions throughout rather than inventing new ones:

- Deep, isolated modules (`registry`, `offset_reader`, `expert_cache`) get unit tests with no model/checkpoint download required, run in the default `pytest tests/` suite.
- Two `@pytest.mark.slow` end-to-end tests (one per architecture), gated by the existing `--run-slow` convention in `tests/conftest.py`, each asserting token-exact output vs. non-streaming generation and measurably lower peak memory against a real downloaded checkpoint.
- A wiring-regression test for both `serve` and `bench` following this repo's own established pattern for class-level monkeypatch installers (`tests/test_deepseek_v32_indexer_gate.py::test_install_fires_on_real_serve_import_path`, written after PR #967 taught this repo that a monkeypatch installer can pass every test while never actually running on the real boot path) — a subprocess imports the real serve-path module (never the patch module directly) and asserts the install actually fired. Verified with a negative control: temporarily broke the real wiring, confirmed the test fails; restored it, confirmed the test passes again.

An independent file-by-file code review of the full diff (registry+offset_reader, expert_cache, patch+streaming-math, CLI wiring — reviewed as four separate passes) found and this PR's history fixes:

- **Path traversal** in the sharded-checkpoint tensor resolver — an adversarial/malformed `model.safetensors.index.json` could point a tensor name at a file outside the checkpoint directory.
- **The lazy-load path was skipping architecture-independent tokenizer/EOS post-load fixups** that a real checkpoint needs to stop generation at the correct token (confirmed via an empirical eager-vs-lazy tokenizer-state diff on both architectures after the fix).
- A directory-vs-file path-resolution bug in the `"stacked"` layout, found by actually installing the package and hitting a live `rapid-mlx serve --disk-stream` with a real request rather than relying on pytest alone (the tests happened to always pass a concrete file path, while the real CLI resolves a checkpoint to its HF-cache snapshot *directory*).
- Missing bounds-checking, missing error-path test coverage, redundant per-tensor header re-parsing (9x → 1x per expert fetch), and a `mx.get_peak_memory()` test-isolation issue where the two slow e2e tests polluted each other's peak-memory readings when run in the same session (fixed properly via `uninstall()`, not just a `reset_peak_memory()` call — the class-level patch was keeping its cache alive via closure even after the test's local references went out of scope).

Both slow e2e tests pass together in a single `pytest` session after these fixes. Manually verified end-to-end via a real running `rapid-mlx serve --disk-stream` instance for both architectures (not just pytest) — real `curl` requests against `/v1/chat/completions` returned correct completions, including one that reproduced a known reference generation from the spike verbatim.

## Test plan

- [x] `pytest tests/` (default suite, no model download) — new tests pass, zero regressions on unmodified files
- [x] `pytest tests/ --run-slow -m slow -k disk_stream` — both e2e tests pass, together and individually
- [x] Manual: `rapid-mlx serve <model> --disk-stream` for both `lfm2_moe` and `qwen2_moe` checkpoints, real chat-completion requests, correct output
- [x] Manual: `--disk-stream` against an unregistered architecture fails clearly at load time, not mid-generation
- [x] Negative control on the wiring-regression test (fails when the real wiring is broken, passes when it isn't)


raullenchai:
Maintainer follow-up pushed in 2565494a after reproducing and adversarial review. It fixes the blocking findings: reject zero-layer and duplicate class-level installs; validate a positive finite cache budget; keep oversized expert bundles out of the LRU so the byte bound is hard; use each projection’s own quantization parameters; contain shard symlink targets while allowing repository-local Hugging Face blobs; and update to Ruff 0.16 formatting. Added regression coverage for each guard. Local validation: Ruff check/format full repo green; focused disk-stream suite 55 passed / 4 slow deselected; full pytest 17,070 passed with only the two live-server tests (no server on :8000) and one order-dependent existing routes pollution failure, which passes in isolation.

raullenchai:
One remaining merge blocker from actually running the advertised slow suite on a clean checkout: both e2e tests resolve their golden token/memory baselines from `Path(__file__).parents[2] / .scratch/moe-disk-stream/...`, i.e. outside this repository. Those files are not in the PR, so both tests skip even when the LFM checkpoint itself is locally cached (`2 skipped` here). Please attach/provide the two baseline JSON payloads from the spike (or commit sanitized fixtures under `tests/fixtures/disk_stream/` and point the tests there). Without that, fresh clones and CI cannot reproduce the claimed token-exact + lower-peak-memory guarantee for this high-risk execution path.

PR #1802, opened 2026-08-10
feat: vendored Muse Glimmer 30B text backbone (native, no mlx-vlm needed)

raullenchai:
## Why

Meta's Muse Glimmer 30B is the hottest open-weights release of the cycle and Ollama shipped day-0 support on their in-house runtime, marketing exactly our positioning (local agents). Our ATEM tool-call + reasoning parsers landed in #1791, but the model itself could not load: mlx-lm has no `muse_glimmer` support and mlx-vlm's port is an unmerged draft (Blaizzy/mlx-vlm#1838). This PR vendors the text backbone so `rapid-mlx serve muse-glimmer-30b-4bit` works today on the standard mlx-lm text lane (BatchGenerator, prompt cache, quantized KV) — chat + tools + agents, text-only. The 1.8B vision tower is dropped at `sanitize`; multimodal routing returns automatically once a future mlx-vlm release ships the arch (find_spec probe, zero code change).

## What

- **`vllm_mlx/models/muse_glimmer.py`** (new): mlx-lm-style architecture module. 52 layers `[sliding ×3, full]` (window 2048), NoPE on full-attention layers, weightless QK-norm + `qk_scale_factor` 3.87 on Q, Afmoe-style sigmoid output gate, centered `1+w` sandwich norms (dual eps 1e-5/1e-8), post-lookup embedding norm (embedding table deliberately unquantized in checkpoint), plain-scale final norm, `output_multiplier` + tanh softcap 20. Weight sanitization strips the `language_model.` wrapper and drops `vision_tower.`/`vision_adapter.`/`vision_projection.`/`multi_modal_projector.`.
- **Registration** (`utils/tokenizer.py`): `sys.modules["mlx_lm.models.muse_glimmer"]` injection in `_register_vendored_archs()` — hy_v3 pattern, incl. find_spec deference to future native mlx-lm support and the `_VENDORED_MODEL_TYPES` promotion-on-success discipline.
- **Lane routing** (`api/utils.py`, `server.py`): new offline probe `mllm_arch_unsupported_but_text_vendored()` — a genuinely multimodal checkpoint whose arch the installed mlx-vlm cannot serve auto-downgrades to the text lane instead of crashing at MLLM load. Generalized the auto-downgrade log message (it hardcoded the hybrid-backbone cause).
- **Aliases**: `muse-glimmer-30b-4bit` / `-bf16` with `is_text_only` state-pin (#393 mechanism), muse parsers, `is_hybrid: false` explicit (#1764 lesson).
[...]

PR #1809, opened 2026-08-10
fix(muse): close the 5 red gates #1802 left for Muse Glimmer

raullenchai:
## Why

#1802 vendored the **Muse Glimmer 30B** text backbone and registered the `muse` tool/reasoning parsers + aliases, but **merged with five muse-only tests red**. Two are real defects; three are unregistered coverage. The model itself serves fine — real 2-turn agent loops on `/v1/chat/completions`, `/v1/responses` (Codex), and `/v1/messages` (Claude Code) all round-trip tool calls cleanly (verified on a live `muse-glimmer-30b-4bit` boot, 2026-08-10) — so these are **completion gaps, not wire bugs**. But two would bite on M5 hardware / in the disconnect-guard path.

Single root cause: muse was half-wired into the cross-cutting gates.

## What

| Fix | File | Red test it closes |
|---|---|---|
| **Real #404 M5 bug** — module did `from mlx_lm…` at import time without calling `_mlx_compat.install()` first, so mlx_lm's import-time thread-local stream capture is unusable on single-stream M5 GPUs. Install the idempotent shim before the imports (deepseek_v4 / hy_v3 pattern; no-op where the original API works). | `models/muse_glimmer.py` | `test_mlx_compat::test_every_mlx_lm_consumer_installs_shim` |
| **Real AttributeError** — #1802 added `_muse_wire_model()` (reads the `_is_muse_wire` flag `__init__` seeds) into the finished-chunk clean path. This test builds an engine via `__new__` (skips `__init__`) and hand-seeds attrs; it lacked `_is_muse_wire` → the muse gate raised. Seed it `False` (the fake engine is not a muse model). | `test_disconnect_guard_aborts_scheduler.py` | `test_batched_engine_publishes_request_id_into_holder` |
| Two muse aliases were absent from the size manifest. Add `Muse-Glimmer-30B-4bit` (19443239525 B, from `estimate_repo_size_bytes`) and `-bf16` (`null` — the mlx-community bf16 repo currently ships only README + .gitattributes, **no weight blobs**, so the size is genuinely unresolvable; `null` is the manifest's documented "unknown"). | `vllm_mlx/model_sizes.json` | `test_model_sizes::test_every_text_alias_has_a_manifest_entry` |
| Register `muse` in `MATRIX_EXEMPT` with a documented reason (verified e2e + 6th family in the tests/integrations agent matrix; parser/model unit coverage; golden_models boot TODO'd on the pr_validate budget) — the lfm/hy_v3 pattern. | `scripts/audit_tool_parser_coverage.py` | `test_tool_parser_coverage::test_tool_parser_matrix_coverage_clean` |
| Add `muse` to `_FIDELITY_EXEMPT` (muse_atem wire, renderer TODO) — the hy_v3 pattern. | `tests/test_tool_call_value_fidelity.py` | `test_tool_call_value_fidelity::test_every_parser_is_covered_or_exempt` |

## Tests

- All **five** previously-red tests now PASS.
- `test_muse_parsers.py` (59) and `test_muse_glimmer_model.py` (15) stay green — the import reorder does not perturb the arch (numeric parity, cache, sanitize all still pass).
- `ruff check` + `ruff format` clean. **No version bump.**

## Notes

Companion to #1808 (adds muse as the 6th family to the agent-gate integration matrix). This PR restores a green `main` for the muse coverage/regression gates #1802 left red. The bf16 alias's empty upstream repo (README-only) is noted but out of scope — the alias resolves and the manifest records it honestly as unknown.

🤖 This PR was written by Claude Code (Opus 4.8).

### DIFF
--- a/scripts/audit_tool_parser_coverage.py
+++ b/scripts/audit_tool_parser_coverage.py
@@ -132,6 +132,19 @@ MATRIX_EXEMPT: dict[str, str] = {
         "tests/test_tool_call_streaming_parity.py."
     ),
     "liquid": "alias of lfm",
+    "muse": (
+        "TODO: add mlx-community/Muse-Glimmer-30B-4bit to golden_models "
+        "when the pr_validate boot budget has a slot for the 30B (~18 GB, "
+        "min_memory_gb=24 — feasible, unlike the Ultra-only Hy3/DeepSeek-V4 "
+        "entries). Wire format (ATEM <atem:function_calls> blocks on "
+        "recipient-routed to=<tool> channels) is verified end-to-end another "
+        "way in the meantime: real 2-turn agent loops on a live "
+        "muse-glimmer-30b-4bit boot over /v1/chat/completions, /v1/responses "
+        "(codex), and /v1/messages (claude-code) all round-trip tool calls "
+        "cleanly (2026-08-10), and muse is the 6th family in the "
+        "tests/integrations agent-gate matrix. Class + parser coverage lives "
+        "in tests/test_muse_parsers.py and tests/test_muse_glimmer_model.py."
+    ),
 }
 
 
--- a/tests/test_disconnect_guard_aborts_scheduler.py
+++ b/tests/test_disconnect_guard_aborts_scheduler.py
@@ -663,6 +663,11 @@ async def test_batched_engine_publishes_request_id_into_holder():
     eng._stream_interval = 1
     eng._is_hybrid_model = lambda: False
     eng._create_output_router = lambda: None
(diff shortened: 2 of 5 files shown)


raullenchai:
Closing as a duplicate — #1806 (fix(muse): restore full-unit validation after #1802) landed on main while this was being prepared and already fixes all five red gates, some more robustly (e.g. the `_muse_wire_model` AttributeError is fixed code-side in `batched.py` via `getattr` + a `_model_name is None` guard, protecting every partial fixture rather than just the one test). Verified main @60abd986 is green on all five. No unique fix here worth salvaging.

PR #1498, opened 2026-08-05
fix(models): retire broken Ministral 3B alias

raullenchai:
## What does this PR do?

Removes the public `ministral-3b-4bit` alias and its catalog metadata so `rapid-mlx models` and Rapid Desktop no longer advertise a checkpoint whose default multimodal route hangs on the first text completion. Historical benchmark evidence stays in `evals/results`; current catalog-derived surfaces and release notes are updated.

The retired short alias now fails fast on stderr before download or model startup. The raw Hugging Face path remains accepted for explicit text-only investigation, local-path precedence remains intact, and its size metadata is retained for disk/download preflight.

## Why is this needed?

Refs #1367. Revalidation on current `main` reproduced the Ministral failure with both supported mlx-vlm lines: 0.6.3 cannot load `PixtralProcessor` without PyTorch/Torchvision, while 0.6.10 reaches ready and then loops on `Model.__call__() missing 1 required positional argument: 'mask'` until the first completion times out. Users selecting the advertised alias therefore get a server that cannot answer.

Gemma e2b was separately revalidated and remains supported: 48/48 coherence cases passed on M2 Pro and 42/42 on M3 Pro across mlx-vlm 0.6.3/0.6.10; original v0.11.4 source also passed 30/30 on M3 Pro. The historical M1 discrepancy remains tracked in #1367 and is not conflated with this removal.

## AI assistance disclosure

Codex performed the reproduction, implementation, and iterative adversarial review. Review converged to zero findings. The maintainer-directed validation used real M2 Pro and M3 Pro servers, the repository coherence gate, Python and Swift tests, lint, GitHub CI, and `pr_validate`.

> By submitting this PR I confirm I can explain the intent, risk, and behavior of every non-generated change in this PR. For generated or historical scorecard content, I can describe how its catalog consistency was verified.

## Test plan

- [x] Reproduced the default Ministral VLM failure with mlx-vlm 0.6.3 and 0.6.10.
- [x] Verified `--no-mllm` is coherent (18/18) and the retired short alias exits 1 before startup.
- [x] Ran affected alias/catalog/CLI suites: 3,059 passed.
- [x] Ran `ToolUseCapabilityCatalogCoverageTests` through SwiftPM.
- [x] Ran final diff-aware suite: 827 passed.
- [x] Ran final full pytest suite hermetically: 15,586 passed, 53 skipped, 6 xfailed, 1 xpassed.
- [x] Ran Ruff check/format and workflow expression/action-pinning checks.
[...]

PR #1815, opened 2026-08-11
fix(mllm): fail closed on non-Value/Runtime generation-step errors (#1367)

raullenchai:
## Why

`MLLMScheduler._step_no_queue` runs the model forward pass via `batch_generator.next()` inside a narrow `except (ValueError, RuntimeError)` that fails all running requests when the batch can't proceed. A checkpoint whose mlx-vlm model class drifts from the batch generator's call convention raises **`TypeError`** from the forward pass — the #1367 repro is `Model.__call__() missing 1 required positional argument: 'mask'` on `ministral-3b`.

`TypeError` is neither `ValueError` nor `RuntimeError`, so it escaped the handler, propagated to `_process_loop`'s outer `except Exception`, and the loop **logged + retried the same poisoned batch every step** (~565 identical `Error in MLLM process loop` lines in one 90s client request). The waiting request's `output_queue` never received an output or the `None` sentinel, so from the caller it was an **unbounded silent hang** (`/v1/status` stuck at `steps_executed: 0, total_requests_processed: 0`), not a slow response or an error.

This is the same class as the F-061 image-decode `OSError` (`tests/test_mllm_corrupt_image.py`), which was patched at its source. Fixing per-exception-type at the source doesn't cover future call-convention drift, so the general fix is to fail closed on any exception out of the generation step.

## What

- Broaden `_step_no_queue`'s `batch_generator.next()` handler from `(ValueError, RuntimeError)` to `Exception`. The **existing** failure path is reused unchanged (emit an error output, evict the requests from the batch generator, `_cleanup_finished`, return), so the client/server error classification — and the `HTTP 400` mapping for client errors — is preserved. `BaseException` (KeyboardInterrupt / SystemExit / GeneratorExit) still propagates for clean interpreter teardown.
- Document `_process_loop`'s outer `except Exception` as a last-resort backstop for step-orchestration bugs only; the generation path can no longer reach it with a poisoned batch to retry forever.

Scope note: this fixes the **hang** (`#1367`'s reported symptom). The `ministral-3b` `mask` signature itself is a model/mlx-vlm-version incompatibility (the alias is already retired via #1498); this change ensures any such drift terminates the request instead of hanging the worker.

## Tests

- New `tests/test_mllm_generation_step_signature_drift.py` (hermetic, no model): a `TypeError` from the forward pass becomes a finished error output instead of propagating; the poisoned request is evicted from the batch generator + every scheduler map so a subsequent step does not re-run it (proves the hang is gone); and the existing `ValueError` client-error path still surfaces its message + `finish_reason="error"`. The two hang tests **fail on `main`** (TypeError escapes `_step_no_queue`) and pass with the fix.
- `tests/test_mllm_corrupt_image.py` and the mllm scheduler suite (`continuous_batching` / `stats` / `executor_cancel` / `usage_tracking`, 68 tests) stay green.

## Notes

- No version bump; no public API change.
- 🤖 AI-authored (Claude Code), human-reviewed.

### DIFF
--- a/tests/test_mllm_generation_step_signature_drift.py
+++ b/tests/test_mllm_generation_step_signature_drift.py
@@ -1,0 +1,147 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression for #1367 — a generation-step exception whose type is not
+``ValueError``/``RuntimeError`` must fail the running batch, not hang.
+
+Background
+----------
+``MLLMScheduler._step_no_queue`` runs the model forward pass via
+``self.batch_generator.next()``. That call used to be wrapped in a narrow
+``except (ValueError, RuntimeError)`` that fails all running requests
+(emits an error output, evicts them from the batch generator, cleans up)
+so the loop advances instead of retrying a poisoned batch forever.
+
+A checkpoint whose mlx-vlm model class drifts from the batch generator's
+call convention raises ``TypeError`` from the forward pass — the reported
+repro was ``Model.__call__() missing 1 required positional argument:
+'mask'`` on ``ministral-3b``. ``TypeError`` is neither ``ValueError`` nor
+``RuntimeError``, so it escaped the handler, propagated to
+``_process_loop``'s outer ``except Exception``, and the loop logged +
+retried the SAME batch every step (~565 identical lines in one 90 s
+client request). The waiting request's ``output_queue`` never received an
+output or the ``None`` sentinel, so from the caller it was an unbounded
+silent hang (``/v1/status`` stuck at ``steps_executed: 0``).
+
+Same class as the F-061 image-decode ``OSError``
+(``tests/test_mllm_corrupt_image.py``). The fix broadens the handler to
+catch every exception from ``batch_generator.next()`` so future
+call-convention drift fails closed.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing as **superseded by #1810** (`430b9a6c`, merged today), which fixes this exact hang more completely.

#1810 added `MLLMScheduler._fail_all_inflight` and routes `_process_loop`'s outer `except Exception` through it, so a `TypeError` escaping `_step_no_queue` (the #1367 `ministral-3b` repro: `Model.__call__() missing 1 required positional argument: 'mask'`) now fails every in-flight request with a sanitized 500-class error and tears down the poisoned batch generator instead of retrying forever. Its regression test `test_process_loop_failure_unblocks_every_inflight_request` drives the async loop with that same `TypeError` side-effect.

My PR broadened the *inner* `except (ValueError, RuntimeError)` in `_step_no_queue` instead. On reflection that's both redundant (the hang is already gone) and worse: a server-side `TypeError` classified as a non-client error would return `error=None, finish_reason="length"` — a silent empty 200 that looks like a truncation — whereas #1810 surfaces an explicit engine error. It would also re-run against a batch generator #1810 deliberately considers poisoned and tears down. Not landing it.

#1810 landed after I branched from `5429aada`; pr_validate's DIRTY-merge gate caught the overlap.

PR #1817, opened 2026-08-11
feat: vendored Ling 3.0 bailing_hybrid backbone (KDA+MLA hybrid MoE, native)

raullenchai:
## Why

inclusionAI's Ling-3.0-tiny (7.9B total / 1.3B active, MIT) is the strongest agent-capable model that fits an 8 GB Mac — 4.2 GB at 4-bit, 131K context, GPQA 73.4 claimed — and it launched with day-0 SGLang/vLLM support while the MLX ecosystem has nothing: mlx-lm's only `bailing_hybrid` work is the unmerged Ling-2.6-era PR ml-explore/mlx-lm#1227 (open, unreviewed since July), transformers has no native port, and the HF remote code imports triton-only `fla` at module level so it cannot even run on macOS. This PR vendors the V3 backbone — the same playbook as `hy_v3` (#1211 vendor) and `muse_glimmer` (#1802) — and one module serves the whole family: Ling-3.0-tiny, Ling-3.0-flash (42L/512E) and Ling-2.6-flash all declare `bailing_hybrid`.

## What

- **`vllm_mlx/models/bailing_hybrid.py`** (new): 24-layer `[KDA ×3, MLA]` hybrid.
  - KDA linear attention with the Ling V3 **safe-gate law** `g = lower_bound · sigmoid(exp(A_log) · (f(x)+dt_bias))` (extracted from fla's `USE_LOWER_BOUND` kernel branch — NOT the softplus law mlx-lm's `compute_g` implements), causal short-conv q/k/v paths, exact-l2norm qk normalization (the rms_norm folding trick's eps placement costs 2e-4 parity — measured), fp32 delta-rule state.
  - DeepSeek-style MLA (q LoRA 256 / kv LoRA 512, 128 nope + 64 interleaved-rope dims, theta 6e6) plus the V3-only head-wise sigmoid output gate; checkpoint names honoured (`attention.*`, `dense`, `model.word_embeddings`).
  - MoE: 128 experts top-8 + shared, sigmoid `noaux_tc` expert-bias group routing (8 groups pick 4), norm_topk_prob, routed scaling 2.5; per-expert weights stacked into SwitchGLU at sanitize; router pinned 8-bit under quantization.
  - Caches: `ArraysCache(4)` per KDA layer (3 conv states + ssm state), `KVCache` per MLA layer. Layer 0 dense (`first_k_dense_replace`), trailing-remainder layers MLA; MTP tensors dropped at sanitize (tiny ships none; flash/2.6 do).
- **Registration** (`utils/tokenizer.py`): `sys.modules["mlx_lm.models.bailing_hybrid"]` injection with find_spec deference to future native mlx-lm support.
- **Auto-config**: `ling[-_.][23]` → `glm47` tool parser + `qwen3` reasoning parser + `is_hybrid` (KDA → hybrid lane). The Bailing V3 chat template's tool wire is GLM-4.7's `<arg_key>/<arg_value>` envelope (fixture-verified incl. the no-newline-after-name rendering) and reasoning is standard `<think>` with template pre-injection.
- **Tests**: 9 unit tests incl. the safe-gate law pinned against the fla formula, tri-cache incremental parity, sanitize stacking/remap/MTP-drop, auto-config routing, glm47-on-bailing-wire fixture; added to the CI apple-silicon list.

## Verification

- **Numeric parity vs the authoritative torch reference** (HF remote code on transformers 4.57, `fla` replaced by a pure-CPU shim implementing the exact kernel semantics — l2norm, safe-gate, delta recurrence): max abs logits diff **1.6e-6** full-prefill and **1.9e-6** token-by-token incremental across all three cache kinds, on an 8-layer tiny config exercising dense+MoE+KDA+MLA.
- Real-weights e2e on Apple Silicon (M2 Pro, official bf16 → mlx 4-bit at 4.507 bpw, 4.2 GB): chat + reasoning demux + tool calls, streaming and non-streaming — results posted as a comment before merge.

Weight hosting for the converted 4-bit (alias + R2 mirror + site) is a deliberate follow-up: the conversion pipeline is proven, the hosting org decision is the founder's.

## Test plan

- [x] `pytest tests/test_bailing_hybrid_model.py` (9 passed, Apple Silicon)
- [x] Torch-reference numeric parity on random weights (1.6e-6 / 1.9e-6)
- [x] `ruff format` + `ruff check` on changed files
- [x] Real-weights serve + chat/reasoning/tool e2e on Apple Silicon (posted below)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,6 +259,7 @@ jobs:
             tests/test_audio_route_registration_gate.py \
             tests/test_forced_alignment_route_hardening.py \
             tests/test_muse_glimmer_model.py \
+            tests/test_bailing_hybrid_model.py \
             -v --tb=short \
             -m "not slow" \
             -k "not Integration"
--- a/tests/test_bailing_hybrid_model.py
+++ b/tests/test_bailing_hybrid_model.py
@@ -177,10 +177,17 @@ def test_detect_model_config_routes_ling():
         assert cfg.reasoning_parser == "qwen3"
         assert cfg.is_hybrid is True
 
-    # Nearby names must not match.
-    for name in ("mlx-community/gemma-4-26b-a4b-it-4bit", "sterling-3b"):
-        cfg = detect_model_config(name)
-        assert cfg is None or cfg.tool_call_parser != "glm47" or "ling" not in name
+    # Nearby names must not be routed to the Ling configuration
+    # (codex r1 #2: the old disjunction was vacuously true).
+    cfg = detect_model_config("mlx-community/gemma-4-26b-a4b-it-4bit")
+    assert cfg is not None and cfg.tool_call_parser == "gemma4"
+    # "sterling" contains the substring "ling" but not at a boundary the
+    # pattern accepts.
+    cfg = detect_model_config("sterling-3b")
+    assert cfg is None or cfg.tool_call_parser != "glm47"
+    assert detect_model_config("acme/sterling-3b") is None or (
+        detect_model_config("acme/sterling-3b").tool_call_parser != "glm47"
+    )
 
 
 def test_glm47_parser_handles_bailing_wire():
--- a/tests/test_bailing_hybrid_model.py
+++ b/tests/test_bailing_hybrid_model.py
@@ -1,0 +1,200 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the vendored Ling 3.0 (bailing_hybrid) backbone.
+
+Pins the contract that:
+
+1. The vendored module registers into mlx-lm's importlib lookup.
+2. The layer schedule puts MLA last in each group (plus trailing
+   remainder layers) and layer 0 stays a dense MLP.
+3. A tiny synthetic config constructs + runs: logits shape and
+   incremental (cache) decode matching full prefill across all three
+   cache kinds (conv states + KDA ssm state via ArraysCache, MLA
+   KVCache).
+4. ``sanitize`` stacks per-expert weights into SwitchGLU layout,
+   remaps ``*_conv1d.weight`` into the nested conv module (with torch
+   depthwise layout transposed), and drops MTP heads.
+5. The KDA safe-gate law matches the fla ``USE_LOWER_BOUND`` kernel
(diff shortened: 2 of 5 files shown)


raullenchai:
## Real-weights verification — Ling-3.0-tiny on Apple Silicon (M2 Pro / 32GB)

Official bf16 (15.8 GB) converted with this branch's vendored arch via `mlx_lm.convert` → **4.2 GB @ 4.507 bpw**. All five acceptance scenarios pass on the final HEAD (fused gated-delta kernel path, head_dim 128):

| # | Scenario | Result |
|---|---|---|
| 1 | Chat, `enable_thinking=true` | `finish=stop`; `reasoning_content` carries 2K chars of structured thinking; content = clean single-sentence Rayleigh answer |
| 2 | Chat, default | `'Hello!'`, `finish=stop` (R12-T2F casual-chat auto-disable applies — the Bailing template has a real thinking switch, so empty reasoning here is the designed budget-saving behaviour) |
| 3 | Non-stream tool call | `get_weather({"city": "Tokyo"})`, `finish=tool_calls`, content empty |
| 4 | Stream tool call | incremental deltas assemble to the same call |
| 5 | Multi-turn tool result | `'The current weather in Tokyo is **22°C**, with **sunny** conditions.'`, `finish=stop` |

Auto-config resolved `glm47` + `qwen3` + hybrid lane from the local path with zero flags.

**Codex rounds**: 7 total. r1 (perf loop → fused kernel; vacuous test; rope wiring) and r3 (retain-all-groups fault, conv quant exemption, k=1 conv state, strict JSON assert) fixed with regression tests; r5 (version-form pattern) fixed. r2/r4/r6 were reference-divergence false positives — each declined with the authoritative evidence pinned inline (LoRA bottleneck and q_proj bias match the torch reference exactly; `mx.put_along_axis` broadcast semantics pinned by an adversarial routing test). r7 (hypothetical `Ling-3.0.1` suffixes) declined: a same-generation patch release would still be `bailing_hybrid`, so Ling routing would be correct, not a misroute.

Weight hosting for the converted 4-bit (HF org choice), alias, R2 mirror and site listing are the deliberate follow-up.

PR #1820, opened 2026-08-11
test(mac): add native XCUITest pixel coverage (#1719)

### LINKED ISSUE #1719: rapid-mac has no XCUITest target: every end-to-end GUI journey runs through a bespoke 1500-line bash harness
## Where we are

The desktop app's GUI testing splits three ways:

| layer | size | runs in CI |
|---|---|---|
| `@Test` unit tests over pure types (`RapidTests`, 131 files) | 1787 tests | yes — `swift test` |
| accessibility-identifier gate on new controls | 1 check | yes — every PR |
| golden flows: real app, AX-driven journeys | 15 flows | **no** |
| committed AX structural baselines | 12 snapshots | **no** |

The first two are in good shape. The identifier gate in particular is ahead of common practice — most projects have no mechanism forcing a new interactive control to be reachable by a stable identifier, and this one fails a PR that adds one without.

The end-to-end layer is the gap, and it is a gap in *tooling choice*, not in discipline.

## The bespoke harness

`apps/rapid-mac/scripts/gui-golden-flows.sh` is ~1500 lines of bash driving a native Swift AX client (`scripts/rapid-ax.swift`), normalising dumps through `scripts/ax-baseline.py`, and diffing against committed baselines. It works, and the AX-identifier-over-coordinates discipline baked into it is correct.

But it is a test framework we maintain ourselves, and it has the failure modes an in-house framework has. From one session of driving it:

* `rapid-ax press` returns `{"success":true}` when the press was delivered — which is not the same as the menu having opened. Acting on that return value produces a green step over a UI that never changed.
* An empty AX result is ambiguous between "the element is absent" and "the walk did not reach it". The harness has grown explicit `walk complete` plumbing to fight exactly this (see the comments in `rapid-ax.swift` about `elementWalkComplete` / `windowListComplete`), which is the right fix — and it is the kind of plumbing a mature framework provides rather than each project rediscovering.
* Assertions can be satisfied by stale state (a prior draft, a previous window) unless every one of them is scoped to the surface under test.

None of these are bad code. They are the standing cost of owning a UI-automation framework.

## What the industry does instead

XCUITest is Apple's supported answer and it targets exactly what we already have: stable accessibility identifiers.

```swift
let app = XCUIApplication()
app.launch()
app.buttons["Sidebar.NewChat"].tap()
XCTAssertTrue(app.textViews["rapid.chat.compose"].waitForExistence(timeout: 5))
```

Every identifier in that snippet already exists in the app today. The migration is not a redesign — the contract the harness relies on is the contract XCUITest consumes.

What we would stop maintaining:
* our own element-waiting and retry semantics (`waitForExistence`, `expectation(for:)`)
* our own "did the walk complete" bookkeeping (the test runner has Accessibility granted implicitly, so reads do not silently half-fail)
* our own press/typing primitives
* our own baseline diffing for the cases where a direct assertion is clearer

What we would keep: the accessibility identifiers, the identifier CI gate, and `scripts/fake-rapid-mlx.sh` for a deterministic backend — a UI test must never assert on real model output.

## Why this is not filed as "just switch"

Two real obstacles, both worth stating up front:

1. **XCUITest needs a full Xcode toolchain.** `xcode-select -p` on the current dev machine is `/Library/Developer/CommandLineTools`, which has neither `XCTest` nor swift-testing — this is already filed as #1638 (`swift test` cannot build `RapidUXTests`). GitHub's `macos-14` runners do ship full Xcode, so CI is the easier half; local runs need the toolchain installed.
2. **The existing flows encode real knowledge.** Fifteen flows worth of hard-won sequencing (model readiness, window lifecycle, streaming settle points) should be ported, not discarded. A rewrite that loses that is a downgrade even if the framework is better.

So the suggested shape is additive: stand up an XCUITest target alongside the harness, port the highest-value journeys first, and let the bash harness shrink as they land rather than deleting it up front.

## Related

* #1638 — `swift test` cannot build `RapidUXTests` because XCTest is unavailable in the CommandLineTools toolchain. Same root cause on the local side.
* Getting the *existing* golden flows into CI is being tracked separately; that is the cheaper, more urgent half. This issue is about what the end-to-end layer should be built on once it runs automatically.

raullenchai:
## What changed

- add a checked-in Xcode UI-testing target alongside the existing SwiftPM and AX suites
- launch the production-shaped `Rapid-MLX Desktop.app` with the deterministic fake sidecar
- port the image-generation journey's missing assertion: capture both thumbnail elements, crop selection chrome, and compare their rendered interior pixels
- run the XCUITest on the existing macOS GUI CI runner and retain `.xcresult` evidence on failure
- keep every existing AX GoldenFlow intact; this is an additive migration

## Reproduction / root cause

`AXUIElement` exposes each gallery button's identifier, label, and bounds but not its bitmap. The existing GoldenFlow can prove that the wire returned two different PNGs and that the gallery bound two different records, yet a rendering regression that paints one bitmap into both buttons still passes. XCUITest's element screenshots close that exact gap.

The repository also had no native UI-testing target, so Apple's waiting, screenshot, and attachment primitives were unavailable to GUI journeys.

## Validation

- `SKIP_SIDECAR=1 RAPID_BUILD_CONFIG=debug ./scripts/build.sh` — pass
- `python3 -m pytest tests/test_rapid_mac_xcui_target.py -q` — 3 passed
- `ruff check tests/test_rapid_mac_xcui_target.py` — pass
- workflow YAML parse and shell syntax — pass
- real XCUITest: pending GitHub's full-Xcode macOS runner (local host has Command Line Tools only)

Closes #1719

### DIFF
--- a/.github/workflows/rapid-mac-ci.yml
+++ b/.github/workflows/rapid-mac-ci.yml
@@ -281,6 +281,23 @@ jobs:
           RAPID_BUILD_CONFIG: debug
         run: ./scripts/build.sh
 
+      # #1719 — the first native XCUITest journey complements (and does not
+      # replace) the AX flows below. XCTest can capture an individual element,
+      # which closes the one gap AX structurally cannot: prove two gallery
+      # records actually draw different pixels rather than the same bitmap.
+      - name: 'XCUITest: image-generation pixels'
+        env:
+          RAPID_XCUI_RESULT_BUNDLE: ${{ runner.temp }}/RapidUITests.xcresult
+        run: ./scripts/run-xcui-tests.sh
+
+      - name: Upload XCUITest evidence
+        if: failure()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4
+        with:
+          name: rapid-xcui-evidence
+          path: ${{ runner.temp }}/RapidUITests.xcresult
+          if-no-files-found: ignore
+
       # One step per flow so a red check names the journey that broke rather
       # than "golden flows failed". Each needs its own output directory: the
       # harness refuses to start on a non-empty one.
--- a/.github/workflows/rapid-mac-ci.yml
+++ b/.github/workflows/rapid-mac-ci.yml
@@ -286,12 +286,14 @@ jobs:
       # which closes the one gap AX structurally cannot: prove two gallery
       # records actually draw different pixels rather than the same bitmap.
       - name: 'XCUITest: image-generation pixels'
+        id: xcui
+        continue-on-error: true
         env:
           RAPID_XCUI_RESULT_BUNDLE: ${{ runner.temp }}/RapidUITests.xcresult
         run: ./scripts/run-xcui-tests.sh
 
       - name: Upload XCUITest evidence
-        if: failure()
+        if: steps.xcui.outcome == 'failure'
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4
         with:
           name: rapid-xcui-evidence
@@ -341,6 +343,20 @@ jobs:
         run: ./scripts/gui-golden-flows.sh --flow window-close-prompt
         env:
           RAPID_GUI_GOLDEN_OUT: ${{ runner.temp }}/golden/window-close-prompt
+
+      # The XCUITest is allowed to fail temporarily only so the established AX
+      # journeys still execute and retain their independent evidence. Restore
+      # the blocking verdict after those flows; a red native journey remains a
+      # red GUI job and can never land.
+      - name: Require native XCUITest
+        if: always()
+        env:
+          XCUI_OUTCOME: ${{ steps.xcui.outcome }}
+        run: |
+          if [[ "$XCUI_OUTCOME" != success ]]; then
+            echo "::error::Native XCUITest failed ($XCUI_OUTCOME)"
+            exit 1
+          fi
 
       # The harness stops at the FIRST mismatched baseline and the job stops at
       # the first failing flow, so one red run reveals exactly one line. With 5
--- a/.github/workflows/rapid-mac-ci.yml
+++ b/.github/workflows/rapid-mac-ci.yml
@@ -344,20 +344,6 @@ jobs:
         env:
           RAPID_GUI_GOLDEN_OUT: ${{ runner.temp }}/golden/window-close-prompt
 
-      # The XCUITest is allowed to fail temporarily only so the established AX
-      # journeys still execute and retain their independent evidence. Restore
-      # the blocking verdict after those flows; a red native journey remains a
-      # red GUI job and can never land.
-      - name: Require native XCUITest
-        if: always()
-        env:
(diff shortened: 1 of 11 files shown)


raullenchai:
Codex review round 3 follow-up:

- Addressed the useful coverage portion by adding `test_swift_source_parent_traversal_resolves_rapid_mac_fixture`. It performs the same four parent traversals and asserts the result is exactly `apps/rapid-mac`, that `scripts/fake-rapid-mlx.sh` exists there, and that the one-level-higher `apps/scripts/...` path does not.
- Dismissing the associated path-count claim with concrete evidence: `#filePath` is the file itself (`.../apps/rapid-mac/Tests/RapidUITests/Tests/ImageGenerationPixelTests.swift`), so the four deletions resolve to `Tests` → `RapidUITests` → `Tests` → `rapid-mac`; they do not resolve to `apps`. The Swift test also independently fails closed with `isExecutableFile(atPath:)`.

Local result: `tests/test_rapid_mac_xcui_target.py` 4/4 passed.

raullenchai:
Codex review round 5 follow-up:

- Fixed exact-binary targeting: the test now constructs `build/Rapid-MLX Desktop.app` from the pinned source root and launches it with `XCUIApplication(url:)`, rather than resolving only by bundle ID.
- Replaced mean-channel comparison with cropped, aligned per-pixel RGB RMS comparison.
- Dismissing the `CODE_SIGNING_ALLOWED=NO` finding with execution evidence: rapid-mac CI run 31456501407 job 93671272908 compiled and launched the runner under that exact setting, executed `testTwoImageRendersDrawDistinctThumbnailPixels`, produced a successful native-XCUITest outcome (failure-evidence upload skipped), and the final native gate passed. This setting therefore does not prevent this macOS runner/host from launching in the supported CI environment.

raullenchai:
Codex review disposition for the latest host/signing findings:

- `Host/main.swift` exiting is not blocking this runner shape: rapid-mac CI run 31457868817 compiled the target, launched the exact production app, and executed `ImageGenerationPixelTests` for 42.225s through multiple XCUI clicks. The failure occurred later at the first image-result assertion, after UI automation was established.
- `CODE_SIGNING_ALLOWED=NO` is likewise empirically supported on the hosted runner: run 31456501407 completed the native XCUITest successfully with this same runner setting, and run 31457868817 again launched and drove the app. No signing or runner-connection error occurred.

Both static findings are therefore dismissed with runtime evidence. The actionable Ruff formatting finding was fixed in 48d1299d.

raullenchai:
Final Codex disposition:

- The repeated host-lifetime blocker is dismissed with current-head runtime evidence: rapid-mac CI run 31460711585 completed `ImageGenerationPixelTests` successfully, the failure-only xcresult upload was skipped, all five AX GoldenFlows passed, and the final native-XCUITest verdict passed. The configured host therefore remained viable for the complete test session on the supported macOS 15 runner.
- The full-Xcode prerequisite nit was actionable and is fixed in 2441fb6f by probing `xcodebuild -version`; Command Line Tools now fail immediately with the explicit prerequisite message.

PR #1831, opened 2026-08-11
feat(mac): add image resolution controls and improve progress readability

osdodo:
## What does this PR do?

  Adds explicit output-resolution controls to the macOS Images tab.

  - Provides 512, 768, 1024, 1280, 1536, and 2048 long-edge presets.
  - Combines the selected resolution with the existing 1:1, 3:4, and 4:3 aspect ratios.
  - Sends the calculated dimensions to the image generation and editing APIs.
  - Keeps 1024 as the default to preserve existing behavior.
  - Uses a responsive two-row layout when the composer is narrow.
  - Improves generation progress readability with an opaque surface, stronger text contrast, and larger timing text.
  - Adds resolution mapping and API dimension-contract tests.
  - Updates the image-generation golden flow and snapshot harness.

  ## Why is this needed?

  The Images tab previously exposed only aspect-ratio controls, while output resolution was fixed and not visible to the user. Users could not choose between faster, lower-memory renders and more detailed
  high-resolution output.

  The translucent generation progress card also allowed bright or detailed images to reduce the contrast of step, elapsed-time, and ETA text, making progress difficult to read.

  This change makes output dimensions explicit and configurable while ensuring generation status remains legible over arbitrary image content.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
+++ b/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
@@ -193,6 +193,9 @@ enum DevSnapshot {
         render(contentView(width: 640, height: 560), to: "\(dir)/content-min.png")
         // Images tab (empty state — no results, catalog not yet resolved).
         render(imagesView(width: 700, height: 640), to: "\(dir)/images-empty.png")
+        // Narrow composer: the canvas controls wrap to the two-row
+        // ViewThatFits layout, which only this width exercises.
+        render(imagesView(width: 420, height: 640), to: "\(dir)/images-narrow.png")
         renderHosted(imagesView(width: 700, height: 640),
                      size: CGSize(width: 700, height: 640),
                      appearance: .darkAqua, to: "\(dir)/images-dark.png")
--- a/apps/rapid-mac/Sources/Rapid/Images/ImageGenViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Images/ImageGenViewModel.swift
@@ -8,7 +8,8 @@ import Observation
 @MainActor
 @Observable
 final class ImageGenViewModel {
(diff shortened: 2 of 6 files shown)


raullenchai:
Maintainer review — thanks @osdodo, this is a well-shaped PR. Summary of what was verified and what changed while landing:

**Verified**
- Server contract match: every preset × aspect lands inside the engine's `parse_image_size` bounds (256–2048, multiple of 16); the two test suites mirror the real constants.
- Live end-to-end on the 0.12.10 engine (`z-image-turbo`): `384x512` (8s), `1280x960` (43s), and worst-case `2048x2048` (240s, no OOM) all return PNGs with exactly the requested dimensions.
- DevSnapshot renders: default chip reads `1024 × 1024` (behavior preserved), the 420pt `images-narrow.png` exercises the two-row `ViewThatFits` fallback, dark mode legible.
- AX identifier gate passes; `Images.Resolution` + per-preset identifiers reachable.

**Changes pushed while landing** (branch was rebased onto `main`, which conflicted with the merged Settings-UI work)
1. Dropped the `DevSnapshot`/`RapidApp` environment-plumbing changes — `main` already fixed the same traps with harness-local throwaway stores (incl. an MCP config pointed at a scratch file so snapshots never touch the user's real `mcp.json`), so only the new `images-narrow.png` render was kept.
2. `runGenerate`/`runEdit` now snapshot `outputSize` at submission — previously it was read after the (possibly minutes-long) warm-up `await`, so changing aspect/resolution mid-warm-up silently retargeted the in-flight request.
3. Resolution menu items carry `.isSelected` for VoiceOver, matching the `aspectPicker` convention, and a test now pins the fresh-view-model default to `1024x1024`.

Deliberately not added: mocked-client request-inspection tests — `ImageGenViewModel` constructs its client internally, and injecting it is a refactor beyond this PR's scope.

PR #1845, opened 2026-08-11
fix(mac): make golden-flow baselines portable across GPU probe and RAM total

raullenchai:
## Why
The four golden flows that capture the main-window footer — `chat-restore`, `slow-stream-stop`, `model-crash-recovery`, `image-generation` — fail the AX structural gate on **every Apple Silicon developer machine**, the platform the app targets. CI stays green because the baselines committed in #1835 were generated on the hosted runner (Intel/sandboxed, 7 GB RAM), so the gate silently only holds on that one environment and is useless as a local dogfood tool.

Two footer gauges read machine-specific state the normalizer never scrubbed:

- **GPU gauge** has no stable cross-machine shape. Apple Silicon publishes a live reading (`AXUnknown desc="GPU 47 percent"`); Intel Macs and sandboxed apps can't probe AGXAccelerator and publish a static note (`AXStaticText help="GPU probe unavailable — …" value=text`) — same footer item, different role, attribute, and text.
- **Memory gauge** names the machine's total RAM (`… out of 32 gigabytes`). The spelled-out unit slipped past the `<size>` rule, so the integer total stayed literal (`7` on the runner vs `32` on a dev Mac).

## What
Fix in `ax-baseline.py`, where the live footer values already collapse (CPU `<percent>`, tok/s `<rate>`):
- Fold spelled-out byte units (`gigabytes`, …) into the size rule so both the memory reading and the total become `<size>`.
- Canonicalize the GPU gauge to one `desc="GPU <gpu>"` line regardless of platform shape (mirrors how `_ROLE_EQUIVALENTS` already unifies a control's role across macOS releases).

The five affected baselines are regenerated; the diff is exactly the two footer lines per file, nothing else.

## Tests
- New regression cases in `tests/test_ax_baseline_os_variance.py` feed **both** platform shapes of each gauge through the normalizer and assert one output. They fail on `main` (GPU/memory lines differ) and pass here.
- `pytest tests/test_ax_baseline.py tests/test_ax_baseline_os_variance.py` → 24 passed.
- All four golden flows PASS locally on Apple Silicon (M2 Pro, 32 GB) with the regenerated baselines.

## Notes
- Normalizer + baseline only; no Swift/product change, no version bump.
- 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.answered.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.answered.txt
@@ -39,8 +39,8 @@ AXApplication title="Rapid-MLX"
       AXUnknown desc="Server status" enabled=true
       AXUnknown desc="Tokens per second: no data yet" enabled=true
       AXUnknown desc="CPU <percent>" enabled=true
-      AXStaticText help="GPU probe unavailable — Intel Macs and sandboxed apps don't expose AGXAccelerator utilisation." value=text enabled=true
-      AXUnknown desc="Memory normal: <version> gigabytes used out of 7 gigabytes" enabled=true
+      AXUnknown desc="GPU <gpu>" enabled=true
+      AXUnknown desc="Memory normal: <size> used out of <size>" enabled=true
       AXButton id="Footer.DesktopVersionPill" desc="Rapid-MLX <version> · up to date" help="Rapid-MLX <version> is the latest release. Click to open Settings → App." enabled=true
     AXToolbar
       AXButton desc="Hide Sidebar" enabled=true
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.transcript-restored.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/chat-restore.transcript-restored.txt
@@ -39,8 +39,8 @@ AXApplication title="Rapid-MLX"
       AXUnknown desc="Server status" enabled=true
       AXUnknown desc="Tokens per second: no data yet" enabled=true
       AXUnknown desc="CPU <percent>" enabled=true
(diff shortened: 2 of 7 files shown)


raullenchai:
### Local `pr_validate` codex_review: stopping after 5 rounds (documented non-convergence)

Every objective gate is green and stayed green across all five rounds:

- `full_unit`: **17153 passed**, 72 skipped, 6 xfailed
- `targeted_tests`: 21 passed · `lint`: clean · `supply_chain`/`test_env_check`: pass
- The four affected golden flows (`chat-restore`, `slow-stream-stop`, `model-crash-recovery`, `image-generation`) **PASS locally on Apple Silicon** with the regenerated baselines; the CI `gui-golden-flows` job re-validates them on the Intel/sandboxed runner — the cross-machine proof this change exists for.
- New regression tests fail on `main` (GPU/memory lines diverge) and pass here.

`codex_review` did not converge — it escalated to a new, more abstract point each round and its latest two are factually wrong about the current code:

- Round 5 asked for "the known pairs (`AXUnknown.description` / `AXStaticText.help`) and full-match." That is exactly what `is_gpu_telemetry()` now does.
- Round 6 then reports it "ignores the element's role … uses prefix matching." The committed code branches on `role == "AXUnknown"` / `role == "AXStaticText"` and uses `re.fullmatch` on each shape — no prefix match, role is checked. The examples it cites (`GPU 47 percent settings`, `GPU probe unavailable options`) fail `fullmatch` and are covered by the negative tests.
- Its remaining ask — a "stable footer identifier or ancestry" — does not exist: the footer telemetry items carry no `accessibilityIdentifier` (confirmed from live AX dumps). Two elements identical in role **and** exact text are, by definition, indistinguishable to a structural fingerprint, and the "GPU N percent" string is only ever the gauge.

Landing per the repo's documented codex non-convergence stopping rule: objective gates green + live behaviour verified + remaining findings are false-positive/unreachable. Merge still gated on the GitHub CI checks (`gui-golden-flows`, `test-matrix`, `test-apple-silicon`, `l1-smoke`) being green and the branch up to date.

Issue #1840, opened 2026-08-11
Images tab is always visible but the bundled sidecar ships without the image extra


Jevin-F:
### App version

0.12.9 (local build from `main`)

### Last version where it worked (if this is a regression)

Not a regression. The same failure string is present in the previously staged 0.12.7 sidecar (`runtime/image_lane.pyc`), so this is long-standing.

### Mac model and memory

MacBook Pro, Apple M4 Pro, 24 GB

### macOS version

26.5.2

### Where in the app does this happen?

Images tab.

### What happened? What did you expect instead?

The Images entry point is rendered unconditionally, but the bundled sidecar is installed without the `image` extra, so image generation cannot work on a stock desktop build. Clicking "Download & start" does nothing — no banner change, no error, no progress.

Expected: either the entry point reflects what the bundled sidecar can actually do, or the missing component is named and installable.

### Steps to reproduce

1. Build the desktop app normally (`scripts/build.sh`), so the sidecar comes from `build-sidecar.sh`.
2. Open the app and go to Images.
3. Click "Download & start" for `flux2-klein-4b`.
4. Nothing happens. No download starts and no error is shown.

### Which model was loaded?

`qwen3.5-4b-4bit` was resident; the requested model was `flux2-klein-4b`.

### Additional context

**The sidecar has no image support.** `scripts/build-sidecar.sh:251-256` installs the engine as `"${RAPID_MLX_SOURCE}[audio-desktop]"` — the `image` extra is not included, so `mflux` is absent. Confirmed on a local bundle: `site-packages` contains no `mflux`, `diffusers`, or `flux` package.

Running the bundled sidecar directly surfaces the reason:

```
Error: image generation requires the `rapid-mlx[image]` Python extra
(`pip install 'rapid-mlx[image]'`).
```

**The entry point is not gated.** `SidebarView.swift:111-117` renders the Images row unconditionally, alongside New Chat and Audio, with no capability check.

**Installing the extra does fix it.** Installing `mflux>=0.18.0` into the bundle's `site-packages` makes `POST /v1/models/load` for `flux2-klein-4b` return `"state":"resident"`, and chat continues to work (verified with a completion request against the same running sidecar).

**Bundling it unconditionally conflicts with the size budget.** `mflux` declares `torch>=2.7.1,<3.0` as a hard runtime dependency, not an extra. Measured on this arm64 bundle, `site-packages` grew from 358 MB to 1.1 GB:

| package | size |
|---|---|
[...]

PR #1847, opened 2026-08-11
fix(mac): surface resident model load rejection on the initiating surface (#1838)

### LINKED ISSUE #1838: Resident model load failures give no feedback on the surface that triggered them
### App version

0.12.9 (local build from `main`)

### Last version where it worked (if this is a regression)

Not a regression — the code path below is present on current `main`. I did not verify older releases.

### Mac model and memory

MacBook Pro, Apple M4 Pro, 24 GB

### macOS version

26.5.2

### Where in the app does this happen?

Any surface that triggers a resident model load. Observed on the Images tab readiness banner ("Download & start").

### What happened? What did you expect instead?

When a resident model load is rejected by the engine, the surface that triggered it shows nothing — no banner change, no error, no spinner. The button appears dead.

The failure reason is available and preserved verbatim through the client, but it only reaches the log pane, which the user has to open deliberately and is never pointed to.

Expected: the reason the engine gave should be shown on the surface that initiated the load.

### Steps to reproduce

1. Run a build whose sidecar cannot serve a given model (e.g. a stock bundle, which has no `image` extra — see #1840).
2. Start the app and let a chat model become ready, so the sidecar is already running.
3. Go to Images and click "Download & start".
4. Nothing changes in the UI. The reason is only visible if the log pane is opened.

### Which model was loaded?

`qwen3.5-4b-4bit` was resident; the requested model was `flux2-klein-4b`.

### Additional context

Code path, all on current `main`:

1. `ServerResidency.swift:171` — a non-2xx response (other than 404/405) maps to `.rejected(detail)`, where `detail` is the engine's own error message decoded from the response envelope.
2. `ServerManager.swift:840-842` — the `.rejected` case appends `"Resident model load declined: \(message)"` to `logLines` and returns `false`.
3. `ReadinessModelStart.swift` — `perform` discards the result: `_ = await server.ensureServing(alias: alias, hfPath: hfPath)`.
4. Callers such as `ImagesView.swift:158` never observe the outcome, so the banner, composer, and model picker are unchanged.

The message is not lost — `logLines` is rendered at `ContentView.swift:1284-1289` — but nothing directs the user there.

Because the sidecar is already running, `ensureServing` takes the in-process `/v1/models/load` path instead of restarting it, so there is no visible process activity either.

A concrete case that reaches this path: on a bundle without the `image` extra, requesting an image model makes the engine fail with `image generation requires the 'rapid-mlx[image]' Python extra (pip install 'rapid-mlx[image]')`. That string is specific and actionable, is carried as far as `.rejected(detail)`, and is then dropped from the UI.

**Suggested direction:** surface `detail` on the view that initiated the load — the readiness banner already renders state for that model. Returning the outcome from `ReadinessModelStart.perform`, or publishing a last-failure reason on `ServerManager`, would let each caller decide how to present it without changing the load path.

raullenchai:
## Summary

Closes #1838. A resident-model load that the engine **rejects** currently produces **no feedback on the surface that triggered it**: the Images banner/composer/model-picker sit unchanged and the button looks dead. Because the sidecar is already running, `ensureServing` takes the in-process `/v1/models/load` path (no process restart), so there is no visible activity either.

The engine's actionable reason is preserved verbatim all the way to `.rejected(detail)` — it's just dropped before it can reach the UI, landing only in the log pane the user is never pointed to.

### Root cause

```
ReadinessModelStart.perform  →  _ = await server.ensureServing(alias:, hfPath:)
                                           │
ServerManager.ensureServing ── .rejected(detail)
  • appends "Resident model load declined: …" to logLines
  • returns false
  • ● result discarded by the caller, never observed by ImagesView/ContentView
```

- `ServerResidency.swift` → non-2xx/non-404/405 response maps to `.rejected(detail)` (the engine's own message).
- `ServerManager.ensureServing` → `.rejected` only wrote to the log.
- `ReadinessModelStart.perform` → discarded the result.
- `ImagesView` / `ContentView` → never observed failure.

### Fix

Publish the rejection so the surface that initiated the load can present it:

- **`ServerResidency.swift`**: new `ResidentLoadFailure { alias, message }`.
- **`ServerManager.swift`**: new `@Observable` `residentLoadFailures` (per-alias dictionary) plus a per-alias attempt token, so a rejection is published to the initiating surface while remaining robust under concurrency:
  - **per-alias**: `residentLoadFailure(for:)` is keyed by the alias that failed, so concurrent loads of *different* models cannot clobber each other's outcome;
  - **latest-attempt-wins (same alias)**: each `ensureServing` mints a per-alias attempt token; when its async load returns it may write/clear the failure only if it is still the newest attempt, so an older attempt that returns last cannot overwrite a newer success (or an old success wipe a newer rejection);
  - a fresh load attempt clears any prior rejection up front, so a stale rejection never outlives the next load, and it is scrubbed like the log before reaching the UI.
- **`ContentView.swift` / `ImagesView.swift`**: render it in the readiness banner, scoped to the alias that actually failed (`.ready`/`.starting`/`.downloading` still outrank a stale rejection, and `failureApplies` keeps it pinned to the failed model).

Net effect: the engine's reason — e.g.

> image generation requires the 'rapid-mlx[image]' Python extra (pip install 'rapid-mlx[image]')

— is shown on the Images banner as an actionable, recoverable message instead of a silent no-op.

### Regression protection

- **New unit tests** — `ResidentLoadFeedbackTests.swift` (serialized suite with process-wide-stub cleanup):
  - a rejected resident load publishes the engine's reason, keyed per alias;
  - a successful load clears a prior rejection;
  - rejections are independent per alias — loading a different, succeeding model does not wipe another model's rejection;
  - **older rejection cannot clobber a newer success** and **older success cannot clear a newer rejection** for the *same* alias (via an in-order gate in the stub that holds the first load in flight so the older attempt returns last — the two directions of latest-attempt-wins).
  Uses a `URLProtocol` stub standing in for the sidecar's in-process residency/load endpoints, so it exercises the real `ServerResidencyClient.load → .rejected` path without a live sidecar. The `residentLoadFailure(for:)` API did not exist on old `main`, so the tests cannot silently rot into a pass.
- **GoldenFlow** — `resident-load-rejected` in `gui-golden-flows.sh`, wired into CI: starts the fake chat sidecar, asks Images to load a model, proves the in-process `/v1/models/load` was attempted and rejected, then asserts the engine's reason actually appears on the Images surface (AX tree) and the banner still offers a recovery action. It waits for the Images model picker to resolve before pressing the readiness action (same sync the `image-generation` flow uses) so the press is deterministic.
- **Fake sidecar** — `fake-rapid-mlx.sh` now serves `/v1/models/residency` + `/v1/models/load` and, with `FAKE_REJECT_IMAGE_LOAD=1`, rejects `fake-image-alias` with the engine's real 422 detail so the flow drives the exact #1838 path.

### Notes on test environment

Swift tests (`RapidTests`) need the **full-Xcode CI runner (macos-15)**; the local machine has only Command Line Tools, which cannot resolve `import Testing` — a pre-existing environment limitation (all existing test targets fail the same way). `swift build --target Rapid` passes locally; the test + GoldenFlow suites run in CI.

This addresses only the feedback surfaced on load rejection (#1838). Gating the Images entry point / bundling the `[image]` extra is tracked separately in #1840.

### DIFF
--- a/.github/workflows/rapid-mac-ci.yml
+++ b/.github/workflows/rapid-mac-ci.yml
@@ -304,7 +304,7 @@ jobs:
       # than "golden flows failed". Each needs its own output directory: the
       # harness refuses to start on a non-empty one.
       #
-      # These five drive the app through `rapid-ax` alone (see
+      # These six drive the app through `rapid-ax` alone (see
       # `flow_requires_peekaboo` in the harness). Most other flows shell out to
       # peekaboo — a third-party install whose bridge socket comes from the
       # Peekaboo app rather than its CLI, with its own permission surface. The
@@ -343,6 +343,11 @@ jobs:
         run: ./scripts/gui-golden-flows.sh --flow window-close-prompt
         env:
           RAPID_GUI_GOLDEN_OUT: ${{ runner.temp }}/golden/window-close-prompt
+
+      - name: 'Golden flow: resident-load-rejected'
+        run: ./scripts/gui-golden-flows.sh --flow resident-load-rejected
+        env:
+          RAPID_GUI_GOLDEN_OUT: ${{ runner.temp }}/golden/resident-load-rejected
 
       # The harness stops at the FIRST mismatched baseline and the job stops at
       # the first failing flow, so one red run reveals exactly one line. With 5
--- a/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
@@ -232,14 +232,29 @@ final class ServerManager {
     /// Bounded to `logBufferCapacity` entries.
     private(set) var logLines: [String] = []
 
-    /// The most recent in-process residency load that the engine rejected, or
-    /// `nil` when there has been no rejection since the last successful load.
+    /// Most recent in-process residency load rejections, keyed by the alias
+    /// that failed. Per-alias rather than a single global slot so two
+    /// concurrent loads of DIFFERENT models cannot clobber each other across
+    /// an `await` (#1838 follow-up): model B's rejection published while
+    /// model A is still loading will not be wiped by A's later success, and
+    /// an older request's rejection cannot overwrite a newer request's
+    /// result for a different model. Each key owns its own outcome, so the
+    /// surface that asked to load that model reads exactly that model's
+    /// result.
     ///
     /// This is *published* (an `@Observable` property), not log-only: the
     /// surface that initiated the load reads it and presents the engine's
     /// reason verbatim, so a rejected resident load is an actionable message
     /// instead of a silent no-op (#1838).
-    private(set) var lastResidentLoadFailure: ResidentLoadFailure?
+    private(set) var residentLoadFailures: [String: ResidentLoadFailure] = [:]
+
+    /// The most recent rejection for `alias`, or `nil` if the last attempt
+    /// for that model succeeded or no rejection has been recorded. Read by
+    /// the surface that initiated the load so it can present the engine's
+    /// reason verbatim rather than only writing it to the log pane (#1838).
+    func residentLoadFailure(for alias: String) -> ResidentLoadFailure? {
+        residentLoadFailures[alias]
+    }
 
     /// True while a start or stop is in flight. The UI disables both
     /// buttons during this window so a second click cannot race the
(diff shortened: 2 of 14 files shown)


raullenchai:
Addressed the latest review round (same-alias concurrency). Summary of what changed since the review:

**1. Same-alias latest-attempt-wins (new)** — commit `3d0b244f`
- `ServerManager` now keeps a per-alias **attempt token** (`residentLoadAttemptTokens: [String: UUID]`), in addition to the per-alias failure dictionary.
- `ensureServing` mints a fresh token at the start of each real load attempt; when the async `residencyClient.load` returns (`.loaded` / `.rejected`) it writes/clears `residentLoadFailures` **only if that token is still the newest for the alias**.
- This closes both interleaving directions the review called out for the same model:
  - an **older rejection** returning after a **newer success** is ignored (no expired banner);
  - an **older success** returning after a **newer rejection** does not wipe the fresh failure.
- Two new deterministic tests reproduce the interleaving via an **in-order gate** in the `URLProtocol` stub: it holds the *first* matching load in flight (on the URL loading thread, never the main actor) so the older attempt returns last, then asserts latest-attempt-wins:
  - `olderRejectionDoesNotClobberNewerSuccess`
  - `olderSuccessDoesNotClearNewerRejection`
  Both are in the `.serialized` suite with `defer { reset() }` restoring the gate state.

**2. PR description corrected** — the body previously still described the single-slot `lastResidentLoadFailure`; it now documents the per-alias + latest-attempt-wins design and the expanded test matrix.

No further review comments open. `build` (Swift tests) and `gui-golden-flows` are running on the new head.

Issue #1849, opened 2026-08-11
Client-correctable prefill-cap rejection surfaces as HTTP 200 + generic 'Internal error during streaming'


raullenchai:
## Problem

When the MLLM scheduler rejects a request for a **client-correctable** reason (prompt exceeds the prefill cap), the streaming route returns **HTTP 200** and a generic mid-stream error event:

```
data: {"id":"chatcmpl-...","choices":[{"index":0,"delta":{"role":"assistant"}}]}
data: {"error": {"message": "Internal error during streaming", "type": "internal_error"}}
data: [DONE]
```

The scheduler produced an actionable message —

```
Total prompt tokens (15993) exceeds the per-batch cap (8192 = prefill_step_size 8192 × 1 request(s)).
For image inputs, downscale the image; for text inputs, shorten the prompt or restart the server
with --prefill-step-size set higher.
```

— and `service/helpers.py` `_disconnect_guard` (line ~3961) catches the `ValueError` and replaces it with the generic envelope. The client sees "internal_error" for what is a 400-class condition with a documented remedy.

## Why it matters

- OpenAI-compatible clients (including our own desktop app) render "Internal error during streaming" — the user gets no hint that shortening the prompt fixes it, and support triage reads it as an engine crash.
- Benchmarks/health checks read HTTP 200 + role delta as success until they notice zero content arrived. This exact shape cost a benchmark cell a misdiagnosis today ("stream produced no tokens").

## Expected

- Errors raised **before the first content token** with a client-correctable cause should map to a 4xx JSON error (message intact), not a 200 stream.
- If the stream has already started, the SSE error event should carry the real message and a distinguishing type (`invalid_request_error` here), reserving `internal_error` for genuine engine faults.

This is the route-layer remainder of the mllm-scheduler error-swallow family (the scheduler-side coarse `except` was fixed earlier; the guard still degrades the message).

## Repro

Same as the prefill-cap issue: 16k-token text request to `mlx-community/gemma-4-26b-a4b-it-4bit` on 0.12.10, `stream=true`.

raullenchai:
Security/correctness follow-up landed in #1860.

[...]

Issue #1848, opened 2026-08-11
MLLM lane rejects text-only prompts >8192 tokens: prefill_step_size acts as a cap, not a step


raullenchai:
## Problem

A **text-only** prompt longer than 8192 tokens sent to an MLLM alias fails: the MLLM scheduler treats `prefill_step_size` as a hard per-batch **cap** instead of a chunking **step**.

```
ERROR:rapid_mlx.mllm_scheduler:Batch generation failed: Total prompt tokens (15993)
exceeds the per-batch cap (8192 = prefill_step_size 8192 × 1 request(s)).
```

The text lane (BatchedEngine) chunks prefill past this size without complaint — the same 16k prompt works on every text alias. Only vision-capable aliases are affected, and only because the request routes through `mllm_scheduler` even when it carries zero images.

## Repro (rapid-mlx 0.12.10, PyPI wheel, M3 Ultra)

```bash
rapid-mlx serve mlx-community/gemma-4-26b-a4b-it-4bit --port 8111
# stream a ~16k-token text-only chat request → mid-stream error event
```

Found by the cross-runtime benchmark: every other runtime in the matrix (oMLX, mlx-lm server, Ollama with `OLLAMA_CONTEXT_LENGTH=20480`) completes the same 16k text cell on this model; rapid-mlx is the only DNF.

## User impact

Long-document chat with any vision model (gemma-4-26b is a default catalog pick) breaks at >8k tokens of context. The desktop app exposes exactly this workflow (attach a doc / long conversation history).

## Expected

Text-only requests on an MLLM alias should prefill in `prefill_step_size` chunks like the text lane does — or route text-only requests to the text scheduler outright (the #1187 text-only autodegrade direction). The error message's own suggestion (`--prefill-step-size` higher) is a workaround, not a fix; the default should handle the model's advertised context.

Related: the error surfacing for this failure is its own bug, filed separately (HTTP 200 + generic "Internal error during streaming").

raullenchai:
Closed by **#1855** (merge commit `1d48af96`) — the text-only prefill-cap DNF is fixed.

## What was delivered
[...]

Issue #1854, opened 2026-08-11
Vision request TTFT is 2.4x stock mlx-vlm on the same model (0.60s vs 0.25s, gemma-4-26b)


raullenchai:
Cross-runtime matrix, vision lane (M3 Ultra, 0.12.10, parity flags, 1024x768 PNG + short prompt, N=3 medians):

| runtime | vision TTFT | vision decode |
|---|---|---|
| mlx-vlm 0.6.3 server | **0.25 s** | 114 tok/s |
| oMLX v0.5.8.dev3 | 0.45 s | 117 tok/s |
| **rapid-mlx 0.12.10** | **0.60 s** | 111 tok/s |
| Ollama 0.32.5 | 0.93 s | 89 tok/s |

Decode is within 3 % of stock — the gap is entirely in time-to-first-token: ~0.35 s of added latency per image message vs the same mlx-vlm stack we sit on. Suspects: image decode/preprocess duplication, MLLM scheduler admission hop, or vision-encoder cache miss on first use per request. Related family: the known B=1 `Scheduler.step()` overhead (perf queue); #1853 (long-context decode) is the text-side sibling.

Repro: `benchmarks/cross-runtime` (branch bench/cross-runtime), `python3.12 run.py --lane parity --models gemma-4-26b --runtimes rapid,mlxvlm --scenarios vision_ttft`. Data: `results/vision-lane/`.

raullenchai:
Fixed in #1856 (merged, squash `e9255e65`).

**Root cause:** rapid's MLLM continuous-batching prefill ran the full VLM forward (`vision_tower` + `embed_vision` + language) once per request and never reused the projected image features. Stock mlx-vlm's server caches them across requests via the model's `get_input_embeddings(vision_cache=…, _image_key=…)` contract, so a **repeated** image skips the vision encoder there (~0.3–0.4s) — rapid never passed those kwargs. rapid's existing `VisionEmbeddingCache` only memoised `prepare_inputs` output (~11ms), not the encoder pass. Decode was already at parity, which is why the whole gap was in TTFT.

The bench (and multi-turn "ask again about the same screenshot") sends the same image every rep, so it exercises exactly the cache-hit path where the two runtimes diverged.

**Fix:** forward mlx-vlm's `vision_cache` + `_image_key` (keyed by image content hash) into the VLM forward for model families that honour the contract (gemma-4 today), so a repeated image reuses cached features.

[...]

PR #1855, opened 2026-08-11
fix(mllm): exempt text-only prompts from prefill cap and surface stream errors (#1848, #1849)

raullenchai:
Fixes **#1848** and partially addresses **#1849** (see the #1849 note below — layer 1 remains a follow-up, so #1849 is intentionally **kept open**).

## Summary

Two coupled MLLM-lane bugs that together made long (~>8k token) **text-only** prompts against any vision-capable alias fail as a DNF, and then swallow the actionable error into a misleading HTTP 200.

### 1. #1848 — text-only prompts >8192 tokens rejected (fixed)

`MLLMBatchGenerator._process_prompts` enforced a hard per-batch cap (`total_prompt_tokens > prefill_step_size × num_requests`). That cap exists purely to bound **vision-merge** memory (#682) — each image request can embed thousands of vision-patch tokens into one forward. But it also counted **text-only** requests, so any single text-only prompt longer than `prefill_step_size` (default 8192) was rejected, even though the chunked text-only prefill path exists (added for #1187/#1193).

**Fix**: only vision-bearing requests (`pixel_values` and `image_grid_thw` present) count toward the cap, comparing their aggregate tokens against `prefill_step_size × num_vision`. An all-text-only batch can never violate the cap — long text-only prompts flow through the existing chunk path, matching the text lane. New `_is_text_only_request` / `_prefill_cap_violation` helpers are shared with `_run_vision_encoding`.

**Review feedback addressed**:
- **Message clarity**: the cap error now reads `Vision-request prompt tokens (…)` instead of `Total prompt tokens (…)`, so a diagnostic can no longer mislead a user into thinking their *text* prompt tripped it.
- **Concurrency / cache-merge concern**: added `test_prefill_cap_exempts_batch_of_long_text_only_prompts` pinning that a batch of multiple >8k text-only prompts never trips the cap, while a long vision prompt among text peers still does (#682 preserved). Validated on real hardware (below).

### 2. #1849 — client-correctable rejection no longer swallowed as a generic 200 (layer 2 done; layer 1 = follow-up)

On the streaming path the MLLM rejection surfaced as a generator exception that `_disconnect_guard` (F-131 hardening) masked into `{"message":"Internal error during streaming","type":"internal_error"}` — HTTP 200 with no actionable message. The non-streaming route already maps the per-batch-cap / failed image-video / chat-template markers to HTTP 400 with the real message.

**Fix (layer 2)**: for known **client-correctable** markers, surface the real message with OpenAI `invalid_request_error` type, while preserving F-131's generic sanitisation for genuine engine faults (no traceback/class-name leak through SSE).

**⚠️ #1849 is NOT fully closed by this PR.** The issue asks for two layers:
1. **Before the first content token** → HTTP **4xx** (intact message).
2. **After the stream has started** → SSE error event with real message + distinguishing type.

This PR implements **layer 2** (the stream has already committed `StreamingResponse` + emitted the role chunk by the time the scheduler rejection surfaces, so we cannot change the HTTP status). Getting a true 4xx requires either a pre-completion preflight or pulling the first real scheduler result before committing the response headers — an architectural change with TTFT / keepalive / disconnect-semantics implications across the guided/strict/plain streaming variants. That is tracked as a **follow-up**; this PR deliberately does **not** claim full closure of #1849, and the issue stays open.

## Verification

- Regression tests added/updated in `tests/test_mllm_batch_generator.py`, `tests/test_sse_keepalive.py`, `tests/test_chat_route_vlm_image.py`.
- Focused suites: **53 passed** (`test_mllm_batch_generator`, `test_sse_keepalive`, `test_chat_route_vlm_image`).
- `ruff check` + `ruff format --check`: clean on all changed files.

### Real-hardware rebench (M3 Ultra / 256 GB, `mlx-community/gemma-4-26b-a4b-it-4bit`, fixed code, stream=true)

- **Single 16,000-token text-only prompt** (tokenizer-verified): HTTP 200, stream `[DONE]`, produced real output (`RB_OK`), 16k-token prompt completed in ~9.5s. Pre-fix this was the only DNF in the cross-runtime matrix (`Total prompt tokens (15993) exceeds the per-batch cap (8192 …)`).
- **Concurrent 3 × 16,000-token text-only streams**: all three completed (HTTP 200, `[DONE]`, 0 error events, unique correct outputs) in ~28s wall — no DNF, no cap rejection, **no OOM at cache merge** with multiple long text prompts batched concurrently.
- **#1849 layer-2 E2E**: a corrupt image produced `{"message":"Failed to process image: …","type":"invalid_request_error"}` in the SSE stream — the real actionable message now reaches the caller instead of the generic "Internal error during streaming".

## CI notes

Full local `tests/` run: 17,198 passed; only `test_event_loop.py` environment-dependent failures (require a live server on `127.0.0.1:8000`, absent on the headless runner). If CI shows the unrelated `hermes` parser-microbenchmark threshold failure on Python 3.12, that is runner noise unrelated to this PR and needs a rerun to green.

### DIFF
--- a/tests/test_chat_route_vlm_image.py
+++ b/tests/test_chat_route_vlm_image.py
@@ -240,7 +240,7 @@ def test_chat_route_maps_per_batch_cap_error_to_http_400():
     """
     engine = _StubMLLMEngine(
         raise_msg=(
-            "Total prompt tokens (2292) exceeds the per-batch cap "
+            "Vision-request prompt tokens (2292) exceeds the per-batch cap "
             "(2048 for 1 request(s)). For image inputs, downscale the "
             "image; for text inputs, shorten the prompt or restart "
             "the server with --prefill-step-size set higher."
--- a/tests/test_mllm_batch_generator.py
+++ b/tests/test_mllm_batch_generator.py
@@ -1640,3 +1640,32 @@ def test_process_prompts_does_not_cap_large_text_only_prompt(monkeypatch):
         f"a 20k-token text-only prompt must pass the cap (then fail on the "
         f"bare-model prefill), but got the cap error: {err_msg}"
     )
+
+
+def test_prefill_cap_exempts_batch_of_long_text_only_prompts():
+    """A batch of MANY long text-only prompts must not trip the cap (#1848).
+
+    Concurrency concern raised in review: with text-only now exempt, several
+    >8k text prompts can batch together. Even though each is prefilled in
+    chunks, they all still need full KV at cache-merge time, so we must not
+    accidentally re-introduce a rejection for multi-long-text batches. The
+    cap is vision-only by design; an all-text-only batch (any count, any
+    length) must never violate it (#1848 / #682 contract preserved).
+    """
+    from vllm_mlx.mllm_batch_generator import _prefill_cap_violation
+
+    # 4 concurrent long text-only prompts, all > prefill_step_size.
+    batch = [_text_request(uid=i, token_count=20000) for i in range(4)]
+    assert _prefill_cap_violation(batch, prefill_step_size=8192) is None, (
+        "a batch of multiple 20k-token text-only prompts must stay exempt "
+        "from the vision-only per-batch cap (no DNF / no cap rejection)"
+    )
+
+    # And when one long VISION request joins them, the cap correctly fires
+    # only because of the vision request, not the text ones.
+    mixed = [_text_request(uid=i, token_count=20000) for i in range(3)]
+    mixed.append(_make_vision_cap_request(uid=3, token_count=10000))
+    assert _prefill_cap_violation(mixed, prefill_step_size=8192) is not None, (
+        "a 10k-token vision request among long text-only peers must still "
+        "trip the cap (#682 preserved)"
+    )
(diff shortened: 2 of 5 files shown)


raullenchai:
## Review response

Thanks for the detailed review. Here's how each point was addressed.

### #1848 (fixed)
- **Message clarity**: cap error text renamed `Total prompt tokens` → **`Vision-request prompt tokens`** so a diagnostic can no longer implicate a user's *text* prompt. Fixture strings updated in `test_sse_keepalive.py` / `test_chat_route_vlm_image.py`, and the `exceeds the per-batch cap` classifier-marker phrase is preserved verbatim.
- **Concurrency / cache-merge concern**: added `test_prefill_cap_exempts_batch_of_long_text_only_prompts` pinning that a batch of multiple >8k text-only prompts never trips the cap, and that a long vision prompt among text peers still does (#682 preserved). This was additionally validated on real hardware: **3 concurrent 16,000-token text-only streams** all completed (HTTP 200, `[DONE]`, 0 error events, distinct correct outputs) with **no OOM at cache merge**.

### #1849 (layer 2 shipped; layer 1 tracked as follow-up)
Per the review, layer 2 (actionable SSE error with `invalid_request_error` after the stream starts) is shipped; **#1849 is intentionally kept open** — layer 1 (HTTP 4xx before the first content token) requires preflight / pre-pull of the first scheduler result before committing `StreamingResponse`, which has TTFT/keepalive/disconnect implications across the guided/strict/plain streaming variants and is beyond this PR's scope. The PR body states this explicitly.

### Note on the conflict
`main` advanced past this PR's base (`#1856`, "cache projected vision features") touching the same two files. The branch was rebased onto current `main`; both the #1854 vision-cache changes and this PR's changes are preserved, and the rebased branch is green. CI on the new head is re-running and will be brought to green before landing.

PR #1856, opened 2026-08-11
fix(mllm): cache projected vision features so repeated images skip the encoder (#1854)

raullenchai:
## Why
A vision chat request whose image **repeats** across turns (multi-turn "ask again about the same screenshot", and the cross-runtime bench which sends the same image every rep) paid the full vision-encoder cost on **every** request, making rapid's vision TTFT ~2.4x stock mlx-vlm (0.60s vs 0.25s on gemma-4-26b, #1854). Decode was already at parity — the gap was entirely time-to-first-token.

**Root cause:** rapid's MLLM continuous-batching prefill runs the whole VLM forward (`vision_tower` + `embed_vision` + language) once per request and never reused the projected image features. Stock mlx-vlm's server caches them via the model's `get_input_embeddings(vision_cache=…, _image_key=…)` contract, so a repeated image skips the vision encoder there (~0.3–0.4s) — rapid never passed those kwargs. rapid's existing `VisionEmbeddingCache` only memoised `prepare_inputs` output (~11ms), not the encoder pass.

**Measured** (gemma-4-e2b, M2 Pro), content TTFT:

| scenario | rapid before | rapid after | stock mlx-vlm |
|---|---|---|---|
| repeated image (multi-turn / bench) | 1.05s | **0.68s** | 0.66s |
| novel image (cold) | 1.10s | 1.10s | 1.05s |

The novel-image path was already at parity and is unchanged — the vision encode is inherent to a first-seen image. The fix closes the cache-hit gap the benchmark and multi-turn conversations exercise.

## What
- Forward mlx-vlm's `vision_cache` + `_image_key` into the VLM forward from `_run_vision_encoding` so `get_input_embeddings` reuses the cached projected features on a repeat. Reuses mlx-vlm's own `VisionFeatureCache` (the model speaks its `.get()/.put()` interface natively).
- Key by image **content** hash (`compute_images_hash`), robust to the per-request temp-file paths `all_images` holds; stashed on the request in `_preprocess_request`.
- **Capability-gated**: only models whose `get_input_embeddings` honours the contract (detected by source inspection — gemma-4 today) get the kwargs; every other model keeps the exact unchanged full-forward path. Future mlx-vlm versions that wire more families in are picked up automatically.
- Bound the feature cache to `min(vision_cache_size, 32)`: each entry pins a projected-features Metal buffer, so it is capped tighter than the pixel cache.

## Tests
- New regression tests in `test_mllm_batch_generator.py`: contract detection (supported / unsupported / missing method), the request-field default, the `__init__` gate (enabled only for a supported model + `enable_vision_cache`), and `_run_vision_encoding` injecting the cache kwargs only for supported **image** requests (not text-only, not unsupported models). `142 passed` across the vision/mllm suite.
- Verified byte-identical generated output across cold-miss and warm-hit requests for the same image (temp=0).

## Notes
- No version bump. No behaviour change for text-only requests or for model families that don't implement the contract.

### DIFF
--- a/tests/test_mllm_batch_generator.py
+++ b/tests/test_mllm_batch_generator.py
@@ -1344,3 +1344,95 @@ def test_run_vision_encoding_no_vision_cache_for_unsupported_model():
 
     assert "vision_cache" not in model.last_call_kwargs
     assert "_image_key" not in model.last_call_kwargs
+
+
+class _CachingBehaviorModel:
+    """Model stub that mimics mlx-vlm's ``get_input_embeddings`` caching dance:
+    on a ``vision_cache`` miss it "encodes" (bumps ``encode_count``) and stores
+    the projected features; on a hit it skips. Lets a test assert that rapid's
+    wiring actually dedups the vision encoder across repeated images end to end
+    — not just that the kwargs are forwarded."""
+
+    def __init__(self):
+        self.encode_count = 0
+        self.language_model = object()
+
+    def get_input_embeddings(self, input_ids=None, pixel_values=None, **kwargs):
+        # Contract markers so _model_supports_vision_feature_cache enables the
+        # feature: names both vision_cache and _image_key in the body.
+        _ = kwargs.get("vision_cache")
+        _ = kwargs.get("_image_key")
+        return None
+
+    def __call__(self, input_ids, cache=None, **kwargs):
+        vision_cache = kwargs.get("vision_cache")
+        image_key = kwargs.get("_image_key")
+        if vision_cache is not None and image_key is not None:
+            if vision_cache.get(image_key) is None:
+                self.encode_count += 1
+                vision_cache.put(image_key, mx.zeros((1, 4)))
+        else:
+            # No cache wired in → the encoder always runs (pre-fix behaviour).
+            self.encode_count += 1
+        return mx.zeros((1, 1, 8))
+
+
+def test_repeated_image_skips_encoder_distinct_image_reencodes():
+    """The wiring dedups the vision encoder: a repeated image reuses cached
+    features (no re-encode); a distinct image encodes again."""
+    pytest.importorskip("mlx_vlm.vision_cache")
+    model = _CachingBehaviorModel()
+    gen = MLLMBatchGenerator(
+        model=model,
+        processor=object(),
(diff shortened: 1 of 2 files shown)


raullenchai:
### Note on the remaining codex "capability detection is lexical" finding

Codex flagged the source-inspection gate across three review rounds (nit → nit → blocking), each time escalating the same core concern — "a lexical `"vision_cache"`/`"_image_key"` match could false-positive and inject kwargs that raise `TypeError` downstream." I've addressed the realistic failure surface and am landing on that basis; recording the reasoning here rather than adding an allowlist:

1. **The TypeError vector is structurally closed.** `_model_supports_vision_feature_cache` now requires **both** `__call__` (where `_run_vision_encoding` passes the kwargs) **and** `get_input_embeddings` (where `__call__` forwards them) to accept `**kwargs`. A false-positive match therefore cannot raise `TypeError` at either hop we control — the extra kwargs are silently absorbed.

2. **No new failure class vs. today.** rapid already forwards processor `extra_kwargs` (`mm_token_type_ids`, `attention_mask`, `image_grid_thw`) through the identical `self.model(input_ids, cache=cache, **kwargs)` call. Any model that would choke on two more keys flowing to a strict downstream call already chokes on those — so every MLLM model rapid supports today already tolerates this shape. The change adds no new surface.

3. **The residual hypothetical isn't a real model.** It would need `get_input_embeddings` that (a) takes `**kwargs`, (b) contains *both* exact quoted contract keys without using them, and (c) force-forwards `**kwargs` to a strict inner call — but (c) contradicts (2). A model that quotes both `"vision_cache"` and `"_image_key"` is, in practice, a model that consumes them.

4. **Auto-detection is deliberate.** An allowlist/version pin would stop future mlx-vlm families that implement the contract from being picked up, and adds a maintenance burden. Structural detection (both hops `**kwargs` + both quoted keys in the body) tracks the actual contract.

Objective gates are green on the final commit: `full_unit` 17176 passed, targeted 41 passed, lint/type-check/validate/test-apple-silicon/test-matrix all green. Behaviour is verified end-to-end (repeated-image TTFT 1.05s → 0.68s at stock-mlx-vlm parity; byte-identical output across cache miss/hit).

Issue #1197, opened 2026-07-24
--kv-cache-dtype is not applied to the live BatchGenerator KV cache


nmorgowicz:
### Rapid-MLX version

0.10.17

### Hardware

Apple M5 Max, 64 GiB unified memory

### macOS version

macOS 26.5.1 (25F80)

### Python version

3.11.15

### Model

`nightmedia/Qwen3.5-9B-DS9-USS-Defiant-1M-q8-hi-mlx`

### Full serve command

rapid-mlx --no-telemetry serve \   nightmedia/Qwen3.5-9B-DS9-USS-Defiant-1M-q8-hi-mlx \   --host 127.0.0.1 --port 18087 \   --max-num-seqs 1 --max-concurrent-requests 1 \   --disable-prefix-cache --no-hybrid \   --pflash off --prefill-step-size 32768 \   --kv-cache-dtype int8 \   --max-tokens 64 --no-mllm --log-level INFO

### Streaming or non-streaming?

Both

### Actual

`--kv-cache-dtype int8` is accepted, resolved, and reported by the server, but it is not applied to the live per-request KV cache used by continuous batching. Instrumenting the live `PromptProcessingBatch.prompt` cache state showed a plain bf16 `BatchKVCache.state`, not a quantized cache, throughout prefill.

The identical source trace also covers the deprecated equivalent forms:

```bash
--kv-cache-quantization --kv-cache-quantization-bits 4
--kv-cache-quantization --kv-cache-quantization-bits 8
```

`--kv-cache-quantization-bits` alone is only a width selector and is not expected to enable quantization.

### Expected

When a user selects `--kv-cache-dtype int8` or `int4`, the continuous-batching live KV cache should transition to a quantized representation after the configured threshold, yielding the advertised memory reduction. If continuous batching cannot support this today, the server should reject the flag or explicitly report that it applies only to retained prefix-cache entries.

### Impact

- Startup/metrics imply live KV quantization is active when it is not.
- VRAM planning based on int8/int4 KV savings is wrong for continuous batching.
- Fresh-server benchmarks with prefix cache disabled receive no memory or quality change from the requested dtype.
- The legacy bit-width form has the same hot-path omission.

### Minimal reproduction

1. Start the server using the command above.
[...]

PR #1857, opened 2026-08-11
perf: fix long-context decode degradation — disk KV checkpoint + quantized live KV defaults (#1853)

### LINKED ISSUE #1853: Long-context decode degrades 2-3x more than every other MLX runtime (16k: -21..-47% vs their -10..-15%)
## Problem

Decode throughput degrades with context length far more steeply than every other Apple-Silicon runtime. Cross-runtime matrix (M3 Ultra, 0.12.10 PyPI wheel, parity lane: prefix cache / PFlash / spec decode off, saturated 512-token generations, N=3 medians):

| model | decode B=1 (128-tok ctx) | decode @16k ctx | degradation |
|---|---|---|---|
| qwen3.5-4b | 150 tok/s | **80 tok/s** | **−47 %** |
| qwen3.5-9b | 106 | 79 | −25 % |
| gpt-oss-20b | 127 | 84 | −34 % |
| qwen3.6-27b | 36 | 21 | −42 % |
| qwen3.6-35b-a3b | 92 | 73 | −21 % |

Same cells, same prompts, same machine:

| runtime | typical degradation 128→16k |
|---|---|
| rapid-mlx | **−21…−47 %** |
| oMLX v0.5.8.dev3 | −10…−15 % (4B: 175→149) |
| mlx-lm 0.31.3 server | ~−14 % (4B: 131→113) |
| Ollama 0.32.5 | ~−10 % (4B: 108→97) |

We are last or second-to-last in every 16k decode cell despite winning most short-context columns. oMLX sits on the same mlx/mlx-lm stack and only loses ~15 % — so this is our scheduler/cache path, not an MLX kernel ceiling.

16k TTFT shows the same signature, milder: 9.30 s vs oMLX 8.61 / mlx-lm 8.76 (4B).

## Where to look

- Per-step KV/cache handling in `Scheduler.step()` at large sequence lengths (known B=1 overhead family; the gap widens with context, suggesting per-token work that scales with sequence length rather than staying O(active batch)).
- Memory-aware cache / hybrid cache bookkeeping on the decode hot path.
- Compare against oMLX's decode loop (`omlx.engine`/BatchedEngine) — same dependency stack, one-third the degradation; also beats stock `mlx_lm.server` B=1 by ~30 %, so their step loop is worth reading.

## Repro

`benchmarks/cross-runtime` (branch bench/cross-runtime): `python3.12 run.py --lane parity --scenarios decode_b1,decode_16k --models qwen3.5-4b --runtimes rapid,omlx`. Full data in `results/parity-full` + `results/parity-v2`, report in `results/FINAL.md`.

raullenchai:
## Why

The 2026-08-11 cross-runtime benchmark (#1853) found rapid last or second-to-last in every 16k-context decode cell: qwen3.5-4b dropped 150 → 82 tok/s (−45%) while oMLX / mlx-lm / Ollama on the same stack lost only 10–15%. Layer-by-layer isolation (raw BatchGenerator → Scheduler → BatchedEngine → stream_chat → live HTTP server, all measured on the same model/prompt) attributed the entire gap to two default-on features, each adding an O(context) cost to every decode step:

1. **Disk KV checkpointing (R15-P1 #296) defaulted to interval 256.** Every 256 generated tokens the scheduler serializes the FULL KV cache (`save_prompt_cache` + fsync) synchronously on the decode thread — ~0.6s per snapshot at 16k on a 4B model. Nothing in the engine reads the checkpoints back: `load_checkpoint` / `scan_checkpoints` have no callers outside the module and its tests (the docstring's "loads checkpoints during startup" integration never landed). The module's own docs already describe 0/opt-in as the intended shape.

2. **Live KV cache quantization (R15 #300 + #1197) defaulted to int4.** The premise was "4×-smaller KV = 4× less decode bandwidth", but the live path (`QuantizedBatchKVCache`) implements quantization as dequant-on-read — it materializes full-precision K/V on every decode step, so per-token cost grows with context instead of shrinking. The #910 numbers behind the default used 292-token prompts, where this is invisible.

## Measurements

qwen3.5-4b, parity server (prefix cache / spec decode / PFlash off), N=3 medians via the `bench/cross-runtime` harness, plus N=2 flag-isolated runs:

| config | decode_b1 (128 ctx) | decode_16k |
|---|---|---|
| shipped defaults (ckpt=256 + int4 KV) | 149.7 | 82.2 |
| checkpoint off | 160.9 | 102.3 |
| checkpoint off + bf16 KV (**new defaults**) | **167.0** | **134.6** |
| in-process mlx-lm BatchGenerator (floor) | 167.7 | 138.2 |
| oMLX / mlx-lm / Ollama (same cell) | 175 / ~131 / 116 | 149 / 113 / 97 |

Isolated dtype A/B at 16k (checkpoint off): bf16 134.6, int4 98.2 (−27%), int8 86.1 (−36%).

Net: 16k decode **82 → 134.6 tok/s (+64%)**, from last place to second (behind oMLX 149); the residual gap is the known B=1 scheduler-overhead family, tracked separately. Post-fix the server is within ~3% of the in-process library floor at both context lengths.

## What changed

- `SchedulerConfig.kv_disk_checkpoint_interval`: 256 → 0 (opt-in), same for the `--kv-disk-checkpoint-interval` serve flag and the bench-path fallback. Feature unchanged when explicitly enabled.
- `DEFAULT_KV_CACHE_DTYPE` / `--kv-cache-dtype`: int4 → bf16. int4/int8 remain explicit opt-ins for memory-constrained hosts (KV 4×/2× smaller); `--reasoning` still pins int8; the sliding-window/MLA auto-downgrade safelist is unchanged.
- Regression tests pin both opt-in defaults (config + CLI), verified red against the pre-fix source.

Closes #1853

## Compatibility: KV memory impact of int4 → bf16

Reviewer concern: bf16 live KV is ~4× larger than int4, so a deployment near its memory ceiling with long contexts could regress on upgrade. Why this does not gate the change:

- **No hard-OOM path.** The engine enforces `gpu_memory_utilization` via `mx.set_memory_limit` (default 0.90) **and** the D-METAL-CAP admission gate, which rejects over-cap requests with 503 + Retry-After instead of letting the allocator crash. A host that no longer fits a given context degrades to backpressure, not a crash loop.
- **The int4 default is weeks old.** bf16 was the default for the project's entire history before R15 #300; "previously working deployments" overwhelmingly means bf16 deployments. The measured cost of keeping int4 is −27% decode at 16k for every user, traded against a memory ceiling that only affects hosts within 4× of their KV limit.
- **One flag away.** Memory-constrained hosts keep the exact prior behavior with `--kv-cache-dtype int4`; the `--reasoning` int8 pin and the sliding-window/MLA safelist are unchanged.
- **Industry default.** Ollama and LM Studio both ship full-precision KV by default; quantized KV is opt-in everywhere else in the ecosystem.

The right long-term fix for wanting int4 back as default is a fused quantized-attention read path (no per-step materialization) — tracked in #1853 discussion, out of scope here.

### DIFF
--- a/tests/test_harmony_parsers.py
+++ b/tests/test_harmony_parsers.py
@@ -1292,6 +1292,9 @@ class TestServeLogLevelFlags:
         # Assert against the real parser (the flag moved from ``main``
         # into ``build_parser`` when the two split for #1853's
         # effective-default tests; source-scraping ``main`` went stale).
+        # Building the parser registers the share subcommand, which
+        # imports ``websockets`` — absent from the no-MLX CI lane.
+        pytest.importorskip("websockets")
         from vllm_mlx.cli import build_parser
 
         args = build_parser().parse_args(["serve", "m", "--log-level", "DEBUG"])
--- a/tests/test_harmony_parsers.py
+++ b/tests/test_harmony_parsers.py
@@ -1289,12 +1289,15 @@ class TestHarmonyCLIIntegration:
 
 class TestServeLogLevelFlags:
     def test_cli_serve_has_log_level_flag(self):
-        import importlib
-        import inspect
+        # Assert against the real parser (the flag moved from ``main``
+        # into ``build_parser`` when the two split for #1853's
+        # effective-default tests; source-scraping ``main`` went stale).
+        from vllm_mlx.cli import build_parser
 
-        source = inspect.getsource(importlib.import_module("vllm_mlx.cli").main)
-        assert '"--log-level"' in source
-        assert 'choices=["DEBUG", "INFO", "WARNING", "ERROR"]' in source
+        args = build_parser().parse_args(["serve", "m", "--log-level", "DEBUG"])
+        assert args.log_level == "DEBUG"
+        with pytest.raises(SystemExit):
+            build_parser().parse_args(["serve", "m", "--log-level", "TRACE"])
 
     def test_module_server_has_log_level_flag(self):
         from pathlib import Path
--- a/tests/test_kv_cache_dtype.py
+++ b/tests/test_kv_cache_dtype.py
@@ -337,11 +337,13 @@ def test_reasoning_wins_over_safelist():
 # ---------------------------------------------------------------------------
 
 
-def test_bf16_default_reason_marks_it_as_default():
+def test_bf16_reason_is_neutral_and_not_a_downgrade():
     decision = resolve_kv_cache_dtype("bf16", model_name="qwen3-4b-4bit")
-    # Operators should be able to see from the startup log that the
-    # default (not a downgrade) produced bf16.
-    assert "default" in decision.reason
+    # argparse cannot distinguish an explicit bf16 from the parser
+    # default, so the reason must not claim either; what matters is
+    # that bf16 resolution is never reported as a downgrade.
+    assert "bf16" in decision.reason
+    assert "default" not in decision.reason
     assert decision.downgraded is False
 
 
--- a/tests/test_kv_cache_dtype.py
+++ b/tests/test_kv_cache_dtype.py
@@ -23,10 +23,18 @@ from vllm_mlx.kv_cache_dtype import (
 # ---------------------------------------------------------------------------
 
 
-def test_default_is_int4():
-    """Locks the R15 #300 contract — int4 is the new default."""
-    assert DEFAULT_KV_CACHE_DTYPE == "int4"
-    assert "int4" in KV_CACHE_DTYPES
+def test_default_is_bf16():
+    """Locks the #1853 contract — bf16 is the default, quant is opt-in.
+
+    The R15 #300 int4 default assumed a bandwidth win, but the live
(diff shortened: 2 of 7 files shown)


raullenchai:
## Merge decision record (codex non-convergence)

Six pr_validate rounds on this PR. Final round (SHA 6721be15): **full_unit 17214 passed · stress/agent matrix 4×3 PASSED (35B bench: PR faster than live-measured base on both cold −3.2% and warm −3.4%) · GitHub CI 15/15 green**.

The only failing step is codex_review, and only intermittently: the int4→bf16 default concern was flagged in rounds 2 and 6 but accepted in rounds 4 and 5 on the identical diff and PR body. The concern is a product tradeoff, not a code defect, and is addressed in the "Compatibility" section above: no hard-OOM path (memory-cap + 503 backpressure), the int4 default is weeks old, rollback is one flag, and full-precision KV is the ecosystem default. Every gate that measures behavior is green — merging on judgment per the repo's non-convergence SOP.

Issue #1852, opened 2026-08-11
Release pipeline: a failed publish step still ships the release — add a secret preflight gate


guo:
## Description

#1851 is the symptom: `mirror-dist` failed on three consecutive tags because `CLOUDFLARE_API_TOKEN` was missing. This issue is the cause it exposed — **a release can be published while the step that describes it fails.**

Two structural gaps let a missing credential survive three releases:

1. The credential requirement was introduced by #1747 and only discovered at the *next tag*. Nothing checks, at merge or at bump time, that the secrets a release workflow references actually exist.
2. `build` attaches the GitHub Release before `mirror-dist` runs, so a downstream failure leaves a shipped release with an unpublished fallback pointer. Nothing retracts it.

Same shape as #1612, inverted: that was a green job doing nothing, this is a red job that doesn't stop anything.

## Context

`rapid-mac-release.yml` references nine secrets and two vars. Two of the nine (`CLOUDFLARE_API_TOKEN`, `CLOUDFLARE_ZONE_ID`) are absent from repo secrets, and the guard that catches this (`:397-404`) only runs after the notarised build, on a tag, once the release is already out.

The existing `release-preflight.yml` is the natural place to catch it: it runs on the `chore: bump version to X.Y.Z` PR — *before* the tag exists — and already has a gate structure (PF-1, G1, G10, G11) with a `preflight-summary` roll-up.

Note the current guard is a presence check (`[[ -n "$TOKEN" ]]`). An expired or under-scoped token is non-empty and passes it. The Cloudflare token needs two capabilities (R2 write on `rapid-desktop-dist`, cache purge on the `rapidmlx.com` zone); a token carrying only one would satisfy any presence check and still fail at `wrangler`.

## Acceptance criteria

- [ ] **PF-2 gate**: preflight fails the bump PR when any `secrets.*` / `vars.*` referenced by `rapid-mac-release.yml` is absent. Derived from the workflow file, not a hand-maintained list, so a future step that needs a new secret extends the gate automatically
- [ ] PF-2 is wired into `preflight-summary` as blocking, and documented in the workflow header + `docs/development/releasing.md`
- [ ] Skips cleanly on fork PRs, where secrets are unavailable to the workflow by design and absence proves nothing
[...]

Issue #1851, opened 2026-08-11
rapid-mac release: mirror-dist has failed on 3 consecutive tags (CLOUDFLARE_API_TOKEN missing)


guo:
## Description

The updater-fallback publishing pipeline introduced by #1747 (merged 2026-08-09, fixing #1612) has never completed successfully. `mirror-dist` has failed on **three consecutive tagged releases** — 0.12.8, 0.12.9, 0.12.10 — each time on a missing repo secret, with `publish-updater-fallback` skipped as a result.

`dl.rapidmlx.com/latest.json` currently advertises 0.12.10 and is *not* stale right now. But `rapid-mac-release.yml` is the only workflow in this repo that writes it, and that job has not succeeded since 0.12.7 — so the pointer is being maintained out of band. That is what makes this easy to keep ignoring: the red release run looks harmless because the symptom keeps getting papered over by hand.

## Context

Three consecutive tagged release runs, identical shape:

| tag | run | build | mirror-dist | publish-updater-fallback |
|---|---|---|---|---|
| `rapid-mac-v0.12.8` | [31391820012](https://github.com/raullenchai/Rapid-MLX/actions/runs/31391820012) | success | **failure** | skipped |
| `rapid-mac-v0.12.9` | [31437871070](https://github.com/raullenchai/Rapid-MLX/actions/runs/31437871070) | success | **failure** | skipped |
| `rapid-mac-v0.12.10` | [31467045177](https://github.com/raullenchai/Rapid-MLX/actions/runs/31467045177) | success | **failure** | skipped |

All three fail at the same guard:

```
##[error]tagged releases require updater fallback publishing; missing: CLOUDFLARE_API_TOKEN
```

The job env confirms the token resolves empty while the account id is populated:

```
CLOUDFLARE_API_TOKEN:
CLOUDFLARE_ACCOUNT_ID: ***
```

`gh secret list` shows only `CLOUDFLARE_ACCOUNT_ID` (last updated `2026-08-09T23:01:08Z`). Both **`CLOUDFLARE_API_TOKEN`** and **`CLOUDFLARE_ZONE_ID`** are absent. The second job requires the zone id as well (`.github/workflows/rapid-mac-release.yml:492-495`), so adding only the API token moves the failure one job later rather than fixing it.

[...]

PR #1859, opened 2026-08-11
ci(release): add PF-2 gate — release secrets present, and usable

guo:
Closes the first and last acceptance criteria of #1852. Context: #1851.

## Why

A workflow can reference a secret that was never configured. Actions does not warn — the expression evaluates to the empty string, and the failure surfaces wherever that value is finally used. For the mac release pipeline that is *on a tag, after the notarised build, once the GitHub Release is already published*.

That is #1851. #1747 introduced two credential requirements (`CLOUDFLARE_API_TOKEN`, `CLOUDFLARE_ZONE_ID`); neither was ever added. `mirror-dist` hard-failed on three consecutive tags — 0.12.8, 0.12.9, 0.12.10 — and each release shipped anyway, with `dl.rapidmlx.com/latest.json` published by hand afterwards. The manual fix-up is what let it survive three cycles: it removed the symptom and left the defect.

The guard inside `rapid-mac-release.yml` was working correctly. It was just placed where nothing can act on it.

## What

A new **PF-2** gate on `release-preflight.yml`, which runs on the `chore: bump version to X.Y.Z` PR — before the tag exists. Two halves.

### 1. Presence — `scripts/check_release_secrets.py`

Parses `rapid-mac-release.yml`, collects every `secrets.*` / `vars.*` reference, fails if any is absent.

- **Derived, not listed.** The required set comes from the workflow file, so a step that starts needing a new secret extends the gate on its own. A hand-maintained list would drift exactly the way the credential did — that is the bug, not a variation of it.
- **References only come from inside `${{ }}`.** Parsed YAML values are walked, so the header comment in `rapid-mac-release.yml` that *documents* the required secrets is not mistaken for wiring, and prose in a `run:` block cannot register a requirement.
- **`GITHUB_TOKEN` is exempt** — injected by Actions, so its absence proves nothing.
- **Keys only.** Actions exposes no way to enumerate configured names, so `toJSON(secrets)` is passed in. The script reads keys and never prints, returns, or logs a value; a test asserts that.

### 2. Validity — `scripts/probe_release_credentials.py`

Presence is not validity: an expired or revoked token is a non-empty string and passes every presence check, including the `[[ -n ... ]]` guard in the release workflow. Cloudflare tokens can carry an expiry date, which makes silent expiry the realistic way #1851 returns.

The blocking/advisory split is the design, because a false block stops releases to punish a token that is scoped *correctly*:

| check | severity | why |
|---|---|---|
| `/user/tokens/verify` reports not-active | **blocking** | every valid token can call this regardless of permissions, so the answer is unambiguous and cannot false-block |
| R2 bucket listing, zone lookup | advisory | the release needs R2 *write* and *cache purge*; neither implies the read permission these use, so a minimally-scoped token can legitimately 403 |
| transport error, timeout, 5xx — anywhere | advisory | "could not check" is not "checked and it is bad"; a Cloudflare blip must not block a release |

Cache-purge permission is deliberately left unproven — there is no dry run, and exercising it on every bump PR to prove it works is the worse trade.

A token-level finding short-circuits the rest, so a dead token or an unreachable Cloudflare yields one finding rather than three restatements of it. Findings are scrubbed of the token before printing: Actions masks registered secrets, but a transport error carries a message we did not compose into a public log, so that mask is the last line of defence rather than the only one.

### Skipped on fork PRs

Actions withholds secrets from forks by design, so an empty context there would fail every name without proving anything. `preflight-summary` accepts `skipped` for PF-2 for that reason only — every other skip path is already excluded by the `is_bump` guard.

## Still out of scope

**Atomic publish.** `build` still attaches the GitHub Release before `mirror-dist` runs, so a downstream failure still leaves a shipped release. Making the release a draft until `publish-updater-fallback` succeeds changes when a release becomes visible and deserves its own review — remains in #1852.

## Verification

The presence half was run against the live repo state at both ends of the #1851 fix:

- **Before** the secrets were added: named exactly `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ZONE_ID`, and did not flag the seven configured secrets or two vars.
- **After**: `release secrets OK — 11 reference(s) across 1 workflow(s), all configured`, exit 0.

25 unit tests total (pure stdlib; the probe takes its HTTP primitive as an argument so the severity split is pinned without network). `ruff check` and `ruff format --check` clean, and the repo's own `check_workflow_expressions.py` and `check_gha_pinning.py` both pass on the modified workflow.

## Note for the reviewer

**PF-2 has not yet executed in CI.** The whole preflight only runs on bump PRs, so on this PR every gate reports `skipping` — a green check here does not exercise the new job. The `toJSON(secrets)` wiring and the probe's live call path first run on the next `chore: bump version to X.Y.Z` PR. Given the subject matter, that caveat seemed worth stating rather than leaving a green tick to imply more than it does.

The gate is blocking. It would have failed every bump PR while the secrets were missing, which is the intent. Those secrets now exist and the token verifies, so it is green today; merging does not block the next release.

### DIFF
--- a/.github/workflows/release-preflight.yml
+++ b/.github/workflows/release-preflight.yml
@@ -14,6 +14,7 @@ name: Release pre-flight
 #   G11  auto-routing escape-hatch registry    macOS-14 (needs MLX)
 #   PF-1 squash subject regex pre-check        ubuntu-latest
 #   PF-2 release secret + var presence         ubuntu-latest
+#        + Cloudflare credential probe (expiry blocks; scope gaps advise)
 #
 # What this workflow does NOT cover (M3-only, runs locally — see
 # releasing.md):
@@ -116,6 +117,17 @@ jobs:
         run: |
           python scripts/check_release_secrets.py \
             .github/workflows/rapid-mac-release.yml
+
+      - name: PF-2 — release credentials are usable, not just present
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          CLOUDFLARE_ZONE_ID: ${{ secrets.CLOUDFLARE_ZONE_ID }}
+          # Read into the environment rather than interpolated into the shell,
+          # so a repository-variable value can never be parsed as script.
+          R2_BUCKET: ${{ vars.RAPID_MAC_DIST_R2_BUCKET }}
+        run: |
+          python scripts/probe_release_credentials.py --bucket "$R2_BUCKET"
 
   g1-release-smoke:
     # G1 — clean-room install + import smoke on macOS-14. Catches the
--- a/.github/workflows/release-preflight.yml
+++ b/.github/workflows/release-preflight.yml
@@ -13,6 +13,7 @@ name: Release pre-flight
 #   G10  MLX upstream cross-chip-family scan   advisory
 #   G11  auto-routing escape-hatch registry    macOS-14 (needs MLX)
 #   PF-1 squash subject regex pre-check        ubuntu-latest
+#   PF-2 release secret + var presence         ubuntu-latest
 #
 # What this workflow does NOT cover (M3-only, runs locally — see
 # releasing.md):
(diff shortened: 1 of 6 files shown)


raullenchai:
Reviewed the failure matrix / fork behavior / secret-leak surface before landing (#1852). Ran an adversarial pass; two robustness gaps were real and are now fixed on this branch (845922ba):

- **429 false-block**: `check_token_active` treated any non-5xx `success:false` (incl. a transient 429 rate-limit) as a BLOCKING dead-token, contradicting the gate's own "only not-active blocks; transient stays advisory" design. Now only 401/403 and a 200 app-layer failure / non-active status block; 429 and other non-2xx are advisory. Guard test added, proven red on pre-fix code.
- **bracket-notation blind spot**: reference parsing matched only `secrets.FOO`, so a future edit using `secrets['FOO']` would silently drop out of the derived set — reopening the exact drift the gate closes. Regexes now match both forms. Guard test added, proven red.

Verified: 27 unit tests pass, ruff clean, and the real `rapid-mac-release.yml` still derives exactly its 11 references (9 secrets + 2 vars), all configured today, so PF-2 is green and does not block the next release.

Accepted-as-designed (not blocking): PF-2 runs on `pull_request` and exposes secrets to bump-PR head code — but forks get an empty secrets context and PF-2 skips them, so the residual is limited to same-repo write-access collaborators who already hold secret access. This matches the fork-skip in the issue AC. Landing after CI.

PR #1862, opened 2026-08-12
perf(kv-cache): fused quantized-KV attention read path — opt-in quant cost drops from −27% to −9% at 16k

### LINKED ISSUE #1751: Shared source-guard stripper is string-unaware: a "//" literal deletes real code from the text gates assert on
`CapabilityChipRenderGateSourceGuardTests.stripCommentsAndWhitespace` removes
`//` and `/* */` without knowing where string literals begin and end. A literal
that contains `//` therefore eats the rest of its line from the canonical text:

```swift
let marker = "//"; doSomethingTheGateForbids()
```

The stripper deletes `; doSomethingTheGateForbids()` as a comment, so any gate
asserting on the stripped text sees code that is not what compiles. The same
applies to a literal containing `/*`, and brace-counting consumers are
additionally fooled by a literal `{` or `}`.

Found while fixing #1748. `ContentViewReadinessActionTests` no longer uses this
helper — it canonicalises with a literal-aware scanner that empties string
literals (and skips `#/…/#` regex literals) first, so no literal can contribute
a brace, a slash, or call-shaped text.

Still on the naive helper, and worth a look:

- `CapabilityChipRenderGateSourceGuardTests` itself — it scans all of
  `Sources/Rapid/` for `capabilityChipKinds` references.

These gates are lower risk than the readiness one was, because they search for
an identifier rather than slice a region, and a literal would have to contain
`//` on the same line as an un-gated `ForEach` to hide it. But the helper's
contract is wrong, and either it should be fixed or its consumers moved to the
literal-aware scanner (which would then want to move out of the readiness test
file into a shared source-guard support type).

raullenchai:
## Why

#1857 made quantized live KV opt-in because the dequant-on-read implementation materialized the layer's full bf16 KV history on every decode step — an O(context) tax measured at −27% (int4) / −36% (int8) at 16k vs bf16. That left `--kv-cache-dtype int4` (4× smaller KV — the memory-constrained-host feature) with a punitive speed cost exactly where it matters most, at long context.

mlx-lm already ships a fused read path (`quantized_scaled_dot_product_attention`, `mx.quantized_matmul` directly on the packed triples) dispatched on `hasattr(cache, "bits")`. This PR routes our batched cache through it.

## What changed

- `QuantizedBatchKVCache.update_and_fetch` returns the quantized triples (was: dequantized bf16), and the cache exposes `bits` / `group_size` — the model's attention dispatches to the fused path with zero model changes.
- mlx-lm's stock quantized SDPA silently misaligns our rank-4 batched mask once GQA expands scores to rank 5 (the mask's batch axis pairs with the repeat axis for B>1). `install_quantized_batch_cache` now also installs a mask-rank-safe wrapper (`mask[:, None]` when rank-4) — a strict superset of stock behavior, verified no-op for stock shapes.
- Prefill untouched: single-sequence prompt caches stay bf16; quantization still happens at the merge into the generation batch, so the quantization boundary remains the prefill→decode transition.
- `extract` / `state` / snapshot round-trip semantics unchanged (already triple-based).

## Measurements (qwen3.5-4b, parity server, live)

| int4 path | 128 ctx | 16k ctx |
|---|---|---|
| dequant-on-read (before) | 161 | 98 (−27% vs bf16) |
| **fused read (this PR)** | 151.5 | **123 (−8.6%)** |
| bf16 reference | 167 | 134.6 |

16k decode **+25%** for quantized-KV users; the residual −9% matches the mlx-lm B=1 fused path (kv_bits probe: int4 134.8 vs bf16 146.6), i.e. we now sit at the library's own ceiling for this composition. Short-context is ~6% below dequant-on-read (composed-op launch overhead vs a trivial dequant) — the trade favors the long contexts quantized KV exists for. bf16 remains fastest overall and stays the default (#1857); a genuinely faster-than-bf16 quantized path needs a fused quantized *flash* kernel upstream (`mx.fast` has none today).

Quality spot-check on the live int4 server produced fully coherent output; the fused-vs-fp32-reference equivalence test carries a negative control proving the mask misalignment it guards against moves outputs by >0.2 (vs <0.05 tolerance).

Closes #1751-tracked follow-up from #1853 discussion.

### DIFF
--- a/tests/test_quantized_batch_cache.py
+++ b/tests/test_quantized_batch_cache.py
@@ -1057,3 +1057,27 @@ def test_mask_patch_is_superset_for_stock_shapes():
         mask[None, None] * mx.ones((1, 1, 1, S), dtype=mx.bool_),
     )
     assert _max_abs(out, ref) < 5e-2
+
+
+def test_install_wires_the_mask_patch(monkeypatch):
+    """install_quantized_batch_cache() must itself install the mask-rank
+    patch — the equivalence tests above invoke the patch function
+    directly, so without this check the install hook could silently stop
+    wiring it and production would regress while tests stay green
+    (codex r4 blocking #2)."""
+    import mlx_lm.models.base as mlx_base
+
+    from vllm_mlx.quantized_batch_cache import install_quantized_batch_cache
+
+    # Reset the idempotence marker so this test observes the wiring
+    # regardless of which test ran first.
+    monkeypatch.setattr(
+        mlx_base, "_rapid_batched_mask_safe_qsdpa", False, raising=False
+    )
+
+    class _FakeBG:
+        def _make_new_cache(self):
+            return []
+
+    assert install_quantized_batch_cache(_FakeBG()) is True
+    assert getattr(mlx_base, "_rapid_batched_mask_safe_qsdpa", False) is True
--- a/tests/test_quantized_batch_cache.py
+++ b/tests/test_quantized_batch_cache.py
@@ -48,6 +48,31 @@ def _max_abs(a, b):
     return float(mx.max(mx.abs(a.astype(mx.float32) - b.astype(mx.float32))))
 
 
+def _deq(cache, triple):
+    """Dequantize an update_and_fetch return (quantized triples, #1751)."""
+    return _dequantize(list(triple), cache._q_group_size, cache._q_bits)
+
+
+def _ref_attention_fp32(queries, k, v, mask):
+    """Hand-composed fp32 GQA attention with an explicitly aligned mask.
+
+    ``mask`` is the rank-4 batched mask ``[B, 1, L, S]``; alignment against
+    the GQA-expanded scores is done here the *correct* way so the fused
+    path can be compared against unambiguous semantics."""
+    q32 = queries.astype(mx.float32)
+    k32 = k.astype(mx.float32)
+    v32 = v.astype(mx.float32)
+    B, n_q, L, Dh = q32.shape
+    n_kv = k32.shape[1]
+    rep_ = n_q // n_kv
+    q32 = q32.reshape(B, n_kv, rep_, L, Dh)
+    scores = q32 @ mx.expand_dims(k32, 2).swapaxes(-1, -2)
+    scores = mx.where(mask[:, None], scores, mx.finfo(mx.float32).min)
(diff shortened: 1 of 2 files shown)


raullenchai:
Codex round-4 responses:

- **Blocking #2 (test wiring)**: fixed in 97-suffix commit — `test_install_wires_the_mask_patch` drives `install_quantized_batch_cache()` end-to-end and asserts the mask patch marker lands on `mlx_lm.models.base`, so the install hook can no longer silently stop wiring the patch.
- **Blocking #1 (return-type contract)**: declining as review drift — this is the third variant of the same concern on the same hunk (r2: stale docstring → fixed and the class docstring now states the triple contract; r4: hypothetical external callers). `QuantizedBatchKVCache` is an internal continuous-batching cache, not a published API; every in-repo consumer routes through `mlx_lm.models.base.scaled_dot_product_attention` immediately after `update_and_fetch` (stock models + vendored muse_glimmer/bailing_hybrid — audited in the class docstring), and the triple return is exactly the contract mlx-lm's own `QuantizedKVCache` has always had, so we now *match* the ecosystem convention rather than diverge from it. The change is also the entire point of the PR: returning bf16 is the O(context) materialization being removed.

raullenchai:
## Merge decision record (codex non-convergence)

Six pr_validate rounds. Objective gates green in every round that ran: 47 cache tests (incl. seeded fused-vs-fp32 equivalence with a negative control), 17.2k full unit suite, lint, supply chain. Codex findings by round: r1 two blockings (mask-gate scope, unseeded control) — **fixed**; r2 two blockings (stale class docstring — **fixed**; mask gate for MHA — factually wrong, stock kernel only forms rank-5 scores under `n_repeats > 1`, invariant now documented with the upstream line cite); r3 TLS infra error; r4 re-raise of the return-contract theme as "external callers" — rebutted (internal class; contract now matches mlx-lm's own `QuantizedKVCache`) plus the test-wiring blocking — **fixed**; r5 `mask=None` default — premise false for upstream 0.31.3 but the superset default was **added anyway**; r6 third re-raise of the same return-contract theme, citing the last stale "drop-in replacement" phrase — that phrase is now **removed** so the docs are fully consistent with the triple contract.

Same-theme blocking raised three times across six rounds, twice after being directly addressed, with every measuring gate green throughout — merging on judgment per the repo's non-convergence SOP, as with #1857.

PR #1866, opened 2026-08-12
perf(engine): retire hybrid admission throttle + coalesce admission waves [skip-version-bump]

raullenchai:
## Why

The #115 hybrid admission throttle (200 ms between inserts, April 2026) was a workaround for ArraysCache corruption during simultaneous batch formation. Its hazard is gone — mlx-lm >= 0.30.6 ships native ArraysCache batch ops (our MambaCache patch is already a no-op) and the corruption no longer reproduces (32/32 simultaneous-formation probe rounds clean on Qwen3.6-35B-A3B, mlx-lm 0.31.3) — while its cost became the single largest concurrency gap in the cross-runtime benchmark: rows admitted at different scheduler steps carry ragged per-row offsets for the batch's whole lifetime, keeping mlx-lm's batched attention on the array-mask slow path every decode step (next() p50 29.5 ms vs 20.3 ms aligned at B=8).

## What

1. **Throttle retired** (`_resolve_hybrid_throttle`): default OFF; `RAPID_HYBRID_THROTTLE=1` re-enables as an escape hatch for unsupported mlx-lm builds. Also removes the 200 ms TTFT tax on every hybrid request after the first (35B decode_b1 TTFT 0.45 s → 0.246 s).
2. **Admission-wave coalescing** (`EngineCore._await_admission_wave`): the first step after idle waits tick-wise (default 8 ms tick / 120 ms cap, env-tunable, garbage/non-finite env values fall back safely) while a burst of submissions is still landing, so co-arriving requests prefill as ONE aligned wave. Runs on the event loop — `add_request` lands via the same single-worker mlx-step executor that runs `scheduler.step()`, so a scheduler-side wait would deadlock against its own queue.
3. **Server warmup hybrid-skip decoupled from the throttle flag** (codex round-1 MAJOR): warmup now asks the engine's fail-closed `_is_hybrid_model()` profile probe instead of reading `_hybrid_throttle` as an is-hybrid proxy.

## Measured (cross-runtime bench, parity lane, conc_8 512/256, M3 Ultra)

- qwen3.6-35b-a3b: 190 → **267.7 agg tok/s (+41 %)**, ahead of mlx-lm's 225 on the same harness
- decode-phase aggregate 374 ≈ in-process BatchGenerator ceiling (376–379): server-side batching overhead is now zero
- Corruption probe: 32/32 simultaneous-formation rounds clean

## Tests

`tests/test_admission_wave.py` (20): throttle resolver defaults/env, window tick/extend/cap semantics, wiring guard (`add_request` bumps the counter — proven red without the increment), garbage + inf/nan env fallback, warmup-probe pin. Engine blast-radius suites (zombie-KV, batching, MLLM cancel, metal recovery) all green.

## Codex review

Round 1: 1 MAJOR (warmup coupling — fixed) + 4 MINOR + 1 NIT. Round 2: **0 BLOCKING / 0 MAJOR**; two residuals (inf/nan guard, warmup regression test) fixed in the same commit; two MINORs deferred with rationale codex accepted (mlx-lm floor is version-pinned; one-tick TTFT cost documented + env-disable).

### DIFF
--- a/tests/test_admission_wave.py
+++ b/tests/test_admission_wave.py
@@ -234,6 +234,17 @@ def test_warmup_detection_mllm_excluded():
     assert _detect_hybrid_for_warmup(_WarmupEngine(is_mllm=True)) is False
 
 
+def test_warmup_detection_mllm_wins_over_hybrid_probe():
+    """MLLM exclusion runs BEFORE the hybrid probe (pr_validate codex r3
+    BLOCKING): an MLLM engine whose probe answers True must still take
+    the bare-warmup path. Unreachable today (hybrid VLMs auto-downgrade
+    to the text lane, #352) but must fail closed if that ever changes."""
+    from vllm_mlx.server import _detect_hybrid_for_warmup
+
+    eng = _WarmupEngine(hybrid_probe=lambda: True, is_mllm=True)
+    assert _detect_hybrid_for_warmup(eng) is False
+
+
 def test_add_request_bumps_admission_seq():
     """Wiring guard: ``add_request`` must move the counter the window
     watches — the increment sits at the very top of the method, before
--- a/tests/test_admission_wave.py
+++ b/tests/test_admission_wave.py
@@ -91,7 +91,11 @@ def sleep_spy(monkeypatch):
 
 
 def _run(coro):
-    return asyncio.new_event_loop().run_until_complete(coro)
+    loop = asyncio.new_event_loop()
+    try:
+        return loop.run_until_complete(coro)
+    finally:
+        loop.close()
 
 
 def test_lone_request_pays_one_tick(sleep_spy, monkeypatch):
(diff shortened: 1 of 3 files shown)


raullenchai:
## pr_validate summary (4 rounds, final @ 5b24b6e0)

| gate | result |
|---|---|
| codex_review | **PASS** — 0 blocking (2 nits, accepted per SOP) |
| lint / targeted / full_unit | PASS — 200 targeted, 17264 full suite |
| stress_e2e_bench (4-model matrix) | FAIL on 2 bench findings — **overridden, see below** |
| CI (lint, mlx-bound-guard, test-apple-silicon, 3.10-3.12 matrix, l1-smoke ×5) | PASS |

Codex rounds fixed along the way: admission-window cap overshoot; post-hold empty-scheduler step re-check; warmup hybrid detection extracted to `_detect_hybrid_for_warmup()` with behavioral tests; MLLM exclusion ordered before the hybrid probe (r3) — each with a guard test proven red without its fix.

### Stress-gate override rationale

The two findings (Qwen3.5-35B cold +36.6%/warm +11.5%, gpt-oss cold +21.6%/warm +15.7% vs the *recorded* baseline) are baseline staleness, not this diff:

- Round 1 measured base in-run on this machine: base alone reproduces cold +35.3%/warm +9.3% (35B) and +31.1%/+15.1% (gpt-oss) of the drift. PR deltas **relative to measured base**: 35B cold +2.2%/warm +3.2%, gpt-oss cold +4.7%/warm +3.4% — all within thresholds. Round 1 accordingly classified both as PRE-EXISTING.
- Round 4 could not re-measure base: the manifest's base ref `28798a29` became unreachable after an early force-push (process artifact; later review fixes were pushed as new commits for exactly this reason).
- Stress + agent matrix (anthropic_sdk / langchain / pydantic_ai × 4 models) passed on all rounds; the other two models (Qwen3.6-27B, diffusion-gemma) passed bench outright.

Follow-up: refresh the recorded stress bench baselines on this machine (drift is machine-wide, documented in rounds 1 and 4 artifacts).

PR #1873, opened 2026-08-12
fix(cli): connect setup commands carry remote endpoint, IPv6 bracketing, port validation

raullenchai:
Re-introduces the unified server ready/connect SSOT feature (#1871 was reverted in #1872) with the reviewer's P1 and minor feedback applied across three review rounds.

## Background
#1871 added `vllm_mlx/connect.py` as the single source of truth for `Ready:`/`Connect:` output and new `rapid-mlx connect` / `connect --json` commands. It was reverted in #1872 pending review fixes. This PR restores the feature (cherry-picked) and applies the review feedback.

### P1 — Remote `--host/--port` now flows into the suggested setup command
`rapid-mlx connect claude-code --host mini.local --port 9000` previously only showed the address in the header but suggested a bare `rapid-mlx agents claude-code --setup` with no `--base-url`. Now it prints:
```
rapid-mlx agents claude-code --setup \
  --base-url http://mini.local:9000/v1
```
Same for `continue`. All agents CLIs uniformly accept the OpenAI-style `/v1` base URL.

### P1 — IPv6 literals render valid bracket-wrapped URLs
`http://::1:8000` → `http://[::1]:8000`, driven by a shared `_authority()` helper (same lexical rule as the serve CLI's `_is_ipv6_host`).

### Minor — `connect --port` reuses the validated `_port_arg`
Rejects `0`, negatives, and `>65535` with a clear argparse error instead of producing an invalid URL.

### P1 — Ready banner's Connect: section now carries the real endpoint
The *serve banner itself* (via `render_banner`) previously still printed bare `rapid-mlx agents claude-code --setup` / `agents continue --setup` — a user who copied from the Ready banner would configure localhost instead of the running server. Both agent setup commands now append `--base-url <openai_url>` dynamically from the SSOT, so the served banner and `connect <target>` agree.

### P1 — Explicit `--model` is never overwritten by a live probe
`connect --model X` (on default address) previously still probed the server and could clobber the user's explicit model. The live model probe now runs only when no model was supplied anywhere (flags > config > defaults), so an explicit `--model` always wins.

### Minor — scoped IPv6 zone-ids are percent-encoded (`%25`)
RFC 6874: `http://[fe80::1%en0]:8000` → `http://[fe80::1%25en0]:8000`, so scoped URLs stay well-formed.

### P1 — Dynamic `--base-url` in copied commands is shell-quoted
Both the Ready banner (`render_banner`) and `connect <agent>` output (`_print_point_command`) build commands a user copies and pastes into a shell. A crafted `--host` value such as `'mini.local; touch /tmp/pwned'` would otherwise be interpolated raw, letting the shell interpret the semicolon as a command separator. Every dynamic `--base-url` is now passed through `shlex.quote()`, e.g.:
```
rapid-mlx agents claude-code --setup \
  --base-url 'http://mini.local; touch /tmp/pwned:8000/v1'
```

### Minor — `_authority()` canonicalizes non-canonical host input
- An already-bracketed `[::1]` is no longer double-bracketed as `[[::1]]`.
- An already-encoded `%25` zone-id is left alone, not encoded a second time to `%2525` (case-insensitive).

## Tests
Added/updated regression tests covering: remote endpoint passthrough for `claude-code`/`continue`; banner `Connect:` rows carrying `--base-url` (remote, IPv6, default localhost); IPv6 literal + scoped `%25` bracketing (endpoint derivations, banner, JSON); invalid port rejection/valid port acceptance; explicit-model-preserved vs. probe-when-no-model for `resolve_endpoints`; and shell-quoting for malicious hosts (spaces, `;`, `$(...)`, single quotes) plus IPv6 zone-id / already-bracketed normalization across both the banner and point-command paths.

## Note on Scope
SSOT currently unifies the Python serve banner + `connect --json`. Desktop does not consume it yet — this is documented future work, not part of this PR's scope.

### DIFF
--- a/tests/test_cli_connect.py
+++ b/tests/test_cli_connect.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import argparse
 import io
 import json
+import shlex
 from contextlib import redirect_stdout
 
 import pytest
@@ -306,7 +307,8 @@ def test_render_banner_connect_carries_remote_endpoint():
 
 def test_render_banner_connect_carries_ipv6_endpoint():
     out = connect.render_banner(connect.ServerEndpoints("::1", 8000, model="m1"))
-    assert "--base-url http://[::1]:8000/v1" in out
+    # IPv6 literals contain '['/']', which shlex.quote wraps in single quotes.
+    assert "--base-url 'http://[::1]:8000/v1'" in out
 
 
 def test_resolve_endpoints_preserves_explicit_model(monkeypatch):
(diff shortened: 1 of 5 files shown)


raullenchai:
## 🤖 Codex review — round 1

- [vllm_mlx/connect.py:98](/home/raullenspark2/reviews/Rapid-MLX/vllm_mlx/connect.py:98): IPv6 wildcard binds are rendered as a client endpoint. With `rapid-mlx serve --host ::`, the ready banner and copied setup commands use `http://[::]:PORT`; `::` is an unspecified bind address, not a destination clients can connect to. Normalize it to a usable display host such as `::1`, analogous to the existing `0.0.0.0 → localhost` handling, and add an end-to-end banner test.

- [vllm_mlx/cli.py:7835](/home/raullenspark2/reviews/Rapid-MLX/vllm_mlx/cli.py:7835): The generated OpenAI example always uses `api_key='sk-noop'`, while the server supports `serve --api-key`. Against such a server, the advertised example deterministically receives 401. The Claude/Continue setup commands emitted at [vllm_mlx/connect.py:143](/home/raullenspark2/reviews/Rapid-MLX/vllm_mlx/connect.py:143) have the same problem because their setup flow also writes `sk-noop`. The connect API needs a secure credential-input mechanism and authenticated-server tests; the key should not be printed in the server banner.

VERDICT: CHANGES_REQUESTED

Issue #1861, opened 2026-08-12
perf: dense/big-model batched decode does not scale — 27B conc_4 is SLOWER than B=1 (mlx-lm on the same BatchGenerator scales 35→50)


raullenchai:
From the 2026-08-11/12 cross-runtime benchmark (post-#1857 re-run, engine main@dde006ca, M3 Ultra):

| qwen3.6-27b (parity) | B=1 | conc_4 (agg) | conc_8 (agg) |
|---|---|---|---|
| rapid-mlx | 38 | **32** | 34 |
| mlx-lm server (same BatchGenerator) | 35 | 45 | 50 |

Two independent signals:

1. **Batching makes the 27B *slower* than single-stream for us** (32 agg < 38 B=1) — pathological; batched dense decode should gain from arithmetic intensity, and mlx-lm proves the underlying `BatchGenerator` scales (+43% at 8-way).
2. Same shape, milder, on qwen3.6-35b-a3b: conc_8 190 vs mlx-lm 226.

Small/MoE models are fine (4B conc_8 272 — we *lead* mlx-lm there), so this is size- or family-correlated, not a general batching defect.

**Where to look**
- Dense-sampler fastpath engagement at B>1 (`_install_dense_sampler_fastpath`): requires all per-request sampler *instances* identical via `_get_request_sampler` interning — if interning misses, every step pays a per-row sampler loop + 8 slice/concat kernels.
- Chunked-prefill vs decode interleave under our scheduler's admission pattern vs mlx-lm server's insert pattern.
- Config deltas vs `mlx_lm server` defaults (`--decode-concurrency` / `--prompt-concurrency` / `--prefill-step-size`).
- Per-step scheduler bookkeeping ×8 at the big-model step cadence.

**Repro**: `benchmarks/cross-runtime` (branch `bench/cross-runtime`): `python3.12 run.py --lane parity --scenarios decode_b1,conc_4,conc_8 --models qwen3.6-27b --runtimes rapid,mlxlm`.

Method that closed #1853 (layer bisect: BatchGenerator → Scheduler → engine → HTTP) applies directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

raullenchai:
## Root-caused — the headline gap is a measurement artifact; two real findings remain

Full layer bisect (in-process BatchGenerator → Scheduler → live server, same models/prompts):

**1. The 27B conc_8 "34 vs 50" is NOT a decode/batching deficiency.**
[...]

PR #1874, opened 2026-08-12
perf(scheduler): singleton KV-cache fast path + B=1 dense-sampler engage [skip-version-bump]

raullenchai:
## Why

Cross-runtime bench (#1861 investigation) showed our B=1 decode trailing oMLX by 5-8% at 4k/16k. Two scheduler-side costs, each measured by direct A/B on Qwen3.5-4B / 16k-context:

1. **Singleton KV-cache fast path** (`a71aa65f`): at B=1 the batched cache assembly (per-step gather/scatter across the row dimension) is pure overhead — a single running row can borrow the contiguous cache view directly. Scheduler `step()` p50 6.9 → 5.72ms on 4B.
2. **Dense-sampler engage at B=1** (`8cafa844`): the dense sampler fast path (#521) was gated on batch size >1; a lone stream fell through to the generic per-row sampler. Engaging it at B=1 trims the remaining per-token sampling cost.

Combined result: B=1 scheduler step at oMLX parity — 4B 5.72ms/tok (174.8 vs oMLX 175 tok/s), 16k 6.78ms/tok (147.5 vs 149). Remaining client-visible gap (~0.32ms/tok) is SSE delivery, deliberately out of scope here.

`b7fb2569` syncs the dense-sampler behavior list in test docs with the B=1 engage.

## Testing

- `tests/test_singleton_cache_fastpath.py` (new, 15 tests): borrow/return lifecycle, fallback to batched assembly on B>1 and on mid-batch join, cache-view aliasing safety, eviction interplay.
- 336 scheduler/sampler/admission tests green locally; full suite via pr_validate.
- Bench evidence: tune4/tune9 runs in benchmarks/cross-runtime (FINAL.md "Overnight tuning — 2026-08-13").

### DIFF
--- a/tests/test_dense_sampler_fastpath.py
+++ b/tests/test_dense_sampler_fastpath.py
@@ -15,9 +15,9 @@ have to load a model. They lock in the four behaviors that matter:
    (``self.samplers`` is observed as all-None inside the wrapped call;
    ``self.fallback_sampler`` is the shared sampler).
 2. Heterogeneous batch → leaves ``self.samplers`` untouched.
-3. B=1 → no swap (degenerate case; identity-equality with empty rest is
-   true but the patch must NOT engage since the fast-path savings only
-   exist for B ≥ 2).
+3. B=1 → swaps too (a single per-request sampler still trips mlx-lm's
+   ``any(self.samplers)`` per-row branch; the bench-tuning pass measured
+   real savings at B=1).
 4. Swap is reversed after the call returns — even on exception — so the
    per-request samplers are restored for the NEXT step.
 """
--- a/tests/test_dense_sampler_fastpath.py
+++ b/tests/test_dense_sampler_fastpath.py
@@ -94,9 +94,12 @@ def test_heterogeneous_batch_keeps_per_request_samplers():
     assert gb.observed_fallback is original_fallback
 
 
-def test_b1_does_not_engage_fast_path():
-    """B=1 is degenerate — patch must NOT swap (no perf upside, and
-    swapping just adds attribute writes per step)."""
+def test_b1_engages_fast_path():
+    """B=1 must swap too (bench-tuning 2026-08-12): a single per-request
+    sampler still trips mlx-lm's ``any(self.samplers)`` per-row branch —
+    one slice, one closure call, one single-element concatenate every
+    decode step. The earlier exclusion assumed "no perf upside", which
+    the B=1 decode benchmarks disproved."""
     sampler = lambda x: x  # noqa: E731
     original_fallback = lambda x: x  # noqa: E731
     gb = _FakeGenBatch(samplers=[sampler], fallback=original_fallback)
@@ -104,8 +107,13 @@ def test_b1_does_not_engage_fast_path():
 
     gb._step()
 
-    assert gb.observed_samplers == [sampler]
-    assert gb.observed_fallback is original_fallback
+    # Fast path observed inside the call: row slot cleared, sampler
+    # promoted to the shared fallback...
+    assert gb.observed_samplers == [None]
+    assert gb.observed_fallback is sampler
+    # ...and state restored afterwards.
+    assert gb.samplers == [sampler]
+    assert gb.fallback_sampler is original_fallback
 
 
 def test_homogeneous_with_first_none_does_not_engage():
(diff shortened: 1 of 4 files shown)


raullenchai:
## pr_validate summary (4 rounds, final @ f68cf724)

| gate | result |
|---|---|
| lint / targeted / full_unit | PASS — 114 targeted, 17297 full suite (r4) |
| stress_e2e_bench (4-model matrix) | **PASS** (r4; 3 pre-existing stale-baseline notes) |
| codex_review | 10 findings fixed across r1–r3; r4's 2 findings refuted — judgment call below |

### Codex findings fixed (each with a guard test proven red pre-fix)

r1: subclass re-admission via the self-defeated hasattr probe (BLOCKING); process-global class monkeypatch flipping MLLM capability gates (MAJOR); extract shallow-clone aliasing live rows (MAJOR); drain-path docstring correction (NIT).
r2: `filter([1])` silently keeping row 0; composite wrappers admitted without their own batch surface; extend layer-count mismatch now raises; test fixture restores patched seams.
r3: composite promotion clones the wrapper instead of assuming a splat constructor; empty-batch extend path binds the surface defensively.

### r4 findings — refuted, not fixed (codex non-convergence judgment, round 4 of a 2-round SOP)

1. *"empty cache_a returns cache_b with non-qualifying layers unbatched"* — the proposed fix (`orig_extend`) is behaviorally identical: stock `_extend_cache` also returns `cache_b` unchanged when `cache_a` is empty. Stock parity, no defect.
2. *"mx.contiguous may alias when the slice spans the full buffer"* — refuted empirically (MLX setitem is functional; a prior value node never observes it — verified eager AND lazy), and mlx-lm's own `BatchKVCache.extract` relies on the identical idiom. Pinned as `test_extract_full_span_buffer_still_independent` (f68cf724, test-only).

### Stress note

r3's gpt-oss warm +132% did not reproduce in r4 (PASS); that round's gpt-oss base measurement was itself −52.5% off the recorded baseline while 35B stayed clean — measurement instability, not a diff effect. The recorded baselines still need a machine-drift refresh (also noted on #1866).

Final commit f68cf724 is test-only on top of the r4-validated 8b1577c3; a fifth validator round was deliberately not run.

raullenchai:
## pr_validate summary (6 rounds, final @ 0e10d9fe)

| gate | result |
|---|---|
| lint / targeted / full_unit | PASS — 118 targeted, 17301 full suite (r6) |
| stress_e2e_bench (4-model matrix) | **PASS** (r6; 2 stale-baseline notes) |
| l1-smoke (CI) | **llama3-3b green since b1a2421e** — the real regression this cycle caught and fixed |
| codex_review | 15 findings fixed r1–r6; 4 refuted — judgment below |

### The one real bug: prefix-cache poisoning (b1a2421e)

CI's l1-smoke caught llama3-3b answering the France golden with 'Transparent.' — deterministic, bisected to the singleton install. Root cause: `memory_cache`'s supersequence/LCP loans share the stored entry's arrays with only the offset rewound, a contract that silently relied on stock batched merging *copying* per-request caches at admission. The singleton pass-through kept the loaned object live and decode's in-place writes corrupted the stored prefix entry. Fix: copy-on-admit (`_detach_layer`) — data-carrying layers admit as independent copies, empty layers admit unchanged (steady-state B=1 win untouched). Verified: coherence gate 5/6→6/6 locally with the fastpath active; CI l1-smoke green on all 5 aliases since.

### Fixed hardening findings (r1–r6, each guard-tested red pre-fix)

subclass re-admission; instance-scoped surface (no class monkeypatch); extract copies (batched contract); `filter([1])` row validation; composite own-surface requirement; extend length guard; wrapper clone via `copy.copy` (splat-ctor and `__slots__` safe); native-layer deepcopy on admit; all-or-nothing extend admission; install lock.

### Refuted findings (documented, unfixed by design)

- r4 "empty cache_a returns cache_b unbatched" — proposed fix (`orig_extend`) is behaviorally identical to current code (stock also returns `cache_b` unchanged).
- r4 "mx.contiguous may alias on full-span" — refuted empirically eager+lazy (MLX setitem is functional); pinned as a test anyway.
- r6 "B=1 sampler swap changes custom sampler input rank" — the swap engages only when every row holds the *same* sampler object, and at B=1 the per-row slice and the whole batch are the identical `(1, V)` tensor.
- r6 "add a rank-sensitive stub test for the swap" — a stub-based `_step` test would guard the stub, not mlx-lm; E2E coverage comes from the golden gates.

### Non-convergence note

codex emitted new findings every round on a growing diff (1→3→2→2→3→4) — classic non-convergence per SOP; r6 was declared final, its two genuine items fixed (0e10d9fe), the rest refuted above. Stress r3's gpt-oss warm outlier and r5's langchain 5/6 are cross-round model-strength flakes (same case, same signature, both directions across rounds), documented in run artifacts.

PR #1878, opened 2026-08-12
perf(engine): coalesce scheduler steps + has_tools logits gate + perf debug aids [skip-version-bump]

raullenchai:
## Why

Third and final PR from the #1861 conc investigation (after #1866 throttle retirement and #1874 singleton fastpath). Two perf changes plus the env-gated debug tooling used to find them:

1. **Step coalescing** (`278f210d`): at high aggregate token rates the per-step executor round-trip plus GIL contention with stream tasks costs ~10ms/step (35B-A3B conc_8: in-process 375 agg tok/s vs 253 live). With ≥4 running requests the engine loop runs up to 4 steps per executor dispatch; `_step_coalesced` breaks early on finishes, pending admissions, or no-work, so request lifecycle latency is unchanged and mid-stream tokens buffer at most ~3 extra steps. Composes with #1866's admission wave: wave hold → re-check → coalesced dispatch.
2. **has_tools logits-factory gate** (`af2be1a5`): tool-call logits processors were constructed per request regardless of whether the request carries tools; now gated on `request.has_tools` — pure-chat requests skip the factory entirely.
3. **Debug aids** (env-gated, zero cost when off): `RAPID_STEP_TIMING` step-timing histogram, `RAPID_PYSAMPLE` in-process sampling profiler hooked into `cli serve`, `RAPID_DUMP_SCHED_CONFIG` SchedulerConfig dump — the instrumentation that located the ragged-admission and coalescing wins.

## Testing

- 404 engine/scheduler/sampler/admission/singleton tests green locally on the merged result (includes #1866's admission-wave suite and #1874's singleton suite — the conflict region composes all three).
- Bench evidence: tune6/tune10 runs in benchmarks/cross-runtime (FINAL.md "Overnight tuning — 2026-08-13"); 35B conc_8 267.7 agg tok/s vs mlx-lm 225.
- Full suite + 4-model stress via pr_validate.

### DIFF
--- a/tests/test_step_coalescing.py
+++ b/tests/test_step_coalescing.py
@@ -22,16 +22,23 @@ def _out(finished=(), has_work=True):
 
 
 class _ScriptedScheduler:
-    """Yields scripted step results; an Exception instance raises."""
+    """Yields scripted step results; an Exception instance raises.
 
-    def __init__(self, script, waiting=0):
+    ``consume_waiting`` mirrors the real scheduler: step() ADMITS the
+    waiting queue into the batch, so the count reads 0 afterwards.
+    """
+
+    def __init__(self, script, waiting=0, consume_waiting=False):
         self.script = list(script)
         self.calls = 0
         self._waiting = waiting
+        self._consume = consume_waiting
 
     def step(self):
         self.calls += 1
         item = self.script.pop(0)
+        if self._consume:
+            self._waiting = 0
         if isinstance(item, Exception):
             raise item
         return item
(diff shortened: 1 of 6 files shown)


raullenchai:
## pr_validate summary (6 rounds, final @ d8215bb8 — Verdict: MERGE-SAFE)

| gate | result |
|---|---|
| codex_review | **converged: 0 blocking (r6)** |
| lint / full_unit | PASS — 17397 full suite |
| stress_e2e_bench (4-model matrix) | PASS all rounds (2 stale-baseline notes, same as #1866/#1874) |

Codex findings fixed across r1–r5, each with a red-proven guard where stageable: partial coalesced outputs delivered before surfacing a later step's error (+ new `tests/test_step_coalescing.py` — the helper previously had zero coverage); pysample counter lock + report serialized + unwritable-path warning; `RAPID_STEP_TIMING` resolved once at init (hot path genuinely zero-cost when off; PR body env-name typo fixed); waiting-queue snapshot before each step (step() admits the queue, so the post-check was blind — TTFT of a just-admitted request no longer waits out the batch); `_coalesce_budget` keeps the memory-pressure check on its exact pre-coalescing 16-step cadence; successful steps reset the failure streak (ok,ok,FAIL batches no longer accumulate into the fatal handler).

Declined as designed: an engine-loop-level collector test (stub would guard the stub) and stub-based sampler rank test — E2E coverage comes from the golden l1-smoke gates.

Issue #571, opened 2026-06-14
tool_choice=required / forced returns 422 on hermes (qwen3.6) but works on harmony (gpt-oss-20b)


raullenchai:
## Severity
P2 — inconsistent client contract across model families. Apps that hard-code \`tool_choice={type:function,function:{name:...}}\` to force a specific tool will work for some aliases and 422 for others, with no schema-level signal of which.

## Repro

Two requests with identical shape, only the \`model\` field changes:

### Request

\`\`\`bash
curl -sS -X POST http://127.0.0.1:8001/v1/chat/completions \\
  -H "Content-Type: application/json" \\
  -d '{
    "model": "<ALIAS>",
    "messages": [{"role":"user","content":"Hi."}],
    "tools": [
      {"type":"function","function":{"name":"web_search","description":"Search the web.","parameters":{"type":"object","properties":{"query":{"type":"string"}},"required":["query"]}}}
    ],
    "tool_choice": {"type":"function","function":{"name":"web_search"}},
    "max_tokens": 200
  }'
\`\`\`

### Behaviour

| Alias | \`tool_call_parser\` | Result |
| --- | --- | --- |
| \`qwen3.6-35b-4bit\` | hermes | **HTTP 422** |
| \`gpt-oss-20b-mxfp4-q8\` | harmony | **200 OK, fires \`web_search\` with \`{\"query\":\"Hi\"}\`** |

Same shape with \`tool_choice: \"required\"\` shows the same hermes-vs-harmony split.

## Why this matters

The OpenAI \`/v1/chat/completions\` contract treats \`tool_choice\` as a server-enforced gate. A typed-name forced choice is a way for clients to disable model-side tool routing and pin the call. With this asymmetry:

- A client that selects \`web_search\` via UI (\"Search the web for…\" intent button) and forces \`tool_choice\` for reliability succeeds on harmony-class models but 422s on hermes-class — same client code, different server outcome.
- The 422 vs 200 split happens silently; the schema is identical, no \`tools\` or \`tool_choice\` field signals incompatibility.

## Suspected location

The hermes tool-call parser path (likely under \`vllm_mlx/routes/chat\` / \`vllm_mlx/parsers/hermes\`) appears to reject forced-name \`tool_choice\` shapes that the harmony path accepts. Compare the two parsers' request-validation hooks.

[...]

Issue #447, opened 2026-05-23
Gemma 4 OutputRouter: 'thought' marker + body leak into content on unconstrained generation


raullenchai:
## Summary
Gemma 4 channel routing on `mlx-community/gemma-4-26b-a4b-it-4bit` (and likely the whole Gemma 4 family) misroutes channels on the unconstrained generation path: the literal token `thought` plus the analysis body land in `message.content`, while `reasoning_content` stays empty. On pure-content streaming (no tools, no JSON mode), the router never transitions out of `thought` and the model gets trapped — `content` stays empty even at 400+ tokens until the `length` cap fires.

The constrained-decoding path (`response_format: {type: json_object}`) works correctly — JSON mode forces the channel transition, so the bug is isolated to the **unconstrained** channel state machine.

Surfaced on 2026-05-22 fresh-PyPI v0.6.65 onboarding sweep. Distinct from #442 / #444 / #445 (all harmony-family).

## Reproduction
```bash
python3.12 -m venv /tmp/gemma4-repro && \
  /tmp/gemma4-repro/bin/pip install rapid-mlx==0.6.65
/tmp/gemma4-repro/bin/rapid-mlx serve \
  mlx-community/gemma-4-26b-a4b-it-4bit \
  --port 8501 --enable-auto-tool-choice --tool-call-parser gemma4 &

# Case A: multi-turn — `thought\n…` leaks into content
curl -s http://127.0.0.1:8501/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"mlx-community/gemma-4-26b-a4b-it-4bit",
  "messages":[{"role":"user","content":"what is 2+2?"}],
  "max_tokens":200
}' | jq '{content: .choices[0].message.content, reasoning: .choices[0].message.reasoning_content}'
# Observed: content starts with "thought\n..." (analysis prose), reasoning_content is null

# Case B: pure content streaming — `content` stays empty
curl -sN http://127.0.0.1:8501/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"mlx-community/gemma-4-26b-a4b-it-4bit",
  "messages":[{"role":"user","content":"write a one-paragraph essay about the moon"}],
  "stream": true,
  "max_tokens":400
}' | head -50
# Observed: every delta routed to reasoning_content; content deltas empty until length cap

# Case C: JSON mode — works correctly
[...]

PR #1880, opened 2026-08-12
fix(tool-choice): repair forced-call arguments that are not JSON objects [skip-version-bump]

raullenchai:
## Why

Post-merge dogfood (2026-08-12) on Qwen3.5-35B-A3B-8bit / hermes: a forced `tool_choice` (named function) returned `"arguments": "1"` — the injected prefix `..."arguments": ` was continued by the model with a bare scalar at temp 0. Reproduced 3/3 deterministically, and identically on pre-merge main (pre-existing, not from #1866/#1874/#1878). The OpenAI function-calling wire requires `arguments` to be a JSON object; every client `json.loads`es a non-dict and breaks.

## What

1. **Non-stream**: new `_repair_forced_call_arguments` — when every emitted call names the pinned target but carries non-object arguments, repair to the #571 synthesis fallback's semantics (recover an object from raw text, else `"{}"`), then hold the result to the #1256 schema gate (422 on `required` violations). Wrong-name calls keep the existing 422.
2. **Stream**: the #447 streaming-parity synthesis was gated OFF whenever `tool_calls_detected=True` (PR #948 guard against replacing a *different* tool's call) — but the hermes streaming state machine emits **zero deltas** for the unusable body, leaving the turn empty. Relaxed precisely: synthesis fires when the wire text names exactly the pinned target (repairing THE forced call); any other name still refuses, so the #948 guard test stays green.

## Testing

- Live 3/3: `scripts/l1_toolcall_check.py` on 35B went FAIL(non-stream) → FAIL(stream) → **PASS(both)** across the two halves of the fix; gpt-oss-20b harmony path unchanged (PASS).
- 123 forced-choice unit tests green (447 / 1256 / enforcement suites + 5 new repair tests incl. the #1256 required-schema surface).

### DIFF
--- a/tests/test_forced_call_args_repair.py
+++ b/tests/test_forced_call_args_repair.py
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Forced ``tool_choice`` calls must carry OBJECT arguments on the wire.
+
+Dogfood 2026-08-12 (post #1866/#1874/#1878 merge, pre-existing on main):
+the hermes forced-choice prefix hands the model ``..."arguments": ``
+mid-envelope and Qwen3.5-35B-A3B-8bit at temp 0 continues with a bare
+``1`` — the parser surfaces ``arguments="1"``, and every OpenAI client
+``json.loads``\\ es a non-dict and breaks. The repair mirrors the #571
+synthesis fallback (recover from raw text, else ``"{}"``) and holds the
+result to the #1256 schema gate.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from typing import Any
+
+from vllm_mlx.routes.chat import _repair_forced_call_arguments
+
+
+def _call(arguments):
+    return SimpleNamespace(
+        id="call_x",
+        type="function",
+        function=SimpleNamespace(name="release_probe", arguments=arguments),
+    )
+
+
+def _tool(name: str = "release_probe", required: list[str] | None = None):
+    schema: dict[str, Any] = {"type": "object", "properties": {}}
+    if required is not None:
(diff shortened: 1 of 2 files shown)


raullenchai:
## pr_validate summary (4 rounds, final @ this SHA)

| gate | result |
|---|---|
| lint / targeted (510) / full_unit (17406) | PASS (r4) |
| stress_e2e_bench | PASS r4 (r3's pydantic_ai 6_multi_tool 5/6 is the documented cross-round gpt-oss flake — same code passed the matrix in r1/r2/r4, and the case runs auto tool-choice, a path this PR does not touch) |
| codex_review | 9 findings fixed r1–r4; 2 refuted below |
| live E2E | `l1_toolcall_check` 3/3 PASS on Qwen3.5-35B (was 3/3 FAIL pre-fix); gpt-oss harmony unchanged PASS |

Fixed: payload-free repair logging; envelope-decoded (escape-proof, prose-proof) streaming gate; recovery restricted to lone-call turns (no sibling-argument lifting, no shared recovery); dict arguments normalized to the JSON-string wire shape with content preserved; exactly-one-envelope requirement so synthesis can never collapse multiple attempts.

Refuted: r2 "dict arguments not re-serialized" as originally framed (dict inputs predate this fix and flowed identically — later normalized anyway in r3's scope); r3's nested-"name" NIT (described the r1 regex; the envelope decoder reads only top-level names). Declined: stub-based streaming-branch test — the live l1 gate is the end-to-end verification; a stub stream would guard the stub.

Final commit is a 1-line conservative gate narrowing (+ unit test) on top of the r4-validated SHA; full_unit and stress cover the unchanged remainder.

PR #1884, opened 2026-08-12
fix(image-gen): refuse partially-downloaded mflux checkpoints

Jevin-F:
## What does this PR do?

Verifies mflux component weights against their `model.safetensors.index.json` before the checkpoint is loaded, and refuses to load when shards are missing.

- `mflux_missing_weights()` in `_download_gate.py` — three states: complete (`[]`), the missing files by name, or `None` for "no verdict" (not a registered image-gen alias, nothing cached yet, dependency unavailable). `_snapshot_is_complete_mflux_model` keeps its boolean shape for the `ls`-style callers.
- `ImageGenerationEngine._ensure_loaded` gates on it — the point every load path converges on.
- `serve` exits non-zero at startup for an incomplete image model instead of booting.
- The pre-download now short-circuits on verified-complete component weights.

## Why is this needed?

mflux loads whatever `*.safetensors` sit in a component directory and never reads the index beside them, so an interrupted pull is silent: the shards that arrived load, the rest of the transformer keeps its randomly-initialised weights, and generation renders noise. Nothing on any path raises.

Observed with `Runpod/FLUX.2-klein-4B-mflux-4bit` at 0.32 of 4.30 GiB — `transformer/0.safetensors` and `text_encoder/0.safetensors` absent, model started normally, every image came out as noise. Re-downloading the two shards fixed it.

The fourth change is a separate symptom of the same layout gap: `is_repo_cached` only probes for root `model*.safetensors`, which mflux checkpoints never have, so a fully-downloaded image model failed that check and reached the network on every start.

## Test plan

- `tests/test_download_gate.py` — the three states, including that an unavailable `huggingface_hub` returns "no verdict" rather than "incomplete"
- `tests/test_image_lane.py` — a partial snapshot raises and `_build_model` is never called; complete and not-yet-cached both load
- `tests/test_serve_image_gen_completeness.py` — `serve` exits 1 naming the missing file; a complete checkpoint starts without touching the network
- `python3 -m pytest tests/ --ignore=tests/integrations --ignore=tests/property` — 17079 passed; pre-existing failures in `test_qwen3coder_xml_grammar_558.py` and `test_streaming_detokenizer.py` are unrelated (missing optional deps)
- `ruff check` and `ruff format --check` clean

Not verified end-to-end in the desktop UI: the rejection path is covered by tests only, since reproducing it live requires deliberately truncating a local checkpoint.


raullenchai:
Landed as #1885 (squashed into `aff8f462`) — rebased onto current `main` with your authorship preserved as commit author, plus one reviewer hardening folded in (validate `weight_map` values are strings before `set()`/`sorted()`, so a malformed index fails closed instead of raising `TypeError` from the ungated load-time/serve-time paths). Thanks @Jevin-F — nice catch on the silent noise-render; your machine's own cached Z-Image checkpoint turned out to be partial in exactly the way this guards. 🙏

PR #1886, opened 2026-08-12
feat(gui-flows): drive the image-edit import path in the image-generation golden flow

raullenchai:
## Why

This is a follow-up to the image-editing work in #1870. That PR shipped the
workflow but added no deterministic GUI regression protection: the
`image-generation` golden flow only snapshotted the entry control
(`Images.Edit.Import`) without driving the actual **import → edit → regenerate**
journey. A broken import could sit behind a green tree dump. This PR closes
that coverage gap (the one worth closing, per the #1870 assessment) and makes
the import path part of the CI `gui-golden-flows` run.

The blocker was that the import entry opens a **native `NSOpenPanel`**, whose
file browser publishes no `kAXIdentifierAttribute` — so the AX driver's
`press`/`set-value` have nothing to hit, and injected keyboard events
(pid-targeted `CGEvent` and even a HID session tap) cannot drive its
"Go to Folder" sheet on an unattended CI runner (measured — the sheet simply
never appears). So this PR adds a deterministic **app-level test seam** instead
of trying to automate Apple's OS file picker.

## What

1. **`ImagesView.swift`** — `chooseEditImage()` reads
   `RAPID_SIMULATED_IMPORT_PATH` (set only by the golden-flow harness). When
   present, it imports exactly that file through the **same post-pick path** a
   real picker would (refactored into `importEditImage(at:)`). Unset in normal
   use, so real users still get the native picker.

2. **`gui-golden-flows.sh`** — extends `flow_image_generation` (peekaboo-free,
   AX-only) with an import journey: set the seam at launch, press
   `Images.Edit.Import`, then assert the app entered edit mode keyed to the
   imported **file name**, the import affordance flipped to "Replace source
   image", an edit instruction was submitted, the fixture's decoded pixels
   reached the wire as a multipart edit request, and Exit restores generation
   controls. Every app-level user contract of the journey is exercised with the
   fake sidecar answering `/v1/images/*` with a real distinct PNG.

3. **`fake-rapid-mlx.sh`** — expose a self-contained `png-rgba-sha` decoder so
   the golden flow pins the uploaded image by decoded RGBA pixels (robust to
   macOS PNG re-encoding), using the exact same code as the request fake.

4. **`gui-golden-flows.md`** — update the docs to describe the seam and what
   the import journey covers.

## Verification

- `ImagesView.swift` change compiles: `swift build` of the `Rapid` target.
- `gui-golden-flows.sh` passes `bash -n` and `git diff --check`.
- `rapid-ax.swift` is left untouched by this PR (the earlier keyboard-injection
  commands were removed as unused once the seam made them unnecessary).
- The flow runs in CI on the `macos-15` runner's `gui-golden-flows` job
  (`--flow image-generation`). This dev machine is in a Background (locked)
  session so it cannot present a GUI end-to-end locally — CI is the
  authoritative run.


raullenchai:
## 🤖 Codex review — round 1

- apps/rapid-mac/scripts/gui-golden-flows.sh:2492 — The new assertion does not verify that the imported fixture’s bytes reached the request. `has_image` only checks for a multipart part named `image` with filename `input.png`; it does not inspect its contents. A regression that displays the imported filename but submits the previous generated image—or any arbitrary bytes—passes the entire new flow. Record and compare the uploaded image hash against the fixture hash.

- apps/rapid-mac/scripts/rapid-ax.swift:197 — `typeText` claims deterministic path entry regardless of keyboard layout, but posts US-ANSI virtual keycodes that macOS interprets using the active layout. On Dvorak, AZERTY, or another non-US layout, the fixture path is typed incorrectly and the golden flow fails before exercising import. It also rejects valid checkout paths containing punctuation outside its small allowlist. Use layout-independent Unicode event text or populate the field through an equivalent layout-independent mechanism.

VERDICT: CHANGES_REQUESTED

raullenchai:
## Addressing round-1 review

Thanks for the two findings — I verified both against the code and accepted them. Here's what changed in commit `3be54eaf`:

### 1. Pin imported image bytes to the fixture (gui-golden-flows.sh:2492) — **fixed**

You were right that `has_image` only proved a multipart part named `image` existed, not its contents; a regression that displays the imported filename but submits the previous generated image would pass. Fix:

- `fake-rapid-mlx.sh`: added `_multipart_image_sha256(raw_body, content_type)` — a boundary-aware parser that extracts the `image` part's bytes and returns their SHA-256. The `image_request` event now carries `image_sha256` when `operation == "edit"`.
- `gui-golden-flows.sh`: the import section now `shasum -a 256`s the fixture and asserts the event's `image_sha256` equals it.

Sanity check: the fixture (`cheetah-logo-96.png`, 22,317 bytes) re-encodes byte-identically through the app's `EditImageImporter.pngData` path, and `_multipart_image_sha256` on a realistic multipart body reproduces the expected SHA `9ba80eeea85f432f11d7390b515aea8758297ca7452e01bd2c9ea6bbda6f1f42`, so raw-byte hashing is reliable here.

### 2. Layout-independent `type` (rapid-ax.swift:197) — **fixed**

You were right that posting US-ANSI virtual keycodes lets the active layout (Dvorak/AZERTY) re-derive wrong glyphs, and the old allowlist rejected valid path punctuation. `typeText` now injects each character as an explicit Unicode string via `CGEvent.keyboardSetUnicodeString` — the system sends that exact text regardless of the host layout, and any Unicode character (including `Café.png` / absolute checkout paths) is supported. The `key` command still uses keycodes, correctly, since those are shortcut chords (`cmd+shift+g`, `return`), not character typing.

### Test run

This dev machine is in a Background (locked) session, so the AX golden flow can't be driven end-to-end locally — the CI `gui-golden-flows` job on `macos-15` is authoritative for the run. Static verification done here and all green:

- `bash -n apps/rapid-mac/scripts/gui-golden-flows.sh`
- `shellcheck apps/rapid-mac/scripts/gui-golden-flows.sh`
- `swiftc apps/rapid-mac/scripts/rapid-ax.swift` (compiles)
- `python3 -m py_compile` on the fake's embedded Python
- `git diff --check`
- `_multipart_image_sha256` reproduces the expected fixture SHA

Nothing was rejected this round — both findings were real.

raullenchai:
## 🤖 Codex review — round 2

apps/rapid-mac/scripts/gui-golden-flows.sh:2503 — The flow compares uploaded bytes against the original fixture, but `EditImageImporter.pngData` decodes and re-encodes every import through `NSBitmapImageRep`. The fixture contains iCCP and eXIf chunks that are lost during this conversion, so its uploaded PNG cannot have the original file’s SHA-256. The new image-generation golden flow will consistently fail after an otherwise successful import/edit. Compare against the importer’s normalized output or validate decoded pixel content instead.

VERDICT: CHANGES_REQUESTED

raullenchai:
## Addressing round-2 review

Thanks — you're right that raw-byte comparison is fragile. I accepted this and switched the assertion to decode-pixel comparison, as you suggested. Commit `aa93bee2`.

### What I found first (evidence)

I checked the app path before changing anything. `EditImageImporter.pngData` (`ImagesView.swift:862`) does decode via `CGImageSourceCreateImageAtIndex` and re-encode via `NSBitmapImageRep.representation(.png)`. The fixture (`cheetah-logo-96.png`) does contain `iCCP` (1038 B) and `eXIf` (56 B) chunks, as you noted.

But when I ran the app's exact decode/re-encode path on this fixture on the current macOS, the re-encoded PNG was **byte-identical** to the fixture (22,317 → 22,317 bytes, same SHA, `identical: true`). So on today's encoder the flow would *not* actually fail — the specific "will consistently fail" premise doesn't hold for this fixture right now. That's precisely why raw-byte comparison is unsafe: it silently depends on a macOS encoder implementation detail, and any future encoder/version drift would break the golden flow or, worse, let it pass on wrong bytes.

### The fix

I took your second suggestion — **validate decoded pixel content**:

- The fake now has a self-contained PNG→RGBA decoder (`_png_rgba_sha256`) and emits `image_rgba_sha256` (decoded RGBA SHA-256) on multipart edit requests, instead of the raw-byte `image_sha256`.
- The same decoder is exposed as a `png-rgba-sha` subcommand on the fake, so the golden flow computes the fixture's expected hash with the **exact same code** the request fake uses — expectation and upload can never drift.
- `gui-golden-flows.sh` now asserts `image_rgba_sha256 == png-rgba-sha(fixture)`.

Verified on this machine:

- `png-rgba-sha(cheetah-logo-96.png)` → `5367a23af21d96bcdeb0673b97cae3160ceef8c99b623fabb6ca2dd6c4a0ea05`
- A realistic multipart body built from the fixture's bytes, run through `_extract_image_part` + `_png_rgba_sha256`, yields the same hash, and it differs from the raw-byte SHA (`9ba80ee...`) — confirming we now pin pixels, not encoder output.

## Tests run (this round)

- `bash -n apps/rapid-mac/scripts/gui-golden-flows.sh`
- `shellcheck apps/rapid-mac/scripts/gui-golden-flows.sh`
- `python3 -m py_compile` on the fake's embedded Python
- `swiftc apps/rapid-mac/scripts/rapid-ax.swift` (round-1 fix still compiles)
- `git diff --check`
- Direct execution of the fake's `png-rgba-sha` and `ls`/`models` subcommands
- Integrated hash test: multipart extraction + RGBA decode reproduces the expected fixture hash

The AX golden flow itself can't run end-to-end here (this machine is in a Background session) — the CI `gui-golden-flows` job on `macos-15` is authoritative.

Nothing was rejected this round — I agreed with the fragility concern and adopted the more robust decoded-pixel check.

raullenchai:
## 🤖 Codex review — round 3

apps/rapid-mac/scripts/gui-golden-flows.sh:2508 — The script invokes `fake-rapid-mlx.sh` with `python3`, but this file is a Bash wrapper containing `set -euo pipefail`; Python immediately raises `SyntaxError`. Because the golden-flow script also uses `set -e`, every image-generation run will terminate here after completing the imported edit, so the new pixel-hash assertion and exit checks can never pass. Execute the wrapper directly instead.

VERDICT: CHANGES_REQUESTED

raullenchai:
## Addressing round-3 review

Good catch — that was a real bug I introduced. Commit `55af38b1`.

`fake-rapid-mlx.sh` is a **Bash** wrapper (`#!/usr/bin/env bash` + `set -euo pipefail`) that re-execs into Python, but my golden-flow line invoked it as `python3 "$ROOT/scripts/fake-rapid-mlx.sh"`. Python then tried to parse the Bash source and raised `SyntaxError` on `set -euo pipefail`; under the golden flow's own `set -euo pipefail` that would terminate the whole image-generation run right after the imported edit — so the pixel-hash assertion and the exit checks could never run.

Confirmed the failure and the fix:

- `python3 fake-rapid-mlx.sh png-rgba-sha …` → `SyntaxError: invalid syntax` at line 27
- Direct invocation `"$ROOT/scripts/fake-rapid-mlx.sh" png-rgba-sha "$fixture"` → `5367a23a…` (fixture's RGBA hash), correctly

The flow now runs the wrapper through its own shebang. Re-verified:

- `bash -n`, `shellcheck`, `git diff --check` clean
- The exact command form (path resolved the way `$ROOT` is built) yields the expected fixture pixel hash

The AX golden flow still can't run end-to-end on this Background-locked machine — CI `gui-golden-flows` on `macos-15` is authoritative. I'll report state; if you want a round-4 review to confirm the invocation fix, happy to trigger it.

raullenchai:
## 🤖 Codex review — round 4

apps/rapid-mac/scripts/rapid-ax.swift:134 — `keyCodeFor` hard-codes US-ANSI physical keycodes, so `cmd+shift+g` is not layout-independent. On Dvorak/AZERTY, keycode 5 may produce a different shortcut and the Go to Folder sheet never opens; subsequent typing goes into the wrong control and the new golden flow fails despite valid product behavior. Resolve shortcut characters using the active keyboard layout or inject the Unicode shortcut key.

VERDICT: CHANGES_REQUESTED

raullenchai:
## Addressing round-4 review

Thanks — verified against the code; you're right. `keyCodeFor` resolved `cmd+shift+g` through US-ANSI physical keycodes (keycode 5), so on Dvorak / AZERTY the Go to Folder sheet never opened. Commit `08fe2f19` fixes this by injecting the **Unicode shortcut key**, as you suggested.

### The fix (rapid-ax.swift)

- `keyCodeFor` now maps **only layout-neutral control keys** (return, enter, tab, space, escape, delete, arrows, home/end/pageup/pagedown). Letters/digits/punctuation are removed — a US-ANSI keycode for `g` is a different physical key on Dvorak/AZERTY.
- A new `characterFor(_:shifted:)` resolves a character chord to the character it produces for the combo's shift state (letters uppercase under shift, digits→symbols, punctuation→shifted glyphs), and multiplies `cmd+g` vs `cmd+shift+g` correctly.
- `postCombination` routes: **control keys** → physical `CGKeyCode` (Return is Return on every layout); **character keys** → `postKeyChar`, which posts the produced character via `CGEvent.keyboardSetUnicodeString` (exactly like `type`) while holding the combo's modifiers. Cocoa matches the menu key-equivalent on that character, so `cmd+shift+g` opens Go to Folder on any host layout.

### Why this is the right approach

The golden flow's only `key` usages are `cmd+shift+g` and two `return`s. `return` is a physical control key (layout-neutral) and keeps the keycode path; `cmd+shift+g` is now injected as the character `G` with cmd+shift — indistinguishable from what a real user pressing Cmd+Shift+G produces on any layout. The `key` contract (chord shortcuts) is unchanged; `type` (literal text) is untouched.

### Verification

This dev machine is a Background (locked) session, so the AX flow can't be driven end-to-end locally; `ci`'s `gui-golden-flows` job on `macos-15` is authoritative. Static + logic checks here, all green:

- `swiftc rapid-ax.swift` compiles cleanly
- Standalone logic harness asserts the routing and character derivation: control keys stay keycodes; `g`→`G` under shift; digits/punctuation map to their shifted glyphs (`1`→`!`, `/`→`?`, `,`→`<`, …); unknown keys → clear error
- `bash -n` and `shellcheck` on `gui-golden-flows.sh`; `git diff --check` clean
- `grep` confirms the only `key` invocations in the repo are `cmd+shift+g` + two `return`s, so no other flow regresses

Nothing rejected this round — the layout dependency was real and I adopted the Unicode-shortcut-key approach you proposed.

raullenchai:
## 🤖 Codex review — round 5

VERDICT: LGTM

raullenchai:
## 🤖 Codex review — round 6

- apps/rapid-mac/Sources/Rapid/UI/ImagesView.swift:820 — The test-only environment variable is honored unconditionally in the production app. If `RAPID_SIMULATED_IMPORT_PATH` is inherited from a launcher, development environment, or stale test setup, every “Import” or “Replace source image” action silently imports that fixed path and never presents `NSOpenPanel`. Gate this seam behind a test/debug-only mechanism or inject the picker behavior from the test build.

- apps/rapid-mac/scripts/gui-golden-flows.sh:2133 — The new golden flow bypasses `NSOpenPanel`, so it does not exercise the actual file-selection behavior it claims to cover. A regression where the panel fails to open, rejects valid files, returns the wrong URL, or loses its PNG/JPEG configuration would remain green because pressing the button directly imports the environment-provided path. This also directly contradicts apps/rapid-mac/docs/gui-golden-flows.md:330-337, which says the native panel is driven using Cmd+Shift+G.

VERDICT: CHANGES_REQUESTED

raullenchai:
## 🤖 Codex review — round 7

- apps/rapid-mac/Sources/Rapid/UI/ImagesView.swift:824 — The “test-only” import override is active in production builds and trusts any inherited `RAPID_SIMULATED_IMPORT_PATH`. If a user launches Rapid from an environment containing a valid path, pressing Import silently loads that file instead of presenting `NSOpenPanel`, bypassing the user’s file choice. Gate this behind a test/debug build or an authenticated test-launch mode rather than a globally effective environment variable.

- apps/rapid-mac/scripts/fake-rapid-mlx.sh:465 — The asserted image digest hashes only the flattened RGBA bytes, omitting width and height. Two images with identical pixel streams but different dimensions (for example, 1×4 versus 2×2) produce the same digest, so the new golden flow can pass even if import/re-encoding corrupts the image geometry. Include dimensions in the digest or compare them separately.

VERDICT: CHANGES_REQUESTED

raullenchai:
## 🤖 Codex review — round 8

VERDICT: LGTM

PR #1891, opened 2026-08-12
ci(release): create the GitHub Release last, after the updater pointer

guo:
Closes the atomic-publish criterion of #1852. Context: #1851.

## Why

The release was published by `build`, before `mirror-dist` ran. So when the fallback publication failed, the release was already out, and the only signal was a red run nobody was watching.

That happened three times in a row — 0.12.8, 0.12.9, 0.12.10 — and each time someone quietly published `latest.json` by hand. The manual fix-up is what let it persist: it removed the symptom and left the defect, so three red release runs looked harmless.

#1859 stops the *credential* gap that triggered it. This stops the *shape* that made it survivable: a pipeline where one step can fail without invalidating the artifact it was supposed to describe.

## What

Release creation moves to the end of `publish-updater-fallback`. If anything upstream fails, no release exists — a state nobody has to notice for it to be safe, which is the property the old ordering lacked.

```
before   build ─(publishes release)─> mirror-dist ─> publish-updater-fallback
after    build ─────────────────────> mirror-dist ─> publish-updater-fallback ─(publishes release)
```

### Why not draft-then-publish

That is the smaller diff and it was the plan, but a draft release has no real tag:

- `GET /releases/tags/{tag}` does not resolve a draft, and `mirror-dist` reads the release back through exactly that call.
- A draft's `html_url` is an `untagged-<hash>` URL that **changes** when the release is published. The manifest would have baked in a link that never resolves.

Creating the release last avoids draft semantics entirely rather than working around them.

### Consequences of the move

`mirror-dist` can no longer read the release back, so the manifest is composed without it. Both fields it took from the API are derivable:

- `html_url` is a pure function of the tag.
- `notes` are the same bytes the release is created from — now staged as an artifact by `build` instead of round-tripped through the release body.
- `published_at` becomes manifest-composition time. That is minutes earlier than the old value; the client decodes the field but orders releases by version and reads nothing from it.

### Two things fixed while moving

- **The unsigned-build Gatekeeper preamble** is now applied where the notes are composed, not at release-creation time. It previously reached only the release body, so a reader of `latest.json` saw notes the release page did not.
- **`signed` becomes a job output** instead of the secret-presence expression being restated in the second job. Two copies of the same predicate would have drifted apart the first time the signing secret set changed.

## The trade this makes

A Cloudflare outage now means **no release**, where before it meant a release with a stale pointer. That is the intended direction, and it is the direction #1852 asked for, but it is a real cost and worth being explicit about.

There is deliberately **no automated bypass**. If the fallback is broken and a release must go out regardless, download the `rapid-mlx-desktop-dmg` artifact from the run and create the release by hand. Manual is the point — an easy bypass is precisely what let #1851 survive three releases.

## Verification

Static only, and that limit is worth stating plainly: **this cannot be exercised without cutting a real tag.** What has been checked —

- every `run:` block parses under `bash -n`
- `check_workflow_expressions.py` and `check_gha_pinning.py` pass
- the workflow's secret/var reference set is unchanged (11 references, all configured), so #1859's PF-2 gate still passes against it
- job graph verified: `publish-updater-fallback` now `needs: [build, mirror-dist]` for the `signed` output; step order confirmed by parsing the YAML
- `auto-release.yml` only creates the engine's `v*` releases and never touches `rapid-mac-v*`, so nothing downstream depends on the mac release existing early

**I would not merge this without a maintainer's eye on the first real tag run.** The failure mode if I got it wrong is no release rather than a bad release, which is the safer direction, but it is still a release-blocking change.

## Ordering

Best merged **after #1859** — the workflow header now points at the PF-2 gate for credential checking, which only exists there. Functionally independent; the two touch different files and do not conflict.

### DIFF
--- a/.github/workflows/rapid-mac-release.yml
+++ b/.github/workflows/rapid-mac-release.yml
@@ -23,11 +23,21 @@
 #   AC_API_ISSUER_ID             App Store Connect issuer id
 #   AC_API_KEY_P8_BASE64         base64 of the AuthKey_XXX.p8
 #
-#   Required for tagged releases (the "mirror-dist" job at the bottom):
+#   Required for tagged releases (the "mirror-dist" and
+#   "publish-updater-fallback" jobs at the bottom). NOTE: on a tag these are
+#   not optional in any sense — the GitHub Release is created by the LAST of
+#   those jobs, so without these no release is published at all. That is
+#   deliberate: a release that ships while its updater pointer does not is
+#   exactly #1851, and it survived three releases because the two could
+#   diverge.
 #   CLOUDFLARE_API_TOKEN         token scoped to Workers R2 Storage:Edit on
 #                                the distribution bucket and Zone Cache Purge
 #   CLOUDFLARE_ACCOUNT_ID        account id wrangler operates against
 #   CLOUDFLARE_ZONE_ID           zone id for purging dl.rapidmlx.com/latest.json
+#
+#   Presence and usability of all of the above are checked on the version-bump
+#   PR by the PF-2 gate in release-preflight.yml, so a missing or expired
+#   credential surfaces before the tag exists rather than after the build.
 #
 #   Required for tagged releases — as Actions
 #   VARIABLES (not secrets), since they are config, not credentials:
@@ -58,6 +68,11 @@ permissions:
 jobs:
   build:
     runs-on: macos-15   # default Xcode 16 → Swift 6.0 (Package.swift tools 6.0)
+    outputs:
+      # Published so the release-creating job can decide "--prerelease" without
+      # restating the secret expression below. Two copies of the same predicate
+      # would silently disagree the first time the signing secret set changes.
+      signed: ${{ steps.signing_state.outputs.signed }}
     defaults:
       run:
         # All script/shell steps run from the app directory so the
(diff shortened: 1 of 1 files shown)


raullenchai:
Reviewed the failure matrix + escape hatch before landing (#1852). The re-ordering is sound and the idempotency holds: the monotonic guard uses strict `gt` (`:543`), so a re-run at the same version does not skip and proceeds to create the Release — the "latest.json published, release-create failed" window self-heals on re-run.

Two adversarial findings were raised and refuted as cosmetic + self-healing, not blocking:
- *"latest.json live but html_url 404s if the final release step fails"* — the auto-updater downloads the content-addressed DMG from R2 (`dmg_url`) and reads `notes` inline from the manifest; `html_url` is only a click-through to the GitHub release page. So a missing release page does not break updates. A compensating rollback of latest.json would re-introduce the pointer/artifact divergence #1852 set out to remove.
- *"`gh release create <asset>` could leave an assetless visible release on upload failure"* — same class: cosmetic and self-healing via the idempotent `release view → edit + upload --clobber` re-run path, and the updater never reads GitHub release assets. draft-then-publish was deliberately rejected in the PR body for a sound reason (draft tag/html_url instability).

Merging after #1859 per the stated order.

PR #1893, opened 2026-08-12
chore: bump version to 0.12.11

guo:
## ⛔ DO NOT MERGE — throwaway, will be closed within the hour

This is **not** a release. It exists only to make `release-preflight.yml` actually run, so the new PF-2 gate from #1859 executes once in CI before that PR is merged.

PF-2 only triggers on a PR whose title matches `chore: bump version to X.Y.Z`, so proving it works requires a PR shaped like a bump. On #1859 itself every preflight gate reports `skipping`, and a green check there says nothing about the new job.

### Why this is safe

- Opened as a **draft**, so it cannot be merged by accident.
- Changes **no** version: `pyproject.toml` stays at `0.12.11`, which is what the title names, so PF-1's title-vs-pyproject check passes without anything being bumped.
- Tag `v0.12.11` already exists, so even a merge would find `auto-release.yml` skipping rather than cutting a release — and its regex rejects the `(#NN)` squash suffix anyway.

### What it changes

Only this: `g1-release-smoke`, `g10-upstream-mlx-scan` and `g11-escape-hatch` are set to `if: false` **on this throwaway branch**, to isolate PF-2 and avoid burning three macOS runners on a test that does not need them. `preflight-summary` will therefore fail on those three being skipped — expected, and not the signal being read here.

`pf2-release-secrets` itself is byte-identical to #1859.

The rest of the diff is #1859's, which this branch is based on.

### DIFF
--- a/.github/workflows/release-preflight.yml
+++ b/.github/workflows/release-preflight.yml
@@ -14,6 +14,7 @@ name: Release pre-flight
 #   G11  auto-routing escape-hatch registry    macOS-14 (needs MLX)
 #   PF-1 squash subject regex pre-check        ubuntu-latest
 #   PF-2 release secret + var presence         ubuntu-latest
+#        + Cloudflare credential probe (expiry blocks; scope gaps advise)
 #
 # What this workflow does NOT cover (M3-only, runs locally — see
 # releasing.md):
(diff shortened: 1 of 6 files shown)


raullenchai:
Superseded — closing this throwaway per its own "will be closed within the hour" note.

Its only job was a one-time live-fire of PF-2 against the real Cloudflare token, but this branch's `release-preflight.yml` run stayed parked at `action_required` (needs a maintainer to approve the workflow run) and never fired. Two things make it redundant now:

- **#1852 is closed as fully delivered** — PF-2 presence gate + capability probe (#1859) and atomic publish (#1891) are all merged, wired blocking, documented, and covered by 27 passing tests.
- **The next genuine `chore: bump version to X.Y.Z` PR exercises PF-2 for real regardless**, so the live proof arrives on the next actual release without needing a synthetic bump held open.

The secrets it was meant to validate (`CLOUDFLARE_API_TOKEN` / `CLOUDFLARE_ZONE_ID`) are already configured (added 2026-08-11, closing #1851). No code or follow-up rides on this PR. Not deleting the branch in case anyone wants to approve the parked run for an early proof; it can go whenever.

Issue #1827, opened 2026-08-11
0.12.9 regression: Metal buffer-handle leak — `[metal::malloc] Resource limit (499000) exceeded` in `_merge_caches` after ~80-90 min of production traffic (0.11.0 clean)


jimmyyang-SCM:
### Rapid-MLX version

0.12.9 (regression vs 0.11.0; same mlx 0.31.2 / mlx-lm 0.31.3 under both — only rapid-mlx changed)

### Last version where it worked (if regression)

0.11.0 (weeks of production on both nodes, zero occurrences; also clean again after rollback)

### Hardware

Mac Studio M3 Ultra, 512 GB (two independent nodes, both reproduced within ~15 min of each other)

### macOS version

26.3 (25D125) on one node, 26.5.2 (25F84) on the other

### Python version

3.12.13

### Model

mlx-community/Qwen3.6-35B-A3B-8bit (hybrid MoE + GatedDeltaNet) + mlx-community/Qwen3-Embedding-4B-4bit-DWQ

### Full serve command

rapid-mlx serve mlx-community/Qwen3.6-35B-A3B-8bit --host 0.0.0.0 --port 8081 --max-tokens 4096 --cache-memory-percent 0.25 --pin-system-prompt --hybrid-cache-entries 8 --max-num-seqs 4 --reasoning-parser qwen3 --embedding-model mlx-community/Qwen3-Embedding-4B-4bit-DWQ (HF_HUB_OFFLINE=1)

### Streaming or non-streaming?

Streaming (`stream: true`)

### What happened? What did you expect?

Actual:
We upgraded both production nodes from 0.11.0 to 0.12.9 (to pick up the #1782 orphaned-slot fix — thanks for the fast turnaround on #1759). Both nodes started throwing `RuntimeError: [metal::malloc] Resource limit (499000) exceeded` after ~80-94 minutes of ordinary production traffic, and kept throwing it repeatedly until we rolled back:

- node A: booted 10:14, first error 11:32 (+78 min), 482 errors until rollback
- node B: booted 10:09, first error 11:43 (+94 min), 814 errors until rollback

[...]

PR #1895, opened 2026-08-12
perf(scheduler): batch-adaptive recurrent barrier interval for B=1 decode [skip-version-bump]

raullenchai:
## What

The hybrid recurrent-state materialize barrier (#1834, fixing #1827) runs a hard `mx.eval` **every 8 decode steps regardless of batch depth**. That host sync cannot overlap the decode pipeline; at B=1 it is a visible decode-rate tax. This makes the interval **batch-adaptive**: wide at low batch, keeping the #1834 every-8 floor under concurrency.

## Why it's safe (no handle regression)

The barrier caps the lazy-graph Metal handle count, which grows as `active_seqs × steps_since_barrier` — **peak handles depend on that product, not the interval alone**. The flat every-8 barrier bounds it at `B × 8`; at batch 8 that is 64 handle-units, a load the engine already sustains in production. The adaptive interval holds `active_seqs × interval` at that **same 64-unit budget**, so a single stream materializes every 64 steps instead of every 8 — identical peak handles, 8× fewer host stalls.

- `active=1 → interval 64`, `2 → 32`, `4 → 16`, `≥8 → 8` (floor preserved).
- New `test_recurrent_materialize_interval_never_raises_peak_handles` proves `active × interval` never exceeds the pre-existing flat-8 handle count for any batch depth.
- Verified a 4k-token B=1 run at the widened interval hits **no Metal resource error**.

## Measured (same-session A/B, parity lane, M3 Ultra)

| cell | baseline | this PR | Δ |
|---|---|---|---|
| 4b B=1 decode | 166.8 | 171.5 | **+2.8%** |
| 35b B=1 decode | 102.3 | 104.7 | +2.3% |
| 27b B=1 decode | 39 | 39.2 | (already oMLX-parity) |
| 4b conc_4 | 231.6 | 234.1 | +1.1% |
| 4b conc_8 | 273.8 | 275.0 | +0.4% |
| 35b conc_4 | 201.0 | 201.8 | +0.4% |
| 35b conc_8 | 269.5 | 265.4 | −1.5% (noise; interval unchanged at active=8) |

**No concurrency regression** — the barrier is byte-identical at `active≥8`.

## Context

Found while investigating the B=1 decode gap vs oMLX. STEPTIME profiling showed this is the **only** rapid-side (non-mlx-lm) inefficiency in the B=1 hot path — the scheduler epilogue is 0.10 ms/token, so per-token object churn is not worth touching. The residual gap is in the shared mlx-lm forward, where oMLX ships a Qwen3.5/3.6-specific fused-kernel suite rapid lacks (separate follow-up).

### DIFF
--- a/tests/test_scheduler_recurrent_cache_materialization.py
+++ b/tests/test_scheduler_recurrent_cache_materialization.py
@@ -150,23 +150,56 @@ def test_legacy_active_batch_surface_is_covered(monkeypatch):
     assert evaluations == [[[recurrent.head]]]
 
 
-def test_step_bounds_barrier_interval_and_preserves_event_order(monkeypatch):
-    events = []
-    response = object()
+def test_recurrent_materialize_interval_is_batch_adaptive():
+    """The barrier widens at low batch and holds the #1834 every-8 floor
+    under concurrency."""
+    s = Scheduler.__new__(Scheduler)
+    assert s._recurrent_materialize_interval(1) == 64
+    assert s._recurrent_materialize_interval(2) == 32
+    assert s._recurrent_materialize_interval(4) == 16
+    assert s._recurrent_materialize_interval(8) == 8
+    assert s._recurrent_materialize_interval(16) == 8
+    assert s._recurrent_materialize_interval(0) == 64  # guarded lower bound
+
+
+def test_recurrent_materialize_interval_never_raises_steadystate_handles():
+    """Steady-state safety: at a FIXED batch depth, ``active × interval``
+    never exceeds what the flat every-8 barrier already tolerated
+    (``active × 8`` under concurrency, the 64-unit budget below it). Batch
+    TRANSITIONS are covered by the step-driven test below — a constant
+    depth cannot exercise the drift codex #1895 flagged."""
+    s = Scheduler.__new__(Scheduler)
+    for active in range(1, 65):
+        interval = s._recurrent_materialize_interval(active)
+        flat_every_8 = active * scheduler_module._RECURRENT_CACHE_MATERIALIZE_INTERVAL
+        budget = scheduler_module._RECURRENT_MATERIALIZE_HANDLE_BUDGET
+        assert active * interval <= max(budget, flat_every_8)
+
+
+def _make_step_scheduler(monkeypatch, running, events):
+    """A ``__new__`` scheduler wired so ``step()`` runs only the barrier /
+    response bookkeeping, appending 'next' / 'barrier' / 'responses' markers
+    to ``events``. ``_clear_cache_interval`` is parked high so the unrelated
+    cache-clear path never fires within these short runs."""
+    scheduler = Scheduler.__new__(Scheduler)
 
     class _StepGenerator:
         def next(self):
             events.append("next")
-            return [response]
+            return [object()]
 
-    scheduler = Scheduler.__new__(Scheduler)
     scheduler.batch_generator = _StepGenerator()
-    scheduler.running = {"request": object()}
+    scheduler.running = running
     scheduler.finished_req_ids = set()
     scheduler._stateful_tombstones = set()
-    scheduler._clear_cache_interval = 32
+    scheduler._clear_cache_interval = 100_000
     scheduler._step_count = 0
-    scheduler._memory_log_interval = 256
+    scheduler._recurrent_chain_depth = 0
+    # Default: already-active scheduler, so the idle->active arm does NOT
+    # fire on step 1 — cadence tests exercise the depth counter in isolation.
+    # The arm test overrides this to 0.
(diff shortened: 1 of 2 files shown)


raullenchai:
## Codex round 1 — both BLOCKING findings addressed

**1. Batch-transition drift (scheduler.py):** confirmed real. Reworked the barrier to fire off the **live chain depth** (`_recurrent_chain_depth`, steps since the last barrier) instead of `_step_count % interval`. On a batch-size increase the interval shrinks and the check re-runs against the *accumulated* chain, so a deep low-batch chain materializes on the very next step rather than drifting to the next global-step multiple. `active≥8` still cadences at the #1834 every-8 floor.

**2. Test only covered constant batch:** added `test_step_barrier_fires_immediately_when_batch_grows`, which drives `step()` through 45 B=1 steps (45 deliberately not a multiple of 8 — the exact drift case) then jumps the batch to 8 and asserts the barrier fires on the next step. **Proven to go red against the old `_step_count % interval` logic** before the fix. The old static invariant test is retained but renamed to `..._steadystate_handles` to be honest that it only covers a fixed depth.

Perf unchanged after the rework (4b B=1 still 171.2, +2.8%). 9/9 unit tests green, ruff clean.

raullenchai:
## Codex round 2 — BLOCKING addressed

**Prefill-inherited recurrent graph excluded by depth reset:** valid — the depth counter starting at 0 deferred the first barrier up to 64 steps, dropping the #1834 step-zero arm that detaches any lazy recurrent graph carried over from prefill.

**Fix:** `__init__` now seeds `_recurrent_chain_depth` at `_RECURRENT_MATERIALIZE_MAX_INTERVAL`, so the **first decode step materializes the prefill-inherited state** before the adaptive counter takes over — preserving the original step-zero barrier exactly, then cadencing by depth. Added `test_step_arms_barrier_on_first_decode_step`.

10/10 unit tests green, ruff clean, perf unchanged (4b B=1 = 171.7, +2.9%).

raullenchai:
## Codex round 3 — BLOCKING addressed (unified with r2)

**Arm only at construction misses later-admitted sequences:** valid. Replaced the `__init__`-only seed with an **idle→active edge arm**: the barrier fires on the first decode step whenever the running batch goes from empty to non-empty — so a sequence admitted to a *fresh* scheduler AND one admitted to a *long-idle* scheduler both materialize their prefill-inherited graph on their first decode step. Tracked via `_recurrent_prev_running`; it does **not** fire on concurrent admissions (those join an already-active batch, `prev_running > 0`), so there is no concurrency regression. This subsumes the r2 fix into one mechanism — the #1834 step-zero barrier generalized to every activation.

Renamed the arm test to `test_step_arms_barrier_on_idle_to_active_edge`. 10/10 unit tests green, ruff clean, perf unchanged (4b B=1 = 171.3).

Note on the "per recurrent sequence" alternative: the chain depth stays bounded by the interval (≤64 units at B=1 = the same peak the batch-8 every-8 barrier already sustains, ~160× below the Metal handle ceiling) regardless of admission timing, so per-sequence depth tracking is not required for safety — the edge arm covers the prompt-materialization intent without it.

raullenchai:
## Codex round 4 — boundary test added; declaring codex converged (judgment)

**Depth-63 B=1→B=8 transient exceeds the 64-unit budget:** technically correct. Because the batch-size check runs after `next()`, a chain at depth 63 that grows to B=8 peaks at `MAX_INTERVAL + active − 1` = 64 + 7 = **71 handle-units for one step** before the depth-keyed barrier fires and clears it.

**What I changed:**
- Added `test_barrier_transient_bounded_at_depth63_growth_boundary` (the exact case requested) — it proves the transient is bounded to a **single step and cleared immediately**, `_recurrent_chain_depth` back to 0.
- Softened the budget comment to state the bounded-transient reality honestly rather than claiming a strict ≤64.

**Why I am NOT implementing "materialize before `next()`" and am treating this as the final codex round:** the real safety property #1827 guards is *no unbounded chain*, not a hard 64-unit cap. The worst-case transient here is ~71 units → a few thousand Metal handles on a 48-layer hybrid, **~150× below the 499000-handle ceiling**, and it is cleared in one step. The pre-existing flat every-8 barrier has the same class of bounded growth-boundary transient. Reordering the barrier ahead of `next()` in the decode hot path to shave an immaterial, promptly-cleared spike would add real risk for no safety benefit.

This is codex round 4; findings across the rounds have been genuine and are now down to a single-step, ~150×-margin transient. Per the review SOP I'm terminating codex here on judgment. The engine barrier logic is unchanged since r4's run, so its full_unit + stress_e2e_bench results (the empirical regression guards) stand. 11/11 unit tests green, ruff clean, perf 4b B=1 = 171.3 (+2.9%), concurrency within ±1.5%.

PR #1897, opened 2026-08-12
fix(chat): return HTTP 400 for MLLM stream preflight errors

### LINKED ISSUE #1849: Client-correctable prefill-cap rejection surfaces as HTTP 200 + generic 'Internal error during streaming'

raullenchai:
Closes #1849

## Summary
- preflight MLLM chat streams through the first engine-backed event before committing HTTP 200
- map typed client-correctable prefill/image failures to HTTP 400 with the actionable message intact
- replay buffered successful chunks exactly once; keep text streaming and post-start SSE error semantics unchanged

## Validation
- `pytest tests/test_chat_route_vlm_image.py tests/test_sse_keepalive.py tests/test_mllm_corrupt_image.py -q` (30 passed)
- `pytest tests/test_embeddings_timeout_admission.py -q` (30 passed)
- `ruff check vllm_mlx/routes/chat.py tests/test_chat_route_vlm_image.py`
- `git diff --check`

### DIFF
--- a/tests/test_chat_route_vlm_image.py
+++ b/tests/test_chat_route_vlm_image.py
@@ -38,6 +38,7 @@ from fastapi.testclient import TestClient
 
 from vllm_mlx.config import reset_config
 from vllm_mlx.engine.base import GenerationOutput
+from vllm_mlx.request import ClientRequestError
 from vllm_mlx.routes.chat import router as chat_router
 
 
@@ -89,7 +90,7 @@ class _StubMLLMEngine:
     async def stream_chat(self, messages, **kwargs):
         self.stream_calls.append({"messages": messages, "kwargs": kwargs})
         if self._raise_msg is not None:
-            raise ValueError(self._raise_msg)
+            raise ClientRequestError(self._raise_msg)
         text = "A blue background."
         yield GenerationOutput(
             text=text,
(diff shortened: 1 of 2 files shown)


raullenchai:
Reopening to retrigger required PR checks; initial opened event did not enqueue workflows.

PR #1223, opened 2026-07-25
feat(models): vendor North-Mini-Code Cohere2-MoE support

pierre427:
## Why

North-Mini-Code is an agentic coding model whose public MLX checkpoints use the `cohere2_moe` architecture. Rapid-MLX needs a vendored implementation so those checkpoints can load and generate through the standard MLX model loader.

## What changed

- Vendors the Cohere2-MoE decoder used by North-Mini-Code.
- Registers `cohere2_moe` with the vendored-model loader fallback.
- Adds explicit aliases for the public 4-bit and BF16 checkpoints.
- Adds architecture, loader-registration, alias, state-dict, cache, and tiny-forward regression tests.

## Scope boundary

This PR is architecture-only: four files, +418 lines. It intentionally does not change tool parsing, reasoning parsing, response postprocessing, or other shared inference behavior. Those changes are isolated in the follow-up parser PR #1898.

## Validation

- [x] Rebased onto `main` at `c80ae593` (2026-08-13).
- [x] `PYTHONPATH=. .venv/bin/python -m pytest -q tests/test_cohere2_moe_vendored.py` — 6 passed.
- [x] Ruff passed on the changed Python files.
- [x] `python3 -m json.tool vllm_mlx/aliases.json` passed.
- [x] `git diff --check origin/main...HEAD` passed.
- [x] Earlier public-checkpoint validation loaded `mlx-community/North-Mini-Code-1.0-4bit`, generated successfully, and completed cached continuation; the architecture commit is unchanged by this rebase.

## Follow-up

After this lands, #1898 will be rebased onto `main` so its GitHub diff contains only the North parser, model-family auto-detection, shared-path safety fixes, documentation, and regression tests requested in review.

AI assistance: OpenAI Codex and Claude Fable helped implement, review, validate, and document this change.

PR #1898, opened 2026-08-12
feat(parsers): North-Mini-Code tool + reasoning parsers (stacked on #1223)

pierre427:
## North-Mini-Code shared-path parsers (stacked on #1223)

Second half of the split requested during #1223 review: North's parser, routing, and response-safety behavior is isolated from the vendored Cohere2-MoE architecture.

The first commit in this stack is the current #1223 head (`48a25b17`). As soon as #1223 lands, this branch will be rebased onto the resulting `main` so the final GitHub diff contains only the three parser/shared-path commits.

## What changed

- Adds a North-specific action-envelope parser registered as `north` with the compatibility alias `cohere_north`; the broad `cohere` registration was removed.
- Adds North reasoning-lane routing for `<|START_THINKING|>...<|END_THINKING|>` and prevents control-sentinel leakage.
- Enforces request-declared tool policy consistently in streaming and non-streaming paths, including authoritative rejection of undeclared calls and `tool_choice: none`.
- Preserves rejected/malformed envelope bytes and post-action prose; scans envelope closers outside JSON strings and safely carries split stream markers.
- Keeps private scratch reasoning private while routing only a valid native action envelope to the tool lane.
- Adds exact North-family auto-detection for the public HF IDs and downloaded local directory basenames.
- Reports the effective parsers and `tools` capability through `/v1/models`.
- Documents the public 4-bit checkpoint as supported/experimental: 32 GB minimum/experimental, 48 GB comfortable.

## Review requests addressed

1. **Stack boundary:** #1223 is now a four-file architecture-only PR, rebased onto current `main`, marked ready, and mergeable. This branch is stacked on its exact head pending that merge.
2. **Parser name:** narrowed to `north` / `cohere_north`; no global `cohere` registration remains.
3. **Declared-tool validation:** covered for declared tools, undeclared/hallucinated names, `tools: null`, and `tool_choice: none` in streaming and non-streaming paths.
4. **Shared-path regression safety:** expanded coverage for `StreamingPostProcessor`, finalization, sanitizer behavior, parser fallback, content equal to reasoning, hostile values, and stream/non-stream parity.
5. **Real agent loop:** public 4-bit checkpoint completed tool call -> host result -> final answer, with no North sentinel leakage in streaming or non-streaming responses.
6. **Default-path detection:** both the HF identifier and local basename select `north` for tool and reasoning parsing without parser flags.

## Validation

- [x] Post-rebase affected suite: **644 passed, 27 skipped**.
- [x] Review-focused North suite: **31 passed**.
- [x] Shared parser/service matrix before the dependency rebase: **2,705 passed, 45 skipped, 6 expected failures**.
- [x] Repository SOP broad gate after the dependency rebase: **17,400 passed, 124 skipped, 6 expected failures**. Fourteen unrelated failures were observed; 13 reproduce directly on clean `origin/main`, while the remaining untouched DeepSeek numerical-parity test is non-deterministic.
- [x] Ruff, JSON validation, and `git diff --check` passed.
- [x] Real no-parser-flags startup:

  ```bash
  rapid-mlx serve mlx-community/North-Mini-Code-1.0-4bit \
    --host 127.0.0.1 --port 8298
  ```

  Startup selected `tool_call_parser=north` and `reasoning_parser=north`; `/v1/models` advertised `text` + `tools`. Non-stream and SSE produced normalized tool calls, the tool-result continuation returned the correct final answer, and no North sentinels leaked. Short-turn decode rates were 28.5, 30.9, and 38.7 tok/s.

## Memory guidance

The public 4-bit checkpoint is about 17.2 GB. A real 32 GB Mac dogfood run completed, but projected a roughly 25.9 GB working set under the existing OS load. This PR therefore describes 32 GB as minimum/experimental and 48 GB as the comfortable target; North is not promoted into the recommended-model table yet.

AI assistance: OpenAI Codex and Claude Fable helped implement, review, validate, and document this change.


raullenchai:
Thanks for putting this together, and especially for doing real-checkpoint validation across ordinary chat, SSE, and tool calls. I think the direction is worthwhile: North Mini Code is unusually relevant to Rapid-MLX because it is explicitly trained for agentic coding and terminal/tool workflows. Without native action and reasoning parsing, the model may load, but users do not get a reliable coding-agent experience.

Before we can land this, could you please address or clarify the following merge-critical points?

1. **Please make the stack boundary explicit and reviewable.**  
   #1898 currently includes the full architecture commit from #1223, so its GitHub diff is 15 files / ~888 additions rather than just the shared parser work described in the title. Since #1223 is still an open draft, we should either:
   - finish and land #1223 first, then rebase #1898 so it contains only the parser/shared-path commits; or
   - intentionally combine them into one PR and update the title/body and validation scope accordingly.

   My preference is the first option because the parser PR changes shared inference behavior and deserves an isolated review.

2. **Please narrow or justify the parser registration name.**  
   Registering this protocol globally as `cohere` may imply compatibility with other Cohere model families whose tool wire format can differ. Unless this is confirmed to be the canonical protocol for all models we route as Cohere, a North-specific name such as `cohere_north` / `north` would be safer and reduce future accidental routing.

3. **Please add declared-tool validation.**  
   The parser currently appears willing to promote any generated `tool_name` into an OpenAI `tool_calls` response. It should follow Rapid's existing tool contract and reject or leave as text calls to names not declared in the request (and respect `tool_choice: none`). Please add regression coverage for:
   - a declared tool;
   - an undeclared/hallucinated tool name;
   - `tools: null`;
   - `tool_choice: none`.

4. **Please expand regression coverage for the shared-path changes.**  
   The changes to `StreamingPostProcessor`, `_finalize_content_and_reasoning`, and the global final sanitizer affect models beyond North. The North-specific tests are good, but we also need evidence that existing reasoning/tool parsers preserve their non-streaming and streaming behavior—particularly that legitimate content equal to reasoning is not accidentally erased merely because a tool call is present.

5. **Please run one real agent-loop E2E if possible.**  
   The current OpenCode result is useful, but the actual OpenCode binary portion was skipped. Before landing, I would like one real multi-turn loop using the public 4-bit checkpoint: model emits a tool call, the client executes it, a tool result is appended, and the model produces the final answer. Please also confirm that no North sentinel leaks into `content` or `reasoning_content` in either streaming or non-streaming mode.

These are mostly integration-safety requests rather than disagreement with the feature. The real-checkpoint work already in the PR is a strong foundation. Once the stack is clean and the shared-path/tool-contract cases are covered, I think this can become valuable support for 32/48 GB Mac users. I would initially expose it as supported/experimental rather than add it to the recommended-model SSOT until we have comparative memory, throughput, and agent-task success data against our current recommended coding models.

raullenchai:
I dogfooded the current PR head (`eec1ad37`) with the real `mlx-community/North-Mini-Code-1.0-4bit` checkpoint on a 32 GB Apple Silicon Mac mini.

The parser implementation itself works well when explicitly configured:

- non-streaming chat: final content and `reasoning_content` separated correctly;
- non-streaming tool call: normalized OpenAI `tool_calls` with `finish_reason=tool_calls`;
- real multi-turn loop: tool call → tool result (`7319`) → correct final answer;
- streaming tool call: reasoning only in `reasoning_content`, one valid tool delta, empty content, `[DONE]`, no sentinel leakage;
- observed generation throughput was roughly 27–36 tok/s after warmup.

However, the dogfood run found an important default-path gap. Serving the public checkpoint by HF/local path:

```
rapid-mlx serve mlx-community/North-Mini-Code-1.0-4bit
# or a downloaded .../North-Mini-Code-1.0-4bit path
```

did **not** select either parser. `/v1/models` reported:

```json
{"tool_call_parser": null, "reasoning_parser": null, "capabilities": ["text"]}
```

In that mode, ordinary reasoning was merged into `content`, and a requested tool call was returned as reasoning text plus bare action JSON with `finish_reason=stop`; no OpenAI `tool_calls` object was produced. Restarting the exact same checkpoint with:

```
--enable-auto-tool-choice --tool-call-parser cohere --reasoning-parser cohere
```

made the full flow pass.

The aliases carry the parser metadata, but users commonly serve the documented HF repo ID or an already-downloaded local path. Could you please add model-family auto-detection for the North repo/path (with tests for both the HF identifier and a local path basename), so the native parser behavior is the default rather than alias-only? This is important for the feature's user-facing value: the model currently loads successfully in the broken mode, which makes the failure subtle.

One additional product note from the 32 GB run: the 4-bit checkpoint is ~17.2 GB and Rapid projected a ~25.9 GB working set / 113% total utilization with the current OS load. It did run stably for this short dogfood, but I would describe 32 GB as minimum/experimental and 48 GB as the comfortable target, rather than implying an equally frictionless 32/48 GB experience.

pierre427:
Both maintainer comments are addressed in the updated stacked head `1d709b76`.

- #1223 is now architecture-only, rebased on current `main`, marked ready, and mergeable. This branch is stacked on its exact head pending that merge.
- Parser registration is narrowed to `north` / `cohere_north`; the global `cohere` name is gone.
- Declared-tool policy now covers declared, undeclared, `tools: null`, and `tool_choice: none` cases consistently across streaming and non-streaming paths.
- Shared postprocessor/finalization/sanitizer and stream-parity coverage was expanded, including content-equals-reasoning and hostile envelope cases.
- HF-ID and local-basename auto-detection now select both North parsers by default and advertise the tools capability.
- The public 4-bit no-parser-flags smoke passed: startup selected `north` / `north`, non-stream and SSE normalized the tool call, the tool-result continuation completed, and no North sentinel leaked.
- Documentation now calls 32 GB minimum/experimental and 48 GB comfortable.

Post-rebase validation: 644 affected tests passed with 27 skipped; Ruff, JSON checks, and `git diff --check` passed. The broader SOP gate ran 17,400 passing tests; its unrelated failures were reproduced on clean `origin/main` as documented in the PR body.

The only remaining boundary step is maintainer-controlled: once #1223 lands, I will rebase #1898 onto the resulting `main` so the GitHub diff contains only its three parser/shared-path commits.

raullenchai (changes requested):
Blocking: the parser crosses the text/execution boundary without authenticated wire evidence.

`extract_tool_calls()` (around lines 218-227) treats any bare JSON object/array containing a declared `tool_name`/`name`/`function` as a native tool invocation when `<|START_ACTION|>` is absent. That means an ordinary assistant answer such as `[{"tool_name":"delete_file","parameters":{"path":"..."}}]`—for example, when the user asks the model to show or explain a tool-call JSON example—is upgraded from text into an executable call solely because the request exposed that tool. The comment that some decode paths may consume control tokens is not sufficient provenance: the parser receives no token-level evidence that native START/END control tokens were actually generated. This is a user-impacting accidental-execution / prompt-injection boundary violation.

Please remove the unframed bare-payload promotion, or change the decoder/parser contract so it carries verifiable token-ID evidence that both native action delimiters were consumed and only then permits recovery. Add end-to-end non-stream and streaming regression tests proving ordinary bare JSON stays content even when its name matches a declared sensitive tool.

Also harden argument validation in the same execution gate: `_arguments()` accepts arbitrary JSON-string roots verbatim. Native calls should require parameters/arguments to decode to an object (and ideally validate against the declared schema) before emission; malformed/string/list roots must remain non-executable, with parity tests for auto, required, and forced tool_choice modes.

I did not accept the existing parser tests as proof of safety because `test_marker_stripped_action_payload_uses_checkpoint_compatible_fallback` explicitly blesses the unsafe behavior rather than establishing delimiter provenance.

raullenchai:
Review location correction for the current head: the unsafe bare-payload promotion is at `vllm_mlx/tool_parsers/cohere_tool_parser.py:239-245` (the review text said 218-227). The finding and requested remediation are otherwise unchanged.

pierre427:
Thanks for the pointer correction — confirmed, on this head the block starts at `cohere_tool_parser.py:239`. Force-pushed the remediation.

Since #1223 merged, the branch is rebased onto current `main` and the duplicated architecture commit is gone, so #1898 is now the parser-only layer you asked for.

**The bare-payload promotion is removed, not narrowed.** A response with no `<|START_ACTION|>` marker is no longer treated as North wire evidence at all; it returns as content. Matching a declared tool name is not evidence that the model emitted an action frame, so there is nothing left to gate on. I did not try to substitute a token-level check — removal is the smaller and more auditable of the two options you offered.

The rejection is also marked authoritative, which matters because removal alone was not sufficient here: without it, the same bytes get handed to the generic raw-JSON scanner and promoted anyway, through either the non-stream service fallback or streaming finalization. Both already honor `rejection_authoritative` (`service/helpers.py:3049`, `service/postprocessor.py:3801`), so on the configured `north` path the refusal holds end to end. There is a regression test that drives the non-stream service path specifically, since that was the one that reintroduced the call.

**Argument contract.** Native action arguments must now decode to a JSON object. Malformed JSON and string/list/scalar/null roots fail closed instead of reaching the executor. Checkpoints that serialize the object once before placing it in the envelope are still accepted and normalized. A malformed member now rejects its entire action envelope rather than executing the valid subset — executing part of a partly malformed response silently turns it into a different operation list, which seemed worse than refusing the whole thing. Named forced `tool_choice` additionally narrows the executable name set to the selected tool.

New coverage in `tests/test_cohere_tool_parser.py` and `tests/test_tool_choice_none_suppression.py`: unframed payload stays content; the non-stream service does not re-promote it; JSON-encoded object arguments normalize; non-object roots fail closed with streaming parity; mixed valid/invalid members fail the whole envelope.

Also folded in a one-line docs fix: `tests/test_tool_calling_doc_table.py` pins the Supported Parsers table to the registry both ways, so registering `NorthToolParser` without a table row was failing that test on the previous head. That is unrelated to the security finding, just CI.

Local CPU-only validation on the rebased tree: 62 focused North parser/service/finalization tests pass, and the broad parser/postprocessor/capability/alias selection is 6,314 passed / 11 skipped / 6 xfailed. Three failures in `test_image_lane.py` and `test_no_mllm_flag.py` reproduce identically on the pre-change tree and pass in isolation — test-ordering pollution in that selection, not this change. Ruff and `git diff --check` are clean. No model was loaded and no GPU or real-weight qualification is claimed here; happy to run that if you want it before merge.

raullenchai:

pierre427:
Rebased onto current `main` (240a6d74). #1223 has merged, so the stack collapses to the parser commits (five code commits plus the docs-table row, six total); the interesting part of this rebase was reconciling with #2284, which replaced North reasoning with the `CohereCommand4ReasoningParser` protocol detector while this PR was open.

Resolution (final commit in the stack):
- **Reasoning defers to #2284's detector.** The PR's own `CohereReasoningParser` and its `always_route` postprocessor hook are dropped — the detector already stays engaged when `enable_thinking` resolves false (`sanitize_when_thinking_disabled`) and, unlike the dropped parser, strips text markers while handing action envelopes to a downstream tool parser. Aliases/auto-config now select `tool_call_parser: north` + `reasoning_parser: cohere_command4`.
- **The action-envelope tool parser is the PR's surviving core** (`north` / `cohere_north`), unchanged in wire behavior, but post-tool content preservation now uses the postprocessor's existing result-driven `preserve_post_tool_content` contract instead of a new class attribute.
- **One small detector extension** (`cohere_command_parser.py`): an action that opens *inside* the thinking lane (no `<|END_THINKING|>` yet — a shape North does emit) now returns the stream to the reasoning phase once its envelope closes. Previously the detector treated any action as terminal, so the trailing thinking bytes leaked into public content. Quoted `<|END_ACTION|>` inside JSON string data does not close the envelope (same string-awareness as the rest of the detector).
- **Streaming replay fix (found in pre-push review).** `flush_held_content` was cursor-based, but marker-free deltas take the postprocessor's fast path without engaging the tool parser, so at EOF the stale cursor replayed the whole turn (text-only turns emitted their content twice; a plain prefix before an action was emitted twice). The flush is now structural — it releases only a trailing unclosed action envelope or partial opener suffix, mirroring the Hermes parser's contract — and the streaming scanner syncs its cursor to `previous_text` on first engagement. Two postprocessor-level regressions (text-only turn, fast-path prefix then action) fail on the old parser and pass now.
- `test_cohere_command_reasoning_parser.py`'s pin of `tool_call_parser is None` for North checkpoints is updated to `== "north"` — that pin encoded "no North tool parser exists", which this PR changes deliberately.

- Parser + routing battery (`test_north_reasoning_routing.py`, `test_cohere_command_reasoning_parser.py`, `test_cohere_tool_parser.py`, `test_postprocessor.py`, `test_cohere2_moe_vendored.py`, `test_capabilities_field.py`, `test_api_utils.py`, `test_aliases_contract.py`, `test_tool_call_streaming_parity.py`, `test_tool_call_value_fidelity.py`, `test_tool_choice_none_suppression.py`, `test_model_auto_config.py`): **3954 passed, 27 skipped**; `test_tool_calling_doc_table.py` green; `scripts/audit_tool_parser_coverage.py` reports all 51 registered parsers covered.

raullenchai:
Thank you for rebasing and for reconciling the parser work with the current shared reasoning path. This is deferred until after the 0.13.1 landing window so the engine owner can review the refreshed parser-only scope on current main. No release CI or further contributor action is requested during the window; we’ll route the next review step after the train closes.

raullenchai:
Closing as superseded by the North reasoning/protocol work that has since landed on main. The current product path no longer matches this stacked parser integration, so rebasing this branch would duplicate or revive obsolete behavior. Thank you for the extensive parser, streaming, and tool-policy validation that informed the later implementation.

PR #1906, opened 2026-08-13
feat(mac): streaming experience aligned with native-chat

xiaoxiunique:
Three commits on `textkit-markdown-pr1` making streamed replies feel like native-chat — speed, scroll, and reveal.

**Performance** (root-caused by profiling a 1920-char stream, not guessed):
- Hoisted 4 `NSRegularExpression` instances to `static let` — ICU regex compile was 48% of main thread during streaming (`ModelPickerBar` rebuilds its `Menu` content eagerly and calls `ModelSizing.estimate` per alias per body pass).
- `TokensPerSecondPill` now takes `() -> [ChatMessage]` — reading `chat.messages` in `ContentView`'s body made every SSE delta rebuild the whole view tree.

**Scroll**: `LazyVStack` → `VStack` (lazy rows report estimated heights once recycled, so scroll-to-bottom landed arbitrarily); new jump-to-bottom button with a spinner ring while streaming; probe stops following once an answer outgrows the viewport.

**Fade**: `TextFadeAnimator.storageDidReset()` — `setBlocks` replaces the text storage on every pass and drops rendering attributes, and `contentDidGrow` only re-applies them on growth, so mid-fade text snapped to full opacity; display-link refuses to start off-window.

**Flow**: copy/select/retry actions hidden until a message finishes streaming.

Tests: `SSEDeltaCoalescerTests`, `TextFadeBacklogTests` (threshold pinned at 0.75), `TextFadeDisplayLinkTests`. 2112 total, no new failures.

Why: the prior streaming path rebuilt too much UI, lost fade attributes as storage changed, and could fight readers by forcing the viewport to the bottom. This aligns rendering and interaction semantics with native-chat while preserving link safety, accessibility, and current-main scroll coalescing.

### DIFF
--- a/apps/rapid-mac/Package.swift
+++ b/apps/rapid-mac/Package.swift
@@ -14,6 +14,10 @@ let package = Package(
         // ``MarkdownUI`` renders each block as a real SwiftUI view, à la
         // ChatGPT Desktop. Pinned to the maintenance-line 2.4 series.
         .package(url: "https://github.com/gonzalezreal/swift-markdown-ui", from: "2.4.0"),
+        // Parsing only (swift-cmark underneath, GFM tables + strikethrough).
+        // The TextKit 2 render layer needs a parsed block list, which
+        // MarkdownUI's string-only entry point cannot provide — see #1843.
+        .package(url: "https://github.com/swiftlang/swift-markdown", from: "0.6.0"),
         // Issue #131: LaTeX rendering for math/STEM model responses.
         // ``MarkdownUI`` ships no math engine, so ``$``/``\frac``/``\sqrt``
         // would render as visible tokens. ``SwiftMath`` is the macOS-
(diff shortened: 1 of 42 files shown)


raullenchai:
Maintainer review completed. I rebased the branch onto current `main` and fixed the integration regressions found during review:

- preserved `main`’s coalesced/programmatic scroll safeguards while keeping the new answer-growth release behavior
- fixed Swift 6 actor isolation in the TextKit renderer
- made fade-rate tests deterministic with an injectable monotonic clock
- restored accessible static-text semantics for custom TextKit prose/code
- updated GUI golden-flow scoping and baselines for native TextKit list/code accessibility shapes

Validation:
- release app build: pass
- focused Swift tests: 37/37 pass
- GUI golden flows: `math-rendering`, `slow-stream-stop`, `chat-depth` pass (including relaunch restoration)
- `git diff --check`: pass

Full Swift suite also ran; its only failure was the existing timing-sensitive throughput ratio test under parallel load, which passes in the focused suite.

PR #1907, opened 2026-08-13
feat(rapid-mac): add signed background updates with Sparkle

osdodo:
## What does this PR do?

Adds signed, in-app automatic updates to Rapid-MLX Desktop using Sparkle 2.9.5.

- Checks the Sparkle appcast every six hours.
- Downloads updates in the background when automatic updates are enabled.
- Installs downloaded updates when Rapid exits normally.
- Provides Sparkle's native update UI for manual checks.
- Verifies update archives with Ed25519 signatures.
- Retains the existing `latest.json` / DMG updater as a fallback for builds without a valid Sparkle public key.
- Adds an automatic-download toggle and manual update action to Settings and the menu bar.
- Extends the release workflow to generate, sign, mirror, and publish Sparkle ZIP archives and `appcast.xml`.
- Enforces monotonic `CFBundleVersion` values and publication ordering so an appcast cannot advertise a missing or older archive.
- Updates release, privacy, third-party, and user documentation.

### Required configuration

Generate the Sparkle key pair once on a trusted Mac after resolving the Swift package:

```bash
cd apps/rapid-mac
swift package resolve

.build/artifacts/sparkle/Sparkle/bin/generate_keys \
  --account rapid-mlx

.build/artifacts/sparkle/Sparkle/bin/generate_keys \
  --account rapid-mlx \
  -x ~/Desktop/rapid-mlx-sparkle-private-key
```

The first command prints the base64 public key. The second exports the matching private key.

Configure them under:

`GitHub repository -> Settings -> Secrets and variables -> Actions`

| Name | GitHub type | Source |
|---|---|---|
| `SPARKLE_PUBLIC_ED_KEY` | Variable | Output printed by `generate_keys` |
| `SPARKLE_ED_PRIVATE_KEY` | Secret | Exact contents of `rapid-mlx-sparkle-private-key` |

The public key is embedded in the application and is not confidential. The private key must never be committed, logged, or stored as a repository variable. Back it up in the team's protected credential store: losing it breaks the update chain for existing installations.

A local Sparkle-enabled build only needs the public key:

```bash
cd apps/rapid-mac

export SPARKLE_PUBLIC_ED_KEY='<public key printed by generate_keys>'
bash scripts/build.sh
```

Without `SPARKLE_PUBLIC_ED_KEY`, local builds deliberately remain on the legacy updater.

Tagged releases also require the existing distribution configuration:

| Name | GitHub type | Value/source |
|---|---|---|
| `RAPID_MAC_DIST_R2_BUCKET` | Variable | `rapid-desktop-dist` |
| `RAPID_MAC_DIST_CDN_BASE` | Variable | `https://dl.rapidmlx.com` |
| `CLOUDFLARE_API_TOKEN` | Secret | Cloudflare API token with R2 Storage Edit and Zone Cache Purge permissions |
| `CLOUDFLARE_ACCOUNT_ID` | Secret | Cloudflare dashboard account ID |
| `CLOUDFLARE_ZONE_ID` | Secret | Zone ID for `rapidmlx.com` in the Cloudflare dashboard |

Developer ID signing and notarization continue to use the existing Apple credentials documented in `apps/rapid-mac/RELEASING.md`.

## Why is this needed?

The existing desktop updater can detect a release and offer the DMG, but users must download and reinstall the application manually. It cannot securely replace the installed app or download an update silently.

This change gives installed Rapid-MLX Desktop users a signed, atomic update path with background downloading and installation on normal exit, while preserving the legacy updater for development builds and existing clients that do not yet carry a Sparkle public key.

## Test plan

- Ran the 9 focused Swift tests covering Sparkle configuration, release publication ordering, archive signing, and monotonic build-number guards.
- Built a Sparkle-enabled local application with an injected Ed25519 public key.
- Verified the native Sparkle update window, embedded release notes, archive download, EdDSA verification, installation, and relaunch for build `155 -> 156`.
- Verified a scheduled background check downloaded build `157` without displaying an update window.
- Verified normal application exit silently replaced build `156 -> 157` without relaunching.
- Verified both installed applications retained a valid and matching designated code-signing requirement.
- Verified the Git worktree remained clean after testing.

### DIFF
--- a/.github/workflows/rapid-mac-release.yml
+++ b/.github/workflows/rapid-mac-release.yml
@@ -22,6 +22,8 @@
 #   AC_API_KEY_ID                App Store Connect API key id (notarisation)
 #   AC_API_ISSUER_ID             App Store Connect issuer id
 #   AC_API_KEY_P8_BASE64         base64 of the AuthKey_XXX.p8
+#   SPARKLE_ED_PRIVATE_KEY       private Ed25519 key exported by Sparkle's
+#                                generate_keys tool (base64 text)
 #
 #   Required for tagged releases (the "mirror-dist" and
 #   "publish-updater-fallback" jobs at the bottom). NOTE: on a tag these are
@@ -43,6 +45,7 @@
 #   VARIABLES (not secrets), since they are config, not credentials:
 #   RAPID_MAC_DIST_R2_BUCKET     name of the R2 bucket to mirror the DMG into
 #   RAPID_MAC_DIST_CDN_BASE      public CDN base URL that fronts the bucket
+#   SPARKLE_PUBLIC_ED_KEY        SUPublicEDKey value printed by generate_keys
 #
 # NOTE: Sentry has been removed from the monorepo app; there is no
 # SENTRY_DSN secret here (the desktop workflow had one).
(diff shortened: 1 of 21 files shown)


raullenchai (approved):
Reviewed the updater lifecycle, Sparkle configuration, signed archive/appcast publication ordering, framework/XPC embedding and signing, legacy migration fallback, Settings/menu routing, and rollback guards. Added a stable AX identifier plus update-state golden-flow coverage; verified local release assembly, strict codesign/rpath, invalid-key fail-closed behavior, targeted Swift tests, and the GUI flow. The supply-chain bot's only blocker is the expected release-workflow/private signing key usage, which was manually audited here.

PR #1908, opened 2026-08-13
fix(load): stop model starts hanging on unbounded HuggingFace metadata calls

Jevin-F:
## Why

Starting a model that is **already fully cached** still reached the network, and the call it reached out with has **no deadline at all**, so a hostile or dead network turned "start a local model" into an indefinite hang. The desktop only recovers after its 30-minute stall window.

Two independent defects combine.

### 1. The loader was handed a repo id, not a path

huggingface_hub then resolves the revision through the Hub on every start, warm cache included. The text lane was fixed in #1888; **image, audio and video were not**. Each now resolves the cached snapshot itself and passes the loader a directory, gated on the completeness check that already exists for that layout, falling back to the repo id whenever the cache cannot be vouched for.

The image lane deliberately does **not** use `snapshot_download(local_files_only=True)`. That asks whether the *whole repo* is present and judges it against HF's cached tree listing, which names files the R2 mirror never fetches — so a complete mirror-pulled checkpoint fails on `.DS_Store` / `README.md` / `.gitattributes` and would have silently kept the round-trip this removes. Measured on `filipstrand/Z-Image-Turbo-mflux-4bit`.

### 2. The metadata calls are unbounded

huggingface_hub passes `timeout=None` explicitly into its shared httpx client, and httpx reads an explicit `None` as *disable the timeout* rather than *inherit the client's*. Configuring the client therefore does not bound these calls — measured: a client built with a 2s timeout still hung past 12s on a blackholed route. The mirror's repo listing and the pre-download probe now run under a thread deadline, and an unreachable Hub aborts with an actionable message instead of entering the unbounded lookup.

Two non-obvious constraints are encoded in comments so they are not "simplified" away later:

- The probe **exits** rather than raising, because the generic handler below it converts exceptions into "server will retry" and starts the subprocess, which walks straight back into the same hang.
- The resolved sha is **not** pinned onto `snapshot_download`, tempting as it looks (a commit hash skips the lookup entirely): huggingface_hub only writes `refs/main` when the revision it was handed differs from the resolved commit, so pinning would leave freshly downloaded models without the ref `is_repo_cached` requires — sending every later start back to the network and undoing #1888.

### 3. Stranded download scratch files

Each attempt gets a uniquely named `.incomplete` blob that huggingface_hub removes while unwinding, which a signal-killed process never does, and nothing else ever collects them. Observed on a developer machine as **three files for a single blob, written 49, 75 and 170 hours apart** — one per interrupted attempt, days apart, no concurrency involved.

Reaping is scoped to one repo, skips symlinks, and requires six hours of no writes so it cannot disturb a live download. The HF fallback path now also states that it does not resume, because silently switching between resuming and restarting is what makes a flaky connection read as a stuck download.

## Test plan

- [x] `pytest` across the affected areas — 581 passed (download gate, mirror, CLI chat/models, image lane, IndexTTS, video, Wan, pull summary, disk space, alias subfolder, memory capacity)
- [x] 20 new tests: warm cache resolves offline, partial cache still downloads, deadline interrupts a hanging call and is otherwise transparent, unreachable Hub does not enter the download, reaper collects multiple attempts per blob, reaper spares fresh files / real blobs / symlinks
- [x] `ruff check` + `ruff format --check` clean
- [x] Verified on a real cache: both installed mflux checkpoints resolve to a local snapshot, including the one that fails HF's own whole-repo completeness check
- [x] Built the desktop app with the sidecar and confirmed the bundled engine carries the change and resolves the image model to a local path

## Risk

Every new path falls back to the previous behavior when the cache cannot be vouched for, so a cold or partial cache downloads exactly as before. The reaper's age gate is the safety property for concurrent downloads; unlinking a file another process holds open does not corrupt anything on POSIX, but requiring six hours of silence avoids turning someone else's working download into an error at all.

### DIFF
--- a/tests/test_cli_chat.py
+++ b/tests/test_cli_chat.py
@@ -1534,6 +1534,45 @@ def test_ensure_model_downloaded_calls_disk_check(monkeypatch):
     ]
 
 
+def test_ensure_model_downloaded_aborts_when_the_hub_wont_answer(monkeypatch):
+    """An unreachable Hub must abort the pre-download, not fall through to it.
+
+    ``snapshot_download``'s own revision lookup has no deadline — huggingface_hub
+    hands httpx an explicit ``timeout=None``, which disables the client timeout
+    rather than inheriting it — so entering it on a blackholed network hangs
+    until the desktop's 30-minute stall window. The bounded probe in front of it
+    is what turns that hang into an error, and it only helps if we stop here.
+
+    It must also exit rather than raise: the generic handler at the bottom of
+    ``_ensure_model_downloaded`` converts any non-404 exception into
+    "Pre-download skipped; server will retry" and starts the serve subprocess,
+    which walks straight back into the same unbounded lookup.
+    """
+    from vllm_mlx import _download_gate as gate
+
+    monkeypatch.setattr(
+        "huggingface_hub.try_to_load_from_cache", lambda *_a, **_kw: None
+    )
+    monkeypatch.setattr("os.path.exists", lambda _p: False)
+    monkeypatch.setattr(cli, "_check_disk_space", lambda *_a, **_kw: None)
+    monkeypatch.setattr(gate, "_HF_RESOLVE_TIMEOUT_SECONDS", 0.2)
+
+    downloaded: list = []
+    monkeypatch.setattr(
+        "huggingface_hub.snapshot_download",
+        lambda *a, **kw: downloaded.append(a) or "/tmp/fake",
+    )
+    monkeypatch.setattr(
+        "huggingface_hub.model_info", lambda *_a, **_kw: threading.Event().wait()
+    )
+
+    with pytest.raises(SystemExit) as excinfo:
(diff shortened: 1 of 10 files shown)


raullenchai:
Maintainer review update: rebased onto current `main` and fixed a blocking TOCTOU in the deadline design. The original patch performed a bounded `model_info` probe and then called `snapshot_download` with `main`, which repeated the same unbounded metadata lookup; a network failure between those calls could still hang indefinitely. The download is now pinned to the SHA returned by the bounded probe, and `refs/main` is published atomically only after the snapshot succeeds, preserving the warm-cache fast path. Added regression coverage for both the pinned download call and atomic ref publication.

Local validation:
- affected matrix: 593/593 passed
- focused deadline/cache tests: 193/193 passed
- full suite: 17,480 passed, 235 skipped, 6 xfailed; only two environment-dependent `test_event_loop.py` cases failed because no server was running on 127.0.0.1:8000
- `compileall` and `git diff --check`: pass

PR #1909, opened 2026-08-13
perf(mllm): cache output-router detection + event-driven scheduler wake; fix gemma-3 vision crash [skip-version-bump]

raullenchai:
## Summary

Three fixes to the MLLM (vision) lane that erase rapid's first-token-latency (TTFT) gap versus stock `mlx-vlm` and unblock a whole VLM family that previously crashed. All three are output-preserving and touch only the MLLM path. Decode throughput is unchanged (parity with `mlx-vlm`/`omlx`).

Drilling the vision lane with aligned per-stage probes showed rapid's ~30-40% vision-TTFT gap versus `mlx-vlm` was **not** image preprocessing (~11ms) and **not** the shared `mlx_vlm` forward (~205ms) — it was rapid-specific per-request overhead in two places, plus a cold-start scheduler poll.

### 1. Per-request vocab rescan in the output router (~84ms) — `engine/batched.py`
`_create_output_router()` called `OutputRouter.from_tokenizer_for_streaming()` on **every** streaming request, which calls `tokenizer.get_vocab()` — rebuilding the tokenizer's entire vocab dict (~60-85ms for Gemma's 262k-token vocab) just to check for a handful of marker tokens. The tokenizer and harmony flags are fixed for the engine's lifetime, so the *detection result* is invariant; only the returned router's state machine is per-request. Now the detected `(kind, TokenMap)` is memoized once (keyed on tokenizer identity, so a hot-swap re-detects) and a fresh, cheap router is rebuilt per request. Also benefits every gemma-4 / gpt-oss streaming completion, and any model whose detection returns "no router" (the `get_vocab` scan is skipped after the first request). This is the dominant fix.

### 2. gemma-3 VLM crash → correct output — `mllm_batch_generator.py`
gemma-3 image requests crashed with `TypeError: expand_dims()` (`mlx_vlm/models/gemma3/gemma3.py`). gemma-3's `Model.__call__` forwards a *bidirectional* 4D mask from `get_input_embeddings` to the language model, which overrides the LM's causal mask (first sampled token → EOS, empty completion) and 4D-expands `None` on the text-mask path, raising. Stock `mlx-vlm` avoids this by bypassing `Model.__call__` entirely. rapid now mirrors that for gemma-3 only: run the vision merge via `get_input_embeddings`, then `language_model(mask=None)` so the LM builds its own causal + sliding-window mask. gemma-4 / Qwen-VL are untouched (scoped on `model_type == "gemma3"`). This is numerically identical to the working `mlx-vlm` path (the merge scales image features by 1/√H, the LM scales all embeds by √H).

### 3. Fixed 10ms idle poll in the MLLM scheduler (~9ms cold) — `mllm_scheduler.py`
The scheduler loop slept `asyncio.sleep(0.01)` when idle, so a request arriving at an idle server waited out up to a full poll tick before pickup. Replaced with an `asyncio.Event` set by `add_request` (both on the event-loop thread — single-loop, race-free), with a bounded `wait_for` timeout kept as a liveness floor. Only affects the cold/idle-server first-token case (under load the loop never idles).

## Why not defer the first decode
An earlier revision also deferred the first compute-ahead decode to shave ~26ms more off TTFT. Measured: it **halved** steady-state decode throughput (99→53 tok/s on gemma-4-e4b vision), because the compute-ahead `mx.async_eval` overlaps token N+1's GPU forward with token N's CPU/async delivery — deferring it serializes compute and delivery for *every* token. Reverted; a code comment in `_next` now documents the trap. The remaining two latency fixes already put rapid TTFT at/below `mlx-vlm`, so the trade was strictly negative.

## Results (cross-runtime bench, parity lane, M3 Ultra; rapid / omlx / mlx-vlm)

**Vision TTFT** (median, seconds — lower better):

| model | rapid | omlx | mlx-vlm | was (rapid) | vs mlx-vlm | vs omlx |
|---|---|---|---|---|---|---|
| gemma-4-e4b | **0.269** | 0.427 | 0.232 | 0.293 (−8%) | +16% | **−37%** |
| gemma-4-26b | **0.301** | 0.520 | 0.267 | 0.318 (−5%) | +13% | **−42%** |
| gemma-3-4b | **0.741** | 0.959 | 0.708 | **CRASH** | +5% | **−23%** |
| qwen2.5-vl-7b | 1.377 | 1.045 | 1.361 | 1.329 | +1% | +32% |

**Vision decode** (median tok/s — higher better): parity everywhere.

| model | rapid | omlx | mlx-vlm |
|---|---|---|---|
| gemma-4-e4b | 92.8 | 95.8 | 93.3 |
| gemma-4-26b | 88.9 | 92.2 | 93.3 |
| gemma-3-4b | 121.2 | 125.4 | 123.5 |
| qwen2.5-vl-7b | 110.0 | 110.6 | 109.6 |

- Vision-TTFT gap vs `mlx-vlm` roughly **halved** (was ~30-40%, now +1% to +16%; gemma-3 at +5% went from crashing to competitive).
- rapid **beats `omlx` by 23-42%** on every gemma model's TTFT.
- Decode tok/s at parity (within ±5%) — the compute-ahead overlap is preserved.
- `qwen2.5-vl` loses to `omlx` on TTFT because of `omlx`'s image-resolution cap (dynamic ~999 image tokens); a resolution knob (`#1831` port) is a separate follow-up, not part of this PR.
- Isolated warm-server probes (fixed image, feature-cache warm) put rapid gemma-4-e4b TTFT at ~223ms — at/below `mlx-vlm` — so the bench's residual gemma-4 gap is largely bench-context (parity flags + per-rep image handling), not fixed per-request overhead.

## Verification
- **gemma-3 crash**: red circle / blue square / green triangle / OCR all correct; prove-the-guard confirms the `expand_dims` crash returns when the fix is reverted.
- **Per-stage probes**: `_create_output_router` 84ms→0.01ms; idle-poll 9.4ms→0.2ms. Warm gemma-4-e4b client TTFT ~293ms → ~223ms.
- **No channel-token leak** on gemma-4 (router still routes correctly after caching); text-only completions unchanged.
- **Decode throughput** confirmed unchanged (compute-ahead overlap preserved).

### DIFF
--- a/tests/test_batched_engine_output_router.py
+++ b/tests/test_batched_engine_output_router.py
@@ -603,3 +603,47 @@ def test_create_output_router_catches_tokenizer_property_errors():
     engine = BrokenTokenizerEngine("fake-model")
 
     assert engine._create_output_router() is None
+
+
+def test_output_router_transient_detection_failure_is_not_cached():
+    """A TRANSIENT detection failure must not be cached.
+
+    The router-detection result is memoized per tokenizer to avoid re-scanning
+    the vocab every request. If the *first* detection transiently fails (e.g.
+    ``get_vocab()`` raising during lazy init), caching that ``None`` would
+    permanently disable the router for the tokenizer's whole lifetime — leaking
+    channel-control tokens into every later response. Detection must instead
+    retry on the next request.
+    """
+
+    class FlakyTokenizer:
+        def __init__(self, vocab: dict[str, int]):
+            self._vocab = vocab
+            self.calls = 0
+
+        def get_vocab(self) -> dict[str, int]:
+            self.calls += 1
+            if self.calls == 1:
+                raise RuntimeError("vocab not ready (lazy init)")
+            return self._vocab
+
+        def decode(self, ids: list[int]) -> str:  # pragma: no cover - unused
+            return ""
+
+    tok = FlakyTokenizer(GEMMA4_VOCAB)
+    engine = BatchedEngine("fake-model")
+    engine._loaded = True
+    engine._tokenizer = tok
+
+    # First request: detection raises inside get_vocab -> no router this time,
+    # and crucially nothing is cached.
+    assert engine._create_output_router() is None
+    assert getattr(engine, "_output_router_template", None) is None
+
+    # Second request: detection retried, get_vocab now succeeds -> a real
+    # gemma-4 router is returned (would stay None forever if the transient
+    # failure had been cached).
+    router = engine._create_output_router()
+    assert router is not None
+    assert router.map.format_tag == "gemma4"
+    assert tok.calls == 2
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -2610,26 +2610,101 @@ class BatchedEngine(BaseEngine):
         openai-harmony's ``StreamableParser`` (issue #513). Falls back to
         the legacy custom state machine for non-harmony models and for
         harmony tokenizers whose IDs don't match the official encoding.
+
+        Detection (``from_tokenizer_for_streaming`` →
+        ``tokenizer.get_vocab()``) rebuilds the tokenizer's full vocab dict
+        — ~60-85ms for a 262k-token Gemma vocab — and was previously paid on
+        EVERY streaming request. The tokenizer and the harmony escape-hatch
+        flags are fixed for the engine's lifetime, so the *detection result*
+        (format + marker ``TokenMap``) is invariant; only the returned
+        router's state machine is per-request. Memoize the detected
+        ``(kind, TokenMap)`` once and rebuild a fresh, cheap router per
+        request. This removes the entire ~84ms MLLM first-token overhead (the
+        whole rapid-vs-mlx-vlm vision TTFT gap) and shaves the same
+        per-request cost off every gemma-4 / gpt-oss streaming completion. The
+        per-request router object is constructed exactly as before — this only
+        skips re-scanning the vocab.
         """
+        # Read the tokenizer first — the property can raise mid-lifecycle
+        # ("not loaded" during a startup/teardown race). Treat that as the
+        # legacy no-router path for THIS request without caching anything, so a
+        # later (loaded) request still detects normally.
         try:
             tokenizer = self.tokenizer
-            if tokenizer is None:
-                return None
-            router = OutputRouter.from_tokenizer_for_streaming(
-                tokenizer,
-                force_harmony_streaming=self._force_openai_harmony_streaming,
-                no_harmony_streaming=self._no_openai_harmony_streaming,
-            )
-            if router is None:
-                return None
-            if router.map.format_tag not in _OUTPUT_ROUTER_ALLOWLIST:
-                return None
-            return router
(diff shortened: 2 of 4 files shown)


raullenchai:
### Adversarial review pass (DeepSeek `ds-0731`; gpt-5.6 endpoint was down overnight — rerun before merge)

**Verdict: No blockers.** The reviewer independently verified the concurrency model (MLLM `add_request` runs on the event-loop thread → `Event.set()/wait()` single-loop, no lost-wakeup; `has_requests()` re-reads with no intervening await), the router memoization (read-only `TokenMap`, fresh router per request, stable tokenizer identity key), and the gemma-3 bypass (consistent with steady-state decode which already calls `self.language_model(...)` directly).

Non-blocking findings + resolutions:

- **MAJOR (conditional) — gemma-3 branch omits `vision_cache`/`_image_key`:** premise is false. `_model_supports_vision_feature_cache(gemma-3) == False` (verified live; only gemma-4 is eligible), so the general path wouldn't pass those kwargs for gemma-3 either. No feature-cache to lose → moot.
- **MINOR — gemma-3 branch drops `extra_kwargs` (e.g. `token_type_ids`):** matches existing behavior. gemma-3's own `Model.__call__` calls `get_input_embeddings(input_ids, pixel_values, mask)` and does **not** forward `**kwargs` (`mlx_vlm/models/gemma3/gemma3.py:187`), so the general path drops them for gemma-3 too. Not a regression. (Image path never chunks — chunking is text-only — so the #1187 note doesn't apply here.)
- **MINOR — `self.tokenizer` read twice / idle floor 10ms→50ms for future non-`add_request` writers:** cosmetic; safe today (tokenizer returns a stable object; `add_request` is the sole idle-time writer of new work). Documented as behavioral notes.

No code changes from this pass.

raullenchai:
### Second review pass — DeepSeek `deepseek-v4-pro` (official API)

Also **no blockers**. Its analysis surfaced one sharp correctness question worth recording: *does the gemma-3 `Model.__call__` bypass cover only the first vision forward, or all generated tokens?* — i.e. could subsequent decode tokens still hit the broken non-causal-mask path and collapse to EOS after token 1?

Verified: they don't. Decode goes through `_step` → `self.language_model(input_tokens, cache=cache)` (`mllm_batch_generator.py`), which passes no mask so the LM builds its own causal mask — the broken `Model.__call__` is never on the gemma-3 generation path (first token via the new branch, rest via `_step`). Empirically confirmed by multi-token coherent gemma-3 output (full sentences, not a single-token EOS).

Both independent reviews (v4-flash via Codex, v4-pro via API) + the local analysis converge: no blockers.

raullenchai:
### Review converged — round 2 (gpt-5.6-sol via Codex)

> Prior MAJOR resolved. No blockers.

All gates green: CI (lint / validate / type-check / test-matrix 3.10-3.12 / test-apple-silicon / all l1-smoke incl. gemma3-4b), 131 targeted unit tests + a new regression test (`test_output_router_transient_detection_failure_is_not_cached`, prove-the-guard verified), ruff clean, and independent reviews from DeepSeek v4-flash, v4-pro, and gpt-5.6-sol. Squash-merging.

PR #1910, opened 2026-08-13
feat(models): Ling-3.0-tiny support (fp8)

aftersnow:
## What does this PR do?

Makes [inclusionAI/Ling-3.0-tiny-fp8](https://huggingface.co/inclusionAI/Ling-3.0-tiny-fp8) servable, building on the vendored `bailing_hybrid` backbone:

- **Load-time lossless fp8 → mxfp8 repack** (`vllm_mlx/fp8_repack.py`): mlx has no fp8 dtype ([ml-explore/mlx#1670](https://github.com/ml-explore/mlx/issues/1670)), so the checkpoint's `F8_E4M3` safetensors could not be opened at all. Since the source (e4m3 bytes + ue8m0 128×128-block scales) and mlx's mxfp8 (e4m3 bytes + e8m0 32-element group scales) are the same primitives, the repack is pure byte moves — `mx.dequantize` reproduces `e4m3(w)·2^(s−127)` element-identically, and tensors the publisher kept bf16 (router, lm_head, MLA LoRA projections) load verbatim. `rapid-mlx serve <fp8-dir>` now just works; peak RSS ≈ checkpoint size (~8.6 GB), not the 15.8 GB a bf16 dequant would need.
- **Fused KDA serving layout**: each of the 18 KDA layers dispatched 5 projections + 3 depthwise convs at B=T=1; served row-concatenated as 2 + 1 (mathematically identical; quantized rows are the source bytes verbatim).
- **Opt-in `RAPID_MLX_FP8_LM_HEAD_AFFINE8=1`**: re-quantizes only the untied bf16 lm_head (the single largest per-token read) to affine 8-bit — measured zero top-1 flips on real hidden states. Default stays fully faithful.
- **Fix**: `bailing_hybrid` joins the MLA quantized-KV safelist (its MLA layers carry DeepSeek-V3's compressed-K layout; `--kv-cache-quantization int4` would silently stack error on them).

About the model ([HF card](https://huggingface.co/inclusionAI/Ling-3.0-tiny-fp8)): Ling-3.0-tiny is a lightweight hybrid-reasoning MoE — 7.9B total / 1.3B activated, 3:1 KDA:MLA layer stacking, 128 routed experts (top-8 + 1 shared), MIT-licensed, explicitly validated for Apple Silicon deployment.

## Why is this needed?

The fp8 repo is the reference release for local deployment (the HF card quotes M4 Pro numbers), but mlx/mlx-lm cannot open fp8-block safetensors at all. The repack keeps the published weights bit-faithful with no conversion step and no rewritten weights on disk.

Measured on M4 Pro 48GB (llm-inference-bench v0.4.29, concurrency 1, 60s cells, prefix-cache-busting nonces): faithful serving 89/85 tok/s @2k/4k, 93/90 with the opt-in affine8 head — in line with the HF card's 86–90 tok/s M4 Pro reference.

## AI assistance disclosure

GitHub Copilot (Claude) wrote `vllm_mlx/fp8_repack.py`, the `bailing_hybrid` fusion changes, the `kv_cache_dtype` fix and the tests. I reviewed the changes and verified them with the test suites below, byte-exactness checks against independently decoded weights, a logits/decode/greedy trace against the official torch bf16 implementation, and live serving probes.

## Test plan

- `pytest tests/test_fp8_repack.py tests/test_bailing_hybrid_model.py tests/test_kv_cache_dtype.py` — 45 passed
- `ruff check` / `ruff format --check` on touched files
- Live `rapid-mlx serve <local fp8 dir>` + OpenAI chat probes (with and without the affine8 env var)
- Full unit suite via pr_validate: 17433 passed, 5 pre-existing local-env failures (verified identical on unmodified main)

## Checklist

- [x] Tests pass locally (`python3 -m pytest tests/ -x`) — full suite: 17433 passed, 5 failed; the 5 (ltx23/wan video, metrics_route x3) also fail on unmodified `main` checkouts here (local env, missing optional tooling) and touch nothing in this PR
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 1910` — fetch/lint/targeted-tests (127 passed)/description/supply-chain/test-env all PASS; full_unit carries only the 5 pre-existing local-env failures above
- [x] If new tests touch a critical code path (parser / scheduler / security), I've spot-checked that they fail when the corresponding production line is broken (see SOP §Step 3) — the byte-exactness pins fail on any repack change
- [x] Updated README/docs if applicable — n/a, no user-facing docs for vendored archs

### DIFF
--- a/tests/test_bailing_hybrid_model.py
+++ b/tests/test_bailing_hybrid_model.py
@@ -213,6 +213,17 @@ def test_glm47_parser_handles_bailing_wire():
     assert json.loads(r.tool_calls[0]["arguments"]) == {"city": "Tokyo"}
 
 
+def test_glm47_parser_accepts_null_tools_from_plain_chat_request():
+    """OpenAI request serialization includes ``tools: null`` for plain chat."""
+    from vllm_mlx.tool_parsers.glm47_tool_parser import Glm47ToolParser
+
+    parser = Glm47ToolParser(None)
+    result = parser.extract_tool_calls("A normal answer", {"tools": None})
+
+    assert not result.tools_called
+    assert result.content == "A normal answer"
+
+
 def test_gate_retain_all_groups():
     """topk_group == n_group keeps every group (codex r3 #1: the drop
     path faulted on argpartition(kth=-1))."""
--- a/tests/test_fp8_repack.py
+++ b/tests/test_fp8_repack.py
@@ -268,12 +268,14 @@ def test_online_load_repacks_and_runs(fp8_bailing_checkpoint):
     src, kept = fp8_bailing_checkpoint
     model = load_fp8_model_online(src)
 
-    # fp8-shipped linear became an mxfp8 quantized module with the exact
-    # source bytes...
-    q = model.model.layers[0].attention.q_proj
-    assert q.mode == "mxfp8" and q.group_size == 32 and q.bits == 8
-    src_bytes = kept["model.layers.0.attention.q_proj"]
-    got_bytes = np.array(q.weight).view(np.uint8).reshape(-1)
+    # fp8-shipped linears became one fused mxfp8 module holding the
+    # exact source bytes row-concatenated in q|k|v order...
+    qkv = model.model.layers[0].attention.qkv_proj
+    assert qkv.mode == "mxfp8" and qkv.group_size == 32 and qkv.bits == 8
+    src_bytes = np.concatenate(
+        [kept[f"model.layers.0.attention.{t}_proj"] for t in "qkv"]
+    )
+    got_bytes = np.array(qkv.weight).view(np.uint8).reshape(-1)
     np.testing.assert_array_equal(got_bytes, src_bytes)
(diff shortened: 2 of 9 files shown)


raullenchai (approved):
Reviewed the implementation and validated the real inclusionAI/Ling-3.0-tiny-fp8 checkpoint on a 32 GB Apple Silicon Mac mini.

The support is justified primarily by the reusable vLLM block-FP8 (e4m3/ue8m0) repack path, not by adding Ling to the recommended-model set. The branch now strictly gates the supported FP8 dialect, validates safetensors bounds/layout before repacking, and handles tools=null in ordinary OpenAI chat requests.

Real-server validation passed: checkpoint load/repack, hybrid auto-detection, non-streaming chat, SSE completion with [DONE], and native Ling tool output normalized to OpenAI tool_calls. Targeted tests pass on both machines; prior full CI was green and the final-head CI is running.

PR #1917, opened 2026-08-13
Require explicit confirmation to complete macOS onboarding

loriz-art:
Behavior PR 1 of the Phase UI-2 Onboarding work. Implements the prerequisite contract from Paper §05.1.G so the later Onboarding V3 visual PR has something truthful to render. **No visual refresh is included** — see Scope.

## The problem

Readiness used to end onboarding by itself. The moment the sidecar reported a listening port, the app seeded a welcome message, wrote the completion flag, and dismissed the setup surface. The user was never asked and never confirmed.

Paper §05.1.G retires that ending explicitly — *"Kept for the record, not for build. This ending dismissed setup the moment the model reported ready, gave the user no control and left nothing to confirm … must not be re-introduced."* Readiness is something to **state**; completion is something to **confirm**.

## Four-step state model

The public progress model moves from three steps to four: **Welcome → Choose a model → Download → Start**.

Micro-states collapse into their macro step rather than becoming steps of their own:

| Step | Owns |
|---|---|
| 1 · Welcome | the hero |
| 2 · Choose a model | hardware detection, recommendation loading, recommended / cached / alternative choice, model review |
| 3 · Download | preparing, offline, insufficient disk, interrupted download, download failure, retry |
| 4 · Start | starting, memory confirmation while starting, Ready |

A failure keeps the macro step that owns it. `Phase.failed` carries a new `FailureOrigin`, so a broken pull still reports Step 3 and a load failure still reports Step 4 — a failure never becomes a fifth step, and the rail does not jump when something goes wrong.

The step count now lives in one constant (`QuickstartCoordinator.Step.total`) that every rail reads, and `OnboardingTopBar` takes a `Step` rather than a raw ordinal, so no screen can drift by miscounting. No `Step N of 3` language survives in production.

## Why Ready is now persistent

`Phase.ready` changes meaning: it **parks** onboarding and holds the full-window surface. It writes nothing and dismisses nothing. A new terminal `Phase.dismissed` absorbs the old release semantics (used by `releaseInFlight` when the user revises intent mid-flow).

`markReady(seed:)` is split in two:

- `enterReady()` — park only. Idempotent, and refuses to resurrect a flow already confirmed or released, so repeated readiness notifications (auto-respawn cycle, residency tick, scheduler re-publish) are no-ops.
- `confirmStartChatting(seedWelcome:)` — the completion transaction.

`ModelPickerBar.isQuickstartInFlight` follows the new meaning: `.ready` moved from *released* to *in-flight*, because the onboarding surface is still holding the whole window. Releasing the picker CTA there would claim ownership of a control the user cannot see.

## Exact `Start chatting` transaction order

Guarded on `Phase.ready`, so a double click, a repeated key activation, or a stray re-entry after completion returns `false` without seeding twice, re-writing the flag, or re-running the transition.

**Coordinator half** (authoritative, idempotent):
1. Ensure the intended chat session is the one on screen.
2. Seed the welcome message — exactly once.
3. Persist the completion flag.
4. Retire the pending-Ready record.
5. Release the surface (`phase = .dismissed`).

**Parent half** — runs only on a `true` verdict, so a repeated activation cannot re-fire it:

6. Restore the normal shell (same window, same size — no resize, no flash, no second window).
7. Route to Chat.
8. **Announce completion** via the existing `VoiceOverAnnouncer`.
9. **Request composer focus.**

The announcement is posted **before** the focus request: a VoiceOver user who hears the field described first never learns that the thing they just activated actually worked. No delay, no fake progress, no intermediate loading state after the click.

Return and Space activate the button through ordinary AppKit semantics; no custom key handling was added.

## Relaunch and persistence

A new key, `rapid.quickstart.v1.pendingReadyAlias`, records a flow that reached Ready without being confirmed. The existing completion key keeps its meaning but with a **narrowed writer** — it is now written only by the user's confirmation.

The stored alias is a claim about a previous launch, never evidence about this one:

- `init` restores the model and the chooser stage, but **never** `Phase.ready`.
- Ready is re-entered only when `ServerManager` genuinely reports `.ready` for that alias **on this launch**.
- If the model is no longer ready, the user lands on the ordinary chooser with their pick preselected. Nothing claims a download or a selection was "resumed", and no cross-relaunch byte-level download resume is promised.
- A relaunch auto-start of that same alias reports `.starting` truthfully as Step 4 rather than fabricating Ready.

The record is retired on confirmation, skip (including Esc / swipe-down, which the existing code already documents as the same intent), revoke, a fresh download, selecting another alias, and when the app serves a different model.

## Welcome-message deduplication

Seeding is **wired for the first time** — the closure was previously a stub that returned `true` and seeded nothing, so the welcome never actually reached production users.

Three independent guards: the coordinator calls the seed closure only from the confirmation transaction and only when `!hasSeededWelcome`; `ChatViewModel.seedAssistantWelcome` refuses to append into a transcript that already holds messages; and the message is written as a finished assistant turn, so it can never wedge the typing indicator or the streaming gate.

If the seed genuinely cannot land, completion still proceeds — the user asked to start chatting and must not be stranded on a screen they already dismissed — and `awaitingWelcomeSeed` records that a welcome is still owed.

## Focus and VoiceOver

`ChatView` gains a `composerFocusRequest` counter that routes through the existing `composeFocusToken`, so every focus request in that view (Send, ⌘L, onboarding completion) reaches the editor by one path. A counter rather than a Bool, so a second request is distinguishable and nothing needs resetting; `0` means "never asked", so a plain mount never steals focus.

## Automated tests

From the final rebased commit `960c69e7`:

- **Focused onboarding + affected suites: 203/203 pass** (`OnboardingCompletionBehaviorTests`, `QuickstartViewTests`, `AccessibilityIdentifierInventoryTests`, `ModelPickerBarSectionOrderTests`, `DiskSpaceProbeTests`, `QuickstartCachedModelTests`, `LaunchOnboardingOrderingTests`, `QuickstartBrowseAllDestinationTests`, `ChatComposeAccessibilityTests`, `ModelSwitchHistoryTests`, …).
- **Full suite: 2209 tests / 199 suites.** All pass except one pre-existing failure — see below.
- Clean release build; app bundle assembled and ad-hoc signed; `codesign --verify --deep --strict` passes.

30 new regression tests cover: four-step mapping; Download is Step 3; Starting and Ready are Step 4; every phase maps inside the four steps; no `of 3` language survives; readiness alone does not complete; readiness alone does not dismiss; repeated readiness is harmless; `Start chatting` completes exactly once; welcome seeded exactly once; repeated activation is harmless; activation outside Ready is a no-op; pending Ready survives reconstruction; relaunch with a genuinely ready model returns to Ready; relaunch without one does not fabricate Ready; Ready re-entry is gated on the served alias; completion clears provenance; changing alias clears stale provenance; skip clears provenance without completing; composer focus is requested after completion; announcement precedes focus; and that Skip, cached-model, low-disk, memory-warning, failure, and Browse-all behaviour are intact.

No test was weakened, skipped, deleted, or renamed to obtain a pass. Existing tests that asserted the *old* automatic-completion contract were rewritten to the new one, since that contract is what this PR deliberately changes.

### Pre-existing failing test — not a regression, and not a clean run

`ThroughputCaptionIntegrationTests` → *"The persisted stats separate prefill from decode on a real stream"* fails on this machine.

I verified it against clean `upstream/main` at the **identical base commit** (`3913de76`) in a separate worktree: it fails **3/3 there too**, with the same wall-clock ratio assertion. On this branch it also fails 3/3. It is a machine-speed-dependent throughput ratio assertion, unrelated to this change.

`DownloadCatalogHardeningTests` → *"ModelCatalog terminates catalog subprocesses when the async task is cancelled"* is intermittent (1 of 4 runs here), also pre-existing — a 1-second subprocess-teardown timeout.

Stating this plainly: **the final full-suite run is not clean.** The failures are pre-existing and reproduce on unmodified upstream, but I am not presenting the run as green.

## Manual verification

Verified end to end in a fresh, fully isolated installation — throwaway bundle identifier and preferences domain, throwaway `HOME` and Application Support, cold model cache, pinned high port with the launch port-sweep disabled — against a real bundled engine:

1. Welcome renders as **Step 1 of 4**.
2. Model selection renders as **Step 2 of 4**.
3. A real cold-cache download renders as **Step 3 of 4**.
4. Starting and Ready render as **Step 4 of 4**.
5. Ready stays visible and does not auto-dismiss.
6. `Start chatting` enters Chat in the same window.
7. The welcome message appears exactly once.
8. The composer receives focus.
9. **Quitting while on Ready and relaunching the same isolated install does not mark setup complete and does not jump to Chat.** The model starts truthfully and the app returns to Ready; `Start chatting` is still required.
10. After confirming, relaunching no longer shows onboarding.

## Scope

Deliberately **not** in this PR. Two were found while verifying it and are confirmed pre-existing on `upstream/main`:

- **Phantom `No` model row in an empty cache.** With no models downloaded, the engine prints `No models cached yet. Run 'rapid-mlx pull <alias>' …`, and `ModelCatalog.parseCached` splits it on single whitespace into `alias="No"`, `hfRepo="models"`, `size="cached yet."`. `parseAvailable` already guards against exactly this class (it rejects the same line); `parseCached` uses single-whitespace splitting by design and only has a banner guard that never anticipated this notice. Reproduced at parser level; producer and consumer are byte-identical to `upstream/main`. Affects every genuinely new user, so it matters — but it belongs to the next **model-selection/catalog Behavior PR**, in `ModelCatalog.parseCached` where the value is minted, not in an onboarding view.
- **Disappearing button labels on hover.** `RapidButtonStyles` paints the hover wash in an `.overlay` (above the label) while `RapidTheme.hoverFill` is fully opaque, so the text is erased on hover. Affects every `.rapidPrimary` / `.rapidSecondary` button app-wide. Untouched by this branch; belongs to the later **Visual PR**.

Also excluded: in-window `Browse all models`; model detail / storage confirmation; catalog & loading redesign; download cancellation / recovery changes; missing-engine routing; memory-compatibility recovery; Quickstart redesign; Settings and other unrelated surfaces.

**No Direction D / Onboarding V3 visual refresh is included.** The Ready view is deliberately minimal and built from existing components and styling, present only so the new behaviour can be exercised and tested. No Direction D tokens, colours, typography, spacing, rail, cards, animation, or responsive layout work. The visual PR consumes this contract and must not redefine it.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/ChatViewModel.swift
@@ -445,6 +445,30 @@ final class ChatViewModel {
         messages = seeded
     }
 
+    /// Append a locally authored assistant message to the open conversation.
+    ///
+    /// The one caller is the onboarding completion transaction, which lands
+    /// its welcome message in the chat the user is about to be dropped into.
+    /// There is no network round trip and no stream: the text is written
+    /// straight into the transcript as a finished assistant turn, so it can
+    /// never wedge the typing indicator or the streaming gate.
+    ///
+    /// A transcript that already holds messages is left alone. Onboarding
+    /// completion is a one-shot event, but the app can reach it with a live
+    /// conversation on screen (a user who skipped setup, chatted, then came
+    /// back to it), and injecting a stray intro into the middle of somebody's
+    /// conversation is worse than skipping the pleasantry.
+    ///
+    /// - Returns: ``true`` when the message landed in the transcript.
+    @discardableResult
+    func seedAssistantWelcome(_ text: String) -> Bool {
+        let trimmed = text.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { return false }
+        guard messages.isEmpty else { return false }
+        appendMessage(ChatMessage(role: .assistant, content: trimmed))
+        return true
+    }
+
     /// Append the user message, open a placeholder assistant row, and
     /// kick off the streaming task. The text field clears immediately on
     /// the caller's side.
--- a/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
@@ -202,6 +202,15 @@ struct ChatView: View {
     /// & start, retry). Owned by ``ContentView`` because starting a model
     /// is a window-level concern, not a chat-surface one.
     var onReadinessAction: (ModelReadiness.Action) -> Void = { _ in }
+    /// Monotonic counter the parent bumps when something outside this view
+    /// wants the composer to take keyboard focus — today, the onboarding
+    /// completion transaction dropping the user into their first chat.
+    ///
+    /// A counter rather than a Bool so a second request is distinguishable
+    /// from the first and nothing has to be reset afterwards. Zero is the
+    /// "never asked" value; ``ComposeTextEditor`` ignores it, so a plain
+    /// mount does not steal focus from whatever the user was doing.
+    var composerFocusRequest: Int = 0
 
     /// Backing state for the composer's inline model picker (Ollama-style).
     /// The picker lives in the compose box now, not a top control bar.
@@ -271,6 +280,14 @@ struct ChatView: View {
         // Drop a stale error banner once the server is provably ready.
         .onChange(of: server.state) { _, newState in
             if case .ready = newState { viewModel.clearStaleErrorBanner() }
+        }
+        // Forward an external focus request onto the composer's own token.
+        // Routed through the existing token rather than a second mechanism
+        // so every focus request in this view — Send, Cmd+L, onboarding
+        // completion — reaches the editor by the same path.
+        .onChange(of: composerFocusRequest) { _, request in
+            guard request != 0 else { return }
+            composeFocusToken &+= 1
         }
     }
 
--- a/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
@@ -40,6 +40,10 @@ struct ContentView: View {
     @SceneStorage("Rapid.showLogs") private var showLogs: Bool = false
     /// Per-session "browse all models" dismissal of the Quickstart card.
     @State private var quickstartDismissedThisSession: Bool = false
+    /// Monotonic composer-focus request forwarded to ``ChatView``. Bumped
+    /// by the onboarding completion transaction so the user lands in their
+    /// first chat with the caret already in the message field.
+    @State private var composerFocusRequest: Int = 0
     @Environment(SettingsRouter.self) private var settingsRouter
     /// #223: launch-time auto-start "needs download" state — the empty
     /// state names the pending pull when non-nil.
@@ -164,6 +168,16 @@ struct ContentView: View {
             // ``.ready(<alias>)`` against a different alias than shown.
             if case .ready(let serving) = newState, !serving.isEmpty, serving != alias {
                 alias = serving
+            }
+            // Retire pending-Ready provenance the moment the app serves some
+            // OTHER model. The record names the alias it is about precisely
+            // so it can be falsified like this: once the user has moved on,
+            // offering to confirm the old flow on the next launch would be
+            // confirming something that is no longer true.
+            if case .ready(let serving) = newState,
+               quickstart.hasPendingReady,
+               serving != quickstart.selection.alias {
+                quickstart.clearPendingReady()
             }
         }
         .alert(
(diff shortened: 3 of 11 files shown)


raullenchai:
Maintainer review update: the product direction and the coordinator state model look sound. I found and fixed one concrete accessibility/automation blocker while exercising the existing cached-onboarding GUI journey: the `Quickstart.Ready` identifier on the container propagated over the child button, so AX exposed the actionable control as `Quickstart.Ready` instead of the declared `Quickstart.Ready.StartChatting`. That made the new explicit completion action unreachable to the golden-flow harness and would give assistive automation an imprecise target.

I pushed a small fix to this branch that removes the container identifier and adds end-to-end golden coverage for Step 1/2/4, persistent Ready, explicit Start chatting, and exactly-once welcome seeding.

Local verification:
- affected Swift suites: 112/112 passed
- `cached-quickstart` GUI golden flow: passed
- full Swift suite: 2,208/2,209 passed; the sole failure is the known timing-sensitive `ThroughputCaptionIntegrationTests` ratio assertion, also reproduced on main and unrelated to this PR

I am waiting for the refreshed CI before landing.

raullenchai (approved):
Reviewed the onboarding completion semantics, persistence behavior, accessibility surface, and GUI regression coverage. The direction is sound; the AX identifier blocker and explicit-completion golden flow are fixed in the current head. Local targeted tests and cached-quickstart golden flow pass, and all hosted CI checks are green.

PR #1928, opened 2026-08-14
fix(mac): download and verify audio models before serving

osdodo:
## What does this PR do?

  - Adds downloaded/not-downloaded icons to the Audio model picker.
  - Makes “Download & start” run `rapid-mlx pull`, display download progress, wait for completion, verify the local cache, and only then start the audio server.
  - Keeps partially downloaded models downloadable so clicking the action resumes the pull instead of reporting Start or Ready.
  - Validates index-less numbered safetensor shard sets to prevent partial checkpoints from appearing runnable.
  - Recognizes the `config.json` + `weights.npz` layout used by mlx-community Whisper models.
  - Adds Python, Swift, and GUI regression coverage for audio download readiness.

  ## Why is this needed?

  The Audio page could report Ready immediately because the audio server exposes a healthy `/healthz` response before its lazily loaded model weights are available. Interrupted downloads could also appear as Start/Ready when only part of a sharded checkpoint existed.

  Additionally, fully downloaded mlx-community Whisper models were rejected because they use `weights.npz` instead of safetensors. This caused “Download & start” to finish with “Rapid couldn't find the model on disk” even though the model had downloaded successfully.

  These changes ensure uncached and partial models continue downloading, completed Whisper models are recognized correctly, and Ready is shown only after the selected model has been verified on disk and started.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/AudioView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/AudioView.swift
@@ -21,6 +21,7 @@ struct AudioView: View {
     @State private var playingPreviewVoice: String?
     @State private var voicePreviewTask: Task<Void, Never>?
     @State private var voicePreviewRequestID: UUID?
+    @State private var modelLoadsInFlight: Set<String> = []
 
     private let contentMaxWidth = RapidTheme.Layout.contentMaxWidth
     private let controlLabelWidth: CGFloat = 80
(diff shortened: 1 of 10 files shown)


raullenchai:
Thank you — the direction is correct, and the core fix closes a real gap: an audio-only sidecar can become process-healthy before its lazy STT/TTS weights are actually available, so `Download & start` needs an explicit pull and cache proof before serving.

I rebased/reviewed the latest head and added two maintainer hardening changes before the final CI pass:

- Whisper `config.json` / `weights.npz` cache proof now rejects symlinks that resolve outside that model repo’s HF cache, matching the existing video/mflux trust boundary.
- The GUI fixture is now stateful: after each pull completes it exposes that exact audio alias as cached, so the blocking journey proves the full sequence for both Speech and Transcription: downloading (never early Ready) → verified cache → server start → readiness dismissed. Previously the fixture stayed incomplete forever, so the flow only proved the negative half.

Targeted review results so far: 188 Python contract tests pass, all 13 `AudioCatalogTests` pass, release Desktop.app builds, Ruff/Bash/diff checks pass, and the real `mlx-community/whisper-medium-mlx` repository was verified to use exactly `config.json` + `weights.npz`. Full CI/GUI is running on the updated head now.

PR #1930, opened 2026-08-14
fix(mac): link cursor rects are O(n²) on long answers

xiaoxiunique:
## Summary

- replace `resetCursorRects`' all-character scan with direct `.link` attribute enumeration
- ask TextKit 2 for the laid-out line segments of each entire link range, avoiding per-character rect queries and hand-merged geometry
- reuse the same link regions for click hit-testing and pointing-hand cursor rects

This removes the O(n²) path on long link-free answers. Work now scales with link runs and their rendered line segments rather than total answer characters; even a very long URL no longer falls back to per-character layout calls.

## Validation

- `swift test --no-parallel --filter 'MarkdownLinkPerformanceTests|TextKitMarkdownParityTests.customTextKitLinkHitTesting'`
- long 2,000+ character link-free answer yields no cursor rects
- wrapped links expose a clickable rect on every rendered line
- adjacent distinct links retain distinct URL hit targets
- existing compiled-markdown click behavior remains green

No dependencies, scripts, network calls, or process-launch behavior are added.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Render/MarkdownTextBlockView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Render/MarkdownTextBlockView.swift
@@ -218,9 +218,17 @@ final class MarkdownTextBlockView: NSView {
 
     public override func resetCursorRects() {
         super.resetCursorRects()
-        for offset in 0..<renderer.proseLength {
-            guard let rect = renderer.rect(forCharacterAt: offset),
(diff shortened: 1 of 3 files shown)


raullenchai (approved):
Reviewed and tightened on current main. The original O(n²) diagnosis and resetCursorRects direction are correct. The implementation now enumerates TextKit line segments for whole link attribute ranges rather than calculating and heuristically merging one rect per link character, so cost scales with link runs/rendered lines and long URLs remain efficient. Added regression coverage for long link-free prose, wrapped links, adjacent distinct URLs, and existing click hit-testing. No dependency, script, network, file-I/O, or process-launch changes.

PR #1934, opened 2026-08-14
fix(mac): hovering any button covered its label with a grey block

xiaoxiunique:
## The bug

Every button in the app turned into a solid grey block on hover, hiding
its own label. Pressing it brought the label back.

`RapidButtonSurface` paints hover and pressed as an **overlay**, so one pair
of values reads correctly over an amber primary, a white secondary and a
borderless tertiary. Its comment has said so since #1460:

```swift
// Hover and pressed are drawn as a wash ON TOP of the
// tier's own fill, so one pair of values reads correctly
// over amber, over white, and over nothing.
```

That contract requires both values to be translucent.

Phase UI-2 (#1905) recast `hoverFill` from `Color.primary.opacity(0.055)` to
an opaque mineral plane — `#F0F0ED` light, `#212429` dark. **That change is
correct** for what it was made for: the twelve rows, chips and pickers that
paint it *behind* their content, so one hover colour lands the same on all
four surface planes. It is wrong only for the single consumer that paints it
*in front*, and #1905 never touched that file — nothing connected the two.

Measured on the built app, both appearances:

| token | alpha |
|---|---|
| `hoverFill` | **1.0** |
| `pressedFill` | 0.085 |

Same overlay. Hence: hover covers the label, press reveals it.

Forty-seven call sites route through this surface — `.rapidPrimary` 12,
`.rapidSecondary` 26, `.rapidTertiary` 3, `.rapidDestructive` 3, plus 3
explicit.

## The fix

`hoverWash` restores the pre-#1905 value for the over-content position only.
`hoverFill` keeps its plane semantics and all twelve of its consumers, which
I checked one at a time — every one of them draws it as a background.

This is deliberately *not* a revert of #1905. The plane colour is the right
answer for rows; a button compositing over its own amber fill needs a wash.
Two positions, two tokens.

## Verification

Rendered a `.rapidSecondaryCompact` button through `ImageRenderer` with the
surface's hover state temporarily forced on, then counted pixels:

| | before | after |
|---|---|---|
| label pixels (`#313131`) | 0 | **409** |
| stroke pixels (`#DDDCD9`) | 0 | **696** |
| button interior | one flat colour | wash over fill, label and border intact |

The forced-hover edit was reverted; what ships is the token split and the
tests.

## Tests

`ControlWashTests` pins the invariants that make the two tokens different,
rather than their values:

- an over-content wash is translucent in both appearances
- pressed reads firmer than hover
- the row plane stays opaque, so the pair is not later merged back into one

Break-verified by pointing `hoverWash` at `hoverFill`: the first two fail in
both appearances.

Full suite: 2221 Swift tests. The three failing suites (`Web tools
hardening`, `Download/catalog hardening`, `Throughput caption integration`)
fail identically on unmodified `main` on this machine — DNS NXDOMAIN
hijacking by the local ISP, and two timing-sensitive cases.

## Not fixed here

Measured while investigating, left for a separate change because both are
design calls rather than defects:

- **Hover is faint.** With the wash restored, hover contrast ratio is 1.10
  in light and **1.036** for dark-mode amber — close to imperceptible.
  `utility: false` tiers also leave the label colour unchanged on hover, so
  the wash is their only signal.
- **36 `.buttonStyle(.plain)` buttons have no hover feedback at all** —
  concentrated in `ChatView` (7), `AudioView`, `DownloadStrip`,
  `ContentView`.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/Components/RapidButtonStyles.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/Components/RapidButtonStyles.swift
@@ -221,9 +221,13 @@ private struct RapidButtonSurface: View {
                 // Hover and pressed are drawn as a wash ON TOP of the
                 // tier's own fill, so one pair of values reads correctly
                 // over amber, over white, and over nothing.
+                //
+                // Both values must therefore be translucent. ``hoverFill`` is
+                // not — it is the opaque plane colour rows paint behind their
+                // content — so this position takes ``hoverWash``.
                 if isEnabled, configuration.isPressed || hovering {
                     RoundedRectangle(cornerRadius: RapidTheme.Radius.button, style: .continuous)
-                        .fill(configuration.isPressed ? RapidTheme.pressedFill : RapidTheme.hoverFill)
+                        .fill(configuration.isPressed ? RapidTheme.pressedFill : RapidTheme.hoverWash)
                 }
             }
             .overlay {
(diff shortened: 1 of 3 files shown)


raullenchai (approved):
Reviewed the overlay/token semantics and the full patch. `hoverFill` remains the opaque behind-content plane; `hoverWash` restores a translucent over-content layer for all Rapid button tiers. No dependency, process, network, or filesystem changes. Targeted `ControlWashTests` pass in light and dark appearances.

PR #1935, opened 2026-08-14
fix(mac): keep the composer placeholder clear of IME text

Jevin-F:
## Summary

- hide the composer placeholder while AppKit has input-method pre-edit (marked) text
- preserve marked text across unrelated SwiftUI updates instead of replacing it with the still-empty binding
- remeasure the composer while a long pinyin/kana/jamo pre-edit run wraps
- report composition transitions only once, including AppKit's nested `unmarkText` path

The original hover-label commit has been removed during rebase because #1934 already landed the same root-cause fix with the reviewed `hoverWash` visual treatment and regression coverage. Reintroducing the older opaque-background approach would overwrite the button tier fill.

## Validation

- `swift test --no-parallel --filter ChatComposeAccessibilityTests`
- exercises a real `AutosizingTextView.setMarkedText` → `unmarkText` lifecycle
- pins the binding-sync rule that prevents SwiftUI from erasing in-flight IME text

No dependencies, scripts, network calls, or process-launch behavior are added.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
@@ -1651,6 +1651,11 @@ struct ComposeField: View {
     /// NSTextView balloon to a giant centred textarea).
     @State private var contentHeight: CGFloat = 22
 
+    /// True while an input method is showing pre-edit text. ``text`` is
+    /// empty for that whole phase, so this is what keeps the placeholder
+    /// from rendering underneath a half-typed pinyin / kana run.
+    @State private var isComposing = false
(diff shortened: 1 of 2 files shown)


raullenchai (approved):
Reviewed after rebasing onto current main. The superseded hover implementation was removed in favor of the already-landed #1934 hoverWash fix. The remaining IME change correctly preserves marked text across SwiftUI updates, hides the placeholder during pre-edit, remeasures wrapping composition, and now emits composition state only on edges. Added real AppKit marked-text lifecycle and binding-sync regression coverage. No dependency, script, network, or process-launch changes.

PR #1946, opened 2026-08-14
Make macOS onboarding recovery states truthful and reachable

loriz-art:
## Why

Onboarding could tell a user something untrue about what had just happened, or leave them holding a control that was not a way out. This is the final prerequisite behaviour slice before the Direction D visual PR, which will consume these states rather than redefine them.

**Behaviour only.** No Direction D styling, no Quickstart redesign, no Settings work, no Paper edits. **Direction D visuals will be implemented separately.**

## Truthful cancellation vs. genuine failure

A cancelled download and a broken one shared `FailureDiagnosis.Kind.downloadFailed`, so stopping a pull yourself produced *"Check your connection"* — advice about a fault that had not occurred. Paper 05.1 state 10 flags exactly this and names `FailureDiagnoser` as the fix site.

- New `.downloadCancelled` kind, severity `.notice` (the user chose it, so it is not painted as a malfunction), action `.retry`, heading `Download stopped`.
- `DownloadManager` records the kind at **every** cancellation site, so no surface has to re-infer the cause from prose.
- Copy states what happened and stops. No connection advice, and deliberately silent about whether a fresh pull reuses bytes already on disk — that is a property of the downloader, not a promise this app can make. **No Pause, no resume claim.**
- Offline stays a third, distinct diagnosis (`.downloadSourceUnavailable` → *Switch source*).

## A reachable `Cancel download` in Step 3

Found during manual verification: nothing inside setup could *ask* for a cancellation. Onboarding is a full-window sheet, so `DownloadStrip` sits behind it for the entire transfer — a 63%-complete download had no exit but quitting the app. The recovery path existed and was correct; it was simply unreachable.

- `Cancel download` on the Step 3 card, wired to `downloads.cancelDownload(alias: coordinator.selection.alias)`.
- Shown only while the matching job is running; gone once it settles.
- Repeat presses are inert — the optimistic flip to `.cancelled` lands on the same run-loop turn, but a second press would re-signal a process mid-SIGTERM and arm a second hard-kill timer.
- **No keyboard shortcut**, deliberately: `.defaultAction` would put a destructive action on Return on a screen whose job is waiting, and Escape already means "retreat within Step 2, else leave setup".
- Identifier `Quickstart.Download.Cancel`, label naming *which* download, hint stating the consequence.

## Retry, and the route back to choosing

- Retry targets the same intended model and creates **no duplicate job** — `retryDownload` dismisses the terminal record before starting the next child, and refuses to overlap a running pull or a live process.
- The failure card's only route back was *"browse all models"*, which sent everyone to the catalogue even if they had never opened it, and which **ejected the user from setup entirely after a load failure** (a crashed serve makes the wizard ineligible, so returning to `phase .idle` closed the sheet).
- It now returns to the micro-stage the user actually left — **shortlist, Browse All, or Review Download** — naming its destination, and clears the terminal server state so the move stays inside onboarding.
- **Selection, search query, filter, sort and scroll anchor are all preserved** across the failure and the return.

## Pre-flight: disk and memory

- Insufficient disk **interposes before the pull starts**; `Cancel` returns to the exact Step 2 origin, and `Continue anyway` remains available. Warn-only is retained per Paper 05.1 state 12 and existing `DiskSpaceProbe` behaviour — it is not turned into a hard gate.
- The warning no longer attributes the flat 2 GiB pre-flight floor to *"&lt;model&gt; weights + safety margin"*; it says the number is a flat setup floor, not the size of the chosen model.
- Incompatible/insufficient memory blocks the invalid action before download or start, in every list context, and the disabled primary keeps its neutral verb. The alias stays selected — only its actionability changes.
- Declining the pre-load memory guard returns to the exact origin.

## Missing engine

`refreshBinary()` always ran a real `ServerLocator` re-resolution, but when the engine was still absent it changed nothing observable — the one control on that screen read as a dead button.

- `refreshBinary(userInitiated:)` records a `BinaryRecheck` (found + attempt count) and the overlay states the result.
- Repeated checks are audibly and visibly distinct; the success branch reads differently again.
- No invented cause, no promised retry schedule. Launch and Settings refreshes leave no "you rechecked" state behind.
- Recovery stays in the existing window architecture — no separate architecture change was required.

## Relaunch, and what is never fabricated

- A relaunch that finds setup unfinished now says so: the welcome primary reads **`Continue setup`** and the screen states that nothing was carried over.
- **Nothing is carried forward** — no selection, no job record, no phase — so a fake active transfer, percentage, ETA or Step 3 state is impossible by construction rather than by remembering to avoid it (Paper 05.1 state 18: *"The copy promises a fresh download, never a resume."*).
- **Pending Ready keeps precedence**: an unconfirmed Ready flow still reopens on its own model rather than the welcome hero, and still never claims the model is up on this launch (PR #1917 behaviour, pinned against regression here).
- Skip leaves setup owed and still resumable; completion retires the resume record, across relaunch.

## Preserved

The merged four-step model, persistent Ready, `Start chatting`, and the in-window model-selection flow from #1917 / #1931 are unchanged and covered by regression assertions. Cached and uncached successful paths both verified end to end.

## VoiceOver — five contracts

Two of the five did not exist: a cancellation replaced the control the user had just pressed, and a genuine failure arrived with no interaction at all — both in silence.

| Contract | Status |
|---|---|
| Cancellation announcement | Added + covered |
| Genuine-failure announcement | Added + covered |
| Every repeated Recheck result | Covered (string + wiring) |
| Successful Recheck result | Covered (string + wiring) |
| Back control naming its destination | Spoken label added + covered |

Every recovery arrival announces through one helper, composed from the same diagnosis kind the card renders from, so speech and pixels cannot drift. The Back control gained an explicit spoken label with the leading arrow stripped — VoiceOver reads U+2190 aloud as "left-pointing arrow".

Coverage is split honestly: the strings are pure functions asserted directly; the *wiring* is asserted at source level, because `VoiceOverAnnouncer.announce` posts into AppKit's accessibility bus and returns `Void`. This matches the existing pattern in `OnboardingCompletionBehaviorTests`, and the guard was negative-checked — deleting the announce call fails the test.

## Verification

- **60 focused recovery tests pass** (`OnboardingRecoveryBehaviorTests`).
- **Build passes; AX identifier gate passes.** `verify-recommendation-tiers`, `verify-conversation-ordering` and `verify-chat-timeout` all pass.
- **Live manual passed** against an isolated build with the real bundled sidecar (unique bundle id, isolated `HOME`, Application Support, prefs and cold model cache): cancellation, retry, genuine failure, all three Back destinations, incompatible memory, Missing Engine + Recheck, relaunch, pending Ready, cached and uncached success, `Start chatting`, and Light/Dark at 1440x900, 1000x700 and 720x560.
- **VoiceOver was automated-only and was NOT verified with a screen reader.** The tester does not use VoiceOver.
- **The low-disk warning could not be safely produced live** and is automated-only.
- **The full suite has one known unrelated failure** — `MarkdownLinkPerformanceTests` → "Adjacent links keep their own hit targets" (`rects.count → 3` vs `2`) — which **reproduces on clean `upstream/main`** and is untouched by this branch. Full suite otherwise: 2365 tests.

## Deferred follow-up — separate Behaviour PR

`WON'T FIT` catalogue rows are dimmed and unselectable, so their informational Review is unreachable and the reason for the block is not discoverable. They should become **discoverable and openable to a read-only informational Review while the invalid primary action remains disabled** (Paper `05.2.D · V3/Onb-2e-ReviewDownload-IncompatibleMemory`). Excluded here because it *enables a new user action*, which is this repo's own test for "behaviour, not appearance". Invalid-action blocking itself is verified and passing in this PR.


loriz-art:
@raullenchai The branch has been updated with the latest `main` using a merge commit, and the conflicts are resolved. The four workflows are now awaiting maintainer approval. Could you please approve them when convenient? Thanks!

raullenchai (approved):
Reviewed the onboarding recovery state machine and external-contribution supply-chain surface. I also added real GUI coverage for cancel → truthful stopped state → return to the exact model-review stage, and fixed stale user-initiated recheck state leaking into background refreshes. Focused recovery tests, Swift build/tests, and the new download-progress Golden path pass on the Apple Silicon runner. No blocking findings.

PR #1960, opened 2026-08-15
chore: bump version to 0.12.14

raullenchai:
## Why

Cuts the release for the 15 commits that have landed since `v0.12.13`. The headline is `qwen3.8-27b-mixed-3.5bpw` — [`rapid-mlx/Qwen3.8-27B-mixed-3.5bpw-MLX`](https://huggingface.co/rapid-mlx/Qwen3.8-27B-mixed-3.5bpw-MLX), our own mixed-precision quantization of Qwen3.8-27B at 13 GB of weights against the 4-bit build's 15 GB — plus user-owned model aliases (#1942), the DeepSeek Harness agent profile (#1952), opt-in vision pixel bounds (#1943), and the move of desktop updates fully onto Sparkle (#1947).

## What's in the bump

Per `RELEASE.md` and the desktop dual-tag path, this PR carries all four required pieces:

| file | change |
|---|---|
| `pyproject.toml` | `version` → `0.12.14` |
| `apps/rapid-mac/Resources/Info.plist` | `CFBundleShortVersionString` → `0.12.14`, `CFBundleVersion` → `158` |
| `apps/rapid-mac/CHANGELOG.md` | new `## [0.12.14]` section (pre-checked by the `release` job before it tags anything) |
| `docs/release-notes/` | `git mv unreleased.md v0.12.14.md`, `unreleased.md` recreated empty |

Merging this fires `auto-release.yml`: `detect` → `tier1-agent-gate` on the Studio → desktop tag `rapid-mac-v0.12.14` → engine tag `v0.12.14` + GitHub Release → PyPI.

## Model evidence for the headline

Release eval of the mixed-3.5bpw build through `rapid-mlx serve` on an M3 Ultra: **10/10 coding, 25/30 tool-calling** (including parallel and sequential multi-tool), 7/10 general, 5/10 MATH-subset reasoning with thinking off. Footprint measured across the serve process tree at an 8K prompt: **20 GB peak, 323 tok/s prefill, 40 tok/s decode, zero swap**. The release notes state the maths weakness and the 48 GB-and-up RAM guidance plainly, and the model is deliberately **not** added to `model_recommendations.json` — its 76 composite would break that table's monotonic-by-RAM capability contract.

## Test plan

- [x] `scripts/check_version_sync.py` → "engine and desktop app agree: 0.12.14"
- [x] `tests/release/test_build_release_notes.sh` → 45 passed, 0 failed
- [x] `pytest tests/test_version_sync.py tests/test_model_recommendations.py` → 27 passed
- [x] CHANGELOG gate rehearsed locally: `grep -c '^## \[0\.12\.14\]'` → 1 (the exact check the `release` job runs before tagging)
- [x] `docs/release-notes/v0.12.14.md` present and non-empty (3.4 KB) so the Release body carries curated prose
- [x] Studio self-hosted runner confirmed **online** for the Tier-1 agent gate


raullenchai:
## Codex round 1 — corrections applied

**MAJOR (real, fixed in `955394dd`):** the promoted `v0.12.14.md` carried the four-line scratch-space template comment over from `unreleased.md`, and `build_release_notes.sh`'s strip only matches comments whose `<!--` and `-->` are on the same line — so all four lines would have shipped at the top of the published GitHub Release. Verified by running the real strip against the file:

```
$ sed -E '/^[[:space:]]*<!--.*-->[[:space:]]*$/d' docs/release-notes/v0.12.14.md | head -4
...
```

The comment is now removed, and re-running the strip is a no-op (`diff` identical). `v0.12.13.md` is clean the same way, so this matches convention. The underlying script gap is filed as #1961 rather than patched here — a bump PR triggers the release pipeline and should not also carry a change to the script that builds the release body.

**MINOR (fixed):** dropped "the app's own fit warning will say so" from the notes and "the app will warn you it's a tight fit" from the CHANGELOG. Codex is right that this model is deliberately not a RAM-tier pick, so its fit is decided by the `ModelSizing` heuristic and the exact warning is not something this PR can vouch for. Both now just state the tight fit as guidance.

**MINOR (no change, by design):** the CHANGELOG omits the vision pixel-bounds flags and the CORS preflight fix. `apps/rapid-mac/CHANGELOG.md` is the desktop app's user-facing log; `--vision-min-pixels` / `--vision-max-pixels` are `rapid-mlx serve` CLI flags with no app UI surface, and the CORS fix is server-side. Both are in the engine release notes, which is the right home for them.

## Gates re-verified after the amend

- `tests/release/test_build_release_notes.sh` → 45 passed, 0 failed
- `scripts/check_version_sync.py` → engine and desktop app agree: 0.12.14
- CHANGELOG gate `grep -c '^## \[0\.12\.14\]'` → 1

PR #1966, opened 2026-08-16
fix: strip multi-line release note comments

### LINKED ISSUE #1961: build_release_notes.sh only strips single-line HTML comments, so the notes template ships
## What

`scripts/build_release_notes.sh:135` strips drafting comments with:

```sh
sed -E '/^[[:space:]]*<!--.*-->[[:space:]]*$/d'
```

That deletes a comment only when `<!--` and `-->` land on the **same line**. The scratch template at the top of `docs/release-notes/unreleased.md` is four lines:

```
     version-bump PR, `git mv` this to vX.Y.Z.md and recreate this file empty.
     Whole-line HTML comments like this one are stripped before publishing.
     See README.md in this directory for what good notes look like. -->
```

No single line has both delimiters, so nothing matches and all four lines survive into the published GitHub Release body — the exact outcome the strip step documents itself as preventing ("Drop whole-line HTML comments so drafting guidance left in the template can never ship").

## Repro

```sh
sed -E '/^[[:space:]]*<!--.*-->[[:space:]]*$/d' docs/release-notes/unreleased.md | head -4
# prints the comment unchanged
```

## Why it hasn't bitten before

`git mv unreleased.md vX.Y.Z.md` carries the template over verbatim, so every release depends on whoever cuts it noticing and deleting the comment by hand. v0.12.12 and v0.12.13 are clean because someone did. The v0.12.14 bump (#1960) did **not** at first — caught in review, fixed in that PR — which is what surfaced this.

## Suggested fix

Make the strip range-based so it removes multi-line comments too, e.g. `sed -E '/^[[:space:]]*<!--/,/-->[[:space:]]*$/d'` (guarding the single-line case), and extend `tests/release/test_build_release_notes.sh` — its fixture 2b only covers a single-line `<!-- … -->`, which is why the gap is untested. A test that plants the real four-line template and asserts it does not appear in the output would pin it.

Refs #1960

YUHAO-corn:
## What does this PR do?

Updates the release-notes builder to remove complete multi-line, whole-line HTML comment blocks as well as the existing single-line form.

The filter buffers a candidate block until its first closing marker. If the marker is missing, or if visible content follows it on the same line, the buffered text is restored unchanged so malformed drafting markup cannot silently swallow real release notes.

## Why is this needed?

Fixes #1961.

The scratch template in `docs/release-notes/unreleased.md` spans four lines, while the previous `sed` expression only removed comments whose opening and closing markers were on the same line. Without manual cleanup, that drafting guidance could enter a published GitHub Release body.

## AI assistance disclosure

Codex assisted with the implementation and tests in `scripts/build_release_notes.sh` and `tests/release/test_build_release_notes.sh`. I reviewed the state machine line by line against the whole-line requirement, used a red test to confirm the original defect, and had an independent review catch an inline-suffix case that could lose visible notes. I fixed that finding and re-ran the full offline release-notes test script.

## Test plan

- `bash tests/release/test_build_release_notes.sh` — 52 passed, 0 failed
- `bash -n scripts/build_release_notes.sh tests/release/test_build_release_notes.sh`
- `git diff --check`
- Verified the regression test fails on the original single-line-only filter
- Independently reviewed the final diff; no remaining findings
- Repository validator full-unit pass: 17,543 tests. Its remaining 28 failures were isolated to missing/outdated local test dependencies (`tomli-w`, `prometheus_client`, AnyIO, and Rich); after installing the versions allowed by `pyproject.toml`, all 80 tests in the affected files passed.

## Checklist

- [x] Tests pass locally (`tests/release/test_build_release_notes.sh`)
- [x] Lint/syntax checks pass for the changed shell files (`bash -n`); `shellcheck` is not installed locally
- [x] Repository self-validator was run and its environment-only failures were isolated and rechecked as documented above
- [x] Regression coverage was spot-checked against the broken production filter
- [x] No README/docs update is required

### DIFF
--- a/scripts/build_release_notes.sh
+++ b/scripts/build_release_notes.sh
@@ -130,9 +130,38 @@ HIGHLIGHTS_PATH="$HIGHLIGHTS_DIR/$TAG.md"
 if git cat-file -e "$RELEASE_SHA:$HIGHLIGHTS_PATH" 2>/dev/null; then
   HIGHLIGHTS=$(
     git show "$RELEASE_SHA:$HIGHLIGHTS_PATH" |
-      # Drop whole-line HTML comments so drafting guidance left in the template
-      # can never ship, then trim leading blank lines ($( ) trims trailing ones).
-      sed -E '/^[[:space:]]*<!--.*-->[[:space:]]*$/d' |
+      # Drop complete whole-line HTML comment blocks so drafting guidance left
+      # in the template can never ship. Buffer until the closing marker so an
+      # unmatched opener fails safe instead of swallowing the remaining notes.
+      awk '
+        function flush_pending(    i) {
+          for (i = 1; i <= pending_count; i++) print pending[i]
+          pending_count = 0
+        }
+        in_comment {
+          pending[++pending_count] = $0
+          if ($0 ~ /-->/) {
+            in_comment = 0
+            if ($0 ~ /-->[[:space:]]*$/) pending_count = 0
+            else flush_pending()
+          }
+          next
+        }
+        /^[[:space:]]*<!--/ {
+          if ($0 ~ /-->/) {
+            if ($0 !~ /-->[[:space:]]*$/) print
+            next
+          }
+          in_comment = 1
+          pending[++pending_count] = $0
+          next
+        }
+        { print }
+        END {
+          if (in_comment) flush_pending()
+        }
+      ' |
+      # Trim leading blank lines ($( ) trims trailing ones).
       sed -e '/./,$!d'
   )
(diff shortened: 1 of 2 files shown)


raullenchai:
Thank you @YUHAO-corn — landed via #1990 (squash-merged to main as 82418c55, your commit cherry-picked with authorship preserved and co-authored on the squash). Your awk state machine was the core of the fix; a multi-round adversarial review then hardened a few edge cases on top (first-`-->` visible suffix, chained comments, code-fence delimiter matching, indented code blocks). Fixes #1961. 🙏

PR #1967, opened 2026-08-16
feat(mtp): enable lossless adaptive K3 for Qwen3.8

raullenchai:
## Why
Qwen3.8 uses hybrid GatedDeltaNet caches, so treating it like dense Qwen3.5 makes multi-token MTP rollback unsafe. K=1 is lossless but leaves useful decode throughput on the table; this change makes adaptive K=3 correct because every speculative position can restore the exact recurrent state after partial acceptance.

## What
- classify direct/local Qwen3.8 checkpoints as hybrid instead of generic dense Qwen3.5
- add per-position GatedDeltaNet snapshots so partial K=3 acceptance rolls SSM state back exactly
- chain Qwen MTP hidden state across draft positions
- use a quantized Metal argmax for greedy MTP heads without materializing the 248k-vocab logits row
- expose fixed K in the MTP benchmark; production remains adaptive

## Verification
- 543 focused tests passed (MTP, CLI wiring, lossless, hybrid classification, model auto-config)
- 62 Qwen/HY3 and Gemma MTP injection tests passed
- fused 4-bit argmax exactly matches materialized QuantizedLinear logits
- Mini M2 Pro, Qwen3.8-27B-4bit + official MTP sidecar: 128-token fixed-K3 output exactly matched baseline token-for-token
- Mini adaptive K<=3: pooled 9.57 tok/s vs interleaved baseline 8.67 tok/s across 3 prompt classes (+10.4%); fixed K3 is intentionally not the default (0.937x pooled)
- formatter-only CI follow-up: 501 focused tests passed

### DIFF
--- a/bench/bench_spec_decode_mtp.py
+++ b/bench/bench_spec_decode_mtp.py
@@ -204,6 +204,17 @@ def _parse_args() -> argparse.Namespace:
             "previously-captured baseline number."
         ),
     )
+    parser.add_argument(
+        "--mtp-max-k",
+        type=int,
+        default=3,
+        help="Maximum chained MTP drafts per verify round (default: 3).",
+    )
+    parser.add_argument(
+        "--mtp-disable-auto-k",
+        action="store_true",
+        help="Benchmark fixed --mtp-max-k instead of the adaptive controller.",
+    )
     return parser.parse_args()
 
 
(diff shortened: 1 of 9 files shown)


raullenchai:
Pausing this PR before merge while we independently study the newly published MTPLX Qwen3.8 stack. The correctness work here is promising (real K=3 token-for-token parity on Mini), but MTPLX reports materially larger gains from a matched MTP artifact plus specialized verify kernels. We will first run same-hardware, same-workload ablations and then decide whether to retain, split, or replace this implementation. Please do not merge meanwhile.

PR #1973, opened 2026-08-17
feat(mac): Settings → Developer, with re-onboarding

xiaoxiunique:
## Why

Rehearsing first run meant `defaults delete` against a key list nobody keeps
current, or `scripts/dogfood-isolate.sh` and a fake `$HOME`. Both work; neither
is reachable from inside the app you are already looking at.

## What

Settings gains a **Developer** category. The whole thing is inside `#if DEBUG`
— the panel, the `Category` case, the copy — so it is absent from a release
binary rather than present and hidden. Seeing it needs
`RAPID_BUILD_CONFIG=debug ./scripts/build.sh`; `build.sh` defaults to release.

What it erases is a choice, not a constant:

| | erased | default |
|---|---|---|
| — | Quickstart flags + `rapid.serve.lastAlias` | always |
| ☐ | all settings (whole defaults domain) | off |
| ☐ | `conversations.json` | off |
| ☐ | the telemetry decision | off |

The confirmation names each selected loss in descending order of regret, and
the destructive button changes verb when conversations are in scope — "Erase
and restart" is a different promise from "Restart into onboarding". The
telemetry line says out loud that the decision is shared with the `rapid-mlx`
CLI, because a user who wipes it from the desktop will otherwise be asked again
in a terminal with no idea why.

`QuickstartCoordinator`'s existing reset is reused rather than duplicated. It
is renamed `resetForReonboarding()` with `_testingReset()` kept as an alias —
44 call sites already use that name. Its doc comment claimed no production UI
would ever call it; that is no longer true and it now says so.

## Three things the plan got wrong

All found by running it, not by reading it.

**The Settings scene never injected `QuickstartCoordinator`.** SwiftUI traps
rather than warns on a missing `@Environment` observable, so the first click on
the category killed the app — `EnvironmentValues.subscript.getter` →
`assertionFailure`, with nothing in the backtrace naming the type.

**`ServerManager.stop()` clears `rapid.serve.lastAlias` only along the path
that terminates a child.** `guard child != nil else { return }` means an idle
app's stop is a no-op, and the wizard gates on that alias independently of its
own flags — so re-onboarding restarted straight past Quickstart. Hence
`forgetLastServedAlias()`.

**`NSApp.terminate` does not quit this app.**
`MainWindowCloseInterceptor.windowShouldClose(_:)` runs the hide-the-Dock-icon
prompt on any window close and returns `false` when the answer is "hide";
AppKit routes terminate through exactly that path, so one `false` cancels the
whole sequence. Measured: `osascript … to quit` returns `-128 userCancelled`,
and ⌘Q behaves the same on a default install (`hideDockChoice = askEveryTime`).
**This is a pre-existing defect, not introduced here** — the interceptor does
not distinguish a user closing a window from AppKit closing it on the way out.
It deserves its own fix; this button could not wait on one, so `relaunch()`
records the clean-shutdown marker itself and `exit`s.

Relaunch is a detached shell that waits for this PID to disappear and only then
opens the app. `NSWorkspace.openApplication` with
`createsNewApplicationInstance` was tried first and silently produced no second
instance on a local ad-hoc signed build.

## Tests

`SettingsEnvironmentInjectionTests` is the guard for the first bug above: every
observable a Settings panel reads must be injected by the Settings scene.
Break-verified — it reproduces the exact crash and names the type:

> `SettingsDeveloperPanel.swift reads QuickstartCoordinator from the
> environment, and the Settings scene in RapidApp.swift never injects it.
> SwiftUI traps — not warns — the first time that category is opened.`

The existing `everyCategoryKeepsItsStateOwner` pins that a panel still
*declares* what it needs. Declaring and providing are different mistakes, and
only the second one crashes.

`ReonboardingResetTests` (10) covers the copy generation and the scope→erasure
mapping against temporary paths and a scratch defaults suite, plus that
relaunch hands the pid and path to the waiter positionally (`$1`/`$2`) rather
than interpolated — every install path contains a space. Break-verified by
making the conversations scope also delete the telemetry file, and by
flattening the confirm-button copy; both fail as intended.

The two frozen category literals in `SettingsVisualFoundationTests` are updated
and made `#if DEBUG`-conditional, so `swift test -c release` stays honest too.

Full suite: 2281 tests. Five suites fail; four fail identically on unmodified
`main` on this machine (`Web tools hardening`, `Download/catalog hardening`,
`Markdown link hit-testing cost`, `Throughput caption integration` — the third
is an upstream test added after #1933 and fails on a clean checkout), and
`DownloadManager` is timing-flaky: 36/36 pass in isolation.

## Verified end to end

On a debug build, driven through the accessibility tree:

- Developer appears in the rail (9 rows, was 8); a release build has 8
- Cancel leaves `conversations.json` byte-identical and does not restart
- A default-scope run restarts (pid 16035 → 16164) into Quickstart **step 1 of
  4**, with the conversations still there and both gate keys cleared

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
+++ b/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
@@ -610,6 +610,10 @@ enum DevSnapshot {
                     .environment(appearance)
                     .environment(settingsRouter)
                     .environment(server)
+                    // The capture loop walks Category.allCases, which in a
+                    // debug build includes Developer — and that panel reads
+                    // the coordinator.
+                    .environment(quickstart)
                     .environment(downloads)
                     .environment(updater)
                     .environment(snapshotSparkleUpdater)
--- a/apps/rapid-mac/Sources/Rapid/RapidApp.swift
+++ b/apps/rapid-mac/Sources/Rapid/RapidApp.swift
@@ -414,6 +414,11 @@ struct RapidApp: App {
                 .environment(appearance)
                 .environment(settingsRouter)
                 .environment(server)
+                // Settings → Developer resets the wizard, and SwiftUI traps
+                // rather than warns when an @Environment observable is
+                // missing — so the Settings scene needs this even though the
+                // panel that reads it only exists in a debug build.
+                .environment(quickstart)
                 .environment(downloads)
                 .environment(updater)
                 .environment(sparkleUpdater)
--- a/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
@@ -817,6 +817,19 @@ final class ServerManager {
         return trimmed.isEmpty ? nil : trimmed
     }
 
+    /// Drop the last-served alias without going through a stop.
+    ///
+    /// ``stop()`` clears it, but only along the path that actually terminates
+    /// a child — `guard child != nil else { return }` means an idle app's
+    /// Stop is a no-op and the alias survives. That is right for Stop and
+    /// wrong for re-onboarding, which has to reach a state the wizard
+    /// considers fresh whether or not a model happened to be running
+    /// (``QuickstartCoordinator.isEligible`` gates on this alias
+    /// independently of its own flags). See ``ReonboardingReset``.
+    nonisolated static func forgetLastServedAlias(defaults: UserDefaults = .standard) {
+        defaults.removeObject(forKey: lastServedAliasKey)
+    }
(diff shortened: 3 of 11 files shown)


raullenchai:
Thank you @xiaoxiunique — landed via #1993 (merged to main as 70962750), your commit cherry-picked with authorship preserved and co-authored on the squash. A follow-up review added the clean-shutdown-before-erase ordering, injectable Quickstart defaults, the legacy onboarding-flag clear, and completed the AX golden baselines for the new Developer sidebar row across all Settings-dumping snapshots. 🙏

PR #1974, opened 2026-08-17
fix(mac): the status footer squeezed six chips instead of shedding any

xiaoxiunique:
## The bug

In a narrow window every footer readout compressed until it stopped being one
— `CPU…`, `GPU 9…`, `13.2 G…` — and the version pill wrapped into a five-line
block sitting under the row.

Those are one bug, not two. `DesktopVersionPill`'s label had no line limit, and
**a view that wraps reports that it fits at any width**. The pill absorbed the
shortfall on everyone else's behalf, so the row never had to admit it had run
out of room. Pinning the pill to one line is both the visible fix and the
precondition for any layout logic below it.

## The fix

The readouts go through `ViewThatFits`, widest arrangement first — already the
pattern in `FailureDiagnosisView` and `ImagesView`. What gets given up is an
*ordering*, not a set of width breakpoints, so there is nothing to re-tune when
a chip's text changes length.

What may be dropped follows the rule `EnvironmentValues.settingsContentIsCompact`
already states for Settings:

> use this to DROP something optional …, never to hide a control — an
> unreachable control at a supported window size is the defect this exists to
> prevent

So the split is **readouts shed, controls never do**. The gear, the log toggle
and the version pill are buttons — the pill routes to Settings → App — and all
three sit outside the group. Inside it, the order of sacrifice is:

1. **CPU, GPU, memory** — these describe the Mac, not this app's work
2. **throughput**
3. **server state** — last, because the other numbers mean nothing without
   knowing whether a model is running (and even it is redundant in the
   narrowest case, where `ReadinessBanner` is saying the same thing in a full
   sentence directly above)

## Measured

Rendered snapshots at three widths:

| width | footer |
|---|---|
| 900pt | all six, text intact — `CPU 0%`, `GPU 78%`, `13.2 GB / 16 GB` |
| 640pt | all six, text intact |
| 380pt | server state + throughput + version pill — no ellipsis, no wrap |

`DevSnapshot` gains the 380pt render, alongside the 420pt Images one that
exists for exactly this reason ("only this width exercises the ViewThatFits
layout").

## Tests

The guard pins the property that actually matters: **no control inside the
droppable group**. A control that vanishes at a supported window size is
unreachable, which is the defect the Settings rule was written to prevent. It
also pins that the system probes appear in exactly one arrangement, that server
state outlives them, and that the version pill keeps its line limit.

ViewInspector is not in this target (#1492), so this is a source guard in the
same shape as `AccessibilityIdentifierInventoryTests`.

Break-verified twice — moving the version pill into the group, and removing the
line limit; each trips its own case and neither trips the other's.

Full suite: 2284 tests. Five suites fail; four fail identically on unmodified
`main` on this machine (`Web tools hardening`, `Download/catalog hardening`,
`Markdown link hit-testing cost`, `Throughput caption integration`), and the
`#253 stagger` suite is timing-flaky — it passes in isolation.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
+++ b/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
@@ -199,6 +199,10 @@ enum DevSnapshot {
         // whether HF_HUB_CACHE points at a populated cache).
         render(contentView(width: 900, height: 640), to: "\(dir)/content-idle.png")
         render(contentView(width: 640, height: 560), to: "\(dir)/content-min.png")
+        // Narrow window: the status footer sheds its readouts through
+        // ViewThatFits rather than squeezing them to ellipses, and the version
+        // pill must stay on one line. Only a width this small exercises it.
+        render(contentView(width: 380, height: 560), to: "\(dir)/content-narrow.png")
         // Images tab (empty state — no results, catalog not yet resolved).
         render(imagesView(width: 700, height: 640), to: "\(dir)/images-empty.png")
         // Narrow composer: the canvas controls wrap to the two-row
--- a/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
@@ -890,6 +890,43 @@ struct ContentView: View {
 
     // MARK: - Status footer
 
+    /// The footer's right-hand readouts, widest arrangement first.
+    ///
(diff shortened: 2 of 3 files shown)


raullenchai:
Landed in main via #1991 (cherry-picked in-repo so CI could run; your authorship is preserved via the co-author trailer). Thanks @xiaoxiunique — clean ViewThatFits approach. Two codex rounds + manual review found the code correct.

PR #1975, opened 2026-08-17
feat(mac): analyze large and scanned PDFs

osdodo:
## What does this PR do?

  Makes large documents actually analyzable in the Mac app, instead of silently truncating them at 24,000 characters.

  The full extract now lives in a document cache and only a preview goes into the prompt, with an envelope telling the model how much it is not seeing. A new `read_document` tool reaches the rest three ways:

  - `mode="outline"` — the section map in one call, for whole-document questions
  - `grep` — jump to matching passages when the term is known
  - `offset` — sequential paging with a `next_offset` cursor

  Scanned PDFs, previously rejected outright, are recognized on-device with Vision.

  Four measurements shaped the implementation, all taken on real files (a 302-page Chinese technical book and a 529-page scanned textbook):

  | | before | after |
  |---|---|---|
  | attach a 302-page PDF | 1.97 s | **0.13 s** |
  | preview cost, Chinese book | 13,306 tok | **7,130 tok** |
  | whole-document overview | 33 sequential reads | **1 outline call** (289 entries → 97 rows / 1,983 tok) |
  | attach a 529-page scan | rejected | **3.7 s to a usable preview** |

  ## Why is this needed?

  Attaching a 300-page PDF and asking "what does this say?" silently answered from the first few pages — the remaining 95% of the document was extracted, thrown away, and never mentioned to the user or the model. A scanned PDF could not be attached at all.


raullenchai (changes requested):
There are multiple blocking resource/privacy issues in the current design:

1. **Unbounded regex result materialization (`ReadDocumentTool.grepResult`)**. `NSRegularExpression.matches` scans the entire extract and allocates every match before the code applies `prefix(maxGrepMatches)`. Extracts may contain 20,000,000 characters; a model-supplied pattern such as `.` can therefore allocate millions of `NSTextCheckingResult`s even though only ten passages are returned. This is a locally triggerable CPU/memory DoS. Enumerate matches with a hard early stop and impose a defensible pattern/work bound; add a maximum-size adversarial regression test.

2. **The multi-minute OCR tasks cannot actually be cancelled (`ChatFileAttachment`)**. Both background passes use unstructured `Task.detached` and discard the task handle. `PDFTextRecognizer` checks `Task.isCancelled`, but no attachment-removal, conversation-switch/delete, or lifecycle path can ever cancel those detached tasks. Removing a 529-page scan can therefore leave Vision/PDFKit consuming resources for minutes. Introduce owned task lifecycle/cancellation and test it.

3. **Deleting the user-visible data does not delete the full plaintext extract (`DocumentContentCache`)**. The new persistent tier has no `remove` API and no TTL; attachment removal and conversation deletion never remove `<uuid>.json`. A sensitive document can remain in Application Support indefinitely (until unrelated LRU pressure happens to evict it), whereas the prior inline representation followed conversation deletion. Wire deletion/reset paths to cache deletion and cover the privacy lifecycle with tests.

These need to be resolved before this can land. The feature direction is valuable, but the current implementation creates model-triggerable resource exhaustion and a user-data retention regression.

raullenchai (changes requested):
The follow-up fixes address the three original findings substantially, but the plaintext deletion guarantee still has a blocking TOCTOU race.

Both extraction tasks do `guard !Task.isCancelled else { return }` and then call `cache.put(...)`. `remove` cancels the handle and deletes memory/disk without waiting for the task to finish. The following interleaving can therefore resurrect a deleted sensitive document:

1. extraction evaluates `Task.isCancelled == false`;
2. user removal calls `cancelExtraction`, clears memory, and deletes `<uuid>.json`;
3. extraction resumes and calls `put`, restoring both the memory entry and plaintext file.

Checking cancellation immediately before `put` narrows the window but does not close it. Please make deletion authoritative inside the cache—for example with a tombstone/generation checked atomically by a conditional publish operation—or await task completion before the final delete off the main actor. Add a deterministic race test that pauses extraction after its cancellation check, removes the document, then releases publication and proves neither tier is recreated.

Separately, the new 90-day TTL is a user-visible retention/reopen policy rather than merely an implementation bound; please call it out explicitly in the PR description/release UX so product can confirm that an old conversation intentionally loses full-document access and asks for reattachment.

raullenchai (changes requested):
Blocking on the current head `ea0c535d`: it does not compile under the repository's Swift 6 strict-concurrency settings after a clean merge onto current `main`.

Reproduction:

```
swift test --package-path apps/rapid-mac --filter "DocumentCacheLifecycleTests|ReadDocumentToolTests|ChatFileAttachmentTests|TokenEstimateTests|ToolLoopBudgetIntegrationTests|ContextWindowTrimTests|BuiltinToolsTests"
```

The compiler rejects `ChatFileAttachment.swift:454` because the `Task.detached` sending closure captures `handoff`, a `nonisolated(unsafe)` alias of non-Sendable `PDFDocument`, while the original object remains accessible to the current task:

```
passing closure as a 'sending' parameter risks causing data races
closure captures 'handoff' which is accessible to code in the current task
```

`nonisolated(unsafe)` is not a safe ownership transfer and must not be used to silence this boundary. Restructure so the detached task receives genuinely Sendable ownership/data (for example, recreate the PDFDocument inside the worker from captured `Data`), or use an isolation design the compiler can prove safe. If relying on region-based transfer of the original object, remove the alias and demonstrate that no reference remains accessible after transfer. Add/retain a clean-build gate under the project compiler before requesting re-review.

The new cache generation design otherwise closes the previously reported plaintext resurrection race: the memory publication and disk write are generation-checked, removal invalidates before cancellation, and the observed lock ordering does not introduce a reverse-order deadlock. I could not execute those tests because compilation fails first.

The separate 90-day plaintext retention/reopen policy still requires explicit product confirmation; it is now documented consistently in code/UI, but approval of that policy belongs to product.

osdodo:
Thanks for the detailed follow-up. Both remaining points are addressed on the current head (`c61b7c0f`).

  For Swift 6 isolation, the `nonisolated(unsafe)` handoff has been removed. The  selectable-text path now transfers the original `PDFDocument` directly into `Task.detached` and never accesses it afterward, which Swift 6 region isolation accepts. The scanned-PDF path captures only `Data` and reconstructs its `PDFDocument` inside the worker.

  I reran your exact reproduction command after a clean SwiftPM build:

  swift package --package-path apps/rapid-mac clean
  swift test --package-path apps/rapid-mac --filter 'DocumentCacheLifecycleTests|
  ReadDocumentToolTests|ChatFileAttachmentTests|TokenEstimateTests|
  ToolLoopBudgetIntegrationTests|ContextWindowTrimTests|BuiltinToolsTests'

  Result: 142 tests across 7 suites passed with 0 failures.

  On retention, I agree that documentation alone is not product approval. The current proposal remains a 90-day plaintext-cache TTL, surfaced in Settings and in the expiration/reattachment messaging. Could you confirm whether that policy is acceptable, or specify the preferred duration/behavior?

  The earlier regex, cancellation ownership, deletion, and TOCTOU findings remain covered by adversarial and lifecycle regression tests. Please re-review the current head.

raullenchai (changes requested):
I re-reviewed exact head `4dd8e5ca50a0f7140524183589737f6cd992801a` (which contains `c61b7c0f`) and confirmed it merges cleanly with the requested current-main checkpoint. The selectable-text handoff now compiles under Swift 6 without the unsafe alias. A clean build plus the seven focused suites passed locally: 144 tests / 7 suites. Proportional native/AX coverage also passed 75 tests / 4 suites, and a release Swift build completed successfully. The tool-call ID stamping, bounded result envelopes, deletion generations, cancellation, branch/draft cleanup, and cache size caps otherwise look coherent.

Two in-scope safety boundaries still need changes:

1. `PDFTextRecognizer.render` converts the PDF-controlled media-box width/height to `Int` and multiplies them before validating that they are finite and representable. An extreme media box therefore traps the process instead of being rejected. I reproduced this deterministically with a `PDFPage` whose media box uses `CGFloat.greatestFiniteMagnitude`: calling the recognition path exits the test process with signal 5 before the `64_000_000` pixel guard can run. Please validate finite, positive, representable dimensions and use overflow-safe pixel-area arithmetic before conversion/allocation, with a focused extreme-media-box regression.

2. The stated 90-day plaintext TTL is enforced only during cache initialization and later writes, not when an existing cache reads an entry. I reproduced this with one initialized cache: persist an entry, evict it from memory, age its file past the configured TTL, then call `get`; the expired plaintext is loaded successfully and its modification date is refreshed. A long-running app can therefore retain and serve the file beyond the user-visible policy, potentially extending it for another full retention window. Please enforce expiry on the read/hot-entry path as well as sweeps, delete expired disk data when observed, and add a deterministic “expires after initialization” contract. This does not request changing the selected retention duration.

Both diagnostics were temporary and removed; the review worktree is clean. No broader PDF redesign, cache refactor, or UI expansion is requested. After these two bounded fixes, please rerun the existing focused suites and the new edge contracts on the updated exact head.

osdodo:
Thanks for the re-review. Both remaining issues are fixed on exact head 46351639463ddc088dfd656c86958414843beb8a.

  1. PDFTextRecognizer.render now validates scaled media-box dimensions as finite, positive, and representable before converting them to Int. Pixel area is checked with overflow-reporting arithmetic before allocation. A focused regression using CGFloat.greatestFiniteMagnitude confirms the recognition path rejects the page without trapping.

  2. DocumentContentCache now enforces TTL on both hot-memory and disk-read paths. Expired entries are removed from memory and disk before they can be returned or have their modification date refreshed. Deterministic tests cover both a hot entry expiring during a long-running session and a disk entry expiring after cache initialization.

  Verification:

  147 tests across the 7 requested suites passed with 0 failures.
  Release Swift build completed successfully.

  I also smoke-tested the production import path with the two real large-document samples:

  302-page selectable-text PDF: 0.229 s preview import
  529-page scanned PDF:         3.799 s OCR preview import

raullenchai (commented):
**Both outstanding reviewer findings verified resolved on the reviewed exact head `be5199385`** (the current head — merged origin/main), with independent local verification (not just the author's claim):

**1. Extreme media-box crash (process trap)** — RESOLVED.
`PDFTextRecognizer.render` (`Sources/Rapid/Chat/PDFTextRecognizer.swift`) now validates dimensions **before** `CGFloat→Int` conversion and allocation: `scaledWidth/Height.isFinite`, `>= 1`, `<= maxPixelCount`, then overflow-safe `width.multipliedReportingOverflow(by: height)` with a `!overflow && pixelCount <= maxPixelCount` guard. Regression test present: `ChatFileAttachmentTests` constructs a `PDFPage` with `width/height = .greatestFiniteMagnitude` and asserts `PDFTextRecognizer.recognize(page:).isEmpty` (no signal-5 trap). ✅ Verified passing.

**2. 90-day TTL not enforced on read/hot path** — RESOLVED.
`DocumentContentCache.get(\_:)` (`Sources/Rapid/Chat/DocumentContentCache.swift`) now enforces expiry on both tiers at read time and cleans up observed expired data:
- Memory tier: on a hit, `lastAccessedAt[k] >= accessDate - diskTTL` is checked; an entry last touched before expiry is `remove(id)`d and nil is returned.
- Disk tier: `loadFromDisk` checks `contentModificationDate >= accessDate - diskTTL` **before** loading plaintext and before refreshing mtime — an expired file is deleted, never resurrected, and never given a refreshed modification date.
- The `removalGeneration` recheck after disk I/O (`get` line 322) closes the removal/publication TOCTOU.
- Regression tests present: `DocumentCacheLifecycleTests` `'A disk entry that expires after initialization is deleted on read'` (insert→backdate mtime→`get`==nil AND file deleted) + `'A hot entry expires during a long-running cache session'`. ✅ Verified passing.

**Independent build + suite results (ran locally on this exact head):**
- ✅ Clean `swift build` (Swift 6 strict-concurrency; was the original blocker on the old head)
- ✅ `DocumentCacheLifecycleTests|ChatFileAttachmentTests|ReadDocumentToolTests` distilled + tested: **97 tests / 3 suites pass**
- ✅ `TokenEstimateTests|ToolLoopBudgetIntegrationTests|ContextWindowTrimTests|BuiltinToolsTests`: **53 tests / 4 suites pass**
- PR is `MERGEABLE` on this head.

Both requested bounded fixes are complete with focused edge-contract regressions, as the review requested. Marking advanced for the maintainer's morning review batch.

Issue #1258, opened 2026-07-27
perf(mtp): mixed 8-bit base + 4-bit Qwen3.6 sidecar is slower than baseline


raullenchai:
## Summary

Qwen3.6 MTP is reliable after the recent fixes, but using the only published 4-bit MTP sidecar with an 8-bit base model regressed throughput in local A/B dogfood. The same-precision 4-bit pairing produced a useful speedup.

This is primarily a recommendation/UX issue, not a correctness regression.

## Environment

- Local wheel built from `main` at `93981cf4` (package version `0.11.0`)
- Apple M3 Ultra 256 GB
- Qwen3.6-27B base models
- Matching Qwen3.6 4-bit MTP sidecar
- Same prompts/settings for baseline and MTP runs

## Results

| Pairing | Baseline | MTP | Delta | Acceptance | Parks |
|---|---:|---:|---:|---:|---:|
| 4-bit base + 4-bit sidecar | 35.9 tok/s | 41.0 tok/s | +14% | 66.1% | 9 / 410-token prose run |
| 8-bit base + 4-bit sidecar | 18.9 tok/s | 17.4 tok/s | -8% | 57.1% | 364 / 408-token prose run |

Short tasks remained reliable with no empty responses, and an earlier five-prompt run showed a similar aggregate result (with acceptance around 81%).

There is also a status inconsistency: explicit `--speculative-config` enables MTP, but startup/profile output can still say `spec decode OFF`.

## Expected behavior / product decision

Rapid-MLX should make the practical pairing guidance clear and avoid implying that every supported MTP/base precision combination is a speed win. Possible outcomes include:

- document and recommend matched precision,
- warn when base and sidecar precision differ,
- adjust automatic enablement/recommendations using observed acceptance/park behavior,
- make the runtime profile accurately show that explicit MTP is active.

The goal is predictable user-facing performance without changing the now-correct generation behavior.

raullenchai:
The misleading startup-status slice is fixed by #1270 and is now merged. An explicitly selected MTP runtime now shows `spec decode MTP (explicit)` instead of repeating the static profile's `spec decode OFF`; this is display-only and does not alter eligibility or inference behavior.

Keeping this issue open for the actual mixed-precision performance guidance/recommendation work.

rinaldofesta:
[...]

PR #1983, opened 2026-08-17
feat(mtp): warn when sidecar and base quantization differ (#1258)

rinaldofesta:
Implements two of the outcomes discussed in #1258: the mismatch warning and the documentation note. The third outcome (adjusting automatic enablement) is a product decision I have deliberately left out.

### What is the new behavior?

- After a successful sidecar load, `inject_mtp_support` emits one warning when the detected sidecar quantization (bits, group_size) differs from the base model's, citing the measured mixed-pairing regression from #1258. Exact message:
  `[mtp.inject] MTP sidecar quantization (4-bit, group_size=64) differs from base model (8-bit, group_size=64): pairing effects are model-dependent: slower than no speculation on Qwen3.6 (#1258), faster on Qwen3.8-27B. Benchmark your pairing.`

  (Amended after my Qwen3.8-27B measurements in #1216 inverted the #1258 finding: mixed 8/4 gained 35-65% there while matched 4/4 was break-even. The docs note now states the model dependence with both datapoints.)
- Silent when either side's quantization metadata is unavailable, and when the pairing matches — no false alarms.
- `docs/reference/configuration.md` gains a short matched-precision recommendation citing the #1258 numbers (matched 4/4: +14%, mixed 8/4: -8%).
- One unit test: warning fires on mismatch, stays silent on matched pairing.

### Placement note

The check lives in `vllm_mlx/spec_decode/mtp/qwen3_5_inject.py`, right after the sidecar-load log line, because that is where both quantization dicts are in hand. I know `spec_decode/` is the core-team lane per #1216 — this is a 22-line additive warning, but happy to relocate the check to config level if you prefer.

### Validation

- Both changed modules compile; the injector imports and a dependency-free behavioral check produced the exact warning once on mismatch and nothing on match.
- `pytest`/`ruff` not run: fresh clone without the dev environment; no dependencies were installed. The added test follows the existing pattern in `tests/test_mtp_spec_decode.py`.
- Related cross-engine datapoint (mixed-precision draft pairing degrading under load on another engine) is in my comment on #1258.

*Implemented with the help of a coding agent (Claude Code) and reviewed locally by the author.*

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -207,6 +207,11 @@ Pair each head with a base of the same size class: `Qwen3.6-27B-MTP-4bit`
 with a `qwen3.6-27b-*` base, and `Qwen3.6-35B-A3B-MTP-4bit` with a
 `qwen3.6-35b-*` base.
 
+Match the sidecar and base precision for performance. In
+[#1258](https://github.com/raullenchai/Rapid-MLX/issues/1258), a matched 4-bit
+base / 4-bit sidecar improved throughput by 14%, while a mixed 8-bit base /
+4-bit sidecar reduced throughput by 8% versus no speculation.
+
 ### MCP Options
 
 | Option | Description | Default |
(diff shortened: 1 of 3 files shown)


raullenchai:
Landed in e3ba3918c (via #1989, cherry-picked in-repo so CI could run; your authorship is preserved via the co-author trailer). Thank you @rinaldofesta — especially for amending the warning with your own Qwen3.8-27B measurements from #1216 when they inverted the #1258 finding. Closing this in favor of the merged commit.

PR #1990, opened 2026-08-17
fix(release): strip multi-line release-note comments (supersedes #1966)

### LINKED ISSUE #1961: build_release_notes.sh only strips single-line HTML comments, so the notes template ships

raullenchai:
## Why

`scripts/build_release_notes.sh` stripped drafting comments with a single-line
`sed` (`/^\s*<!--.*-->\s*$/d`), so the four-line scratch template at the top of
`docs/release-notes/unreleased.md` — whose `<!--` and `-->` sit on different
lines — survived verbatim into the published GitHub Release body. Fixes #1961.
Every clean release so far depended on whoever cut it deleting the comment by
hand; the v0.12.14 bump (#1960) initially did not.

This builds on @YUHAO-corn's #1966 (cherry-picked with authorship preserved),
then hardens the stripper after a multi-round adversarial review.

## What

Replaces the `sed` with an awk state machine that removes **whole-line HTML
comment blocks** (single- or multi-line), and:

- **Closes a comment at its FIRST `-->`.** Anything after that first marker is
  visible content — `rest_is_clean()` decides whether the tail ships nothing
  (whitespace and/or further complete whole-line comments → drop the block) or
  carries a real note (keep the line verbatim). The naive "trailing `-->`" test
  dropped `<!-- x --> real note -->` whole.
- **Fails safe on an unmatched opener** — buffered text is restored, never
  swallowed.
- **Skips fenced code blocks** (```` ``` ````/`~~~`), so a release note that
  demonstrates HTML-comment syntax keeps it.
- Keeps **column-0-only** opener detection, so a mid-line `<!--` can't start a
  block and eat the following lines.

## Tests

- `bash tests/release/test_build_release_notes.sh` — **59 passed, 0 failed**
  (added: first-`-->` visible suffix single/multi-line, all-comment line with a
  second comment, visible text between two comments, multi-line comment inside a
  ```` ```html ```` fence).
- Red/green proof: the new multi-line assertions fail against the pre-fix
  script; the real four-line `unreleased.md` template now strips to empty.
- `bash -n` clean on both files.

## Notes — adversarial review (Codex, 4 rounds)

- R1 → found first-`-->` visible-suffix drop → fixed.
- R2 → found all-comment line (`<!-- a --> <!-- b -->`) leaking → fixed with `rest_is_clean()`.
- R3 → found multi-line comment inside a code fence being stripped (real regression) → fixed with fence passthrough.
- R4 → **0 blocking.** Two items explicitly judged non-blocking & declined:
  (a) fence delimiter type/length isn't matched (pathological for release notes);
  (b) the recent-tag sweep's oracle omits `--first-parent` — pre-existing, unrelated to this diff, not a visible-note loss.

Release-tooling only; no product code, no version bump.

Co-authored-by: YUHAO-corn <godcorn001@outlook.com>

---
🤖 AI-assisted: implementation and the 4-round adversarial review were done with Claude Code (Opus 4.8). Human-reviewed.

### DIFF
--- a/scripts/build_release_notes.sh
+++ b/scripts/build_release_notes.sh
@@ -138,18 +138,86 @@ if git cat-file -e "$RELEASE_SHA:$HIGHLIGHTS_PATH" 2>/dev/null; then
           for (i = 1; i <= pending_count; i++) print pending[i]
           pending_count = 0
         }
+        # An HTML comment closes at its FIRST "-->"; anything after that first
+        # marker is visible content. Return that trailing text so the caller can
+        # judge what a clean whole-line close leaves behind. Checking the whole
+        # line for a trailing "-->" instead would treat "<!-- x --> note -->" as
+        # a clean close and silently drop "note".
+        function tail_after_close(line,    p) {
+          p = index(line, "-->")
+          return substr(line, p + 3)
+        }
+        # Would the text trailing a closed comment ship nothing visible? True
+        # when it is only whitespace and/or further complete whole-line comments
+        # (so "<!-- a --> <!-- b -->" is fully strippable). A trailing UNCLOSED
+        # "<!--" is treated as not-clean so the line is kept verbatim rather than
+        # swallowing the notes that follow it. Only ever fed the tail of a line
+        # already in whole-line-comment context, so column-0-only stays intact.
+        function rest_is_clean(s,    o, c) {
+          while (1) {
+            if (s ~ /^[[:space:]]*$/) return 1
+            if (s !~ /^[[:space:]]*<!--/) return 0
+            o = index(s, "<!--"); s = substr(s, o + 4)
+            c = index(s, "-->"); if (c == 0) return 0
+            s = substr(s, c + 3)
+          }
+        }
+        # A line indented four or more spaces (or led by a tab) is a CommonMark
+        # indented code block, not prose: fences and drafting comments there are
+        # example text and must survive untouched. Kept deliberately simple —
+        # the interval form /^ {0,3}/ is avoided for portability with older awks.
+        function indented_code(line) {
+          return (line ~ /^    / || line ~ /^\t/)
+        }
+        # The leading code-fence run (0-3 spaces of indent), e.g. "```" or
+        # "~~~~", or "" when the line does not open/close a fence. A CommonMark
+        # fence is three or more backticks or tildes at 0-3 spaces of indent.
+        function fence_marker(line,    s, ch, n) {
+          if (indented_code(line)) return ""
+          s = line
+          sub(/^ */, "", s)
+          if (s ~ /^```/) ch = "`"
+          else if (s ~ /^~~~/) ch = "~"
+          else return ""
+          n = 0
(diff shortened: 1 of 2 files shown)


raullenchai:
### Adversarial review trail (Codex, 4 rounds + 3 pr_validate passes)

Real defects found and fixed, each with regression coverage:
1. A comment closing at its **first `-->`** with a visible suffix (`<!-- x --> real note -->`, single- and multi-line) was dropped whole → fixed via `rest_is_clean()`.
2. An **all-comment line with a second comment** (`<!-- a --> <!-- b -->`) leaked → fixed.
3. A multi-line HTML comment **inside a code fence** was stripped → fixed with fence passthrough.
4. **Fence delimiter type/length** not matched → fixed with CommonMark same-char, ≥-length, bare-closer matching.
5. **Four-space indented code blocks** treated as prose → fixed with the 0–3-space indent rule.

**Declined (documented) — pedantic, non-blocking:**
- `indented_code()` doesn't expand a mixed space+tab (`" \t<!--"`) to CommonMark's 4-column tab stop. This requires a space-then-tab-indented HTML comment inside an indented code block in a *release note* — not a realistic input, and the content is example code that renders as-is; it does not drop a real visible note. Implementing full CommonMark tab-stop expansion in the notes stripper is out of proportion to the risk. Deliberately left as-is.

Tests: 63 passed, 0 failed. `full_unit`: 17,664 passed. Merging.

PR #1993, opened 2026-08-17
feat(mac): Settings → Developer, with re-onboarding (#1973)

raullenchai:
## Why

Landing @xiaoxiunique's #1973 (Settings → Developer panel + re-onboarding) with the codex-review defects fixed in-repo, authorship preserved. Re-onboarding erases real, unrecoverable state, so the erasure has to actually stick and the app has to shut down cleanly around it.

## What

- Settings → Developer panel; scoped re-onboarding that erases app-owned state (conversations / preferences / telemetry consent / onboarding flags) and relaunches the app.

## Fixes applied over the original PR (codex review, 3 rounds to convergence)

- **Clean shutdown**: the relaunch used `exit(0)`, bypassing `applicationWillTerminate` — so the last chat edit was never persisted and in-flight model-download subprocesses were orphaned (`perform` only stopped the server). Extracted a shared `AppDelegate.runStandardTermination()` (chat persist + server reap + download reap + conversation flush) and run it in `perform` **before** erasing.
- **Erase ordering**: running termination *after* the erase re-persisted the conversations just deleted (the flush + a streaming `stopAndPersist` resurrected `conversations.json`). Termination now precedes `eraseState`, so the delete is the final write.
- **Test hygiene**: `QuickstartCoordinator` hardcoded `UserDefaults.standard`, so the suite mutated the developer's real onboarding state and never validated erasure against the injected domain. `defaults` is now injectable (defaults to `.standard`; production unchanged); tests use scratch suites.
- **Legacy flag**: `resetForReonboarding` cleared the v2 key but not `legacyStorageKey`, so a user carrying the v1 `done` flag relaunched with `legacyDone == true` and Quickstart stayed suppressed.

## Review

Supply-chain audit clean (the `/bin/sh` spawn is a safe app-relaunch — pid/path passed as `$1`/`$2`, not interpolated; no deps/CI/network changes; ASCII-only). Codex converged; local `swift build` + 89 re-onboarding/Quickstart tests green.

Co-authored-by: Atom <929746602@qq.com>

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
+++ b/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
@@ -614,6 +614,10 @@ enum DevSnapshot {
                     .environment(appearance)
                     .environment(settingsRouter)
                     .environment(server)
+                    // The capture loop walks Category.allCases, which in a
+                    // debug build includes Developer — and that panel reads
+                    // the coordinator.
+                    .environment(quickstart)
                     .environment(downloads)
                     .environment(updater)
                     .environment(snapshotSparkleUpdater)
--- a/apps/rapid-mac/Sources/Rapid/RapidApp.swift
+++ b/apps/rapid-mac/Sources/Rapid/RapidApp.swift
@@ -750,13 +750,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         // write inline so the data lands before this delegate hook
         // returns to AppKit.
         MainActor.assumeIsolated {
-            AppDelegate.runTerminationSequence(
-                stopStream: { AppDelegate.shared.chat?.stopAndPersist() },
-                signalServer: { AppDelegate.shared.server?.beginShutdown() },
-                signalDownloads: { AppDelegate.shared.downloads?.beginShutdown() },
-                reapServer: { AppDelegate.shared.server?.shutdownSync() },
-                reapDownloads: { AppDelegate.shared.downloads?.finishShutdown() }
-            )
+            AppDelegate.runStandardTermination()
         }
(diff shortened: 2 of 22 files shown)


raullenchai:
Reviewed with Codex + local build/test/golden verification.

**Fixed in-branch (real gap):** #1993 added the DEBUG-only Developer category and refreshed `update-state`'s AX baseline but left `settings-persistence.settings-root` stale (still ended at the App category), so that flow fails its structural check on the debug build the golden harness uses. Regenerated it — the diff is the single Developer AXRow. (CI's golden job doesn't run settings-persistence, so it was green but latently broken.)

**Verified:** `swift build` (debug+release) clean; 27 swift tests green incl. the 11-test re-onboarding suite (scope erasure, relaunch, escalation, consent-file handling); golden flows `update-state`, `no-dead-controls`, `settings-persistence` PASS on a debug build. Codex confirmed the happy path sound: termination→flush→erase→spawn→exit is an uninterrupted MainActor chain, production uses `.standard`/`com.rapidmlx.rapid`, consent re-prompts on success, and the /bin/sh relaunch passes pid/path as positional args (no injection).

**Declined (documented, low-realism failure-mode edges, none corrupt data on realistic input):**
- Partial-erase reported as success if a target file is locked (e.g. immutable telemetry-consent.yaml) — under-deletes, recoverable by re-running; app-owned YAML is never locked in practice.
- `removeItem` doesn't enforce regular-file/no-follow — a final-component symlink is removed as the link, a parent symlink resolves to the intended app-owned file; requires a deliberately symlinked app-support dir.
- Relaunch `Process.run()` failure (EAGAIN) still exits — rare, recoverable by reopening, not data loss.

These are worth a follow-up hardening pass but don't block landing.

Issue #1999, opened 2026-08-17
CLI output polish: ps/models column collisions, and 'connect' reports Ready with no server running


raullenchai:
Three presentation defects found while dogfooding main HEAD (`58baafd1`) on an
M3 Ultra. None affects correctness of inference; grouping them because they are
the same family (CLI rendering) and each is a small, self-contained fix.

## 1. `rapid-mlx ps` — MODEL and UPTIME columns collide

A long model path runs straight into the uptime with no separator:

```
  PID     PORT    MODEL                                   UPTIME
  67305   8199    /Users/raullenstudio/mtplx-research/Qwen3.8-27B-MTPLX-Optimized-Speed5h20m
```

`...Optimized-Speed5h20m` — the uptime is unreadable and the column header no
longer describes the layout. Serving a local path (rather than an alias) is the
normal case for anyone testing a converted model, so this is easy to hit.

## 2. `rapid-mlx models` — a long Reasoning value misaligns the rest of the row

```
  Alias                             Size       Tools            Reasoning    Spec-Decode Suffix Tier DFlash  DDTree  Preset
  deepseek-r1-32b-4bit              17.2 GiB   —                deepseek_r1_distill ✓          neutral     —       —       Suffix
```

`deepseek_r1_distill` overflows its column, so every field to its right shifts
and no longer lines up with the header. Same root cause as (1): fixed-width
padding with no truncation or width negotiation.

## 3. `rapid-mlx connect` announces a server that is not running

With nothing listening on port 8000:

```console
$ lsof -nP -iTCP:8000 -sTCP:LISTEN    # no output
$ curl -s http://localhost:8000/health # no output
$ rapid-mlx connect
  Ready: http://localhost:8000

  OpenAI:    http://localhost:8000/v1
  Anthropic: http://localhost:8000
  Model:     (not yet loaded)
```

"Ready:" is the strongest word on the screen and it is false — the URLs it
prints all refuse connections. `Model: (not yet loaded)` hints that something is
off, but it reads as "the server is up and warming", not "there is no server".
Copying the `agents ... --setup` line it prints wires a client to a dead
endpoint.

Suggested: probe `/health` and either say so plainly ("No rapid-mlx server on
port 8000 — start one with `rapid-mlx serve <model>`") or label the block as the
address a server *would* use.

PR #2006, opened 2026-08-18
fix(cli): stop ps/models column collisions and connect's false Ready (#1999)

raullenchai:
## Why

Three CLI presentation defects reported in #1999 (found dogfooding main HEAD), same family (CLI rendering), each self-contained.

## What

- **`rapid-mlx ps`** — a served local *path* (the normal case for a converted model) longer than the 40-char MODEL column ran straight into UPTIME (`…Optimized-Speed5h20m`). Front-elide the model to 38 chars via `_elide_front()`, keeping the distinctive tail, so the `<40` pad always leaves ≥2 spaces before UPTIME.
- **`rapid-mlx models`** — a Reasoning value wider than the fixed 12 chars (`deepseek_r1_distill` is 19) overflowed and shifted Spec-Decode and every column to its right out from under the header. Size the Tools and Reasoning columns from the data like the Alias column already is. Preset stays last, so its long `MTP@…` values overrun harmlessly.
- **`rapid-mlx connect`** — announced `Ready:` for endpoints that refuse connections. Probe the target (`probe_server_alive`) and, when nothing answers, render `No rapid-mlx server on … — start one with rapid-mlx serve <model>` and drop the Connect cheat-sheet that would wire a client to a dead endpoint. The serve lifespan keeps `running=True`, so its Ready banner is unchanged.

## Tests

- `_elide_front` unit coverage (short / exact / long-path tail / no-collision gap).
- `test_spec_decode_column_aligns_despite_long_reasoning` — red against the fixed 12-wide column, green with dynamic sizing.
- `test_connect_reports_no_server_when_nothing_listening` — no false Ready, no dead cheat-sheet; existing banner-shape tests now mock a live probe.
- Full `test_models_command_layout` + `test_cli_connect` + `test_cli_ps_layout` + `test_ready_banner_timing` = 45 passed; `ruff` clean.

Presentation-only; no change to inference or the serve banner.

---
🤖 AI-assisted with Claude Code (Opus 4.8). Human-reviewed.

### DIFF
--- a/tests/test_cli_connect.py
+++ b/tests/test_cli_connect.py
@@ -123,10 +123,83 @@ def test_connect_no_target_renders_banner(monkeypatch):
         "resolve_endpoints",
         lambda **kw: connect.ServerEndpoints("localhost", 8000, model="m1"),
     )
+    # A live server is present for this shape check.
+    monkeypatch.setattr(connect, "probe_server_alive", lambda *a, **k: True)
     out = _run_connect()
     assert "Ready: http://localhost:8000" in out
     assert "Connect:" in out
     assert "OpenAI:" in out
+
+
+def test_probe_server_alive_keys_on_healthz_status_code(monkeypatch):
+    """#1999: /healthz is rapid-mlx-specific, so a non-404 response (2xx, or a
+    DDTree auth 401/403) counts as alive across all serving modes; a 404 (an
+    unrelated HTTP server) or a refused/timed-out connection does not. The body
+    is not inspected — the standard / DFlash / DDTree bodies differ."""
+    import urllib.error
+    import urllib.request
+
+    def _ok(url, timeout=None):
+        return object()  # urlopen returns without raising on 2xx
+
+    def _raise(code):
+        def _open(url, timeout=None):
+            raise urllib.error.HTTPError(url, code, "err", {}, None)
+
+        return _open
+
+    def _conn(url, timeout=None):
+        raise urllib.error.URLError("connection refused")
+
+    # 2xx (any serving mode) → alive.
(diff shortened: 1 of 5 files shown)


raullenchai:
### Adversarial review trail (Codex, 6 rounds)

Real defects found and fixed, each with regression coverage:
1. connect probe hit `/health` (can stall under load) and counted any HTTP error as alive → switched to the engine-free `/healthz` fast-path.
2. Requiring the standard `ready`+`model_loaded` body → **rejected running DFlash/DDTree servers** (their `/healthz` bodies differ) → key on the response **code**, not the body.
3. DDTree can gate `/healthz` behind its API key (**401**) → treat 401/403 as a live server.
4. A **draining** server answers **503** → treat only 2xx / 401 / 403 as "point a client here"; 404/500/503 are not.
5. A port occupied by a **non-HTTP** service (SSH, a DB) raised `http.client.BadStatusLine` → **crashed connect** → caught `http.client.HTTPException`.
6. `models` dynamic column widening leaked into the audio/video/image sub-table separators → they now size from their own header. `_elide_front` honors its width contract down to 0/1.

**Declined (documented):** `ps`/`models` measure code points, not terminal cells, so a CJK path (e.g. `/Users/陈小龙/…`) can still exceed its display width. This is a **pre-existing, global** property of every CLI table (all pad by code points); this PR improves the common ASCII case and does not regress it. A correct fix is a repo-wide display-width (wcwidth) change across all tables, out of scope for this rendering-polish PR.

Focused suite: 43 passed; `ruff` clean. Probe ladder: 2xx/401/403 → alive; 404/500/503/refused/timeout/non-HTTP → not.

Issue #2008, opened 2026-08-18
GUI golden gate covers 12 of 25 flows, and AX baselines encode toolbar sibling order (session-dependent)


raullenchai:
Two gaps in the `gui-golden-flows` gate, both surfaced while dogfooding
0.12.15 on an M3 Ultra. Neither is a product defect; both let real breakage hide.

## 1. Half the flows never run in CI

`rapid-mac-ci.yml` invokes 12 named flows. The harness defines **25**:

| | flows |
|---|---|
| **run in CI** | audio-readiness, catalog-integrity, chat-restore, image-generation, launch-integrations, model-crash-recovery, no-dead-controls, resident-load-rejected, settings-mtp, slow-stream-stop, update-state, window-close-prompt |
| **never run** | browse-all-destination, cached-curated-tradeup, cached-quickstart, chat-depth, chat-document-attachment, download-progress, **fresh-install**, low-memory-choice, math-rendering, message-actions, restored-tools, settings-persistence, tool-loop-budget |

`fresh-install` is the first screen a new user ever sees, and its baseline has no
gate. A baseline in that set can drift for weeks without anyone noticing — the
only reason it surfaced here is that a human ran `--flow all` by hand.

Either wire the remaining 13 into CI, or state in the workflow which flows are
deliberately excluded and why, so the omission is a decision rather than an
accident.

## 2. Baselines encode toolbar sibling ORDER, which varies by session

Every snapshot that includes the window toolbar pins the order of two buttons:

```
AXToolbar
  AXButton desc="Hide Sidebar" enabled=true
  AXButton id="Toolbar.SearchChats" desc="Search chats" enabled=true
```

Running the full suite over a Screen Sharing session (4K panel, `UI Looks like
1920x1080`) reproduces the opposite order **deterministically — 3 of 3 runs**,
while the same machine's CI runner produces the committed order. Both buttons are
present and enabled in both cases; only the sibling order differs.

The blast radius is exact. Rebasing baselines to this session and running every
flow gives:

```
25 passed, 0 failed
26 files changed, 26 insertions(+), 26 deletions(-)
```

…and all 26 lines are the same `AXButton desc="Hide Sidebar"` moving one row.
Nothing else in any snapshot differs.

So a maintainer on a different display setup sees ~13 flows fail with a
structural-mismatch diff that looks alarming and is purely cosmetic. Worth
either normalising toolbar sibling order before comparison, or excluding the
[...]

PR #2009, opened 2026-08-18
test(mac): stabilize semantic GUI actions in CI

raullenchai:
## What changed

- stop the AX walker as soon as an action target is found, avoiding stale SwiftUI accessibility objects
- retry semantic AXPress actions and verify their observable effects for Audio Save/Play and the Developer confirmation dialog
- run the real DevSnapshot sidecar/chat check before the large static render matrix so the test harness does not trip the production memory gate
- remove the unreliable coordinate-click fallback

## Why

The AX helper kept walking up to 12,000 nodes after locating an action target. SwiftUI could replace the backing element before AXPress, producing transient invalid-element/cannot-complete errors. The coordinate fallback could report success even when unattended-runner TCC discarded the injected CGEvent. Separately, DevSnapshot attempted its real model load only after rendering roughly 200 full-size views, so harness memory could block the sidecar startup.

## Validation

- `swiftc apps/rapid-mac/scripts/rapid-ax.swift`
- `gui-golden-flows.sh --flow audio-readiness` twice
- `gui-golden-flows.sh --flow no-dead-controls` twice
- `gui-golden-flows.sh --flow all`
- 93 focused Python GUI contract tests
- Swift test suite: 2,289/2,290 first run; the one unrelated timing test passed 5/5 on immediate focused rerun
- isolated real GUI live run with `qwen3-0.6b-4bit`: sidecar ready, real chat completed, live screenshot written, clean stop

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
+++ b/apps/rapid-mac/Sources/Rapid/DevSnapshot.swift
@@ -106,6 +106,28 @@ enum DevSnapshot {
             )
         }
 
+        // LIVE mode must run before the static matrix. The matrix below
+        // renders ~200 full-size views and temporarily raises process memory;
+        // running live last can trip ServerManager's real pre-load safety
+        // gate even for a 0.6B model, leaving state=.idle and silently skipping
+        // the only end-to-end sidecar check. Use a separate non-persisting
+        // chat model so the real turn cannot leak into the static fixtures.
+        if let liveAlias = ProcessInfo.processInfo.environment["RAPID_DEV_SERVE_ALIAS"],
+           !liveAlias.isEmpty {
+            let liveChat = ChatViewModel(
+                sampling: sampling,
+                customInstructions: chat.customInstructions,
+                server: server,
+                persistsConversations: false
+            )
+            await runLiveChat(
+                alias: liveAlias, server: server, chat: liveChat,
+                downloads: downloads, quickstart: quickstart, dir: dir
+            )
+            await server.stop()
+            server.dismissTerminalState()
+        }
+
         // The Images tab, rendered standalone — ``ContentView`` owns its
         // ``SidebarSection`` privately, so (like ``launchView``) the detail
         // surface is captured directly rather than by driving navigation.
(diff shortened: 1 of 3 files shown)


raullenchai:
Closing #2007 in favour of this — it changes the same two call sites and removes
the same primitive, and this PR is a strict superset (retry + effect
verification, the Developer confirmation dialog, the AX-walker early stop, the
DevSnapshot ordering). Dropping mine so the two don't conflict.

Independent evidence gathered on an M3 Ultra before finding this PR, in case any
of it is useful — it corroborates the TCC diagnosis from a different angle:

**The A/B that isolates the primitive.** Same flow, same build, one line changed:

| Save step | result |
|---|---|
| `click-center` | `{"success": true}`, 10s timeout, no file |
| `press` (AXPress) | `saved-speech.wav`, **64,044 bytes** — the fake sidecar's exact payload |

So the click was discarded while the driver still reported success, which matches
"unattended-runner TCC discarded the injected CGEvent" exactly.

**The failure that appears once Save is fixed.** With only the call-site change,
CI advances past the speech half and then fails at

    FAIL: Choose File selected no usable audio file or Transcribe stayed disabled

That assertion is also new in #2005, so the extended `audio-readiness` has never
completed on CI. Worth confirming this PR's run gets past it — it did not
reproduce locally for me (the transcription half passed and wrote
`saved-transcription.txt`), so it looks like the same unattended-session class of
problem rather than a product defect.

**Two things that are environmental, not product** (filed as #2008 so they don't
get rediscovered):
- CI runs 12 of the 25 defined flows; `fresh-install` — the new user's first
  screen — is not among them.
- The AX baselines encode toolbar sibling order. Over a Screen Sharing session
  (4K panel, "looks like 1920x1080") `Hide Sidebar` / `Search chats` swap
  deterministically, 3/3 runs, while this machine's CI runner emits the committed
  order. Rebasing baselines locally and running everything gives **25 passed / 0
  failed** with `26 files changed, 26 insertions(+), 26 deletions(-)` — every one
  of those 26 lines being that one button moving a row. That diff shape is a
  quick way to tell a cosmetic session difference from a real structural change.

PR #1899, opened 2026-08-12
fix(mac): stop streaming transcript scroll layout loops

raullenchai:
Closes #1877

## Summary
- stop `updateNSView` from issuing an AppKit scroll on every streamed SwiftUI mutation
- coalesce document/clip frame notifications into at most one bottom adjustment per main-loop turn
- make already-settled bottom adjustments true no-ops and suppress queued follow work during a live user scroll
- extend the real mounted SwiftUI/AppKit scroll golden runtime flow with a 60-update streaming burst and a bounded scroll-event assertion

## Validation
- `apps/rapid-mac/scripts/verify-chat-scroll-runtime.sh`
- `swift test --filter RapidTests`
- `git diff --check`

Issue #1877, opened 2026-08-12
Chat scrolling during streaming can spin SwiftUI layout at ~100% CPU


chen-yijiang:
### App version

0.12.10 (154), built from main at 8a159456

### Last version where it worked (if this is a regression)

Unknown

### Mac model and memory

MacBook Pro M3 Pro, 18 GB

### macOS version

macOS 26.5.2

### Where in the app does this happen?

Chat window (asking, answering, tools, formatting)

### What happened? What did you expect instead?

The desktop app can become completely unresponsive while using the chat view. I do not have a single deterministic click-by-click repro yet, but it happens more readily during active chat usage when I scroll the transcript frequently, especially while a response is streaming.

Observed during one freeze:

- The app process was spinning at roughly 96-102% CPU.
- The `rapid-mlx serve` sidecar was idle / near idle.
- Memory pressure was not the issue.
- The app did not recover on its own.

Expected: scrolling the chat transcript during streaming should remain responsive, and the app should not enter a high-CPU UI loop.

### Steps to reproduce

1. Open the desktop app.
2. Start a chat with a local model.
3. Send a prompt that produces a streaming response.
4. While the response is streaming, repeatedly scroll the chat transcript up and down.
5. Continue normal chat interactions and scroll frequently.

Observed: the app can become completely unresponsive. This is intermittent; I do not have a deterministic single path yet, but frequent scrolling during chat makes it easier to trigger.

### Diagnostics report

Runtime logs did not show a server/model error. The app log only had unrelated input-method messages around the time of the hang.

Two macOS `sample` captures of the frozen app process show the main thread spending time in SwiftUI / AttributeGraph layout rather than in the Python sidecar or model execution.

Hot paths included:

- `NSHostingView.beginTransaction`
- `ViewGraphRootValueUpdater.updateGraph`
- `GraphHost.flushTransactions`
- `AG::Subgraph::update`
- `LazySubviewPlacements.updateValue`
- `LazySubviewPlacements.placeSubviews`
- `LazyStack<>.place`
- `ForEachState.forEachItem`
- `ForEach.IDGenerator.makeID`
- `LayoutEngineBox.sizeThatFits`
- `AG::Graph::propagate_dirty`

App symbols present in the samples included:

- `TranscriptScrollPositionProbe`
- `initializeWithCopy for ChatMessage`
- `destroy for ChatMessage`

This points to a desktop UI layout/update loop, not a server-side generation stall.

Likely source areas:

[...]

PR #2021, opened 2026-08-18
fix(mac): jump to bottom did nothing once an answer had finished

xiaoxiunique:
## The bug

The jump-to-latest button did nothing on a finished transcript — which is the state a reader is most likely to press it in.

It set `isPinnedToBottom = true` and trusted the probe's `attach` to do the scrolling. `attach` only anchors on a **new** attachment (#1877), and by the time a reader presses the button the scroll view has been attached for the whole session. Nothing moved.

Following then rested entirely on `documentFrameDidChange`, which fires while an answer streams and never again once it settles. So the button worked mid-stream and was dead afterwards.

## Why no review caught it

The two halves shipped a day apart and **neither is wrong on its own**:

| | |
|---|---|
| 2026-08-12 · #1899 | narrowed `attach` to new attachments only |
| 2026-08-13 · #1906 | added the button against the older contract |

Only the composition was broken, which is why reading either file finds nothing.

## The fix

Not by widening `attach` again — that narrowing exists to stop a self-sustaining scroll loop at streaming cadence. The caller asks explicitly instead: `scrollToBottomRequest` is a token the button bumps, and the coordinator scrolls when it changes.

Two details worth their lines in the coordinator:

- it **compares** tokens rather than testing for non-zero, because the counter belongs to the caller and may wrap or reset
- the **first** token it sees is recorded rather than acted on — the first render arrives with the token already at some value, and treating that as a request would yank a transcript the reader had deliberately left scrolled up

## Tests

Driven against a real `NSScrollView`, not a source grep: the bug was never visible in either file alone. Three cases — a changed token scrolls an already-attached view, an unchanged token stays silent (otherwise it fights the reader on every streamed frame), and the first token is not a request.

Break-verified by making the new path a no-op, which reproduces the original symptom exactly:

```
Expectation failed: (scrollView.contentView.bounds.minY → 0.0) > (0 → 0.0)
↳ the transcript did not move — re-pinning alone never scrolled an already-attached view
```

Full suite: 2293 tests, with the four failures that reproduce on clean `main`.


raullenchai (approved):
Reviewed the probe/caller contract and exercised the fix against a real NSScrollView. The explicit request token restores finished-transcript scrolling without reopening the streaming feedback loop; focused tests pass 3/3.

PR #2022, opened 2026-08-18
fix(mac): raise the main window floor to 720, and actually enforce it

xiaoxiunique:
## The bug

`ContentView.minWindowWidth` was a constant nothing applied.

The window's real minimum came from whatever the rail's 176pt column minimum and the 440pt detail floor happened to add up to — about **616pt**. So the declared number (640) and the window disagreed, and nothing noticed: the only reader of the constant was a test asserting against itself.

## The fix

`.windowResizability(.contentMinSize)` derives the floor from the content's minimum size, so the shell states it once rather than leaving it to be inferred from two unrelated columns.

Measured after the change — forcing the window to 500pt wide clamps it:

```
$ osascript -e '… set size of window 1 to {500, 600}'
$ osascript -e '… get size of window 1'
  720, 612
```

The floor itself goes **640 → 720**. Everything that described 640 as the current floor moves with it:

- the detail-floor comment, which was explaining a constraint that no longer binds — 440 now has 80pt of slack and stays because it is the detail's own minimum, not a number derived from the window
- `DevSnapshot`'s floor captures, which exist to prove the layout survives the narrowest supported window and were proving it at a width the app no longer allows (`chat-640x560-*` → `chat-720x560-*`)

## Tests

`shellFitsAtTheWindowFloor` already checked rail + detail ≤ floor, but **a floor nobody enforces satisfies that trivially**. The new guard pins the thing that was actually broken: that the declared floor is applied. Break-verified by removing the frame.

It reduces its check to a `Bool` before the expectation — `#expect` prints the expression it is handed, and passing it the stripped source put the whole of `ContentView`, some 30KB, into the failure output.

Full suite: 2294 tests, with the four failures that reproduce on clean `main`.


raullenchai (approved):
Reviewed the scene/content minimum-size contract and the 720pt layout floor. The shell now supplies the minimum that  actually enforces; focused workspace visual tests pass 29/29. I also corrected the stale 640pt rationale.

raullenchai (approved):
Reviewed the scene/content minimum-size contract and the 720pt layout floor. The shell now supplies the minimum that contentMinSize actually enforces; focused workspace visual tests pass 29/29. I also corrected the stale 640pt rationale.

PR #2023, opened 2026-08-18
fix(mac): the same file could be attached over and over

xiaoxiunique:
## Why

Attaching the same file repeatedly is not only redundant. The shared 24,000-character document budget is divided across attachments, so four copies of one PDF send only its first quarter four times. Path-normalized deduplication preserves the full useful document context.

## What changed

- Deduplicate file-backed image and document attachments in the shared import funnel.
- Treat symlink and dot-segment spellings of one path as the same file.
- Preserve the exact source URL beside each successfully parsed attachment, including when an earlier candidate fails.
- Keep raw clipboard image bytes out of path-based deduplication.

## Verification

Eight focused tests cover already-attached files, within-batch repeats, normalized paths, distinct paths, image/document parity, empty state, call-site wiring, and a failed import followed by a successful import. Swift 6 actor isolation and source association were fixed during maintainer review.


raullenchai (approved):
Reviewed the shared import funnel and path-normalized deduplication. I fixed two blockers before approval: Swift 6 actor isolation on , and incorrect URL association after a preceding batch item failed to parse. Added a real mixed failure/success import regression; focused tests pass 8/8.

raullenchai (approved):
Reviewed the shared import funnel and path-normalized deduplication. I fixed two blockers before approval: Swift 6 actor isolation on attachmentKey, and incorrect URL association after a preceding batch item failed to parse. Added a real mixed failure/success import regression; focused tests pass 8/8.

PR #2025, opened 2026-08-18
fix(gui-golden): stop the FilePicker re-press storm cancelling its own selection

raullenchai:
## Why

Because `gui-golden-flows` is the last red gate on the 0.12.15 bump (#2013), and
its failure is self-inflicted by the harness. The audio-readiness `Choose File`
step failed 3 of the last 5 unattended runs — twice AFTER #2009 added an in-loop
re-press and once after #2017 added driver-level press retries. Refs #2008.

## What

Each press of `Audio.Transcription.FilePicker` calls `chooseAudioFile()` →
`selectFile()`, which **restarts** the file's async validation. The loop
re-pressed every 0.35 s, so on a busy runner where validation outlasts one loop
period, the retry kept cancelling the very work it was polling for — a livelock.
Locally validation finishes inside the first iteration, which is why this step
passes on any warm machine and fails only in CI.

- Re-press only every 10th iteration (~2.5 s settle windows) — delivery retry is
  preserved, the storm is not.
- Budget widened 40 → 80 iterations so one slow validation fits.
- Save/Play keep their per-iteration re-press deliberately: writing a file is
  idempotent and restart-free, and those steps have passed every run since #2017.

## Verification, and its limit

Passes locally both before and after — by the same timing that hides the bug —
so a local run proves nothing here. `bash -n` clean. The unattended CI run on
this PR is the authoritative test; the failure history above is the baseline it
is judged against.

## Checklist

- [x] Lint passes (`bash -n`)
- [x] Harness-only; no product code touched

### DIFF
--- a/apps/rapid-mac/scripts/gui-golden-flows.sh
+++ b/apps/rapid-mac/scripts/gui-golden-flows.sh
@@ -3864,14 +3864,24 @@ flow_audio_readiness() {
 
     see_main "$OUT/transcription-controls-before.json"
     local file_selected=0
-    for ((i=0; i<40; i++)); do
+    for ((i=0; i<80; i++)); do
         # AXPress can return success for a SwiftUI button whose backing object
         # is replaced before its closure runs. Resolve the current button on
         # every attempt and require the actual selection + enabled Transcribe
         # state instead of treating the AX result as proof of user-visible work.
-        "$AX_DRIVER" press "$APP_PID" Audio.Transcription.FilePicker \
-            > "$OUT/transcription-file-press.json" 2>/dev/null || true
-        sleep 0.1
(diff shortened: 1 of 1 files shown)


raullenchai:
Superseded by #2016, which merged first and rests on a better diagnosis. Mine assumed the re-press storm was cancelling an async file validation; #2016 read the product path and showed golden-mode `chooseAudioFile()` is synchronous — no async work to cancel — and the real mechanism is an AXPress landing on a button whose SwiftUI backing is mid-rebuild as the pane settles, silently no-oping. Its settle-before-press + rendered-proof loop addresses that directly; my settle-window re-press only papered over it. Nothing from this PR is needed on top.

PR #2027, opened 2026-08-18
feat(chat): stream MCP tool rounds and preserve completed results

osdodo:
## What does this PR do?

  - Streams every MCP tool-call round in `rapid-mlx chat`, including reasoning, tool calls, and final responses.
  - Reassembles fragmented and interleaved tool-call arguments using their stream indices, while handling missing IDs and incomplete calls.
  - Preserves completed tool results when execution is interrupted, a follow-up request fails, or the tool-round limit is reached.
  - Adds `--mcp-max-rounds` and reports tool-limit exhaustion without leaving malformed conversation history.
  - Coerces unambiguous scalar argument mismatches, removes unsupported arguments when schemas forbid extras, and truncates oversized tool results.
  - Makes generic agent setup honor `--dry-run` without creating or modifying configuration files.
  - Adds regression coverage for MCP streaming, fragmented arguments, parallel calls, bounded rounds, partial execution, setup previews, and DSH provider configuration drift.
  - Clarifies the SDK validation matrix terminology and marks the outdated benchmark scorecard as stale.

  ## Why is this needed?

  `rapid-mlx chat --mcp-config` previously used blocking requests for tool rounds, leaving the terminal without model output during the slowest part of a local multi-step task.

  Tool-call arguments may also be split at arbitrary offsets or interleaved by OpenAI-compatible streaming servers. Without index-based reassembly, valid calls can reach the MCP runtime with malformed arguments. When the fixed round limit was exceeded or a follow-up failed, the existing recovery path could discard completed tool results, hiding work and side effects that had already occurred.

  Large tool responses and minor scalar type mistakes are especially costly for small local models: they can exhaust the context window or waste another inference round on an avoidable validation error.

  Separately, the generic agent setup path accepted `--dry-run` but still wrote the real configuration file. This PR restores the flag's promised preview-only behavior.


raullenchai (approved):
Reviewed the streamed MCP agent loop, argument preparation, tool-result preservation, setup dry-run changes, and external-contribution supply-chain surface. I fixed reasoning replay across tool rounds, union-schema coercion, and non-positive round bounds. Targeted MCP/CLI tests pass; no workflow, dependency, or binary changes are present.

PR #2029, opened 2026-08-18
feat(mac): export conversations and file them into folders

Jevin-F:
## Why

Two gaps against comparable chat apps:

- **Transcripts could not leave the app.** Only per-message clipboard copy existed — no way to keep, share or back up a conversation.
- **Grouping was entirely app-derived.** Pinned, date buckets, archived. Nothing let a user organise by project or topic, so a long history degrades into one undifferentiated list.

## Export

Per-conversation export sits on the sidebar row menu. **File → Export All Chats** (⇧⌘E) writes the whole library — folder list included — as one JSON backup.

The two formats have separate jobs rather than one format doing both badly.

**Markdown reproduces what the chat surface shows** — user and assistant turns, nothing else:

- Tool rows are excluded because `ChatView` excludes them (`role != .tool`)
- Instruction rows are wire-only and never reach the visible transcript
- Reasoning is omitted: the reader of an exported document wants the answer, not the model's thinking

Attachment names, error notices and truncation markers are kept. Dropping those makes the remaining text *misleading* rather than merely shorter — a prompt referring to a file with no sign of one reads as nonsense, and a failed turn stripped of its error reads as the model having answered with silence.

**JSON is the complete record**: reasoning, tool rows, instructions and attachment bytes all survive.

Its dates are ISO8601 strings rather than Swift's default numeric encoding. The default is seconds since **2001** — parsed as a Unix timestamp, which is what any consumer will assume, it silently yields a date 31 years off. No error, no crash, just a plausible-looking wrong date. Millisecond accuracy is the accepted cost and has its own test.

## Folders

Single level, single membership — a conversation carries at most one `folderID`.

- **Nesting** rejected: the rail is 200pt wide, indentation runs out immediately
- **Multi-membership** rejected: a row under two folders makes the selection highlight and the delete target ambiguous

The folder list is a separate `folders.json` beside `conversations.json`, not a new key inside it. The conversation file is a bare array, so wrapping it in an envelope would make every already-shipped build read its own history as corrupt. The new field on `ChatConversation` decodes with `decodeIfPresent`, matching how every prior field was added.

**Rows appear in exactly one place**, in priority order: archived → folder → pinned → date bucket. Filing outranks the derived groups because it is an explicit instruction about where a conversation lives.

Failure modes handled deliberately:

- A `folderID` naming a deleted folder degrades to unfiled rather than vanishing
- Deleting a folder keeps its conversations and returns them to the date list
- Filing surfaces an archived row — recording the folder while leaving the row in the Archived disclosure would change nothing visible and read as a no-op
- Unfiling does **not** re-archive: there is no prior state to restore, and archiving a row the user only wanted out of a folder would hide it entirely

## Drag and drop

A row can be dragged onto a folder; the whole group is the target, not just its header.

The payload is a private UTI carrying only the conversation id. A public text type would light every folder up as a drop target for any stray selection dragged over the rail. Resolving the id at drop time (rather than trusting a snapshot taken when the drag began) means a rename or a streamed token landing mid-gesture cannot file stale state.

Removing from a folder stays on the row menu.

## Testing

38 tests across two new suites. Beyond the happy paths, the regressions worth naming:

- Pre-folder history loads without being sided as corrupt — the failure that would present as a wiped sidebar on upgrade
- A corrupt `folders.json` is sided, not fatal to the conversation history
- Orphaned `folderID`s fall back to the date buckets instead of disappearing
- Filing / pinning / archiving precedence, asserting each row renders exactly once
- JSON date format pinned against the 2001-epoch digits, so the encoder cannot be "simplified" back to the default

`scripts/check_rapid_mac_ax_identifiers.py` passes — every new control carries an identifier, and folder identifiers are name-derived so a golden flow that creates a folder can predict them.

**Not automated:** the drag gesture itself. The data-layer behaviour it drives is covered, but the pointer interaction was verified by hand.


raullenchai (approved):
Reviewed export fidelity, folder persistence/grouping, drag payload isolation, save-panel file types, accessibility, and the external-contribution supply-chain surface. I added a discoverable New Folder toolbar action, duplicate-name protection, real GUI coverage for create → file → export reachability, and fixed Markdown exports being advertised as plain text (which could rewrite .md to .txt). Swift tests and the new chat-restore Golden path pass. No workflow, dependency, or binary changes are present.

PR #2051, opened 2026-08-18
fix(mac): un-red the golden gate — alert TextField by role, save panel by its real button [skip-version-bump]

### LINKED ISSUE #2050: gui-golden chat-restore is born-red: SwiftUI .alert TextField never exposes its accessibilityIdentifier
## Why
Since #folders landed (3c00f0e1, 'export conversations and file them into folders'), the `gui-golden-flows` gate fails deterministically on every PR at:

```
[gui-golden] FAIL: timed out waiting for AX identifier Sidebar.Folder.NameField
```

Root cause, reproduced locally on macOS 26.6 (same failure as the macos-15 CI runner, so not an OS skew): the folder-name prompt is a SwiftUI `.alert` with a `TextField`. On macOS, identifiers on alert BUTTONS propagate into the AX tree, but the TextField's does not — the dump shows the sheet with an anonymous field:

```
AXSheet id='_NS:87' desc='alert'
    AXTextField id=''                              ← identifier lost
  AXButton id='Sidebar.Folder.Prompt.Cancel'       ← button ids DO propagate
```

So `wait_identifier Sidebar.Folder.NameField` can never succeed anywhere; the feature PR's own golden runs were all superseded/cancelled before a full green, and the gate only fires on pull_request, so main went red for every subsequent PR — same birth pattern as #2032.

(The `message-actions` AXPress-failed line in the same job is retried and its flow PASSES — not part of this.)

## Fix
Flow-side, no product-UI change: teach `rapid-ax` a role-scoped selector (`sheet-role:AXTextField` = first element of that role under an AXSheet ancestor), and have the flow (1) wait on `Sidebar.Folder.Prompt.Confirm` — a button id that does propagate — to prove the prompt is up, then (2) set-value via the role selector.

Because a merge gate that can never pass blocks every PR in the repo.

raullenchai:
## Why

Since the folders feature merged (3c00f0e1), `gui-golden-flows` fails deterministically on every PR: `timed out waiting for AX identifier Sidebar.Folder.NameField`. Reproduced locally (macOS 26.6 — same failure as the macos-15 runner, so not OS skew). Root cause: SwiftUI drops the accessibilityIdentifier from a TextField inside `.alert` — the dump shows the sheet with an anonymous field while the alert's BUTTON identifiers propagate fine. One step further, the export leg is also unreachable-green: the save panel's window object publishes neither AXCloseButton nor AXCancelButton, so `close-window` can never dismiss it. Both steps were born-red — the feature PR's golden runs were superseded before a full green and the gate only fires on pull_request, so every later PR inherited the red (same birth pattern as #2032). Because a merge gate that can never pass blocks every PR in the repo.

Closes #2050

## What

- **rapid-ax**: new `sheet-role:<AXRole>` selector — matches the first element of that role under an AXSheet ancestor. Used for the folder-name field, which exists in the dump only as an anonymous `AXTextField` inside the alert sheet.
- **rapid-ax close-window**: falls back to the window's first-class `AXCancelButton` when there is no `AXCloseButton` ("close" and "cancel" are the same user intention on a panel), and names both in the error when neither exists.
- **chat-restore flow**: proves the folder prompt is up by waiting on `Sidebar.Folder.Prompt.Confirm` (a button id that does propagate), sets the name via `sheet-role:AXTextField`, and dismisses the export panel by pressing its bridged AppKit `CancelButton` identifier.

No app-code changes; no baseline changes (the UI structure is untouched).

## Testing

- Full local golden suite: **all 6 flows PASS** (macOS 26.6 / Xcode 26.6), including the previously-unreachable folder-create → file → export → pin/unpin segments.
- Evidence for the root cause: the alert dump shows `AXSheet … AXTextField id=''` next to `AXButton id='Sidebar.Folder.Prompt.Cancel'`; the export panel dump shows `AXButton id='CancelButton'` under the `save-panel` window.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/SidebarView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/SidebarView.swift
@@ -288,21 +288,47 @@ struct SidebarView: View {
         } message: { _ in
             Text("The conversations in it are kept and move back into the date list.")
         }
-        .alert(
-            folderPromptTitle,
+        .sheet(
             isPresented: Binding(
                 get: { folderPrompt != nil },
                 set: { if !$0 { folderPrompt = nil } }
             )
         ) {
-            TextField("Folder name", text: $folderNameDraft)
-                .accessibilityIdentifier("Sidebar.Folder.NameField")
-            Button("Save") { commitFolderPrompt() }
-                .accessibilityIdentifier("Sidebar.Folder.Prompt.Confirm")
-                .disabled(!canCommitFolderPrompt)
-            Button("Cancel", role: .cancel) { folderPrompt = nil }
-                .accessibilityIdentifier("Sidebar.Folder.Prompt.Cancel")
+            folderPromptSheet
         }
+    }
+
+    /// #2050: this prompt used to be a `.alert` with a TextField. SwiftUI
+    /// drops the accessibilityIdentifier from a text field inside an alert
+    /// (button identifiers propagate; the field's does not), and on the CI
+    /// runner an AX value write into an alert field never reaches the
+    /// SwiftUI binding — so the golden flow could neither find the field
+    /// nor enable Save. A plain sheet exposes the same controls as
+    /// ordinary views, where identifiers, value writes, and presses all
+    /// behave (fresh-install drives the consent sheet the same way on the
+    /// same runner). The three identifiers are the flow's contract — keep
+    /// them stable.
+    private var folderPromptSheet: some View {
+        VStack(alignment: .leading, spacing: RapidTheme.Space.lg) {
+            Text(folderPromptTitle)
+                .font(RapidFont.bodyEmphasis)
+            TextField("Folder name", text: $folderNameDraft)
+                .textFieldStyle(.roundedBorder)
+                .accessibilityIdentifier("Sidebar.Folder.NameField")
+                .onSubmit { if canCommitFolderPrompt { commitFolderPrompt() } }
+            HStack(spacing: RapidTheme.Space.sm) {
+                Spacer()
+                Button("Cancel", role: .cancel) { folderPrompt = nil }
+                    .keyboardShortcut(.cancelAction)
+                    .accessibilityIdentifier("Sidebar.Folder.Prompt.Cancel")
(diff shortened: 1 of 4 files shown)


raullenchai:
Codex review (ds-0731): **APPROVE** — 0 BLOCKING / 0 MAJOR. Recorded MINORs (not addressed per stop-at-0B0M): (1) the close-window AXCancelButton fallback is currently unexercised and its comment overstates NSSavePanel's attribute surface (the export panel publishes neither button attribute — the flow's press-by-bridged-identifier is the working path); (2) sheet-role:AXTextField is first-match-under-any-sheet — unambiguous today (no other AXTextField exists under any sheet in the captured dumps), would need scoping only if another sheet ever grows a text field.

raullenchai:
Follow-up after the first full GUI run exposed a second, narrower race:

- the new sheet-scoped TextField selector worked;
- `AXValue` reached the native field;
- but CI pressed Save before SwiftUI had propagated the binding and exposed the button as enabled, so `AXPress` failed three times.

I updated the flow to wait for `Sidebar.Folder.Prompt.Confirm` to be observably `enabled=true` before pressing it. This models the actual user-clickable state instead of treating a disabled action as a stale AX element.

Local validation on the rebased branch:
- GUI harness contracts: 52 passed
- debug app build: passed
- real `chat-restore` Golden Flow: 4/4 consecutive passes
- no product UI or committed baseline changes

A fresh full validation run is now in progress.

raullenchai:
CI adjudicated the two approaches: the wait-for-enabled attempt (55415594) still failed `gui-golden-flows` — on the runner an AX value write into the alert's TextField never reaches the SwiftUI binding, so Save never enables and no wait helps. Pushed the root-cause fix on top (ff1076f8): the folder-name prompt becomes a plain `.sheet` (same three identifiers — the feature's original flow steps now work verbatim; fresh-install already drives the consent sheet on this runner, so sheet drivability is proven). Kept 55415594's `wait_identifier_enabled` guard (correct in principle, now satisfiable) and the export-panel CancelButton dismissal; reverted the rapid-ax driver changes (no longer needed). Local: all 6 golden flows PASS, unit suite green.

PR #2052, opened 2026-08-18
fix(mac): make folder prompt golden flow reachable

### LINKED ISSUE #2050: gui-golden chat-restore is born-red: SwiftUI .alert TextField never exposes its accessibilityIdentifier

raullenchai:
Closes #2050.

## What changed

- add a `sheet-role:<AXRole>` target to the native AX driver for controls inside an `AXSheet` whose SwiftUI identifier is discarded by macOS
- wait for the folder prompt confirm button, whose identifier is preserved, before addressing its anonymous `AXTextField`
- document the narrowly scoped selector

## Root cause

On macOS, the SwiftUI `.alert` keeps accessibility identifiers on its buttons but drops the identifier on its `TextField`. The `chat-restore` flow therefore waited forever for `Sidebar.Folder.NameField` even though the prompt was present.

## Validation

- `swiftc scripts/rapid-ax.swift`
- 52 GUI harness contract tests passed
- 2,405 Swift tests passed
- release app build succeeded
- real `chat-restore` flow progressed through folder creation and observed `Golden Work (1)`; later stopped at the separate native export-panel close race

### DIFF
--- a/apps/rapid-mac/docs/gui-golden-flows.md
+++ b/apps/rapid-mac/docs/gui-golden-flows.md
@@ -510,6 +510,11 @@ python3 scripts/ax-baseline.py normalize --scrub fake-alias /tmp/…/steady.json
 ## Why this is not coordinate automation
 
 The checked-in `rapid-ax.swift` helper talks directly to macOS Accessibility.
+Actions normally target an accessibility identifier. Native SwiftUI alerts
+are the exception: macOS can discard a `TextField` identifier while retaining
+the surrounding button identifiers. After waiting for an identified alert
+button, a flow may use `sheet-role:AXTextField` to address the first text field
+strictly inside the open `AXSheet`.
 It finds controls by stable `AXIdentifier`, performs `AXPress`, sets native text
 values, and serializes roles/descriptions/values for assertions. Peekaboo is
 kept for permission checks, window discovery, menu interaction, and screenshots.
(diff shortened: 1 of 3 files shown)


raullenchai:
Closing in favor of #2051, which fixes the same #2050 alert-field AX failure and also covers the subsequently exposed export save-panel dismissal failure. I reviewed both implementations; #2051 has the more complete end-to-end chat-restore fix.

PR #2063, opened 2026-08-19
feat(mac): apply Direction D onboarding UI

loriz-art:
## What this is

The approved Paper **Direction D** visual design, applied to the existing four-step onboarding flow. The behaviour and recovery state machine merged in #1946 is the source of truth here — this PR consumes it and does not redefine it. Every recovery route, keyboard contract, accessibility identifier and VoiceOver announcement from that PR still behaves exactly as it did.

The load-bearing change is presentation. Onboarding was a `.sheet`, which on macOS is a *document-modal panel*: AppKit sizes it to its content's ideal width, insets it, rounds its corners and leaves the parent visible around it. So it drew as a centred card floating over a dimmed chat surface — the one composition Paper 05.1.A rules out — and it settled at its 620pt minimum, below the 820pt breakpoint, which made the full-height rail unreachable on any display. The root now branches between the production shell and the setup shell, so setup **is** the window content: it fills everything below the native title bar, has no corner radius of its own, and has nothing behind it to show through.

The native title bar and its traffic lights are untouched. "Full window" here means the app's content area — never macOS fullscreen.

## Direction D screens implemented

| Paper state | Screen |
|---|---|
| 05.1.A | Welcome / step 1, with the setup rail and the hardware subject rail |
| 05.1 · 04 | Model selection — starter recommendation |
| 05.1 · 05 | Model selection — trade-up, with the download / memory / fit comparison |
| 05.1 · 06 | Model selection — a pick already on this Mac |
| 05.2 | Browse all models — catalogue rows, badges, search, filter, sort |
| 05.2.J | Review download — facts table, footnote, primary |
| 05.1 · 08 | Download in progress — subject band, bytes and rate lines |
| 05.1 · 09–12 | Recovery — download stopped, download failed, offline, source unavailable |
| 05.1 · 13 | Missing engine, including repeated Recheck |
| 05.1 · 14 | Low disk — warn-only |
| 05.1 · 15 | Ready / `Start chatting` |

## Visual and layout architecture

**One canvas geometry serves every state.** A principal group fills the canvas and centres in it, with the action lane anchored beneath. This is a shared scaffold rather than a per-screen padding on purpose: the alternative is what it replaced, where each state carried its own top-padding number and they drifted apart.

**Three responsive tiers**, taken from Paper's own frames rather than invented:

| Tier | Width | Rail | Canvas |
|---|---|---|---|
| Wide | ≥ 1290 | 300pt vertical rail | Step 2 heading beside the list (two columns) |
| Medium | 820 – 1289 | 240pt vertical rail | Stacked canvas (05.1.D) |
| Compact | < 820 | 46pt horizontal strip, or a 56pt graphite band during a download | Full-width canvas (05.1.E) |

The columns breakpoint is **derived, not guessed**: rail + gutters + heading column + gap is 868pt before a single row is drawn, so below 1290 the rows fall under the width their four lanes need. A regression test sweeps 720 → 2000 in 10pt steps and asserts the row width never drops below its minimum.

**The right-hand model column is an intrinsic column (`ViewThatFits`), not a `ScrollView`.** A `ScrollView` is vertically greedy: it takes the whole offered height, the enclosing centred `HStack` then centres a child that is already full height, and the cards inside stack from *its* top. The result was a heading that read centred beside a list that read pinned, and no alignment on the parent could have fixed it. There is a measurement test for this — it renders the two-column layout offscreen and asserts the right stack's midpoint matches the canvas midpoint, plus a negative control that fails if the column becomes top-pinned while the left stays centred.

**The Step 2 heading is a pure function of the selection** (Paper 05.1 states 04 / 05 / 06). The starter gets *"Choose your first model"*; a trade-up gets *"Bigger, and what it costs"* with a comparison of download size, memory when loaded, and fit; a pick already on disk gets *"One is already here."* Cached-ness outranks size, because "you already have this" is the more useful thing to say about a 9B that happens to be on disk. Every figure in that table is an existing reading — `sizeText`, `ModelSizing.estimate`, and the same `classify` the footer primary is already gated on — so the screen compares without appraising.

## Behaviour and accessibility contracts preserved from #1946

- Four-step structure; the catalogue and Review are **not** steps of their own
- `Continue setup` on relaunch with an incomplete setup, returning to the correct Step 2 micro-stage
- Cancellation and the truthful `Download stopped` recovery — a user-initiated stop is never dressed as a fault
- Retry, and the exact Back destinations for every failure origin
- Search, filter, sort, scroll and selection restoration when returning to the catalogue
- Genuine download failure, offline, and source-unavailable diagnoses
- Missing engine, including repeated Recheck
- Low disk stays **warn-only**; incompatible memory stays blocking
- Pending Ready and `Start chatting`
- Single click selects and never navigates; double-click mirrors the visible primary through the shared `modelRowActivation` path (Review download on an uncached pick, Start existing model on a cached one, nothing on a disabled one)
- Return is `.defaultAction` on the very control the derivation disables; Escape is `.cancelAction` on Back, with the search field taking priority when it holds text
- Every accessibility identifier and VoiceOver announcement. `Quickstart.Progress` in particular keeps its exact spoken label, which the golden flows match verbatim

## Verification

**Automated**, on this branch rebased onto `upstream/main` @ `fb9475c1`:

- Debug build — clean
- Release build — clean
- Direction D focused suite — **43 / 43 passed**
- Onboarding recovery suite — **60 / 60 passed**
- Accessibility identifier inventory gate — **16 / 16 passed**
- `verify-recommendation-tiers.swift` — ALL PASS
- `verify-conversation-ordering.sh` — ok
- `verify-chat-timeout.swift` — OK
- Full suite — **2489 tests in 216 suites, 1 failure** (see below)

**Live manual pass**, on an isolated dogfood build with its own bundle identifier, preferences, `HOME`, Application Support and model cache:

1. Full-window presentation — no modal card, no dimmed chat behind, native title bar and traffic lights intact
2. All three responsive tiers, resized live across both breakpoints
3. Welcome → model selection → Browse All → Review → download → Ready, end to end
4. All three Step 2 selection narratives (starter, trade-up, already-on-disk)
5. Cancel download, and the `Download stopped` recovery copy and routes
6. Download failure and retry, returning to the correct origin
7. Catalogue state restoration (search, filter, sort, scroll, selection) on Back
8. Missing engine and repeated Recheck
9. Low-disk warning — warn-only, primary still enabled
10. Relaunch with an incomplete setup → `Continue setup` → correct micro-stage

VoiceOver was not manually exercised; its contracts are covered by the deterministic accessibility regression tests added in #1946 and preserved here.

## Known unrelated test failure

`MarkdownLinkPerformanceTests` → *"Adjacent links keep their own hit targets"* fails. **This reproduces on clean `upstream/main` @ `fb9475c1`** in a fresh worktree with none of this branch's changes applied, so it is pre-existing and unrelated to onboarding.

## Deliberate deviations from Paper

Each of these is stated in the code, next to the thing that deviates.

1. **Medium-tier headline sizing.** Paper 05.1.D steps the headline down at medium as well as compact. Here it holds its desktop size through wide *and* medium and steps once, at compact. Continuously scaling the headline with window width made the same screen read as a different screen during a resize.
2. **True centring.** Paper's frames sit roughly 5% above true centre — an optical lift. This centres truly. The lift is not reproducible across three tiers without becoming a magic number per tier, which is the failure mode the shared canvas exists to prevent.
3. **State 07 is not implemented.** Its trigger overlaps state 04 in every real first run, so it would be unreachable.
4. **Missing engine keeps its current presentation ownership.** Moving it into the setup shell would change which surface owns the window, not how it looks — a behaviour change, out of scope here.

## Deferred

**`WON'T FIT` informational Review is not implemented.** Oversized catalogue rows remain dimmed and unselectable. They gained the Paper badge and the memory reason, which is visual, but opening their read-only Review would enable a new user action and belongs in its own behaviour PR.

## Scope

No Paper edits — the document was read only. No Settings, Chat, Images, Audio, Launch, About, Update, menu-bar, Dock, notification or global-shortcut surfaces touched. The diff is 9 files, all under `apps/rapid-mac`: 4 sources and 5 test files.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ContentView.swift
@@ -77,61 +77,13 @@ struct ContentView: View {
         // picker lives inline in the compose box (see ChatView) and the
         // model comes up on first send (implicit lifecycle). Search belongs
         // to the sidebar column beside macOS's native collapse control.
-        VStack(spacing: 0) {
-            // #1588: this recovery path existed since the app was introduced
-            // but was never mounted, so a failed Finder replacement was
-            // detected and then silently discarded.
-            FailedReplaceBanner()
-            NavigationSplitView {
-                SidebarView(
-                selection: $section,
-                chat: chat,
-                onNewChat: {
-                    chat.newConversation()
-                    section = .chat
-                },
-                onSearchChats: {
-                    showConversationSearch = true
-                },
-                onSelectConversation: { id in
-                    chat.selectConversation(id)
-                    section = .chat
-                },
-                server: server
-            )
-            // v1.0: the rail paints an explicit warm surface rather than
-            // inheriting the system sidebar material. The material is a
-            // cool translucent grey that fought the warm canvas beside
-            // it — the two planes read as belonging to different apps.
-            .background(RapidTheme.surfaceSidebar)
-            .navigationSplitViewColumnWidth(
-                min: SidebarView.columnMinWidth,
-                ideal: SidebarView.columnIdealWidth,
-                max: SidebarView.columnMaxWidth
-            )
-            } detail: {
-                detailArea
-                    .frame(maxWidth: .infinity, maxHeight: .infinity)
-                // Detail floor dropped 520 → 440 alongside the narrower
-                // rail, back when the window floor was 640 and the old pair
-                // (190 sidebar + 520 detail) over-committed it by 70pt —
-                // which is what forced horizontal clipping instead of
-                // graceful compression. The floor is 720 now, so this has
-                // 80pt of slack; it stays at 440 because it is the detail's
-                // own minimum, not a number derived from the window.
-                .frame(minWidth: 440, minHeight: Self.minWindowHeight)
-                    .background(RapidTheme.surfaceCanvas)
+        Group {
+            if quickstartVisible {
+                // Setup owns the window (Paper 05.1.A). See ``onboardingShell``.
+                onboardingShell
+            } else {
+                productionShell
             }
-            // Background pulls are process-wide, not chat-only.  Keep their
-            // progress visible whichever sidebar destination is selected.
-            DownloadStrip(downloads: downloads)
-            if showLogs {
-                LogDrawer(server: server)
-                    .frame(minHeight: 100, idealHeight: 150, maxHeight: 220)
-                    .accessibilityIdentifier("ContentView.LogDrawer")
-                    .transition(.move(edge: .bottom).combined(with: .opacity))
-            }
-            statusFooter
         }
         // `.windowResizability(.contentMinSize)` derives the window's floor
         // from this, so the shell states it once here rather than leaving it
@@ -243,14 +195,6 @@ struct ContentView: View {
         .sheet(isPresented: firstRunSheetPresented) {
             firstRunSheet
         }
-        // Presented from the split view — NOT from `mainArea` — so onboarding
-        // covers the whole window. Rendered inside the detail pane it left the
-        // sidebar exposed and clickable, which reads as "the app is already
-        // running, this panel is just one more surface" rather than "finish
-        // setup first".
-        .sheet(isPresented: quickstartSheetPresented) {
-            quickstartSheet
-        }
         .onChange(of: settingsRouter.quickstartReturnGeneration) { _, _ in
             quickstartDismissedThisSession = false
         }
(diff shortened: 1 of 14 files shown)


raullenchai (approved):
Reviewed the full Direction D onboarding redesign and the rebased integration. Fixed a live progress regression where measured bytes could still read PREPARING and disappear when an incorrect total was discarded. Added wide/medium/compact + Step 2 AX baselines to the existing fresh-install Golden Flow without changing CI hooks on the external PR.

Validation: 2,509 Swift tests; focused Python GUI contracts; local fresh-install/download/cached/browse/low-memory/crash journeys; hosted PR validation, full CI, rapid-mac build, and all 25 GUI Golden Flows are green.

PR #2106, opened 2026-08-19
fix(markdown): stop the code fence flickering while it streams

xiaoxiunique:
## Problem

A fence marker reaches the parser one keystroke before it *is* a fence. Compiling a growing prefix of

```
Here is code:\n\n```swift\nlet x = 1\nprint(x)\n```\n\nDone.
```

shows three separate flickers, all from that one cause (`T(n)` = text block of n chars, `C(n|lang)` = code block):

| prefix | blocks | what the reader sees |
|---|---|---|
| n=16 | `T(13) T(1)` | a lone backtick renders as its own text block |
| n=20 | `T(13) C(0\|sw)` | index 1 turns from text into code — SwiftUI swaps the whole representable at that slot |
| n=24 | `T(13) C(0\|swift)` | the language changes, re-running syntax highlighting |
| n=48 | `T(13) C(19\|swift)` | the closing backticks were being displayed *as code content* until they closed the fence |

The study that prompted this described one flicker; measuring found three, by a different mechanism than described.

## Fix

`MarkdownCompiler.withoutFormingFence` drops a trailing fence marker whose line is still being typed. The card then appears once, already knowing its language, and no stray backticks are ever shown inside it.

Scope is deliberately narrow:

- **Last line only, streaming only.** The settled row recompiles with `isComplete` true (`ChatView` swaps the streaming renderer for the settled renderer), so nothing stays hidden.
- **Real fences only.** A run of 3+ backticks (with or without an info string), or a line that is nothing but one or two backticks. An inline code span being typed — `` `foo `` — is left alone.

## Tests

6 cases in `StreamingFenceStabilityTests`, each break-verified by reintroducing the bug and confirming the specific test fails.

## Why this matters

This prevents code answers from visibly changing component type and losing apparent characters while tokens arrive, because a forming fence no longer reaches the Markdown parser before its line is complete. The assembled-app Golden Flow samples the live stream and guards that user-facing behavior.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Compile/MarkdownCompiler.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Compile/MarkdownCompiler.swift
@@ -25,6 +25,50 @@ struct MarkdownCompiler: Sendable {
         self.postProcessors = postProcessors
     }
 
+    /// Drop a trailing fence marker that is still being typed.
+    ///
+    /// Three separate flickers, all from the same cause — the parser is shown
+    /// a fence that is one keystroke old. Measured by compiling a growing
+    /// prefix of "Here is code:\n\n```swift\nlet x = 1\nprint(x)\n```\n\nDone.":
+    ///
+    ///     n=16   T(13) T(1)          a lone backtick renders as its own text block
+    ///     n=20   T(13) C(0|sw)       index 1 turns from text into code — SwiftUI
+    ///                                swaps the whole representable at that slot
+    ///     n=24   T(13) C(0|swift)    the language changes, re-running highlighting
+    ///     n=44   T(13) C(21|swift)
+    ///     n=48   T(13) C(19|swift)   the closing backticks were being displayed
+    ///                                as code content until they closed the fence
+    ///
+    /// Holding the marker back until its line is finished removes all three:
+    /// the code card appears once, already knowing its language, and no stray
+    /// backticks are ever shown inside it.
+    ///
+    /// Only the LAST line is considered, and only while streaming. The settled
+    /// row re-compiles the whole message with ``isComplete`` true (see
+    /// ``StreamingMarkdownStore/finish()``), so nothing stays hidden.
+    ///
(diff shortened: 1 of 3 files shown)


xiaoxiunique:
Updated after review. Three defects, all found by mutating the source and re-running the suite.

**The doc comment named the wrong guarantor.** It said nothing stays hidden because `StreamingMarkdownStore.finish()` re-compiles with `isComplete: true`. It does not — `finish()` flushes and then discards the result. The real guarantor is `ChatView` swapping to `TextKitMarkdownView(content:)` the moment a message leaves `.streaming`, which compiles the raw `message.content` with `isComplete` defaulting to true. The behaviour was right; the stated reason was wrong, which is worse than no comment.

**The info-string guard was over-broad.** It rejected whitespace, but CommonMark forbids only *backticks* in a backtick fence's info string. So ```` ```swift {highlight} ```` and ```` ```py file=a.py ```` were real fences that kept flickering — the fix missed exactly the info strings people write. The backtick half stays: ```` ``` a ``` ```` is a paragraph, not a fence.

**CRLF silently opted out.** Swift reads `\r\n` as one `Character`, which `lastIndex(of: "\n")` never matches, so the whole fix was a no-op on CRLF streams. Now searches on `isNewline`.

**Two tests were not testing what they claimed**, both proven:

- `inlineSpanIsUntouched` — all three inputs began with a non-backtick, so all three bailed at the `ticks.isEmpty` guard and never reached the span-vs-fence rule they were named for. Deleting that rule left the suite green. Added the four inputs that actually reach it.
- `settledRenderKeepsEverything` — compiled with `isComplete: true`, so it never ran the new code at all. It passed with the function body replaced by `return ""`. Replaced with a floor assertion on the *streaming* render, plus a mid-fence case that pins the `isComplete` gate itself.

The suite now kills every mutation tried: no-op, `return ""`, and each of the three guards deleted individually.

**Out of scope, stated plainly in the doc comment rather than left implied:** `~~~` fences and fences inside blockquotes still flicker. A tilde info string may itself contain backticks, so it needs its own validation rather than sharing this one.

PR #2107, opened 2026-08-19
feat(markdown): typeset inline math, and bridge the LaTeX models write

xiaoxiunique:
Independent of #2106 — both touch `MarkdownCompiler.swift`, but in different regions, so they merge in either order.

Three layers, all needed before `$x$` renders as mathematics.

## 1. Compile — the sentinel

Inline math was turned back into `$…$` text and handed to the markdown parser. That is exactly what running the segmenter *first* was meant to avoid: `_` is emphasis syntax, so the subscript in `$x_1$` is gone before any math code sees it.

It now carries a private-use sentinel through parsing and swaps back afterwards. Nothing in the sentinel is markdown-significant, so it survives as one contiguous piece of a single run — a formula keeps working inside `**bold**`, inside a list item, and inside a table cell, inheriting the styling and link of the run it came from.

## 2. Render — an image attachment, not a view

`NSTextAttachmentViewProvider` hosts its view inside an `NSTextView`. `MarkdownTextBlockView` draws fragments into its own context instead, so a hosted view would be created and never shown — the same constraint `TypingDotAttachment` documents. ChatGPT can take the view route because its message block *is* a text view; ours is not.

So the formula is rasterised once and carried as an image, which TextKit lays out and draws like any other attachment: it flows with the sentence, wraps with it, and needs nobody to chase its frame.

The cache is keyed on latex **+ point size + colour**. `LaTeXMarkdownView` records what happens when the colour is left out: MarkdownUI's inline image provider keyed on the parsed nodes, an appearance change never re-fired it, and since the glyph colour is baked into the bitmap the reader was left with black-on-black formulas after switching to Dark.

Two measured details, both with a precedent in this repo:
- `fittingSize`, not `intrinsicContentSize` — #131 established that `MTMathUILabel` overrides the former on macOS.
- `cacheDisplay(in:to:)` after `layoutSubtreeIfNeeded()`. Rendering the layer directly crashes *inside SwiftMath* (`MTMathUILabel.swift:238`, force-unwrapped display list) rather than returning an empty bitmap, because only `_layoutSubviews` builds that list.

## 3. Compatibility — why display math was broken

Reported separately: display math rendered as centred monospaced `$$…$$` source. The segmenter was **not** at fault — it recognises all three `$$` shapes (tight, inner spaces, inner newlines). The cause is one layer down: `\mod` is not in SwiftMath's table, and **one unknown command fails the whole formula**, so `MathView` falls back to showing the source.

Measured against 46 commands common in chat output, **15 fail**: `\mod \bmod \pmod`, `\begin{align}` `\begin{equation}` `\begin{array}`, `\operatorname`, `\boxed`, `\dots`, `\tag` `\notag`, one-column `\begin{cases}`, `\lVert`. Fixing only `\mod` would break again on the next message containing `\begin{align}`.

`LaTeXCompatibility` uses two mechanisms:

- **Registration**, through SwiftMath's own `add(latexSymbol:)` extension point. Preferred wherever it fits: the parser takes the longest run of letters after a backslash, so teaching it `mod` cannot misfire on `\models`, whereas a textual substitution would.
- **Source rewriting** for what registration cannot express — environment names, and commands that take an argument.

Every rewrite is lossy and each says what it gives up (`gather` loses centring, `array` loses column alignment, `\boxed` loses its rule). The original source is kept everywhere a person sees it: `MathView`'s fallback text, its accessibility label, and the inline renderer's prose fallback all show what the model actually wrote.

## Tests

`InlineMathCompileTests` (8), `InlineMathRenderTests` (5), `LaTeXCompatibilityTests` (13).

Each guard was break-verified — the bug reintroduced, the specific test confirmed to fail. That caught one of my own tests being decorative: the nested-environment case I first wrote had `&` in every row, so it never reached the guard and passed against an implementation with no guard at all. Replaced with `\begin{cases} \begin{matrix} 1 \\ 2 \end{matrix} & a \\ 0 & b \end{cases}`, a formula that already renders and that the unguarded code would break.

Two guards that measurement justified after the fact:
- Without the command-name boundary check, `\operatornamewithlimits{argmax}` is silently rewritten to `\mathrm{w}ithlimits{argmax}` — different mathematics, no error.
- Without the escaped-brace skip, `\boxed{\left\{ x \right. }` is left unwrapped and falls back to source.

## Not verified

Rendering was verified at the parse level (SwiftMath accepts the bridged source) and by unit test. I have not visually confirmed the typeset output in a running app.

Pre-existing on `main`, unrelated: `MarkdownLinkPerformanceTests` "Adjacent links" and `WebToolsHardeningTests` NXDOMAIN both fail on a clean checkout of the base commit (checked in an isolated worktree).

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/Markdown/LaTeXCompatibility.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/Markdown/LaTeXCompatibility.swift
@@ -169,10 +169,19 @@ enum LaTeXCompatibility {
     // MARK: - Commands that take an argument
 
     private static func rewritingArgumentCommands(_ source: String) -> String {
+        // `\left`/`\right` resolve their operand through SwiftMath's separate
+        // `delimiters` table, which is a `let` with no extension point — so
+        // registering `\lVert` above does nothing for the `\left\lVert …
+        // \right\rVert` spelling, which is the one written for any norm taller
+        // than a single symbol. Substituting the plain double bar is the only
+        // route, and loses the same asymmetric spacing the registration does.
+        var source = source
+            .replacingOccurrences(of: "\\left\\lVert", with: "\\left\\|")
+            .replacingOccurrences(of: "\\right\\rVert", with: "\\right\\|")
         // `\operatorname` sets an upright multi-letter name and adds operator
         // spacing around it. `\mathrm` gives the upright letters; the spacing
         // is lost.
-        var source = rewritingCommand(source, "operatorname") { "\\mathrm{\($0)}" }
+        source = rewritingCommand(source, "operatorname") { "\\mathrm{\($0)}" }
         // The box around a model's final answer is decoration. Losing the
         // rule keeps the answer.
         source = rewritingCommand(source, "boxed") { $0 }
--- a/apps/rapid-mac/Sources/Rapid/UI/Markdown/LaTeXCompatibility.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/Markdown/LaTeXCompatibility.swift
@@ -1,0 +1,311 @@
+import Foundation
+import SwiftMath
+
+/// Bridges the LaTeX models write to the subset SwiftMath parses.
+///
+/// ``MathView`` falls back to showing the raw `$$…$$` source in monospace
+/// when SwiftMath cannot parse a body — deliberately, so the reader is not
+/// left with a hole. But a model writing ordinary number theory trips that
+/// fallback on its first line: `\mod` is not in SwiftMath's table, and one
+/// unknown command fails the whole formula, not just the token. Measured
+/// against 46 commands common in chat output, 15 failed.
+///
+/// Two mechanisms, chosen per command:
+///
+/// * **Registration**, via SwiftMath's own `add(latexSymbol:)` extension
+///   point, for commands that are simply missing from its table. Preferred
+///   wherever it fits, because the parser reads the longest run of letters
+///   after a backslash as the command name — teaching it `mod` cannot
+///   misfire on `\models`, whereas a textual substitution would.
+/// * **Source rewriting**, for anything registration cannot express: a
+///   command that takes an argument, or an environment name.
+///
+/// Every rewrite is lossy in some way and each one says what it gives up.
+/// The bar is that the reader sees typeset mathematics that means what the
+/// model meant, not that the output is pixel-identical to LaTeX's.
+///
+/// The original source is *not* replaced anywhere it is shown to a person:
+/// ``MathView``'s fallback text and accessibility label, and the inline
+/// renderer's prose fallback, all keep what the model actually wrote.
+@MainActor
+enum LaTeXCompatibility {
+
+    /// The body to hand SwiftMath. Returns `latex` unchanged when nothing
+    /// applies, which is the common case.
+    static func normalized(_ latex: String) -> String {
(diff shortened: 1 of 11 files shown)


xiaoxiunique:
Updated after review. Two of these were reachable from message content.

**Crash (fixed).** `Int("-1")` parses and `-1 < latex.count` is true, so a literal sentinel carrying a negative index reached `latex[slot]` and trapped — process death in the renderer. Reproduced (`Fatal error: Index out of range`, signal 5) with:

```
Value $x$ and \u{E000}-1\u{E001} tail
```

`U+E000` was a poor choice: it is the first Nerd Font Pomicons codepoint, so it arrives in pasted terminal output and in model text about glyphs. Three changes rather than one, because the bounds check alone would still have left the substitution bug below:

- bounds check is now `latex.indices.contains(slot)`
- the sentinel moved to plane 15, which no font ships glyphs in
- any sentinel surviving into the source is dropped before segmentation

**Content substitution (fixed).** A well-formed literal sentinel was read as ours, and the reader's own text replaced by whichever formula sat at that index. Closed by the stripping pass above.

**Black-on-black in Dark (fixed).** The key held the `NSColor`. The app renders with `.labelColor`/`.textColor` — dynamic catalog colours that compare equal *and hash equal* in every appearance — so the Light bitmap was served straight back in Dark. That is the exact `LaTeXMarkdownView` failure the key was added to prevent, reproduced by the key meant to prevent it. It now holds the resolved components. The existing test could not have caught it: `.black` vs `.white` differ under either key. Added one using `.labelColor` under `performAsCurrentDrawingAppearance`.

**Sentinels in code blocks (fixed).** `LaTeXSegmenter` skips backtick fences; swift-markdown also recognises `~~~`, which it does not. Fixed in `restoringInlineMath` rather than in the segmenter, so it repairs any construct the two disagree about, not just tilde fences.

**`\left\lVert … \right\rVert` (fixed).** `\left`/`\right` resolve through SwiftMath's separate `delimiters` table — a `let` with no extension point — so registering `\lVert` never reached the spelling models actually write for a norm. Substituted instead.

**Decorative tests.** Both confirmed by mutation and removed or corrected: the `normalized("\dotsb")` assertion cannot fail for an implementation that does no textual substitution, and `unparseableDeclines` does not cover the `label.error` guard because the size check declines first. Also corrected the claim on `InlineMathAttachment.latex` that fading and hit testing consult it — nothing does.

Every fix was break-verified. One note on method: after moving the sentinel to plane 15 and adding the stripping pass, the bounds check became unreachable through `compile`, so it was code no test could fail. Rather than drop a guard against process death, the new test drives `expandingSentinels` and `restoringSourceSpelling` directly — restoring `slot < latex.count` now crashes the test run.

**Accepted, not fixed** — each is real, none is a regression from this PR:

- Inline math does not participate in the streaming fade. `TextFadeAnimator` fades via `.foregroundColor` over a range, which an image attachment ignores, so a formula pops in at full opacity. Cosmetic and visible on every streamed formula; wants its own change.
- `~~$x$~~` loses its strike — the math branch returns before the strikethrough and inline-code styling.
- Registration overwrites SwiftMath's reverse symbol map (`U+2026` now maps to `dots` rather than `ldots`). Only `latexSymbolName(for:)` reads it and nothing calls that, so the effect today is nil.
- An unclosed inline code span (`` The var `$x$ is set ``) now typesets where `main` left it literal — the segmenter's pre-existing behaviour for an unterminated span, surfaced by this change rather than caused by it.

PR #2110, opened 2026-08-19
fix(launch): harden agent configuration targets and authentication

osdodo:
## What does this PR do?

  This PR validates and fixes the Launch page integrations that were exercised in this branch.

  - Updates Cline through the config files it actually reads, covering both legacy and next bundle layouts.
  - Adds an OpenHands adapter that updates encrypted provider settings through the running Agent Canvas API.
  - Replaces Cursor's ineffective config writer with an accurate manual setup guide.
  - Removes the obsolete Continue.dev launch integration.
  - Propagates the active `RAPID_MLX_API_KEY` into agent guides and configs instead of rendering `not-needed`.
  - Covers Aider, LangChain, OpenCode, Qwen Code, Kilo Code, PydanticAI, Smolagents, and DeepSeek Harness authentication.
  - Adds safe config merging, preflight validation, backups, atomic writes, restrictive credential permissions, and shell quoting.
  - Updates the macOS Launch page, CLI output, documentation, and regression coverage.

  This PR intentionally covers only the agents validated here. Remaining Launch page agents will be validated in follow-up PRs.

  ## Why is this needed?

  Several Launch page actions reported success but did not produce a usable agent configuration:

  - Cline was being pointed at stale or incomplete configuration locations.
  - Cursor ignored the generated `cursor.aiprovider.*` settings because they are not part of Cursor's real configuration schema.
  - OpenHands stores its API key encrypted, so directly writing its settings file cannot create a credential the application can read.
  - Agent guides rendered `not-needed` even though the desktop server requires its per-launch bearer, causing immediate HTTP 401 responses.
  - A malformed later Cline config could leave earlier files partially updated.
  - Generated shell commands exposed credentials through an intermediate `/usr/bin/env` argv and did not consistently quote dynamic values.

  The changes make the displayed commands match the configuration and authentication mechanisms used by the current agent versions.

  ## AI assistance disclosure

  Codex assisted with implementation and review across `vllm_mlx/launch/`, `vllm_mlx/agents/`, the CLI/ integration registry, the macOS Launch page, related tests, and documentation.

  I reviewed the resulting behavior and manually validated OpenHands with Agent Canvas using Rapid-MLX on port 8000 and Agent Canvas on port 8010. Codex also ran the targeted Python and Swift regression suites, lint checks, and directly inspected the generated DeepSeek Harness guide.

  ## Test plan

  - Ran 196 targeted Python tests covering launch adapters, agent config merging, first-class setup, CLI connection output, DeepSeek Harness, and first-run guides.
  - Ran 14 Swift tests covering Launch page commands, integration catalog decoding, and integration ordering.
  - Ran Ruff against the changed launch, agent, integration, and test modules.
  - Ran `git diff --check`.
  - Manually validated OpenHands through a running Agent Canvas instance.
  - Verified that the DeepSeek Harness guide emits a correctly shell-quoted `RAPID_MLX_API_KEY`.


raullenchai (changes requested):
Thanks for doing the real-client validation here. The underlying problems are worth fixing — several current Launch actions can report success while leaving an unusable configuration — but this PR is not safe for the imminent release in its current form. It changes 41 files across credential handling, multi-file config writes, CLI compatibility, the macOS Launch surface, and multiple agent profiles. The repository PR gate currently reports **DO NOT MERGE** with the blocking findings below.

### Blocking

1. **Do not send the OpenHands session credential to an arbitrary `OPENHANDS_URL`.** `vllm_mlx/launch/openhands.py::_base_url()` accepts the override verbatim, then `_is_openhands()` / the PATCH request attach the local `X-Session-API-Key`. Validate the URL before attaching the key (at minimum loopback-only for the default trust model; any remote mode needs an explicit secure policy and HTTPS). Add negative tests for non-loopback HTTP/HTTPS URLs, userinfo, malformed URLs, redirects, and redirect targets. Redirects must not forward the credential off-origin.

2. **Make the Cline three-file update actually transactional.** Parsing all files first only prevents parse-time partial writes. `_write_legacy()` replaces `globalState.json` and then `secrets.json`; `_write_next()` replaces `providers.json` afterwards. Permission, disk, fsync, or replace failure can leave a mixed configuration while the command reports failure. Stage and validate all replacements before commit, then rollback every already-replaced target if any commit fails. Test injected failures at each replacement boundary, including restoration of prior contents and permissions.

3. **Preserve authenticated OpenHands setup when users follow the printed guide later.** `agents/profiles/openhands.yaml` prints `rapid-mlx launch openhands ...` without carrying `{api_key}` / `RAPID_MLX_API_KEY`. The Launch-page assignment is scoped only to the guide-rendering process, so re-running the printed command writes the default `sk-noop` and receives 401 from the desktop server. Render a safely shell-quoted key-bearing command or provide a secure copy/action flow, and cover the actual two-step UX in a regression test.

4. **Do not remove a released CLI command without a compatibility path.** Removing `continue-dev` from `ADAPTERS` changes the previously documented `rapid-mlx launch continue-dev` into an unknown-client error. Keep a deprecated alias for at least one release that exits with a clear migration message to Kilo Code, and test both direct invocation and `launch list` behavior.

5. **Make the Cursor manual flow actionable without weakening secret handling.** The generated guide says “paste your RAPID_MLX_API_KEY” but does not identify how a CLI user obtains it; in the GUI the key exists in the separate masked row. Clarify both entry points and test the rendered guide. Do not simply print an unmasked long-lived credential to stdout.

### Release recommendation

Please target this for the next release after the security and rollback behavior is fixed and dogfooded against fresh and existing Cline/OpenHands installs. I recommend splitting the compatibility-policy changes (removing Continue, demoting Cursor) from the credential/config correctness fixes so each can be reviewed and reverted independently. The direction is useful; the current blast radius and confirmed failure modes are too high for a release-bound merge.

PR #2112, opened 2026-08-19
feat(chat): name conversations, and offer what to ask next

xiaoxiunique:
Two cheap background completions at the end of a turn — the fourth item on the UX list.

**Titles.** The sidebar row is the first user message truncated to 42 characters. For "explain how euler's theorem works" that is a label you cannot pick out of a list of ten. After the first exchange the model now names the conversation instead — once, and never again.

**Follow-ups.** After every settled answer, three short questions the reader might ask next, as chips under that answer. Tapping one sends it.

Both are always on; neither is a setting.

## Titles need their own flag, not `hasCustomTitle`

`persistActive` re-derives the title from the first user turn on *every* save, so anything else needs a gate or the next streamed token reverts it — the regression `hasCustomTitle` was added for.

Reusing that flag was the obvious move and is wrong. It means the **user** owns the name. A generated title carrying it would be indistinguishable from a rename, and the generator could no longer tell "the reader named this, leave it" from "I named this, leave it". So `hasGeneratedTitle` is separate, and a rename still wins outright in both orderings.

## The parsers are generous going in, strict coming out

A small local model asked for a title returns it labelled, fenced, quoted, numbered, or buried in a paragraph. The first four are stripped. The last is refused: truncating prose to 42 characters reads worse than the reader's own opening words, which is what the row already shows.

Three rules beyond the obvious, each defensible on its own and not merely convenient:

- A markdown table row is a row whatever else it is.
- A trailing colon means the model wrote a lead-in to a list and stopped — "Three things, in order:" names nothing.
- More than one sentence is the answer, not a name for it. This is crude and takes `Dr. Who and the daleks` with it; that is an accepted loss, because a false reject keeps the derived title and a false accept puts a paragraph in the sidebar.

For follow-ups, **requiring every line to end in a question mark** is the highest-leverage decision here. One predicate removes preambles, markdown table rows, code, numbered-list residue and prose paragraphs, while compliant output passes by construction. Fewer than three survivors renders nothing — one lonely chip reads as a bug, and the whole feature is garnish.

## The rail holds its height before it has anything to show

`TranscriptScrollPositionProbe` follows any document growth while the reader is pinned, and its release valve `releaseIfAnswerOutgrewViewport` is gated on `isStreaming` — already false by the time a settled-turn footer could appear. Chips arriving a second after the answer would therefore yank a pinned reader with nothing to stop them, and the same growth flips `isPinnedToBottom` for a frame, popping the jump-to-bottom button and popping it back.

So the rail mounts at its final height in the same layout pass that adds the stats caption and the action row, and the chips fade in *inside* that reservation. Filling it changes no height, posts no frame notification, and cannot move anybody. **The probe needs no changes at all** — the hazard is gone by construction rather than guarded against.

That is also why the row scrolls sideways with `.lineLimit(1)` instead of flowing: a flowing layout makes the height depend on the text, and the text comes from a model.

## Not every lane will batch us

Skipped when the engine reports a non-text modality. The `--mllm` lane runs one request at a time, so a background call there is not batched alongside the reader's turn — it is queued in front of their next one.

The signal is `ModelResidencySnapshot.modality`, which is the engine's own answer per alias. `ModelBrandStyle.modelType(forAlias:)` is a second, name-based classifier already used to decide whether image parts go on the wire, but it treats the whole `qwen3.5-` and `gemma3-` families as vision — gating on it would have killed the feature on the most common models.

Two server constraints are satisfied by construction rather than by discipline: `ChatStreamClient.Request` cannot express a stop sequence, and the engine's batch-compatibility key is `(frozenset(stop_token_ids), bool(ignore_eos))` — a mismatch requeues behind the whole running batch. `enableThinking: false` is mandatory for the other reason `SamplingConfig` documents: thinking on spends the entire 24-token budget on a reasoning trace and returns empty content.

## Failure is silent, everywhere

No retry, no banner, no log the reader sees — the contract `ServerProfileFetcher` states for its own background fetch, for the same reason: there is nothing they could do about it. Every rejection leaves the derived title and an empty rail.

Cancelled on send, on conversation switch, on delete, on stop, and on termination. The five existing `inflight?.cancel(); inflight = nil` pairs are folded into one `cancelInflightWork()` — strictly less code than a registry, and it removes the "somebody added a third task and missed site four" class.

## Streaming, not a one-shot

`stream: true` stays hardcoded. A `stream: false` path would be the first desktop code to use the server's non-streaming branch, and `scripts/fake-rapid-mlx.sh` answers every chat request with `Content-Type: text/event-stream` and never reads the `stream` field — so a JSON-expecting client would decode-fail in the one harness that can drive this end to end, and fall back silently forever.

## Tests

41 new. Every guard was break-verified by reintroducing the defect and confirming the specific test fails: the title gate, both ownership guards, the one-shot guard, the follow-up anchor check, the `"Stopped."` clause, and the modality check.

**On golden-flow baselines.** The plan for this change asserted zero baseline churn on the grounds that the fake server's replies all fail the parsers. That reasoning was wrong — it checked two of the seven shapes. Measured, four of seven produced a title before the three quality rules above were added, and one (`math`) still does. `BackgroundAssistFakeServerTests` now pins exactly what each fake shape parses to, so a parser change that would move a baseline fails there first, with a name, instead of surfacing as a mystery diff.

Ran `chat-depth`, `math-rendering`, `message-actions` and `slow-stream-stop`: all pass, no baseline changes. `chat-restore` fails on `search-results`, but **it fails identically on a clean `main` build** (checked in a separate worktree) — the search panel moves in the AX tree, unrelated to this. `all` needs peekaboo, which is not installed here.

## Not in this change

- The fake server cannot produce a title or three questions on demand, so the harness cannot exercise the happy path. Teaching it a `shape:followups` marker plus a `flow_follow_ups` that presses `ChatView.FollowUp.0` introduces new baselines rather than mutating existing ones, which is a better second PR than a bigger first one.
- Chips do not participate in the streaming fade (`TextFadeAnimator` works on rendering attributes; these are not text).
- Identifiers are `ChatView.FollowUp.<index>`, deliberately **not** `ChatView.Message.` — `gui-golden-flows.sh`'s `transcript_only` scopes its slice by that prefix, and a chip inside it would freeze model-authored text into a snapshot.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Chat/BackgroundCompletionClient.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/BackgroundCompletionClient.swift
@@ -110,7 +110,18 @@ struct BackgroundCompletionClient {
         defer { timeout.cancel() }
 
         do {
-            try await work.value
+            // `work` is unstructured, so it does not inherit cancellation from
+            // whoever is awaiting it, and `work.value` is not itself a
+            // cancellation-aware suspension point. Without this handler,
+            // cancelling the caller left the request running for the whole
+            // deadline — measured at 6.3s against a server that never
+            // answered, holding an engine admission slot in front of the turn
+            // the reader had just asked for.
+            try await withTaskCancellationHandler {
+                try await work.value
+            } onCancel: {
+                work.cancel()
+            }
         } catch {
             return nil
         }
--- a/apps/rapid-mac/Sources/Rapid/Chat/BackgroundCompletionClient.swift
+++ b/apps/rapid-mac/Sources/Rapid/Chat/BackgroundCompletionClient.swift
@@ -1,0 +1,128 @@
+import Foundation
+
+/// One completion, taken for the app's own purposes rather than the reader's.
+///
+/// Both background features — naming a conversation and proposing follow-up
+/// questions — need the same thing: ask the model something short, get a
+/// string back, and if anything at all goes wrong, get nothing back and say
+/// nothing about it. That contract is the whole type.
+///
+/// ## Why it streams
+///
+/// The wire body's `stream` is hardcoded `true` (``ChatStreamClient``), and a
+/// one-shot `stream: false` path would be the first desktop code to use the
+/// server's non-streaming branch — untested here, and worse, invisible to the
+/// one harness that can drive this end to end: `scripts/fake-rapid-mlx.sh`
+/// answers every chat request with `Content-Type: text/event-stream` and never
+/// reads the `stream` field at all, so a JSON-expecting client would decode-fail
+/// in the golden flows and silently fall back forever. Accumulating deltas also
+/// inherits what ``ChatStreamClient/send(_:bearerToken:onEvent:)`` already
+/// owns: bearer injection, the pre-stream retry, HTTP-status bodies drained
+/// into a readable error, the SSE line cap, `[DONE]`, and both cancellation
+/// shapes. For a ≤96-token reply the framing overhead is noise.
+///
+/// ## Why it can never be loud
+///
+/// There is no error surface, no retry, and no log the reader sees — the same
+/// contract ``ServerProfileFetcher`` states for its own background fetch:
+/// there is no UI affordance for "the title call failed" because there is
+/// nothing the reader could do about it. Every failure is `nil`, and every
+/// caller treats `nil` as "leave things as they are".
+@MainActor
+struct BackgroundCompletionClient {
+
+    /// Where to send, resolved at call time.
+    ///
+    /// Never cached: ``ServerManager/activeBearer`` rotates on every `start()`
+    /// and is cleared on stop and on crash, so a handle held across a restart
+    /// is a guaranteed 401. The alias comes from the `.ready(alias:)` state
+    /// for the same reason it must not be chosen freely — naming a model that
+    /// is not resident would issue a real `/v1/models/load`, and with the
+    /// assistant replacement group that evicts the model the reader is using.
+    struct Target {
+        let port: Int
+        let bearer: String?
+        let alias: String
+    }
+
+    /// Test seam. Production leaves this nil and takes
+    /// ``ChatStreamClient``'s shared ephemeral session.
+    var session: URLSession?
+
+    /// The reply text, or nil.
+    ///
+    /// `deadline` is a ceiling on the whole call, separate from the client's
+    /// inactivity timeout: a background request that is merely slow is one we
+    /// would rather abandon than let sit in the engine's batch.
+    func complete(
+        _ messages: [ChatMessage],
+        target: Target,
+        maxTokens: Int,
+        temperature: Double,
+        topP: Double = 0.9,
+        deadline: Duration
+    ) async -> String? {
+        var client = ChatStreamClient(
+            baseURL: ChatStreamClient.loopbackURL(port: target.port),
+            session: session
+        )
+        client.requestTimeout = 20
+
+        let request = ChatStreamClient.Request(
+            alias: target.alias,
+            messages: messages,
+            temperature: temperature,
+            topP: topP,
+            maxTokens: maxTokens,
+            // No tools: a background call that advertises them can finish
+            // with `tool_calls` and no content at all.
+            tools: nil,
+            // Mandatory. On a hybrid model with thinking on, the reasoning
+            // trace alone exhausts a 24-token budget and the reply comes back
+            // `finish_reason: "length"` with empty content — the failure
+            // ``SamplingConfig`` documents. Both features would be
+            // permanently dead on Qwen 3.x and its relatives.
+            enableThinking: false,
+            forcedTool: nil
+        )
+        // Note what is NOT here: any stop sequence. ``Request`` cannot express
+        // one, which is the point — the engine's batch-compatibility key is
+        // `(frozenset(stop_token_ids), bool(ignore_eos))`, and a request whose
+        // key differs from the live batch is requeued until every running
+        // request drains. Being unable to set it is a stronger guarantee than
+        // remembering not to.
+
+        let sink = TextSink()
+        // Two tasks racing rather than a task group: the send closure
+        // captures a main-actor client, which a group's `sending` parameter
+        // will not accept. Cancelling the work closes the socket, which the
+        // engine's disconnect guard turns into an abort that frees its
+        // admission slot — so an abandoned call stops costing anything.
+        let work = Task { @MainActor in
+            try await client.send(request, bearerToken: target.bearer) { event in
+                if case .content(let delta) = event { sink.text += delta }
+            }
+        }
+        let timeout = Task {
+            try await Task.sleep(for: deadline)
+            work.cancel()
+        }
+        defer { timeout.cancel() }
+
+        do {
+            try await work.value
+        } catch {
+            return nil
+        }
+
+        let text = sink.text.trimmingCharacters(in: .whitespacesAndNewlines)
+        return text.isEmpty ? nil : text
+    }
+
+    /// The accumulator has to be a reference: ``ChatStreamClient/send``'s
+    /// handler is an escaping `@MainActor` closure, which cannot capture a
+    /// mutable local.
+    @MainActor private final class TextSink {
+        var text = ""
+    }
+}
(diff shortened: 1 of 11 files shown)


raullenchai (changes requested):
Blocking UI-state leak in the follow-up failure path:

scheduleBackgroundAssist sets followUpAnchorID before starting the request, and ChatView mounts FollowUpSuggestionRail whenever that anchor still matches the last message. The rail deliberately reserves a fixed nonzero height even in .idle.

However, publishFollowUps clears only followUp when the reply is nil or parses to no questions; it does not clear followUpAnchorID. Thus a timeout/network failure (nil), empty completion, or ordinary unparsable model output leaves a permanent blank spacer below the final answer. Cancellation after the rail enters .pending can strand it similarly because the task exits without publishing.

Please clear the anchor whenever no chips will be shown (including cancellation), and add assertions for followUpAnchorID == nil in the nil/unparseable/cancel paths. The current test named “Nothing parseable leaves the rail empty” only checks the enum and therefore misses the rendered blank rail.

xiaoxiunique:
Rebased onto current `main` (no conflicts — nothing upstream touched `ChatViewModel`, `ChatView` or `ConversationHistory`), and force-pushed with the follow-up fix that was missing here.

**The reported defect.** Sending "hi" titled the conversation `3 to 6 words`.

The instruction was "Reply with ONLY the title: 3 to 6 words, no quotes…", and the bundled 1.2B model returned the constraint as the answer — repeatably, along with `Chat conversations` for the other half of the sentence. Measured against that model rather than reasoned about: a model this size lifts a noun phrase out of the instruction and hands it back, so the instruction must not contain one that could pass for an answer. The wording now uses a verb where it had a noun, and the length constraint is off the sentence that says what to output.

Same three exchanges, three samples each, before and after:

| input | before | after |
|---|---|---|
| 解释一下欧拉定理 | `3` `3` `3` | a correct Chinese noun phrase |
| write a python function to reverse a list | `3` `3` `3` | `Function reverse` |
| how do I deploy a docker container | `3` `3` `3` | `Deploy command` |

Two parser rules sit behind it. Words that name the container rather than the contents — "Chat", "Conversation", "Topic" — are refused, because "Chat" in a list of chats says less than the reader's own "hi" does; a bare greeting returns exactly that on every sample, so a greeting now keeps its derived title. And a title that is a length specification is refused *in its general form*, since a denylist of echoed phrases only holds until the instruction is reworded again.

**A defect never measured at all.** Follow-ups had their language instruction obeyed on 10 of 20 turns, so a reader writing Chinese regularly got three English chips. A `CRITICAL:` clause took it to 12 of 20; adding "in the user's voice" collapsed it to 8 of 20 parseable and 0 of 8 correct. Prompting does not settle this, so the last word is a deterministic script match — suggestions in a different writing system from the conversation are dropped, which is the failure this feature already uses everywhere else. Roughly one Chinese turn in three now shows no chips rather than wrong ones.

**Also dropped:** the Chinese literals this landed with — the refusal prefixes, the label, the generic-title words. None had ever been observed; the model answers a Chinese greeting with English `Chat`. Removing speculative branches and keeping the source in English were the same edit. Four tests that covered only those branches went with them.

xiaoxiunique:
Rebased onto `main` (was 25 commits behind).

The rebase reported no conflicts but did not compile: `main` has since added `branches`, `activeLeafID` and `branchChoices` to `ChatConversation.CodingKeys`, and this branch's `hasGeneratedTitle` case was dropped without a conflict marker. Restored, along with its line in the hand-written `encode(to:)` — that one would not have failed to compile, only failed to persist. `The flag round-trips` covers it; verified by deleting the line and watching it go red.

Full suite on the rebased branch: 2726 tests, no failures other than a pre-existing DNS test that needs network.

xiaoxiunique:
**1. Clearing the anchor on nil / unparseable output — `05568f9e`.**

`publishFollowUps` routes every failure through `clearFollowUps()`, which clears the anchor as well as the state: anchor no longer matching the last message, a nil reply, and output that parses to no questions.

**2. Including cancellation — `348fe3ee`.**

Of the six `cancelInflightWork()` call sites, three cleared the rail and three did not: `send` cleared `followUp` but left the anchor; `stop()` and `stopAndPersist()` cleared neither. The assist's own continuation clears both, but only once it gets to run — nothing was synchronous with the cancellation itself.

The clear is now folded into `cancelInflightWork()` and the three explicit calls are gone. That is the argument the helper was created for, stated in its own doc comment — *"the class of bug this removes is somebody added a third task and missed site four"*. A fourth thing to remember at six call sites is the same bug with a different noun.

**3. Assertions — `05568f9e` for nil/unparseable, `348fe3ee` for cancellation.**

`Nothing parseable puts the rail away` asserts `followUpAnchorID == nil` and starts from `.ready` rather than `.idle`; the previous version asserted `.idle` on a model that was already `.idle`, so it passed with the entire body of `publishFollowUps` replaced by a no-op.

`cancellingClearsTheRail` covers `stop`, `stopAndPersist` and `newConversation` as three separate routes, because a single-route test cannot see three-of-six coverage. Verified by removing the folded call: all three fail, including `newConversation`, whose explicit call this commit deletes.

Full suite: 2727 tests, no failures other than a pre-existing DNS test that needs network.

raullenchai:

xiaoxiunique:
Rebased onto `87344238`. Head is `219b225d`; the PR reports `MERGEABLE`.

## Conflicts

Two, both in the first commit.

**`ChatViewModel.swift` — `send()`.** The incoming side carried the app-side forced-tool routing (`forcedToolForUserTurn`, `webSearchQuery`, `forcedWebSearchQuery` threaded through `beginAssistantTurn`). That code is not part of this PR — it arrived with the old base from #1694, and #2244 (`refactor(mac): use native schema-driven tool routing`) has since removed it. Dropped rather than reapplied. What this PR contributes at that point is two lines, `cancelInflightWork()` and `followUp = .idle`, and those are what the resolution keeps.

`beginAssistantTurn` therefore keeps `main`'s signature unchanged — no `forcedWebSearchQuery` parameter.

**`ChatView.swift` — `sendSuggestion(_:)`.** Kept. Two adjustments to match current `main`:

- the branch's `private var supportsImageInput: Bool { ModelBrandStyle.supportsImageInput(forAlias: alias) }` is dropped; `ChatView` now has a catalog-backed stored `supportsImageInput`, and the chip passes it through to `send`
- `isImportingFiles` moved onto `attachmentDraft`; the guard reads `attachmentDraft.isImportingFiles`, as the four neighbouring call sites do

## Validation

| | result |
|---|---|
| `BackgroundAssistGatingTests\|BackgroundAssistParsingTests` | 56 tests, green on three consecutive runs |
| `Chat\|Conversation\|FollowUp\|Title\|ToolKit` | 280 tests, green |
| full suite | 3028 tests, 6 distinct failures |

The six: four `DMGPresentationScriptTests` (`pyexpat` failing to load on a Homebrew Python 3.14), `The injected clock/time-zone pins today's local date, time, and zone` (`TimeZone.abbreviation(for:)` returns `GMT-7` where the test expects `PDT`; `DateFormatter` with `zzz` still returns `PDT`), and `ModelCatalog terminates catalog subprocesses when the async task is cancelled`. The first five reproduce on a clean `main`; the last passes on its own three times out of three and only appears under full-suite parallelism.

## One thing left as it was

Commits 1 and 2 do not build in isolation: `hasGeneratedTitle` is used in commit 1 but its `CodingKeys` entry lands in commit 3. That split predates this rebase — before it, commit 1 failed with 63 errors against current `main`. I have not reordered the hunks; say the word if you would rather each commit stood alone.

raullenchai (approved):
Initial review complete across all four commits. `clearFollowUps()` folded into `cancelInflightWork()` fixes the blocking anchor leak from the prior review. `hasGeneratedTitle` separate from `hasCustomTitle` is the right call. Modality-based lane gating correctly replaces name-based classification. Script-matching for CJK is measured against the bundled model. 41 tests with break-verified guards. No blocking issues remaining. Ready for PR validation and Codex review.

raullenchai:
@Mergifyio queue

raullenchai:
@mergifyio queue

raullenchai:
Superseded by repository-owned integration PR #2698 because the managed queue intentionally excludes fork heads. The integration preserves all four original commits and their authorship, rebases them onto current main, and records fresh focused and full Swift validation. Further review and queue status will continue on #2698; thank you for the contribution and the detailed measured evidence.

raullenchai:
Superseded by #2698, which preserved this PR’s user-facing conversation-title and follow-up-suggestion intent, rebased it onto current `main`, passed independent review and the full shared Mac/GUI merge candidate, and merged as 36ef0530.

PR #2130, opened 2026-08-19
fix(responses): re-attach MCP namespace to function_call for Codex dispatch

### LINKED ISSUE #2114: Codex namespace flattening loses MCP dispatch identity in Responses function calls
> **Bug Report template:** This issue was submitted through `gh`, so GitHub did not attach the issue-form label. All required Bug Report fields are included below.

### Rapid-MLX version

0.12.15

Unknown. This appears to be an incomplete round-trip in the namespace flattening added by #993 rather than a model regression.

### Hardware

Mac Studio, Apple M3 Ultra, 256 GB unified memory

### macOS version

macOS 26.5.1 (25F80)

### Python version

Python 3.12.13

### Model

PocketAiHub/Qwen3.8-27B-Abliterated-MTPLX-Optimized-Speed

### Full serve command

```shell
cd <LOCAL_MODEL_DIRECTORY>
rapid-mlx serve . --host 127.0.0.1 --port 8002 --served-model-name rapid-qwen38-27b --reasoning-parser qwen3 --tool-call-parser qwen3 --gpu-memory-utilization 0.35 --hybrid-cache-entries 12 --resident-memory-limit-gb 80 --speculative-config '{"method":"mtp","model":"./mtp.safetensors","num_speculative_tokens":3}'
```

This is the exact invocation; only the user-specific absolute model directory is redacted. All behavior-affecting flags and relative file locations are shown.

Client: Codex CLI 0.147.0 using the Responses API.

Streaming (`stream: true`)

Codex groups some function tools using its internal Responses wire shape:

```json
{
  "type": "namespace",
  "name": "mcp__example",
  "tools": [
    {"type": "function", "name": "lookup", "description": "...", "parameters": {"type": "object"}}
  ]
}
```

`normalize_responses_tool_types()` accepts this shape by flattening the namespace children into top-level functions. However, the namespace identity is discarded. When the model calls the function, Rapid-MLX emits a bare Responses item:

```json
{"type":"function_call","name":"lookup","arguments":"{...}","call_id":"..."}
```

Codex needs the originating namespace in order to dispatch the call to the correct MCP server:

```json
{"type":"function_call","namespace":"mcp__example","name":"lookup","arguments":"{...}","call_id":"..."}
```

Without it, Codex receives `lookup` as an unsupported unqualified call instead of routing it to `mcp__example`.

This is not specific to a particular MCP server or tool name. MCP tool names only need to be unique within one server, so two namespaces may legally contain the same child name. Flattening without a reversible mapping loses dispatch identity and can also introduce silent name collisions.

PR #993 fixed request rejection for Codex namespace groups, but its end-to-end check covered Codex's built-in `multi_agent_v1` functions, whose bare names can be recognized by the orchestrator. Arbitrary MCP namespaces expose the missing reverse mapping.

### Minimal reproduction

1. Start Rapid-MLX with the command above.
2. Send a Codex-shaped Responses request directly, without requiring an MCP transport:

   ```shell
   curl -N http://127.0.0.1:8002/v1/responses \
     -d '{
       "model": "rapid-qwen38-27b",
       "stream": true,
       "input": "Call lookup with q=ping. You must use the provided tool.",
       "tool_choice": "required",
       "tools": [
         {
           "type": "namespace",
           "name": "mcp__example",
           "description": "Example MCP namespace",
           "tools": [
             {
               "type": "function",
               "name": "lookup",
               "description": "Look up a value",
               "parameters": {
                 "type": "object",
                 "properties": {"q": {"type": "string"}},
                 "required": ["q"],
                 "additionalProperties": false
               },
               "strict": true
             }
           ]
         }
       ]
     }'
   ```

3. Inspect the streamed `function_call` item. Rapid-MLX emits `name: "lookup"` but no `namespace`.
4. In a real Codex 0.147.0 session with any MCP server exposing that tool, the same response produces `unsupported call: lookup` rather than an MCP `tools/call` request.

Actual response shape:

```json
{"type":"function_call","name":"lookup","arguments":"{\"q\":\"ping\"}","call_id":"..."}
```

Expected Codex-compatible response shape:

```json
{"type":"function_call","namespace":"mcp__example","name":"lookup","arguments":"{\"q\":\"ping\"}","call_id":"..."}
```

For the collision case, send two groups that both contain `name: "lookup"`:

```json
[
  {
    "type": "namespace",
    "name": "mcp__alpha",
    "tools": [{"type": "function", "name": "lookup", "description": "Alpha lookup", "parameters": {"type": "object"}}]
  },
  {
    "type": "namespace",
    "name": "mcp__beta",
    "tools": [{"type": "function", "name": "lookup", "description": "Beta lookup", "parameters": {"type": "object"}}]
  }
]
```

After the current `flattened.extend(sub_tools)` normalization, both engine-visible tools have the same name and neither result can be routed unambiguously.

### Suggested implementation

Please preserve namespace identity across the entire request/response lifecycle instead of attempting to infer it from the bare function name after generation.

One general approach:

1. While converting each namespace child for the engine, assign a request-local, collision-free synthetic name.
2. Store a request-local mapping:

   ```python
   engine_name -> (namespace_name, original_function_name)
   ```

3. Use the synthetic name in the flattened engine tool schema. This handles two namespaces containing the same child name, and a top-level function colliding with a namespace child.
4. When parsing tool calls, decode through that mapping and emit the original `name` plus an optional `namespace` field on the Responses `function_call` item.
5. Omit `namespace` for ordinary top-level function tools so public Responses behavior remains unchanged.

This likely requires:

- changing namespace normalization/conversion to return or carry request-scoped identity metadata rather than only mutating `tools`;
- adding `namespace: Optional[str]` to the internal Responses function-call output model;
- applying the reverse mapping in both streaming and non-streaming output builders before serialization.

Using a plain `dict[function_name, namespace]` is not sufficient because duplicate child names are valid across namespaces. The engine-visible alias must itself be unique.

Suggested regression coverage:

- one namespaced function round-trips with `namespace` restored;
- two namespaces with the same child name remain distinct;
- a top-level function and namespace child with the same name remain distinct;
- namespace restoration works for both streaming and non-streaming Responses;
- multiple/parallel tool calls retain the correct request-local mapping;
- ordinary top-level functions remain unchanged;
- the existing `multi_agent_v1` and hosted-tool filtering tests from #993 continue to pass.

```shell
# Reproduced on Rapid-MLX 0.12.15 with the curl command above.
# The response.created payload already shows the namespace flattened away:
"tools": [{"type":"function","name":"lookup", ...}]

# The completed call still has no namespace:
event: response.output_item.done
data: {"type":"response.output_item.done","output_index":0,"item":{"type":"function_call","id":"...","call_id":"...","name":"lookup","arguments":"{\"q\": \"ping\"}","status":"completed"},"sequence_number":5}

# Codex-side tool output
unsupported call: lookup
```

Related: #993

raullenchai:
## Why
Codex CLI submits MCP tools grouped under `{"type":"namespace","name":"mcp__x","tools":[{"type":"function",...}]}` entries. `normalize_responses_tool_types` flattens each namespace into its child function tools so the allowlist gate accepts the request (PR #993), **but it discarded the namespace identity**. When the model then called the function, Rapid-MLX emitted a bare `function_call` (`name:"lookup"`, no `namespace`), so Codex could not route the call back to the originating MCP server and reported `unsupported call: lookup`. Closes #2114.

## What
- `normalize_responses_tool_types` now **returns** a `{function_name: namespace}` mapping, built during the existing Pass-1 flatten (still mutates `tools` in place).
- The route captures it and threads it through **both** output paths: non-stream (`_non_stream` → `openai_to_responses` → `_build_tool_call_output_item`) and stream (retry wrapper → `_stream_responses` → the inline emitter, both `added` and `done` items).
- `ResponsesOutputItem` gains optional `namespace: str | None = None`, omitted under `exclude_none`.
- Name resolution centralised in `_flatten_tool_name`, tolerant of malformed input (non-dict `function`, non-string `name`) so bad shapes reach the 400 validation path instead of 500-ing.

## Collision safety
MCP names are unique only within one server. A bare name is attached **only** when it resolves to exactly one namespace; collisions omit the attachment, degrading to the pre-#2114 unqualified shape rather than mis-routing.

## Tests
Mapping build + all collision cases + malformed-shape hardening; non-stream emission; streaming `added`+`done` items via the fake-engine route harness. 15 assertions fail on pre-fix source and pass after; no-regression guards hold both ways. 108 responses adapter+SSE tests pass; ruff clean.

## Notes
Real user report on 0.12.15 (M3 Ultra) with a standalone `/v1/responses` reproduction. No version bump.

### DIFF
--- a/tests/test_responses_adapter.py
+++ b/tests/test_responses_adapter.py
@@ -21,6 +21,7 @@ from vllm_mlx.api.models import (
     Usage,
 )
 from vllm_mlx.api.responses_adapter import (
+    _build_tool_call_output_item,
     _convert_status,
     _convert_text_format,
     _convert_tool_choice,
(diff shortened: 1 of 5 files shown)


raullenchai:
Merging via the required `tests` context (SUCCESS on the latest SHA). The non-required `validate` check is red due to a **CI-side pr_validate environment fault, not a regression**:

- Its scorecard lists **134 responses tests as "pre-existing (also fail on main)"** — unmodified main code failing en masse in that runner, which is only possible if the CI validate env itself is degraded. The same scorecard shows `codex_review skip (codex CLI not found)`, `diff_coverage skip (diff-cover not installed)`, and `test_env_check: 2 missing plugin(s) recovered` — a degraded toolchain.
- The 2 flagged "regressions" (`TestResponsesStreamNamespaceReattachment`) **pass cleanly locally** (verbose: 2 passed) and in the full suite.

Authoritative green signals: local `pr_validate` = **MERGE-SAFE** (codex 0 blocking, full_unit **17896 passed / 0 failed**, lint clean, 100% diff coverage), and the required apple-silicon `tests` matrix = **SUCCESS** on this SHA. `validate` is not a required context; only `tests` gates merge here.

PR #2131, opened 2026-08-19
feat(mac): open WON'T FIT rows to a read-only Review

loriz-art:
Implements the behaviour deliberately deferred from #1946 and #2063, plus one small window-default change that unblocks the layout #2063 shipped.

---

## 1 · WON'T FIT informational Review

A catalogue row this Mac cannot run carried `.disabled(!available)`. That blocked the invalid action — correctly — by making the whole row take no click, which also meant the user could see `WON'T FIT` and the memory reason and had **no way to ask why**, or to read the model's actual figures. Paper 05.2.D · `V3/Onb-2e-ReviewDownload-IncompatibleMemory` is explicit that this is the wrong trade:

> Opening the detail of a WON'T FIT row is allowed and informational — the user asked what this model is, and refusing to answer would be worse than answering. The primary is disabled, not hidden, so the shape of the screen never changes between a model that can start and one that cannot.

So the row is a live control again, and the refusal moved to the primary of the screen it opens.

The change is in the **pure derivation**, not in the views. `OnboardingModelSelection.primary` gained one branch that answers differently by context:

| Context | Incompatible selection |
|---|---|
| shortlist / catalogue | `Review download` · **enabled** · action `.reviewIncompatible` |
| review | `Download & start` (or `Start existing model` if cached) · **disabled** |

The label stays `Review download` from a list so the catalogue's control never relabels itself between rows — Paper's stability rule. Inside Review the verb is the one the model *would* have taken, greyed, because there the control's job is to name what is being withheld.

Review's incompatible variant reuses the Direction D shell exactly: the flat subtitle *"This model cannot run on this Mac."*, an error-tint callout, the `Memory when loaded` row recoloured, and one added `Usable for a model` row. Same composition, same fact table, same footer lane — Paper keeps the shape identical between a model that can start and one that cannot, because the user compared them from the same list a moment ago. There is **no "Free memory and retry"**: only the Step 4 pre-load guard has a live `MemoryProbe` reading to act on.

**One sentence goes beyond Paper's copy, deliberately.** Paper's callout names the usable pool, but the ceiling `ModelSizing.classify` actually enforces is 75% of that pool. On a 32 GB Mac a 21 GB model is refused while the same screen calls 25.6 GB usable — which reads as a contradiction. The footnote now names the real limit, sourced from a new `ModelSizing.largestFittingGB(on:)` derived from the band itself so the two cannot drift apart. A test fails if the band changes without it.

Two tensions inside Paper, resolved and stated rather than silently picked:

- 05.2.C says WON'T FIT "is the only non-selectable state"; 05.2.D says opening its detail is allowed. The 05.2.D behaviour note is the later, specific decision.
- 05.2.D's own companion caption still reads "Rows that will not fit stay listed and stay unselectable". Not reproduced in the product — it would now be false.

## 2 · Safety invariant: incompatible models never download or start

This is the property the `.disabled` existed to protect, and it is now stated **more strongly** than the `.disabled` ever stated it.

Before, it held because the row swallowed clicks. Now it holds because the one derivation every input funnels through cannot produce a commit for such a row, in any context. `Action` gained an `isCommit` property; only `.startExisting` and `.downloadAndStart` are commits, and `activatePrimary` re-derives and refuses a disabled primary before it switches. So "this can never become a download" is a property of the type rather than a check repeated at every call site.

Concretely: no path from a WON'T FIT row reaches `DownloadManager`, `DiskSpaceProbe`, or `ServerManager.start`. Covered by a matrix over `{cached, uncached} × {shortlist, catalogue, review}`, plus a source guard that the execution route sits behind the commit arm.

Click, double-click and Return all resolve through the same `activatePrimary` path — no hidden bypass was added. Single click still only selects and records the scroll anchor.

## 3 · Cached but incompatible

**Incompatibility outranks cached-ness**, and the branch is ordered to make that structural rather than incidental. A model already on disk that cannot be loaded is still a model that cannot be loaded — having been downloaded is not evidence about memory, and `startExisting` is the shorter route into exactly the same `ServerManager` load.

So a cached incompatible row keeps its `WON'T FIT` badge (which replaces `ON THIS MAC` rather than stacking with it), and Review shows a **disabled `Start existing model`** — not a download it does not need, and not an enabled start.

## 4 · Accessibility

- New identifiers `Quickstart.Review.Incompatible` and `Quickstart.Review.UsableMemory`, pinned in the inventory gate — along with the five pre-existing Review fact identifiers that had never been pinned.
- The disabled primary carries a hint: *"Unavailable. Needs ≈ 45.7 GB of memory. This Mac has 32 GB, of which roughly 25.6 GB is usable for a model."* macOS announces a disabled control as "dimmed" and stops there; here the why is the entire point of the screen.
- The unrunnable row carries a hint stating the verdict and that opening it downloads nothing. The label already carries alias, memory reason and badge.
- Runnable rows get **no** hint. No single sentence is true of both a cached row (which starts) and an uncached one (which opens Review), and their behaviour is the norm the rest of the list establishes.
- Keyboard reachability comes from the row being an enabled `Button` again — a disabled one is out of the focus chain. There is no arrow-key list navigation in the catalogue for any row; that is pre-existing and unchanged.
- The default hint is empty rather than absent, matching the existing `identifier ?? ""` pattern in `OnboardingFactTable`. The AX dumper skips empty help, so the golden-flow baselines are unaffected.

## 5 · Fresh-install window default: 1200×820 → 1440×900

Direction D enters its two-column layout at `OnboardingD.columnsMinWidth == 1290`, so the old default sat **90pt below its own breakpoint**: every first run met Step 2 in the medium *stacked* tier, and the composition Paper 05.1.A specifies was in practice unreachable without a manual resize.

This is a **default and nothing more**. It applies only where SwiftUI has no frame to restore — `AppDelegate.attachMainWindow` sets the `Rapid.MainWindow` autosave name, which restores a saved frame synchronously and therefore still wins for every returning user. Nothing resizes the window when onboarding appears or changes state, no AppKit resizing machinery was added, `columnsMinWidth` is untouched, and narrowing still reaches the medium and compact tiers exactly as before. Frame clamping is unchanged.

The number lives in a named `MainWindowDefaults` rather than inline, because it is not a taste choice — it exists to clear the breakpoint — and the test should assert the value the scene actually uses. A source guard stops a future edit hardcoding a width back inline and leaving the test passing against a number nothing reads.

## 6 · Verification

**Automated**, rebased onto `upstream/main` @ `123e5d6f`:

- Debug build — clean
- Release build — clean
- `OnboardingWontFitReviewTests` (new) — **24 / 24**
- `OnboardingDirectionDTests` — **46 / 46**
- `OnboardingRecoveryBehaviorTests` + `Step2ModelSelectionBehaviorTests` — **106 / 106**
- Accessibility identifier gate — **16 / 16**
- `verify-recommendation-tiers.swift` — ALL PASS
- `verify-conversation-ordering.sh` — ok
- `verify-chat-timeout.swift` — OK
- Full suite — **2582 tests in 225 suites, 1 failure** (see §7)

Three existing tests asserted the old contract and were **rewritten rather than deleted**. Each now states the claim that actually mattered — "never reaches a commit" — instead of the mechanism that used to enforce it.

**Live manual pass**, on an isolated dogfood build (unique bundle id, throwaway `HOME`, isolated Application Support and model cache, production preferences untouched):

1. Uncached WON'T FIT row → selectable → informational Review ✓
2. Review clearly explains why the model cannot run ✓
3. Disabled primary cannot download or start ✓
4. Back preserves Browse All query, filter, sort, scroll anchor and selection ✓
5. Cached-but-incompatible row remains WON'T FIT ✓
6. Cached incompatible Review shows the disabled `Start existing model` ✓
7. Click and Return on that disabled primary cause no progress, no server start, no navigation ✓
8. Compatible model flow unchanged ✓
9. Fresh install opens at 1440×900 in the wide two-column layout ✓
10. A previously saved narrower frame still wins ✓
11. Light, Dark, and all three responsive tiers ✓

Items 9 and 10 were also measured programmatically via `CGWindowListCopyWindowInfo`: a fresh bundle id with no `Rapid.MainWindow` key opened at exactly 1440×900; rewriting that key to a 1000×760 frame and relaunching restored 1000×760, not the default.

## 7 · Known unrelated failure

`MarkdownLinkPerformanceTests` → *"Adjacent links keep their own hit targets"* fails in the full suite.

**This reproduces on clean `upstream/main` @ `123e5d6f`** in a fresh worktree with none of this branch's changes applied. It is pre-existing and unrelated to onboarding, and is deliberately not touched here.

## Scope

11 files, all under `apps/rapid-mac` — 6 sources and 5 tests. No Paper edits; the document was read only. No Settings, Chat, Images, Audio, Launch, menu-bar, Dock or update surfaces touched. No generated files, screenshots, personas or preference data.


raullenchai (approved):
Reviewed the behavior and safety invariant on current main. The incompatible path only navigates to read-only Review; download/start remain behind a disabled commit action. Verified clean integration with current main, 24/24 WON’T FIT tests, 46/46 Direction D tests, and reran the unrelated timing-sensitive full-suite failures successfully. No blocking findings.

PR #2139, opened 2026-08-19
fix(cli): size the agents-list name column to the widest alias

### LINKED ISSUE #2138: agents: 16-char alias name overflows the name column and skews the table
## Found via dogfood (release 0.12.16)

`rapid-mlx agents` renders a ragged row for `deepseek-harness`:

```
  name            client               GitHub  tools  recommended models
  claude-code     Claude Code            140K  FC     ...
  deepseek-harness DeepSeek Harness         1K  FC     ...   <- client/GitHub/tools shifted 1 col right
```

## Root cause

`cli.py` formats the name column with a hardcoded `{p.name:<15}` (header and rows). `deepseek-harness` is **16** chars — one past the width — so Python's `<15` neither truncates nor pads it. The name eats its own separator space and every later column on that row shifts one character right. Any alias name >= 15 chars trips this.

## Fix

Size the name column to the widest alias (`max(15, max(len(name)))`), floored at 15, and use it in both the header and the rows. Auto-accommodates any future long name instead of bumping the constant.

raullenchai:
## Why

Found while dogfooding release 0.12.16. `rapid-mlx agents` renders a ragged row for the 16-char alias `deepseek-harness`:

```
  name            client               GitHub  tools  recommended models
  claude-code     Claude Code            140K  FC     ...
  deepseek-harness DeepSeek Harness         1K  FC     ...   <- shifted 1 col right
```

The name column used a hardcoded `{p.name:<15}` (header + rows). Because `deepseek-harness` is one char past 15, Python's `<15` neither truncates nor pads it — the name eats its separator space and every later column on that row slides one character right. Any alias name ≥ 15 chars trips this.

Closes #2138.

## What

- `cli.py` `agents_command`: compute `name_w = max(15, max(len(p.name) for p in profiles))` once and use `{…:<{name_w}}` for both the header and each row. Floored at 15 so short rosters keep the familiar layout; grows automatically for any future long alias instead of needing the constant bumped.

## Tests

- New `tests/test_agents_list_alignment.py`: renders a short + a 16-char name and asserts the client column starts at the same offset on the header and both rows. Confirmed it **fails** on pre-fix source (offset 18 vs 19) and **passes** after.
- 54 agents-adjacent tests (`test_agent_first_class_setup`, `test_deepseek_harness_profile`, `test_first_run_guide`, new file) green.

## Notes

Human-facing table formatting only — no profile data or command behavior changes. AI-authored (Claude Opus 4.8), human-reviewed.

### DIFF
--- a/tests/test_agents_list_alignment.py
+++ b/tests/test_agents_list_alignment.py
@@ -1,0 +1,53 @@
+"""Regression: the `rapid-mlx agents` roster must keep its columns aligned
+even when an agent name is as wide as (or wider than) the name column's old
+hardcoded 15-char width.
+
+`deepseek-harness` is 16 chars. With the pre-fix `{p.name:<15}` the name
+overflowed its field, consumed the single separator space, and shifted the
+client / GitHub / tools columns one character right *on that row only* — a
+ragged table. The column is now sized to the widest alias, so every row lines
+up under the header.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+from vllm_mlx import cli
+
+
+def _profile(name: str, display: str) -> SimpleNamespace:
+    return SimpleNamespace(
+        name=name,
+        display_name=display,
+        stars=1000,
+        needs_function_calling=True,
+        recommended_models=["qwen3.5-9b-4bit"],
+        kind="agent",
+    )
+
+
+def test_agents_list_client_column_aligns_for_max_width_name(monkeypatch, capsys):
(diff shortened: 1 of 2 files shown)


raullenchai:
Addressed the codex NIT: the header divider was still fixed at 78 chars while the table grows with `name_w`. Now derived as `78 + name_w - 15` so the rule tracks the widened columns. Alignment test + lint green; re-running validation on the final diff.

PR #2144, opened 2026-08-20
feat(perf): blocked-seq Metal kernel for GDN prompt prefill

raullenchai:
## Why

Qwen3.5/3.6/3.8 hybrid checkpoints run their linear-attention (Gated DeltaNet) prefill scan through mlx-lm's stock kernel, which re-reads the same k/q rows from device memory once per `Dv/4`-slice threadgroup — ~32x redundant traffic at long chunk lengths. A blocked-sequential restructuring (threadgroup-staged k/q/v token blocks, register-resident state, `simd_shuffle_down` reductions, no barriers in the hot loop) computes the **exact same recurrence** materially faster at the op level.

Measured per-layer-call (Qwen3.8-27B GDN geometry `Hk=16 Hv=48 Dk=Dv=128`, bf16, M3 Ultra):

| T | stock | this kernel | speedup | state rel-err |
|---|---|---|---|---|
| 2048 | 3.93ms | 2.67ms | 1.47x | 8.7e-08 |
| 4096 | 8.05ms | 5.11ms | 1.57x | 9.5e-08 |
| 8192 | 18.00ms | 9.99ms | 1.80x | 9.4e-08 |
| 16384 | 42.36ms | 19.89ms | 2.13x | 7.0e-08 |

**Honest end-to-end framing**: because chunked prefill caps each scan call at `prefill_step_size` (2048) and quantized matmuls dominate total prefill on these models, the server-level TTFT win is modest — measured **-1.2%** on Qwen3.8-27B (6K-token prompt, 18.86s→18.63s; same at 16K) and **-2.6%** on Qwen3.6-35B-A3B (2.80s→2.73s), medians of 3-4 runs, cache-busted, isolated single-server protocol on M3 Ultra. This is a real, exact, zero-risk op-level improvement, not a TTFT transformation.

Kernel design adapted from the oMLX project (https://github.com/jundot/omlx, Apache-2.0), which ships it default-on for the same architectures.

## What

- `vllm_mlx/gdn_prefill.py` — the Metal kernel (`mx.fast.metal_kernel` JIT, no build step) + `install()` that wraps `mlx_lm.models.gated_delta.gated_delta_kernel` (the module-global dispatch all GDN GPU paths funnel through). Fast path is strictly shape-gated: scalar gating, prompt ≥64 tokens, `Dk == 128`, `Dv % 32 == 0`, fp32 state, no mask — everything else falls through to the stock kernel unchanged. `RAPID_MLX_GDN_PREFILL=0` opts out.
- `vllm_mlx/scheduler.py` — call `install()` at module import, next to the existing `ensure_mamba_support()` patch precedent.

## Tests

`tests/test_gdn_prefill_kernel.py` (14 tests, Metal-gated):
- numerics vs the stock Metal kernel (bf16, T=64/100/512) and vs the pure-ops reference (fp32 end-to-end — pins the algorithm with no rounding slack)
- chunked-prefill state-chaining invariant (two chained 64-token calls == one 128-token call)
- every fallback gate (masked / vectorized gating / decode step / wrong Dk / non-fp32 state)
- `install()` idempotency, env opt-out, and wrapper fallback-to-stock for ineligible calls

All 14 pass on M2 Pro; kernel verified on M3 Ultra via the microbench above. Decode (T=1) provably unaffected: E2E decode tok/s unchanged (36.1 vs 36.4 within noise).

## Notes

- The MLLM vision lane uses mlx-vlm's own qwen3_5 copy and is intentionally untouched; this covers the default text lane.
- AI-authored (Claude Opus 4.8) as part of an overnight cross-engine benchmarking session (rapid-mlx vs mlx-lm vs omlx on Studio M3 Ultra); benchmark artifacts retained on mac-studio `~/work/benchout/`.

### DIFF
--- a/tests/test_gdn_prefill_kernel.py
+++ b/tests/test_gdn_prefill_kernel.py
@@ -1,0 +1,318 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Correctness gates for the blocked-seq GDN prefill kernel.
+
+The kernel computes the exact same sequential recurrence as mlx-lm's stock
+``gated_delta_kernel`` — any numeric divergence beyond fp32-accumulation
+noise is a bug, so these tests compare against both the stock Metal kernel
+and the pure-ops reference on random Qwen-shaped inputs. Fallback gates are
+exercised so masked / vectorized-gating / decode-step calls provably keep
+the stock path.
+"""
+
+from __future__ import annotations
+
+import importlib
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+if not mx.metal.is_available():  # pragma: no cover - CI runners have Metal
+    pytest.skip("Metal GPU required for GDN kernel tests", allow_module_level=True)
+
+from mlx_lm.models import gated_delta as gd
+
+from vllm_mlx import gdn_prefill
+
+
+def _stock_kernel():
+    """The unwrapped mlx-lm kernel, regardless of whether install() ran.
+
+    Another test module importing ``vllm_mlx.scheduler`` installs the
+    wrapper at collection time; comparing against the module attribute
+    would then be fast-path-versus-fast-path. The wrapper carries the
+    true stock implementation on ``._stock``.
+    """
+    fn = gd.gated_delta_kernel
+    return getattr(fn, "_stock", fn)
+
+
+# Qwen3.8-27B GDN geometry (scaled-down head counts, same head dims).
+B, HK, HV, DK, DV = 1, 2, 6, 128, 128
+
+
+def _inputs(t_len, dtype=mx.bfloat16, seed=7, g_ndim=3):
+    mx.random.seed(seed)
+    q = (mx.random.normal((B, t_len, HK, DK)) * 0.1).astype(dtype)
+    k = (mx.random.normal((B, t_len, HK, DK)) * 0.1).astype(dtype)
+    v = (mx.random.normal((B, t_len, HV, DV)) * 0.1).astype(dtype)
+    if g_ndim == 3:
+        g = mx.sigmoid(mx.random.normal((B, t_len, HV)).astype(mx.float32))
+    else:
+        g = mx.sigmoid(mx.random.normal((B, t_len, HV, DK)).astype(mx.float32))
+    beta = mx.sigmoid(mx.random.normal((B, t_len, HV)).astype(mx.float32))
+    state = mx.zeros((B, HV, DV, DK), dtype=mx.float32)
+    mx.eval(q, k, v, g, beta, state)
+    return q, k, v, g, beta, state
+
+
+def _max_rel_err(a: mx.array, b: mx.array, atol: float = 1e-3) -> float:
+    """Elementwise stabilized relative error (max over elements).
+
+    Dividing the global max error by the global max magnitude would let a
+    small-magnitude element be badly corrupted while the metric stays
+    tiny; normalizing per element (with an absolute floor at the data's
+    noise scale — inputs are O(0.1..1)) catches that: real corruption is
+    O(1) under this metric, while fp32 accumulation-order noise on
+    near-zero elements stays under 1e-3.
+    """
+    a32 = a.astype(mx.float32)
+    b32 = b.astype(mx.float32)
+    rel = mx.abs(a32 - b32) / (mx.abs(a32) + atol)
+    return float(rel.max())
+
+
+class TestBlockedSeqNumerics:
+    @pytest.mark.parametrize("t_len", [64, 100, 512])
+    def test_matches_stock_kernel_bf16(self, t_len):
+        q, k, v, g, beta, state = _inputs(t_len)
+        y_fast, st_fast = gdn_prefill.gated_delta_blocked_seq(q, k, v, g, beta, state)
+        y_stock, st_stock = _stock_kernel()(q, k, v, g, beta, state, None)
+        mx.eval(y_fast, st_fast, y_stock, st_stock)
+        # Both kernels accumulate in fp32; outputs round to bf16 so allow
+        # one-ulp-of-bf16 divergence on y, near-exactness on the fp32 state.
+        assert _max_rel_err(y_stock, y_fast) < 2e-2
+        assert _max_rel_err(st_stock, st_fast) < 1e-3
+
+    def test_matches_ops_reference_fp32(self):
+        # fp32 end-to-end pins the algorithm itself (no rounding slack).
+        q, k, v, g, beta, state = _inputs(96, dtype=mx.float32)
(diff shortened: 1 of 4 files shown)


raullenchai:
Addressed both review findings: (1) `_eligible()` now treats `state=None` as eligible (the fast path zero-inits fp32 state exactly like the stock update path) with wrapper-path and gate tests; (2) `gated_delta_blocked_seq()` validates `Dk == 128` / `Dv % 32 == 0` and raises `ValueError` instead of silently mis-computing, with tests for both rejects. Also registered `RAPID_MLX_GDN_PREFILL` in `ALLOWED_RAPID_MLX_ENV_VARS` (non-routing kernel-selection toggle, same shape as `RAPID_MLX_DISABLE_FUSED_SAMPLER`). 19 tests green. Note: full_unit's other 3 failures (`cohere2_moe` shim, North-Mini-Code manifest/profile) reproduce on pristine main — pre-existing from the upstream vendor commits, not this diff.

raullenchai:
Round-2 findings addressed: (1) numeric tests now resolve the true unwrapped kernel via a `._stock` attribute the wrapper carries — verified by running the suite with `vllm_mlx.scheduler` imported first (wrapper pre-installed), all 19 green; install() also detects an already-bound wrapper and never double-wraps (new regression test). (2) install() retrieves `gated_delta_kernel` via `getattr(..., None)` and degrades to no-fast-path instead of breaking scheduler import on older mlx-lm. (3) `_eligible` now pins the full kernel layout contract — `k.shape == q.shape`, single InT dtype across q/k/v, `beta.shape == g.shape`, matching [B,T] — before taking the fast path.

raullenchai:
Round-3 findings: (2) and (3) addressed — `_eligible` now requires the GQA head map invariant (`Hk > 0`, `Hv >= Hk`, `Hv % Hk == 0`) and, for a provided state, the exact `[B, Hv, Dv, Dk]` fp32 layout the kernel's pointer arithmetic assumes; both covered by new fallback tests (21 total, all green). Finding (1) — `T` as a kernel input being a pointer misread as a scalar — is refuted empirically: `mx.fast.metal_kernel` declares 0-d scalar inputs as `constant T&` references, so bare `T` in the source is a valid scalar; the numeric suite exercises this on real Metal hardware including `t_len=100` (a non-multiple of the 32-token block, so the `min(TB, T - t0)` partial-block path) and chunk-chaining at T=64/128 — all byte-level assertions pass on both M2 Pro and M3 Ultra. This is also the exact input convention the upstream oMLX kernel ships in production. Formatting applied (ruff format).

raullenchai:
Round-4 findings all addressed: (1) the decode_step fallback test now truncates every tensor to one token so the minimum-token gate alone determines the fallback; (2) `gated_delta_blocked_seq()` validates a caller-provided state's fp32 dtype and exact `[B, Hv, Dv, Dk]` shape and raises `ValueError` (new reject test); (3) `_eligible` requires `g.shape == (B, T, Hv)` exactly (beta already pinned equal to g). 22 tests green, lint/format clean.

raullenchai:
Round-5 findings addressed with a terminal consolidation: (1) `gated_delta_blocked_seq()` is now the single validation authority for the kernel's full layout contract — k mirrors q (shape+dtype), v is [B,T,Hv,Dv] in q's dtype, Hv a positive multiple of Hk, g/beta exactly [B,T,Hv], fp32 state of exact shape — every violation raises `ValueError`; `_eligible` mirrors the same invariants purely for stock-fallback routing. This closes the validated-where category entirely: no call path can reach the Metal launch with an unchecked layout. (2) install tests now snapshot/restore the gdn_prefill module globals (`_installed`/`_original_kernel`) alongside the patched function via an autouse fixture — verified order-independent by running TestInstall before the numeric classes and with `vllm_mlx.scheduler` pre-imported (wrapper pre-installed); all 22 green in every ordering.

raullenchai:
Round-6 addressed: (1) the wrapper now catches any fast-path exception (e.g. a Metal JIT rejection on an exotic GPU), logs it, and permanently degrades to the stock kernel — a broken fast path can never crash a request; covered by a new test that injects a failure and verifies both the immediate fallback and that the fast path stays dead afterward. (2) [NIT] `_max_rel_err` is now an elementwise stabilized relative error with an absolute floor at the data's noise scale (atol=1e-3 for O(0.1..1) tensors) — real corruption reads O(1) under this metric while fp32 accumulation-order noise on near-zero elements stays under the thresholds; tolerances recalibrated coherently (fp32/state <1e-3, bf16 y <2e-2). 23 tests green in all three orderings.

raullenchai:
Round-7 addressed: (1) install() now eagerly compiles AND executes both kernel variants (bf16/TB=32, fp32/TB=16) on tiny probe inputs — with `mx.eval` inside the try — before rebinding anything; a Metal rejection keeps the stock kernel bound and returns False, so lazy-eval failures can no longer escape into a production request (the runtime catch remains as belt-and-braces). (2) the degradation test now saves the real fast-path function before injecting the failure, restores it behind a call counter, and asserts the counter stays zero — proving `fast_path_dead` persisted rather than a second exception merely falling through. 23 tests green.

raullenchai:
## Review-convergence disposition (8 rounds)

Across 8 adversarial review rounds this PR accumulated **17 findings: 15 fixed, 1 refuted with hardware evidence** (the `T` scalar-input claim — `mx.fast.metal_kernel` declares 0-d inputs as `constant T&`; verified by green numeric tests at non-block-aligned lengths on two machines), and round-8's 2 findings are recycled escalations of the already-addressed lazy-eval class. Disposition for those two, with rationale:

**"Probe doesn't cover every production specialization"** — partially adopted: the probe now runs the actual flagship geometry (`Hk=16, Hv=48, Dv=128`) for both dtypes with `mx.eval` inside the try. The residual demand (probe *every* possible Hk/Hv/Dv) is declined on technical grounds: specializations of the same `(TB, InT)` variant differ only in integer address constants — threadgroup-memory and register footprint depend solely on `Dk` (fixed 128), `DB` (fixed 32), `TB`, and `InT`, all exercised by the probe — so a specialization cannot fail compilation when its probed variant succeeded. The runtime catch-and-permanently-degrade guard remains as belt-and-braces.

**"Force evaluation inside the wrapper"** — declined: an `mx.eval` per GDN layer call would insert a synchronization barrier into the prefill hot path, destroying the lazy pipelining the kernel exists to accelerate — the fix would cost more than the (already twice-mitigated) risk.

Full-unit status: the only failures are `test_healthz_perf_budget` (known CPU-contention flake, passes isolated) and 3 pre-existing failures reproduced on pristine main (`cohere2_moe` shim, North-Mini-Code manifest/profile — introduced by the upstream vendor commits, tracked separately). Everything else in pr_validate is green; 23 kernel tests pass in three orderings, including with the wrapper pre-installed.

Landing per the documented stopping rule for non-converging reviews on large diffs.

PR #2147, opened 2026-08-20
feat(mac): keep replaced answers as switchable branches [skip-version-bump]

osdodo:
## What does this PR do?

  Regenerate, Retry and editing a prompt no longer discard what they replace.
  Each appends a sibling under the same parent, and a `‹ 2/3 ›` control on the bubble steps between the alternatives.

  Mechanics:

  - `ChatMessage` gains a single `parentID`. Siblings are derived by scanning rather than mirrored in a `childrenIDs` array, so the two halves of a bidirectional link cannot drift apart. Sibling order comes from `createdAt`.
  - New `MessageTree` holds all the tree arithmetic. Every entry point is total: cycles, orphans and dangling leaf pointers degrade to a readable transcript rather than trapping, on the same principle as `ChatMessage.Role.unknown`.
  - `ChatViewModel.messages` stays the linear visible **path**, with off-path nodes in `branchedAway`, re-joined only at persist time. This is what leaves the ~74 streaming call sites that address messages by index untouched.
  - `ChatConversation` splits storage: `messages` keeps holding the active path, alternatives go in a new `branches` key.

  Also in scope, because both are only reachable once branches exist:

  - **Delete message** — takes the whole subtree, and the confirmation quotes a turn count that includes branches the user cannot currently see.
  - **Per-fork position memory** — returning to a branch lands where you left it instead of jumping to its deepest tip.

  ## Why is this needed?

  Regenerating an answer destroys the previous one, with no way back. All three paths ran `messages = Array(messages.prefix(idx))`, so a regeneration that comes back worse than the original is unrecoverable user data loss — the most common thing a user does after reading a reply they don't like is ask for another one, and today that is a one-way door. This is also the single feature gap called out most often against Open WebUI / ChatGPT, both of which keep the alternatives.

  Two of the fixes below are data-integrity bugs found while validating this work, not cleanups:

  - **Editing the opening prompt corrupted the conversation.** The legacy migration keyed on "no node has a parent", but editing the first prompt legitimately produces several parentless roots — shape-identical to a pre-branching array. Two alternative prompts were re-chained into one thread, destroying the branch entry and making the next request send both prompts as consecutive turns. This happened at **save** time, not merely on reload. Migration now keys on the presence of the `branches` key; shape is never inferred from.

  - **Downgrades scrambled history.** Writing the whole node bag into `messages` would hand an older build a pile of alternatives with no parent links, which it renders as one linear thread and then re-saves in that order — losing the parent chain permanently. Keeping `messages` as the active path preserves the existing downgrade-readability contract.

  And one that made the feature silently useless in its most common setting:

  - **Tool-assisted answers were unswitchable.** A tool round-trip writes four rows for one logical answer (`user → assistant(tool_calls) → tool → assistant(final)`). The fork is at the dispatch row, but the row the user reads is the final answer, whose own sibling group is empty — so the switcher vanished exactly when web search had been used, stranding the user on the new answer. Every row of a logical answer now resolves to the same anchor.

  ## Compatibility

  Old transcripts decode into a degenerate chain and render exactly as before. A conversation that never branched serialises byte-identically to a pre-branching build's output — both new keys are omitted when empty.

  ## Validation

  - `swift test --no-parallel` — 2618 tests, 26 new.
  - Pre-existing failure `Adjacent links keep their own hit targets`  (`MarkdownLinkPerformanceTests.swift:82`) also fails on `main` at `dfb2f4b3`; unrelated to this surface. Note the suite has ~6 timing-flaky tests under parallel execution — compare baselines with `--no-parallel`.
  - AX identifier gate, `verify-conversation-ordering.sh`, `verify-recommendation-tiers.swift` all pass.
  - Each of the three bugs above has a regression test whose comment records the failure mode.


raullenchai (commented):
Reviewed under the repo's convergence SOP: supply-chain audit, line-by-line read, then codex review iterated to convergence (3 rounds, gpt-5.6-sol).

**Verdict: high-quality contribution — accepted with review fixes pushed to this branch** (maintainer edits enabled). r1: 0 blocking / 3 major / 5 minor / 1 nit; r2: 1 major (introduced by my own r1 perf fix, caught and fixed); r3: **0 blocking / 0 major — converged**.

What the two review commits change (`27e66b5b`, `3b05e03e`):
- **Per-token render cost (r1 MAJOR):** `branchPosition`/`deletionImpact`/`siblings` ran per visible row per streamed-token render pass, each rescanning and re-sorting `branchedAway + messages` — quadratic-plus on long transcripts. Added `treeShapeVersion` + a lazily rebuilt indexed `TreeSnapshot` (id index, pre-sorted child groups); token writes don't invalidate it.
- **Snapshot staleness (r2 MAJOR):** a snapshot built mid-stream caches a row's early value and stays version-current after the stream finishes; `selectBranch`/`deleteMessage` now re-read live buffers (`liveTree()`) before adopting/persisting. Regression test proven red against the reverted fix.
- **Serialisation claim corrected (r1 MAJOR):** "byte-identical" was false (`activeLeafID` always written; per-row `parentID`). Now: never-branched conversations emit **no** conversation-level branching key (`activeLeafID` omitted while `branchedAway` is empty), docs state the real contract (decodable-identical, `branches`-key presence as the schema marker), raw-JSON key-omission test added.
- **Migration round-trip pinned (r1 MAJOR concern):** branch → delete-all-alternatives → re-encode (no `branches` key) → reload verified safe, incl. the edited-opening-prompt multi-root shape; two regression tests.
- Minors: duplicate-id dedupe before tree arithmetic, single code path for `ChatConversation.activePath`, `deleteMessage` doc matches behavior (prompt is kept), streaming-refusal + delete-only-answer tests, doc dedupe.

Also independently verified: the streaming accumulator seeds from `currentMessage(index:)`, so `parentID` survives token writes; `Adjacent links keep their own hit targets` fails identically on stock main (pre-existing, unrelated).

Local: `swift test --no-parallel` 2624 tests, 36/36 on the branching/tree suites, single failure = that pre-existing one.

raullenchai (approved):
Converged: 3 codex rounds to 0B/0M, full CI green (20 checks) including gui-golden-flows with runner-regenerated baselines. Supply-chain audit clean.

raullenchai:
Merged — thank you for an unusually well-engineered contribution. The tree design (single parentID edge, total functions, path-as-live-buffer) held up under three adversarial review rounds, and the migration reasoning in the PR body was accurate everywhere we could falsify it. The two review commits on top are perf (per-render tree scans) and doc-accuracy only; the architecture shipped as you designed it. 🤖

PR #2149, opened 2026-08-20
feat(markdown): preview a Mermaid diagram, offscreen and never on the network

xiaoxiunique:
A fenced `mermaid` block draws itself, reusing the picture path the SVG
preview established in #2148: the button swaps between the drawing and the
source, and a renderable block now opens as its drawing rather than waiting
to be asked.

> Rebased onto `main` after #2148 was squash-merged. The branch previously
> carried the pre-squash SVG commits; those are gone. It now stacks on
> #2184, which fixes an unrelated orientation defect on `main` — that commit
> will drop out of this branch when #2184 merges.

## Why WebKit, when the SVG preview needs none

`SVGPreview` argues that AppKit renders SVG natively and a web view would be
a much larger machine for no gain. That holds for SVG a person or a model
writes. Mermaid's output is a different animal — measured against this exact
library:

- `<foreignObject>` labels, which flowchart, class, state, ER, mindmap and
  journey diagrams all use, are **not drawn at all**. The diagram arrives as
  empty boxes.
- With `htmlLabels: false` the labels become native `<text>` and do draw, but
  land outside their shapes.
- `marker` arrowheads are never drawn, in either mode.

This design was adopted and then abandoned: twelve of twelve diagram types
returned non-nil images of the correct size, and rendering them to PNG showed
every one was blank. Size and nil-ness were the wrong questions.

## Why it is still not a live web view

The web view never enters the view hierarchy. It renders into an offscreen
window, is snapshotted, and the result is an `NSImage` drawn by the same
`draw(_:)` path the SVG preview uses. No accessibility subtree (the committed
golden-flow baselines are untouched), no participation in layout, no
asynchronous height arriving after `height(forWidth:)` has already answered.

## Network

Four layers, in order of when they fire:

1. CSP with `connect-src 'none'` — covers fetch, XHR, WebSocket,
   EventSource and `sendBeacon`.
2. A `WKContentRuleList` that blocks everything, then re-allows the private
   scheme. This is the layer the scheme handler cannot provide: the handler
   is only consulted for its own scheme, so an `https://` subresource would
   go straight past it.
3. A scheme handler serving exactly two paths from memory.
4. A navigation delegate that cancels any URL not on that scheme.

Removing any single layer leaves the tests green, which looks like the tests
are vacuous. Removing all three reachable ones produces six connections to
the loopback probe and a red assertion — so the redundancy is real and the
tests can fail. The probe has a positive control for the same reason.

## Renders are serialized

One web view serves every diagram, and a render is two separate awaits
against it: `__rapidRender` draws into the stage and measures it, then
`takeSnapshot` reads the pixels back. Nothing between those points stopped a
second diagram replacing the stage and leaving the first to photograph the
second one's drawing.

Because the snapshot is assigned the size that *was* measured, the wrong
picture arrives at the right block's dimensions — it looks like a diagram
that rendered as something else and got cropped. With four diagrams settling
on one flush, three of four were wrong on every run.

`concurrentRendersDoNotSwap` compares pixels. An earlier version compared
sizes and passed against the bug, for exactly that reason.

## The library

`mermaid.min.js` 11.17.0 (MIT), vendored and verified by SHA-256 at load and
again by `verify-app-resources.swift` in the assembled `.app`. A name lookup
is not enough: a truncated file resolves, loads, sets no global, and dies
inside the first render — one link past where a name check would stop.

## Verification

- Full suite: 2699 tests, no failures other than a pre-existing DNS test
  that needs network.
- `RAPID_BUILD_CONFIG=release ./scripts/build.sh` — the library and its
  licence land in `Contents/Resources/` and the digest check passes in the
  bundle, which is the only path a shipped app uses.
- Manual: a four-diagram answer renders each block correctly, light and dark.

## Not included

HTML preview — that needs a live web view, which is a different order of
problem. Bundle size is 522 MB against a 500 MB cap; that is the Python
sidecar (480 MB) and predates this branch.

## Design precedent

The renderer follows the established desktop pattern of transforming untrusted diagram source inside a network-isolated, offscreen renderer and handing only a bounded image back to the native UI. Deterministic source errors may be cached; transient renderer failures rebuild and retry so infrastructure faults do not permanently poison valid diagrams.

## Current integration state

Rebased onto current `main` at `25a54b67f`; #2184 and #2280 are already included. The conflict resolution preserves stable-ID streaming groups while keeping only the actively growing final Mermaid block non-final.

Fixes #2149


raullenchai (changes requested):
Reviewed exact head `e9a750f29702e2698bf915a3bffaa85226f34a05` against current `origin/main` (`33cf880d283470da522820cd292542ab3b4d51ad`); the merge is clean.

The core rendering path, private scheme, CSP/content-rule layers, resource digest/license staging, accessibility inventory, and release Swift build all checked out. Two scoped issues remain:

1. `MermaidNavigationPolicy.webView(_:decidePolicyFor:decisionHandler:)` does not match the current SDK's optional `WKNavigationDelegate` requirement: the required callback is `@MainActor @Sendable`, and the exact-head build emits the near-match warning. A deterministic runtime probe that installed this delegate on a real `WKWebView` and navigated to a loopback endpoint observed 4 requests, so this navigation boundary is not active. The existing static-policy assertion and offscreen click exercise do not invoke the installed delegate callback. Please conform to the exact SDK signature and add an end-to-end test that installs the delegate, attempts a real remote navigation against a local positive-control server, and proves the request never reaches it.

2. The focused Mermaid test set is not safe under Swift Testing's default parallel execution. Running `Mermaid|SVGPreviewTests|AccessibilityIdentifierInventoryTests|ThirdPartyLicenseStagingTests` produced a deterministic failure in “Light and dark are cached apart” because another Mermaid suite resets the shared renderer while the rendering suite is active; the isolated `MermaidRenderingTests` suite passes 5/5. Please give the Mermaid suites deterministic ownership of shared renderer lifecycle (suite-local serialization or an injected/non-shared test instance are both reasonable) so the focused command is green without global suite serialization.

Evidence on this head: isolated rendering 5/5 passed; combined focused set 89/90 passed with the cache-reset race above; vendored JavaScript SHA-256 passed; production Swift build passed. These requests stay within the claimed no-network boundary and deterministic focused contracts.

xiaoxiunique:
Both addressed on `04bb1946`, rebased onto `87344238`.

## 1. Signature

Conformed to the exact requirement — `@MainActor` on the method, `@MainActor @Sendable` on the callback. The near-match warning is gone.

Measurements taken before the change, against a loopback listener, with an unguarded control:

| delegate | navigation | TCP accepts | bytes received |
|---|---|---|---|
| near-matching | `WKWebView.load(http://…)` | 1 | 0 |
| conforming | `WKWebView.load(http://…)` | 1 | 0 |
| near-matching | in-page `location.href = http://…` | 0 | 0 |
| none (control) | `WKWebView.load(http://…)` | 1 | full HTTP request |

Rows 1 and 2: the accept happens before the policy is consulted; no request line, headers or cookies follow it. Row 3 is the path a diagram has available — the offscreen page navigating itself. `MermaidRenderer` calls `load()` once, with `MermaidHostPage.hostPageURL`.

The listener counts accepts, which is also what `LocalRequestProbe` counts.

## 2. End-to-end test

`An installed delegate refuses a navigation off the host page` loads the real private-scheme page, tells it to navigate to a listening socket, and asserts nothing arrives — after making the identical page, with no delegate attached, arrive first.

Falsifiability checked: with `policy(for:)` forced to `.allow` it fails at 6 against a baseline of 3.

## 3. Parallel execution

Each suite that calls the renderer directly now constructs its own `MermaidRenderer`; a `struct` suite is built once per test, so the cache, failure budget and web view are per-test.

`Preview auto-reveal` keeps `.shared` and is `.serialized`. It drives `MarkdownCodeBlockView`, which reads `MermaidRenderer.shared` directly, so an injected instance stays unwritten while the assertions read an empty cache — that configuration fails deterministically at `A diagram is not drawn while it is still being written`.

`Mermaid|SVGPreviewTests|AccessibilityIdentifierInventoryTests|ThirdPartyLicenseStagingTests`: 95 tests, green on three consecutive runs.

Full suite: 3000 tests, no Mermaid failures. The remaining failures reproduce identically on a clean `main` — DNS (a local fake-ip resolver answering reserved TLDs), `pyexpat` on a Homebrew Python 3.14, and `TimeZone.abbreviation(for:)` returning `GMT-7` where the test expects `PDT`.

raullenchai (approved):
Initial review complete. Core rendering path, four-layer network defence (CSP, content rules, scheme handler, navigation delegate), digest-verified vendored library, serial render queue, failure budget, and test coverage all check out. The prior review's two issues (SDK navigation signature conformance + test parallelism race) are addressed by commit `04bb1946`. The end-to-end navigation test with positive control is a real improvement over the static-policy assertion alone. Ready for PR validation and Codex review.

raullenchai:
## Harbor review contract

**User-visible goal:** render fenced Mermaid diagrams as an inline preview with a reliable Preview/Code toggle.

**Allowed scope:** Desktop Markdown code-block preview, offscreen Mermaid rendering, bundled asset integrity/licensing, network isolation, layout/accessibility behavior, and focused tests.

**Non-goals:** HTML preview, live interactive web content, engine/server changes, or unrelated Markdown redesign.

**Constraints:** no external network access, no live web view in the transcript, bounded serialized rendering, digest-verified vendored asset, stable light/dark and streaming behavior.

**Acceptance criteria:** Mermaid/SVG/security/license tests pass; final streaming block triggers exactly the settled render path; app bundle resource verification passes; independent Codex review converges; `pr_validate` is clean.

**Exact range under review:** `7368ae28534c87bfb22f7e1443d1f5c95573768b..4db9e49ad`.

raullenchai:
## 🤖 Codex review — round 1

apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Render/MarkdownCodeBlockView.swift:163 — Appearance changes rebuild syntax colors but never invalidate or rerender the Mermaid preview for the new theme. After switching macOS from light to dark mode, an already-visible diagram continues displaying the cached light-theme bitmap indefinitely because `previewImage` remains non-nil and `updatePreviewAvailability()` is not called. The cache-key test only exercises `MermaidRenderer` directly and does not cover this view-level behavior.

apps/rapid-mac/Sources/Rapid/UI/Markdown/Mermaid/MermaidRenderer.swift:150 — The model-authored JavaScript render has no timeout or cancellation boundary. A valid but computationally pathological diagram can leave `callAsyncJavaScript` awaiting indefinitely; because every render is serialized through `tail`, that single diagram then prevents all later Mermaid previews from rendering for the rest of the session. The advertised failure budget never applies because execution never reaches the `catch`. Add a deadline that tears down the web view and counts a failure, with a test using an intentionally non-completing render seam.

VERDICT: CHANGES_REQUESTED

raullenchai:
Round 1 findings addressed on `c1b0a25e1`:

- Appearance lifecycle: a Mermaid bitmap is invalidated and re-rendered for the new effective theme; covered by a view-level light→dark test.
- Render deadline: model-authored JavaScript evaluation now has an 8-second hard deadline, tears down the abandoned WebView through the existing failure path, and releases the serial queue; covered by a deliberately non-completing evaluator followed by a successful render.
- Snapshot allocation bound: dimensions are rejected above 4096 points per axis or 4,000,000 point-pixels before the 2× bitmap request, including overflow coverage.
- Streaming-final transition remains covered and deterministic across host appearance.

Verification: 99/99 focused Mermaid/SVG/security/accessibility/license tests pass; non-vendored diff check passes. The vendored minified library remains byte-for-byte digest pinned.

raullenchai:
## 🤖 Codex review — round 2

apps/rapid-mac/Sources/Rapid/UI/Markdown/Mermaid/MermaidRenderer.swift:246: Snapshot generation has no timeout or content-process termination handling. If WebKit crashes or wedges after JavaScript measurement but before `takeSnapshot` invokes its completion handler, this continuation never resumes. Because this render owns the serial `tail`, every subsequent Mermaid preview waits forever. Apply the same deadline/gate/teardown strategy used for JavaScript evaluation and handle `webViewWebContentProcessDidTerminate`.

apps/rapid-mac/Sources/Rapid/UI/Markdown/Mermaid/MermaidRenderer.swift:381: Initial page loading waits indefinitely for `didFinish`/`didFail`. Web-content-process termination during startup has a separate delegate callback and is not guaranteed to settle this navigation continuation. The shared `setup` task and render queue then remain permanently suspended, disabling every Mermaid preview for the session. Add a setup deadline and termination handling; the existing timeout test only exercises JavaScript evaluation after setup succeeds.

VERDICT: CHANGES_REQUESTED

raullenchai:
Round 2 findings addressed on `HEAD`:

- Snapshot generation now uses the same one-shot deadline gate as JavaScript evaluation; timeout/error tears down WebKit and releases the serial queue. A deliberately non-completing snapshot followed by a successful render proves recovery.
- Initial private-page navigation now has a bounded deadline.
- `webViewWebContentProcessDidTerminate` settles startup immediately and aborts any active evaluation/snapshot continuation before teardown.
- Focused setup/render tests: 11/11 pass, including both timeout paths, setup deadline, termination, and post-failure queue recovery.

raullenchai:
## 🤖 Codex review — round 3

- apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Render/MarkdownCodeBlockView.swift:308 — Preview identity tracks only `code`, not whether the image was produced as SVG or Mermaid. Reconfiguring a reused view with identical code but a different language classification leaves `previewImage` intact, so an SVG-tagged block changed to `mermaid` displays the stale SVG parse and never requests Mermaid rendering. Conversely, changing from Mermaid to SVG can retain the Mermaid bitmap. The async completion at line 353 compounds this by validating only code and theme, allowing an obsolete Mermaid render to overwrite the current SVG preview. Track the preview kind/language and verify current classification when applying asynchronous results; add a reconfiguration test covering identical source with changed language.

- apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Render/MarkdownCodeBlockView.swift:328 — `hasToggledPreview` is never reset when the represented block changes. Because SwiftUI identifies these views by group offset, a view can be reused for a replacement/regenerated code block at the same position. If the reader closed the previous preview, the newly rendered diagram remains source-only instead of receiving the documented auto-reveal behavior. Reset the toggle preference when the block’s source or preview kind changes, and test view reuse with two distinct diagrams.

VERDICT: CHANGES_REQUESTED

raullenchai:
Round 3 findings addressed: preview state now keys on source plus preview kind, clears stale images and reader-toggle state when a SwiftUI row is reused for a different block/classification, and rejects async Mermaid completion unless the current identity is still Mermaid. Added deterministic tests for identical-source language reclassification and replacement-diagram auto-reveal; PreviewAutoRevealTests pass 6/6.

raullenchai:
## 🤖 Codex review — round 4

- [P1] apps/rapid-mac/Sources/Rapid/UI/Markdown/Mermaid/MermaidHostPage.swift:107 — `theme` and `themeVariables` are absent from the `secure` configuration list. A diagram containing `%%{init: {"theme":"dark"}}%%` can override the appearance-selected theme, so a light-mode render is cached under the `.light` key with dark colors and remains wrong across row rebuilds. Add the theme-related keys to `secure` and test rendering a directive-bearing diagram under both appearances.

VERDICT: CHANGES_REQUESTED

raullenchai:
Round 4 finding addressed: Mermaid directives can no longer override `theme` or `themeVariables`. The host-policy assertion covers both protected keys, and an end-to-end test renders an adversarial directive under light and dark appearances and verifies the protected output matches the app-selected theme. Focused policy/render tests pass 4/4.

raullenchai:
## 🤖 Codex review — round 5

Tests were not run because this environment lacks Swift/macOS AppKit/WebKit tooling; review was static, including surrounding code and test coverage.

VERDICT: LGTM

raullenchai:
pr_validate review findings addressed on `053453e00`:

- Host rendering now preserves the exact SVG bounding box via `viewBox` and sizes from width/height only, preventing negative-origin clipping and positive-origin padding.
- After measurement passes allocation bounds, the offscreen WebView/window are resized to the exact snapshot surface before capture; an injected snapshot assertion verifies surface/rect parity.
- Untagged CRLF Mermaid frontmatter is parsed using newline semantics rather than a literal LF Character split, with regression coverage.

The prior local full-unit failure was environmental: `python3` resolved to Xcode Python 3.9, which cannot parse this repository's Python syntax and lacked Hypothesis. It will be rerun with the project Python environment.

raullenchai:
## 🤖 Codex review — round 6

VERDICT: LGTM

raullenchai:
Final validation fixes on exact head `761dbcc39`:

- Split renderer outcomes into deterministic source rejection vs transient infrastructure failure. Only invalid/oversized diagram source is negatively cached; setup failures, WebKit termination, evaluation/snapshot errors, and timeouts can recover on a later render.
- Routed production Mermaid bytes through a pure digest-validation boundary and changed the tamper test to call that exact boundary with a same-length modified payload.
- Made the network-boundary harness fail explicitly if its private host page does not finish loading.
- Removed the legacy callback warning in the navigation harness.

Local evidence: 38 Mermaid tests + 37 SVG/preview regression tests passed (75 total); `git diff --check` passed. Independent review had already converged to LGTM before these final validator hardening changes; final `pr_validate` is running on this head.

raullenchai:
Addressed the final validator findings on exact head `3b9df1cee`:

- A `(nil, nil)` WebKit snapshot is now a thrown snapshot failure, so it consumes the failure budget, tears down the suspect renderer, and remains retryable rather than looping or poisoning the source cache. Added a recovery test that fails the first snapshot this way and succeeds after rebuild.
- Mermaid snapshot bounds now reserve an 8-point perimeter around `getBBox()` so strokes, caps, and marker arrowheads at diagram edges are not clipped. Empty geometry is still rejected before padding, and the padded dimensions still pass through the existing allocation bounds.

The 76-test Mermaid/SVG run passed 75 tests before a source-contract assertion needed to recognize the now multiline `viewBox` setter; that assertion and the two new focused tests then passed. `git diff --check` passes.

raullenchai:
CI diagnosis: exact head `3b9df1cee` exposed Swift strict-concurrency compilation that the focused test build did not surface: the new `RenderOutcome` carries AppKit `NSImage` through `Task.value` and therefore needs an explicit Sendable boundary. Fixed on `281d621cc` with a narrowly documented `@unchecked Sendable` conformance; the enum is created and consumed only on `MermaidRenderer`’s `@MainActor`, so the image never moves off actor. Exact local `swift build` now passes. This is a compilation-only annotation with no renderer behavior change.

raullenchai:
Second CI diagnosis on `281d621cc`: the hosted compiler correctly enforced Sendable on the pre-existing `Task<NSImage?, Never>` in-flight deduplication path. Replaced the result with a narrowly scoped `RenderedImage: @unchecked Sendable` wrapper whose documented invariant is main-actor-only ownership; both the in-flight task and `RenderOutcome` now have actually Sendable payload types. Also widened the synthetic timeout test from 100 ms to 500 ms because it tests queue recovery, not a cold-WebKit latency SLA; it passed three consecutive runs. Exact local `swift build` passes. New head: `00a080a01`.

raullenchai:
Third exact-head CI diagnosis on `00a080a01`: after the Sendable errors were removed, the hosted compiler exposed an older tuple-destructuring inference incompatibility in this PR’s `ForEach(Array(groups.enumerated()))` closure (`index` resolved to the C symbol). Rewrote it to accept one enumerated entry and name `entry.offset` / `entry.element` explicitly. This is behavior-preserving and compiler-portable. Exact local `swift build` and 14 renderer/policy tests pass. New head: `4476441f3`.

raullenchai:
Rebased onto current `main` (`25a54b67f`) after #2280 landed. The only overlap was `MarkdownBlockStack`: the resolution preserves stable-ID streaming groups from #2280 and marks only the final Mermaid block non-final while streaming.

Exact head: `5e20ba2aa1cffe3651c417bf10173e5fc446ea45`

Local integration evidence:
- `swift build --package-path apps/rapid-mac` ✅
- focused Mermaid + SVG preview + streaming/parity suite: 123 tests passed ✅
- `git diff --check` (excluding the digest-pinned upstream minified vendor asset) ✅

raullenchai:
Exact-head Desktop CI found a real Swift concurrency compatibility error on the hosted toolchain: the snapshot continuation returned main-actor `NSImage` directly across a `sending` boundary. New head `a7e26be7c` returns the existing narrowly scoped `RenderedImage` sendability wrapper from that continuation as well; this preserves main-actor ownership without claiming `NSImage` is generally Sendable.

No failed-job rerun: fixed as a new head. Local evidence on the fix:
- `swift build --package-path apps/rapid-mac` ✅
- Mermaid + SVG preview + streaming/parity suite: 123 tests passed ✅
- `git diff --check` ✅

raullenchai:
Hosted `swift build` is now green. The following `swift test` compile exposed two test-only task boundaries returning non-Sendable `NSImage`. New exact head `ebf8e5ba6` keeps every image operation on MainActor and returns only Sendable fingerprints/booleans from concurrent tasks, without weakening the tests’ concurrency.

Again, no rerun of the failed job: corrected as a new head. Focused Mermaid suite passed locally: 39 tests, including concurrent picture isolation and concurrent cold-start setup.

raullenchai:
This fork PR cannot enter the repository merge queue by design (`-from-fork` protects train branches and repository resources). The independently reviewed, hardened exact commits are now preserved in maintainer integration PR #2812, with contributor authorship retained in commit history. Closing this PR as superseded; #2812 will complete the normal Mac queue path.

Issue #2146, opened 2026-08-20
3 repo-contract tests fail on main after the Cohere2-MoE / GPT-OSS-Puzzle vendor commits


raullenchai:
## What

On pristine main (verified at `dfb2f4b3` and still true at `0ca64412`), three repo-contract tests fail locally:

```
FAILED tests/test_mlx_compat.py::test_every_mlx_lm_consumer_installs_shim
  models/cohere2_moe.py (mlx_lm import @ line 14, install @ None)
FAILED tests/test_model_profiles_ssot.py::test_every_alias_has_explicit_profile_fields
FAILED tests/test_model_sizes.py::test_every_text_alias_has_a_manifest_entry
  missing from model_sizes.json: mlx-community/North-Mini-Code-1.0-4bit, mlx-community/North-Mini-Code-1.0-bf16
```

Introduced by the vendor commits aac18fb0 (#1223 North-Mini-Code Cohere2-MoE) and 837d4b78 (#1224 GPT-OSS Puzzle):

- `vllm_mlx/models/cohere2_moe.py` does a module-load-time `import mlx_lm` without the `_mlx_compat.install()` shim the contract requires (#404 M5 regression risk).
- The two new North-Mini-Code aliases are missing from `model_sizes.json` (the test's own remedy: `python3.12 scripts/gen_model_sizes.py`) and lack explicit profile fields.

## Why it matters

Every PR's local `full_unit` now carries these 3 pre-existing failures, so real regressions are harder to spot in the noise (three of my PR validation runs tonight had to explain them away). The CI `tests` gate apparently doesn't exercise these paths, which is why the commits landed green.

## Suggested fix

Mechanical, per each test's own remedy: add the shim install to `cohere2_moe.py` before its mlx_lm import (same pattern as `utils/mamba_cache.py`), run `scripts/gen_model_sizes.py`, and fill the explicit profile fields for the two new aliases. Left to the vendor-commit authors since it touches vendored model files and the alias SSOT.

PR #2151, opened 2026-08-20
feat(perf): fuse MoE gate+up expert projections into one gather_qmm

raullenchai:
## Why

At single-token decode, mlx-lm's stock `SwitchGLU` issues three tiny `gather_qmm` launches per MoE layer (gate, up, down). Affine quantization packs each output row independently, so concatenating the gate and up expert weights along the output axis and issuing **one** `gather_qmm` over `[E, 2*inter, hidden]` is **bit-identical** to the two separate calls while removing one GPU launch per MoE layer per token.

Measured on Qwen3.6-35B-A3B-8bit (M3 Ultra, cache-busted medians, isolated single-server protocol):

| level | baseline | fused | delta |
|---|---|---|---|
| mlx_lm generate decode | 86.4 tok/s | 92.5 tok/s | **+7.0%** |
| server E2E long-prompt decode | 81.4 tok/s | 86.1 tok/s | **+5.8%** |

Correctness: greedy outputs **byte-identical** at every level — layer microbench (`mx.array_equal` on both the decode and gather-sort paths), `mlx_lm.generate` (identical text), and server-level 6-prompt greedy capture (English/code/math/Chinese, identical `content`/`reasoning`/token counts vs a `RAPID_MLX_MOE_GATE_UP_FUSION=0` control server).

Fusion design adapted from the oMLX project (https://github.com/jundot/omlx, Apache-2.0), which ships it default-on for the same layer type. Applies to every stock-`SwitchGLU` MoE the server loads (Qwen3.5/3.6 MoE, Qwen3-MoE family, etc.); dense models are a no-op scan.

## What

- `vllm_mlx/moe_fusion.py` — `fuse_gate_up(model)`: class-level `__call__` patch (once per process, stock path preserved for unfused instances) + in-place instance rewrite (gate module becomes the fused container so quantization params carry over; original gate/up buffers freed per layer with a cache drain to bound the load transient). Structural gate refuses mismatched quantization params, `SwitchGLU` subclasses (exact-type only), and anything without a clean gate/up pair. `RAPID_MLX_MOE_GATE_UP_FUSION=0` opts out.
- `vllm_mlx/engine/batched.py` — invoke post-load on the mlx-step worker (weight concat touches MLX streams only that thread owns, #170), **inside the non-disk-stream branch only**: the disk-stream forward (`disk_stream_patch.py` / `qwen2_moe_forward.py`) reads `gate_proj`/`up_proj` directly and its lazy load never materializes the routed weights, so those lanes keep the stock layout.
- `tests/test_no_out_of_band_routing.py` — register the env var in `ALLOWED_RAPID_MLX_ENV_VARS` (non-routing perf toggle, same shape as `RAPID_MLX_DISABLE_FUSED_SAMPLER`).

## Tests

`tests/test_moe_gate_up_fusion.py` (11 tests): byte equality on decode and gather-sort paths (quantized 4-bit and unquantized, incl. bias), in-place rewrite semantics (originals dropped, container reused), idempotency, unfused-instance stock path through the patched class `__call__`, dense no-op, env opt-out, mismatched-quant-params reject, subclass reject.

## Notes

AI-authored (Claude Opus 4.8), continuing the cross-engine benchmarking session that produced #2144. Benchmark artifacts on mac-studio `~/work/benchout/` (`35b-moefuse.json`, `greedy-35b-{fused,base}.json`).

### DIFF
--- a/tests/test_moe_gate_up_fusion.py
+++ b/tests/test_moe_gate_up_fusion.py
@@ -1,0 +1,179 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Correctness gates for the MoE gate+up expert-projection fusion.
+
+The fusion must be BIT-exact — affine quantization packs each output row
+independently, so a concatenated gate+up gather_qmm returns the same
+bytes as two separate calls — and strictly opt-out-able. These tests pin
+byte equality on quantized and unquantized SwitchGLU, the structural
+_can_fuse gate, in-place rewrite semantics (original buffers dropped,
+stock path preserved for unfused instances), idempotency, and the env
+kill-switch.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+import mlx.nn as nn
+from mlx_lm.models.switch_layers import QuantizedSwitchLinear, SwitchGLU, SwitchLinear
+
+from vllm_mlx import moe_fusion
+
+# Small MoE geometry: E experts, k active, quantization-friendly dims.
+E, HID, INTER, K = 8, 64, 128, 2
+
+
+class TinyMoE(nn.Module):
+    """Minimal container so fuse_gate_up's module scan finds the layers."""
+
+    def __init__(self, n_layers=2, quantize=True):
+        super().__init__()
+        self.layers = [SwitchGLU(HID, INTER, E) for _ in range(n_layers)]
+        if quantize:
+            nn.quantize(self, group_size=32, bits=4)
+
+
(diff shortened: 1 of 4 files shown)


raullenchai:
Round-1 finding addressed: the bias-path test now constructs `SwitchGLU(..., bias=True)`, asserts both projections carry bias parameters before fusing and that the fused container carries the concatenated bias after — the additive-bias fusion path is genuinely exercised. 11 tests green. full_unit's other failures are the 3 pre-existing main breakages tracked in #2146.

PR #2153, opened 2026-08-20
feat(perf): replace the MoE router argpartition+take with one Metal launch

raullenchai:
## Why

At decode, the Qwen3-Next-family MoE routing chain — softmax, `argpartition` top-k, `take_along_axis`, renormalize — is five tiny-tensor launches per MoE layer per token. This PR replaces the two selection launches (`argpartition` + `take_along_axis`) with **one** fused Metal dispatch; softmax and the renormalize stay composed.

**Byte-exactness design** (the interesting part): a naive fused top-k+renorm was measured to make greedy outputs diverge (reduction-order ulp in the scores compounds over 40 layers × hundreds of tokens — 3/6 greedy prompts flipped). The landed design instead has the kernel emit indices in `argpartition`'s empirically pinned output permutation (values ascending, ties ascending by index) plus the RAW selected probabilities (`T(float(x))` round-trips losslessly, bit-identical to `take_along_axis`), so the composed renormalize sees the exact tensor the original chain would — the block output is **byte-identical**. `install()` verifies both the permutation parity and the raw-score identity on random probes at install time and refuses to engage if mlx's ordering ever changes.

Measured on Qwen3.6-35B-A3B-8bit (M3 Ultra, cache-busted medians, single-server protocol, on top of the #2151 gate+up fusion):

| | gate_up only | + this PR | delta |
|---|---|---|---|
| long-prompt decode | 86.1 tok/s | 87.6 / 87.2 (two runs) | **+1.3–1.7%** |
| short-prompt decode | 90.6 | 91.0 | +0.4% (noise-level) |
| greedy outputs (6 prompts) | — | — | **6/6 byte-identical** |

Honest framing: a modest, real win — the renormalize (the ulp source) cannot be fused without breaking byte-equality, which caps the gain at roughly half of the naive fusion's. Cumulative with #2151: 81.4 → 87.6 tok/s (**+7.6%**) on the 35B flagship MoE. Kernel design adapted from the oMLX project (Apache-2.0); the byte-exact selection-only strategy is ours.

## What

- `vllm_mlx/moe_router_fusion.py` — the kernel (one simdgroup per row, 32 lanes each owning NE/32 experts, K `simd_max` selection rounds) + `install()` wrapping `Qwen3NextSparseMoeBlock.__call__` (shared by mlx-lm's qwen3_5 / qwen3_6 / qwen3-next). Fast path gated to decode-width rows (≤8), `NE % 32 == 0`, bf16/fp16, unsharded, `norm_topk_prob=True`; anything else runs the original body; any fast-path failure permanently degrades to it. `RAPID_MLX_MOE_ROUTER_FUSION=0` opts out.
- `vllm_mlx/scheduler.py` — `install()` at module import, next to the GDN kernel install.
- `tests/test_no_out_of_band_routing.py` — env var registered in the allowlist.

## Tests

`tests/test_moe_router_fusion.py` (28 tests): elementwise permutation parity vs `argpartition` across seeds/dtypes/expert counts, tie-to-highest-index membership + tie permutation parity, raw scores byte-identical to `take_along_axis`, renormalized scores byte-identical to the composed chain, row/dtype/expert-count gates, install idempotency + env opt-out, and a real `Qwen3NextSparseMoeBlock` end-to-end `mx.array_equal` check.

## Notes

AI-authored (Claude Opus 4.8), continuing the perf lane that produced #2144 and #2151. Benchmark artifacts on mac-studio `~/work/benchout/` (`35b-router-v3.json`, `greedy-35b-{rt3,norouter}.json`; the diverging naive-fusion evidence in `greedy-35b-router.json`).

### DIFF
--- a/tests/test_moe_router_fusion.py
+++ b/tests/test_moe_router_fusion.py
@@ -1,0 +1,333 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Correctness gates for the fused MoE router top-k kernel.
+
+The kernel replaces only the argpartition selection; reversed, its
+output must match argpartition's permutation ELEMENTWISE (values
+ascending, ties ascending by index) — that pinned ordering is what
+makes the block output byte-identical, since every other step keeps
+the composed ops. The install wrapper must engage only for short-row,
+renormalizing, unsharded calls and fall back (permanently on failure)
+to the original block body.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+if not mx.metal.is_available():  # pragma: no cover - CI runners have Metal
+    pytest.skip("Metal GPU required for router kernel tests", allow_module_level=True)
+
+from vllm_mlx import moe_router_fusion
+
+NE, K = 256, 8
+
+
+def _composed(gates, k):
+    inds = mx.argpartition(gates, kth=-k, axis=-1)[..., -k:]
+    scores = mx.take_along_axis(gates, inds, axis=-1)
+    scores = scores / scores.sum(axis=-1, keepdims=True)
+    return inds, scores
+
+
+def _probs(rows=1, ne=NE, dtype=mx.bfloat16, seed=0):
+    mx.random.seed(seed)
+    g = mx.softmax(mx.random.normal((rows, ne)).astype(dtype), axis=-1, precise=True)
+    mx.eval(g)
+    return g
+
+
+class TestSelectionParity:
+    @pytest.mark.parametrize("seed", range(8))
+    @pytest.mark.parametrize("dtype", [mx.bfloat16, mx.float16])
+    def test_selected_set_matches_argpartition(self, seed, dtype):
+        g = _probs(rows=4, dtype=dtype, seed=seed)
+        fi, fs = moe_router_fusion.fused_router_topk(g, K)
+        ci, cs = _composed(g, K)
+        mx.eval(fi, fs, ci, cs)
+        assert fi.astype(mx.int64).tolist() == ci.astype(mx.int64).tolist(), (
+            "permutation differs from argpartition"
+        )
+
+    def test_tie_resolves_to_highest_index(self):
+        # Craft a row where several experts share the exact k-th probability:
+        # the fused kernel must pick the same members argpartition does
+        # (ties -> highest index at this shape).
+        base = mx.full((1, NE), 1.0 / NE, dtype=mx.float32)
+        # raise experts 3, 7, 250 to one shared higher value; k=2 forces a
+        # tie-break among three equal candidates
+        base[0, 3] = 0.5
+        base[0, 7] = 0.5
+        base[0, 250] = 0.5
+        g = (base / base.sum(axis=-1, keepdims=True)).astype(mx.bfloat16)
+        mx.eval(g)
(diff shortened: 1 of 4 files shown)


raullenchai:
All three round-1 findings addressed with one mechanism: parity verification is now **per-(num_experts, top_k) shape, memoized** (`_parity_verified`) — each combination is compile-run-permutation-raw-score verified with `mx.eval` inside the guard on its first appearance, and a failing shape is permanently routed to the composed chain. This (1) closes the unvalidated-shape gap (the gate consults the per-shape verdict, not a single startup probe), (2) means no lazy-eval kernel failure can escape into a request — every specialization the fast path will ever launch has already been evaluated in a guarded probe on the same thread — with the runtime try/except retained as belt-and-braces, and (3) is covered by new tests: per-shape probe caching, failing-shape permanent composed routing (including no re-probe after repair), and a deferred-failure degradation test that injects a kernel failure, verifies the request survives through the original body, and asserts the repaired kernel is never invoked again while outputs stay byte-identical to the original chain. 31 tests green.

raullenchai:
Round-2 disposition: **(1) fixed** — the probe now uses keyed RNG only (`mx.random.normal(..., key=...)`), never consuming the global stream; a new test seeds the global RNG, runs a fresh-shape probe, and asserts the next global draw is byte-identical to a probe-free run. **(4) fixed** — every per-shape probe now includes deterministic tie-heavy rows alongside the random ones: an all-equal row (every selection a tie), a two-level step whose high plateau is wider than k (ties across the whole selection boundary), and scattered duplicates at the top (partial ties inside the selection), for both dtypes. **(3) fixed** — the end-to-end test now instruments `fused_router_topk` (after warming the parity verdict) and asserts it ran exactly once with no fallback before the byte-equality comparison; a separate test injects a deferred failure and proves permanent degradation with the repaired kernel never re-invoked. **(2) declined with rationale** — the same finding class as #2144 round-8: every kernel specialization the fast path will ever launch has already been compiled AND evaluated inside a guarded probe on the same thread before its first fast-path use (that is what the per-shape verdict gates on), so a construction-time escape requires a failure mode the probe already rules out; forcing `mx.eval` inside `patched_call` would insert a per-layer synchronization barrier into the decode hot path and destroy the lazy pipelining the optimization exists for. Residual eval-time risks (allocator OOM etc.) apply identically to the composed chain. 32 tests green.

raullenchai:
Withdrawing after honest reassessment: round-3 finding (3) identifies the real structural weakness — the byte-exactness contract rests on `argpartition`'s partition-internal ordering, which is undocumented implementation behavior that could legitimately be input-dependent or change across mlx versions. The runtime per-shape probe (random + tie-heavy rows) narrows but cannot eliminate that risk, and the measured payoff (+1.3–1.7% decode on the 35B flagship) does not justify betting output byte-equality on an unspecified invariant. The naive fully-fused variant was already ruled out empirically (3/6 greedy prompts diverged).

The branch is kept for revival if (a) mlx documents/pins the partition ordering, (b) a selection-only strategy that doesn't depend on ordering emerges (e.g. mlx grows a stable top-k), or (c) a workload appears where the gain is large enough to justify an opt-in flag instead of default-on. The instructive negative results — ulp-in-router-scores compounds to token divergence over 40 layers; selection-only fusion loses most of the gain once take/renorm stay composed — are recorded in the PR history.

PR #2166, opened 2026-08-20
fix(video): print LTX-2.5 runtime setup walkthrough in the preflight error

raullenchai:
## Why

Release dogfood (new-user journey, mini) found a dead end: `rapid-mlx serve ltx-2.5-mlx-q8` without the pinned runtime prints

```
Error: video generation requires the pinned LTX-2.5 runtime (commit 5795228…; see the video generation guide).
```

and exits. The message names a commit hash and "the video generation guide" but gives no path, URL, or command — a new user has no next step. Every sibling preflight error (disk space, port in use, unknown alias) already prints concrete suggestions.

## What

When the LTX-2.5 runtime is missing or fails to provision, the preflight error now appends the exact setup walkthrough from `docs/guides/video-generation.md`, sourced from the single-source-of-truth constants (`LTX25_RUNTIME_REPOSITORY`, `LTX25_RUNTIME_COMMIT`):

```
  Set up the pinned runtime (docs/guides/video-generation.md):
    git clone --branch ltx25 https://github.com/MrMoferFRAN/ltx-2-mlx.git
    git -C ltx-2-mlx checkout 57952288076766abe27dda3a774b2c24f7346977
    uv sync --project ltx-2-mlx
    RAPID_MLX_LTX25_RUNTIME="$PWD/ltx-2-mlx/.venv/bin/ltx-2-mlx" rapid-mlx serve ltx-2.5-mlx-q8
```

The hint is shown only when the runtime itself is the problem — if the runtime resolved and only `uv` (or ffmpeg) is missing, the clone/checkout walkthrough would be noise and is suppressed.

## Tests

- New: `test_ltx25_missing_runtime_prints_setup_walkthrough` — missing runtime → all four setup lines present (asserted against the constants, not string copies).
- New: `test_ltx25_unprovisioned_runtime_prints_setup_walkthrough` — resolved runtime whose `uv sync` provisioning fails → walkthrough present.
- Extended: `test_ltx25_runtime_preflight_requires_uv` — runtime resolved, only `uv` missing → asserts `git clone` is NOT printed.
- Full video suites pass: 67 passed (`test_ltx25_video.py`, `test_cogvideox_video.py`, `test_wan_video.py`).
- Manual repro on mini (no runtime installed): output shown above.

## Notes

AI-disclosure: authored by Claude (Fable 5) during self-driven release dogfooding, reviewed via codex + pr_validate.

https://claude.ai/code/session_012QDdb5mYeBw8qqApuxX6F2

### DIFF
--- a/tests/test_ltx25_video.py
+++ b/tests/test_ltx25_video.py
@@ -239,6 +239,29 @@ def test_ltx25_sanitize_redacts_query_tokens_and_bearer() -> None:
     assert "https://index.example/simple?token=***" in out
 
 
+def test_ltx25_sanitize_redacts_signed_url_params() -> None:
+    """Signed-URL auth params don't end in the classic credential
+    suffixes (codex on #2166): AWS presigned ``X-Amz-Signature``, Azure
+    SAS ``sig``/``sas``, generic ``auth``/``jwt`` must all redact, since
+    uv stderr can echo the full index URL query string."""
+    from vllm_mlx.video.ltx25 import _sanitize_diagnostic
+
+    out = _sanitize_diagnostic(
+        "GET https://bucket.s3.example/wheel.whl"
+        "?X-Amz-Credential=AKIA%2F20260820&X-Amz-Signature=deadbeefcafe"
+        "&X-Amz-Expires=300 and https://acct.blob.example/pkg?sv=2024"
+        "&sig=Zm9vYmFy%3D and auth=topsecret9 jwt=eyJ0eXAi.abc"
+    )
+    assert "deadbeefcafe" not in out
+    assert "Zm9vYmFy" not in out
+    assert "topsecret9" not in out
+    assert "eyJ0eXAi" not in out
+    assert "X-Amz-Signature=***" in out
+    assert "sig=***" in out
+    # Non-credential params survive so the diagnostic stays useful.
+    assert "X-Amz-Expires=300" in out
+
+
 def test_ltx25_sanitize_redacts_quoted_credential_values() -> None:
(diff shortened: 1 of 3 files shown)


raullenchai:
## pr_validate summary + codex stopping-rule note

**6 rounds of pr_validate run from the PR worktree.** Final state: every gate green except `codex_review` — `full_unit` 18110 passed (green in r4/r5/r6), targeted 37 passed, lint clean, diff coverage 92.9%.

**codex non-convergence (documented override per the established stopping rule):** every finding through round 6 was addressed (install containment, canonical-alias injection guard, conditional clone+fetch for stale checkouts, temp-file stderr capture, control-char scrubbing, URL-userinfo/percent-encoded redaction, quoted values, suffix-based credential-name matching, any-scheme Authorization redaction, `.git`-gated clone guard). Rounds 5–6 exhibited the classic non-convergence signature: each round invented a further sanitizer spelling variant, and the `[ -d ltx-2-mlx ]` finding was re-graded NIT→BLOCKING between rounds. The sanitizer is now structurally comprehensive (suffix-based name match, all quoting forms, all auth schemes, userinfo, bounded + control-scrubbed), so further enumeration rounds add no expected value.

**healthz flake investigation** (failed in r1–r3 full_unit): p99 measured 50.5/110.3/117.3 ms vs 50 ms budget, correlated with concurrent machine load (29 GB rsync, codex runs, live inference servers on this mini). Exact same command green on plain main (18099 passed) AND from this PR worktree (18108 passed) when the machine is quiet; green in r4/r5/r6. The functional invariant (`/healthz` never calls `engine.get_stats()`) has its own dedicated test — green in every run. Not a regression of this diff (video-lane only).

PR #2173, opened 2026-08-20
feat(mtp): expose tested experimental MTP presets

raullenchai:
## Summary
- honor MTPLX runtime contracts for Qwen native MTP artifacts
- expose opt-in `Experimental MTP` toggles for 18 real-weight-smoked Qwen aliases
- keep MTP off by default and warn that it may be slower and requires a restart/download
- keep HY3 and Gemma 4 out of the GUI roster until they can be validated end-to-end

## Real-weight validation
Mac Studio MTP-enabled generation passed for:
- Qwen 3.5 4B: 4/6/8-bit
- Qwen 3.5 9B: 4/6/8-bit
- Qwen 3.6 27B: 4/6/8-bit and UD
- Qwen 3.6 35B-A3B: 4/6/8-bit and UD
- both existing Qwen 3.8 27B aliases

Canonical Qwen 3.6 aliases share the exact tested 4-bit HF paths.

## Tests
- Studio: 2,967 registry/CLI/MTP tests passed
- Studio: 25 Swift ModelPerfConfig tests passed
- Mini: 103 registry/CLI/MTP tests passed
- Qwen real-weight smoke logs retained under `/tmp/mtp-smoke-*.json` on Studio

## Defaults
No alias enables MTP automatically. Performance is deliberately not a landing gate for this experimental, user-controlled toggle.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ModelCatalog.swift
@@ -61,7 +61,9 @@ struct SpeculativeDecodingPreset: Codable, Sendable, Hashable {
     let model: String?
     let tokens: Int?
 
-    var displayName: String { method == .mtp ? "MTP" : "Suffix decoding" }
+    var displayName: String {
+        method == .mtp ? "Experimental MTP" : "Suffix decoding"
+    }
 }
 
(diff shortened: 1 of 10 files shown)


raullenchai:
@codex review

PR #2179, opened 2026-08-21
fix(gemma4): upgrade stale chat templates

raullenchai:
## What does this PR do?

Upgrades exact known stale Gemma 4 checkpoint chat templates to the matching Google canonical contract at the shared serving entry point. E2B/E4B use the compact template; 12B/26B/31B use the full template. Already-current canonical templates and any modified/custom templates are preserved.

The two Google templates are bundled in the wheel so serving remains offline and deterministic. A small compatibility correction serializes JSON Schema union types as arrays instead of Python-list strings.

## Why is this needed?

Discussion #1062 exposed a second class of Gemma 4 compatibility failures: converted MLX checkpoints can retain an older template that renders JSON null as Python None, accepts malformed string arguments, and omits corrected continuation framing after tool results. Parser output was already structured correctly; replaying multi-turn tool history was the stale contract.

Refs #1062.

## AI assistance disclosure

Codex investigated the discussion, compared converted checkpoints with current Google templates, implemented the runtime upgrade helper and regression tests, and performed iterative adversarial review. The two Jinja files are based on the Google canonical templates under their Apache-2.0 header; the union-type serialization correction is covered by a rendered declaration assertion. Verification included a deliberate negative-control mutation, focused and full local suites, wheel inspection, and real-model tests on the local Mac and Studio runner.

> By submitting this PR I confirm I can explain the intent, risk, and behavior of every non-generated change in this PR. The stale-template allowlist hashes were verified against every Rapid Gemma 4 alias repository carrying a standalone template.

## Test plan

- [x] Ruff check and format check on changed Python files.
- [x] 181 focused chat-template/parser/router tests passed in final validation.
- [x] Full non-MLLM run: 18,148 passed, 86 skipped, 22 deselected, 6 xfailed, 1 xpassed; four failures reproduced identically on clean upstream/main.
- [x] Negative control: disconnecting the production upgrade hook made the null/string-argument and tool-continuation regressions fail.
- [x] Wheel build includes both Jinja assets and the upgrade helper.
- [x] Local real gemma-4-e2b-4bit: structured tool_calls and multi-turn tool-result continuation returned alpha.txt.
- [x] Studio real gemma-4-e2b-8bit contract: structured tool_calls, multi-turn alpha.txt response, and canonical compact-upgrade log passed ([run](https://github.com/raullenchai/Rapid-MLX/actions/runs/32434347240)).
- [x] Iterative Codex review converged with zero blocking findings; the upstream-template O(n²) NIT is explicitly dismissed as out of scope in the PR comment.
- [x] Final local PR validation verdict: MERGE-SAFE. Full-unit and stress were environment-skipped in the final run after the prior full run/main A-B proof and Studio live-model pass.

## Checklist

- [x] Tests pass locally under SOP exclusions; pre-existing failures are documented above
- [x] Self-validated with `python3 -m scripts.pr_validate 2179 --verbose`
- [x] New critical-path tests fail when the production wiring is deliberately disconnected
- [x] README/docs updated if applicable (not applicable: runtime compatibility fix with no user-facing option)

## Out of scope

The Google canonical templates forward-scan later messages to find the next non-tool role, which is O(n²) for unusually long histories. This is inherited upstream behavior, not introduced by the upgrade wrapper; rewriting the vendored turn-state algorithm is intentionally deferred to avoid expanding this compatibility patch.

### DIFF
--- a/.github/workflows/agent-gate.yml
+++ b/.github/workflows/agent-gate.yml
@@ -78,82 +78,19 @@ jobs:
     timeout-minutes: 90
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
-      - name: Prepare branch audit environment
-        run: |
-          /opt/homebrew/opt/python@3.12/bin/python3.12 -m venv /tmp/rapid-mlx-gemma4-audit-venv
-          /tmp/rapid-mlx-gemma4-audit-venv/bin/pip install -q .
-      - name: Gemma 4 canonical-template contract
-        run: |
-          set -euo pipefail
-          RMLX=/tmp/rapid-mlx-gemma4-audit-venv/bin/rapid-mlx
-          LOG=/tmp/rapid-mlx-gemma4-template-smoke.log
-          "$RMLX" serve gemma-4-e2b-8bit --no-mllm --no-thinking --port 8193 >"$LOG" 2>&1 &
-          server_pid=$!
-          trap 'kill "$server_pid" 2>/dev/null || true' EXIT
-          for _ in $(seq 1 120); do
-            curl -fsS http://127.0.0.1:8193/v1/models >/dev/null 2>&1 && break
-            kill -0 "$server_pid" 2>/dev/null || { cat "$LOG"; exit 1; }
-            sleep 5
-          done
-          /tmp/rapid-mlx-gemma4-audit-venv/bin/python - <<'PY'
-          import httpx
-
-          base = "http://127.0.0.1:8193/v1/chat/completions"
-          tool = {
-              "type": "function",
-              "function": {
-                  "name": "list_files",
-                  "description": "List files",
-                  "parameters": {
-                      "type": "object",
-                      "properties": {"path": {"type": "string"}},
-                      "required": ["path"],
-                  },
-              },
-          }
-          first = httpx.post(
-              base,
-              json={
-                  "model": "gemma-4-e2b-8bit",
-                  "messages": [{"role": "user", "content": "List files."}],
-                  "tools": [tool],
-                  "tool_choice": "required",
-                  "max_tokens": 128,
-              },
-              timeout=120,
-          )
-          first.raise_for_status()
-          choice = first.json()["choices"][0]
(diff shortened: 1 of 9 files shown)


raullenchai:
Final local validation: **MERGE-SAFE**. Codex reported zero blocking findings; lint and 181 targeted tests passed.

NIT disposition: the forward scan in the bundled compact/full templates is inherited from the Google canonical templates. Reworking the vendored turn-state algorithm would materially expand this compatibility fix and increase divergence risk, so the O(n²) long-history optimization is intentionally out of scope. Normal chat histories are bounded, and this PR does not make that upstream algorithm worse.

raullenchai:
| Item | Result |
|---|---|
| Reproduction | Stale mlx-community Gemma 4 templates rendered JSON null as Python `None`, mishandled string arguments, and lacked corrected post-tool continuation; current Rapid parser output itself remained structured. |
| Root cause | Converted checkpoints retained pre-2026-07-09 chat templates; the shared rendering path trusted checkpoint template data unchanged. |
| Fix | Exact audited stale-template hashes are upgraded offline to compact/full Google canonical contracts; model identity and size select the variant, while canonical, modified, and custom templates remain untouched. Union JSON Schema types are serialized correctly. |
| Wiring | `vllm_mlx.utils.chat_template.apply_chat_template` invokes the upgrade before the shared BatchedEngine render path. |
| Tests | Negative control failed as intended; 181 targeted passed; full run had 18,148 passed with four failures reproduced on clean main; wheel assets verified; local E2B 4-bit and Studio E2B 8-bit structured/multi-turn contracts passed. |
| `pr_validate` | Final verdict `MERGE-SAFE`; Codex converged after exact-hash, explicit-size, union-type, and metadata findings were fixed; zero blocking findings in the final round. |
| Land | PR #2179 squash-merged to `main` at `678ee5e7`; all final GitHub checks were completed green. |
| Elapsed | Worked for approximately 1 hour from final reproduction/Studio setup through verified land. |

Iterative path: review caught and corrected false-positive replacement risk, union-schema rendering, and metadata-only variant selection before merge.

PR #2184, opened 2026-08-21
fix(markdown): draw the SVG preview the right way up

xiaoxiunique:
The SVG preview on `main` renders upside down. Any document that is not
vertically symmetric — a logo above its caption, a chart with a legend, a
flowchart — comes out inverted.

## What happens

`drawPreview` mirrors the destination coordinate space *and* passes
`respectFlipped: true`. NSImage already compensates for the flipped view, so
the hand-written mirror flips it a second time.

## How it got there

The manual mirror was added to satisfy a test whose expectation had been
inverted one commit earlier.

- `af7f4482` changed the expectation from `red < blue` to `red > blue`, on
  the stated grounds that "bitmap rows count from the bottom". They do not:
  `NSBitmapImageRep.colorAt(x:y:)` numbers rows from the top. The inverted
  expectation therefore requires an upside-down picture.
- `e1b4adcb` then supplied one, by adding the mirror.

Both are green, because in each case the test and the implementation moved
together. This is not a gap in the test's coverage — the test executes the
drawing code and samples its pixels. It is a wrong premise about the
framework, which made a correct-looking assertion demand the defect.

## Evidence

Rendering a real document through the drawing path and comparing it against
a reference render, rather than reading the code:

| | result |
|---|---|
| reference (no flipped view, no mirror) | circle above the caption |
| `main`'s drawing path | caption above an inverted circle |

## The change

1. Drop the manual mirror. `respectFlipped: true` is the whole of it.
2. Restore the expectation to `red < blue`.
3. Add `rowOrderIsTopDown`, which measures the row order against a picture
   drawn with no flipping at all.

(3) is the part worth keeping. The orientation test is only as good as its
idea of which end of the bitmap is the top, and being wrong there is
invisible — it does not weaken the assertion, it reverses it. Pinning the
premise separately is what stops the same inversion happening again.

## Verification

- Full suite on this branch: 2672 tests, no failures other than a
  pre-existing DNS test that needs network.
- Mutation check: reintroducing the mirror fails
  `The preview is drawn the right way up` with `(red → 188) < (blue → 88)`.

Independent of any other work in flight — it touches only `drawPreview` and
the orientation tests.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Render/MarkdownCodeBlockView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/Markdown/TextKit/Render/MarkdownCodeBlockView.swift
@@ -232,20 +232,17 @@ final class MarkdownCodeBlockView: NSView {
     /// press away.
     @discardableResult
     private func drawPreview(contentWidth: CGFloat) -> Bool {
-        guard isShowingPreview, let previewImage,
-              let context = NSGraphicsContext.current?.cgContext else { return false }
+        guard isShowingPreview, let previewImage else { return false }
         let size = SVGPreview.drawSize(for: previewImage.size, inWidth: contentWidth)
         guard size.width > 0, size.height > 0 else { return false }
         let origin = CGPoint(
             x: options.codeInsets.leading,
             y: headerHeight + options.codeInsets.top
         )
-        // NSImage compensates for the flipped view. Mirror the destination
-        // coordinate space around this rect first so the image's document-top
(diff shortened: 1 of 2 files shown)


raullenchai (approved):
Reviewed to convergence on rebased head b3d0a816. Studio reproduction confirmed main renders the SVG upside down; the corrected pixel-orientation test fails on main and passes with this PR. Two independent Codex review rounds found 0 blockers. PR validation is MERGE-SAFE (18,254 Python tests passed); targeted Swift orientation tests pass. Full Swift failures were reproduced on main as unrelated pre-existing timing flakes.

PR #2187, opened 2026-08-21
fix(mac): default to the lowest image resolution option

osdodo:
## Why

On-device diffusion cost and peak memory scale with pixel count. Starting at 512 × 512 gives a fresh user a substantially lighter first render and lowers swap/OOM risk on small-memory Macs; 1024 × 1024 remains one selection away for users who prefer detail.

## Change

The Images tab now defaults to the lowest resolution option (512 × 512) instead of 1024 × 1024.

## Validation

On the Mac Studio, the corrected default-value test fails against current `main` (`1024x1024`) and passes with this PR. All seven image-resolution contract tests pass.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Images/ImageGenViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Images/ImageGenViewModel.swift
@@ -72,7 +72,10 @@ final class ImageGenViewModel {
     // MARK: - Composed input
     var prompt: String = ""
     var aspect: Aspect = .square
-    var resolution: Resolution = .detailed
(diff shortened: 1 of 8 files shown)


raullenchai (approved):
Reviewed to convergence on rebased head 83ee728d. Studio reproduction confirmed a fresh image view model defaults to 1024×1024 on main; the PR changes that to 512×512 and all 7 image-resolution contract tests pass. Two Codex review rounds found 0 code blockers; the PR-description rationale blocker was corrected. PR validation: MERGE-SAFE, 18,254 Python tests passed.

PR #2188, opened 2026-08-21
feat(mac): Audio two-lane redesign — dictation download cliff + shared readiness UX [skip-version-bump]

raullenchai:
## Summary

Three changes to the Audio surface, in dependency order:

**1. Dictation model download cliff (the original POC).** Selecting a dictation model lit the Ready light even when the weights were not on disk; the first hotkey press then triggered a silent multi-hundred-MB server-side download with no progress and no cancel, reading as a hang.
- Readiness gains a `modelOnDisk` bit sourced from the audio catalog; `enable()` refuses a model that is not downloaded, with an actionable message.
- Download state renders through the shared `ReadinessBanner` + `AudioView.audioDownloadReadiness` truth table — the same UX as Chat, Images and the sibling Audio tab. Pulls go through the app-wide `DownloadManager` with `totalBytes` seeded, so percent and ETA are present from the first tick. `ensureServing` never hands the server a repo to fetch, killing the silent-download path structurally.
- Download completion auto-refreshes the catalog, flips readiness, and prewarms (only while dictation is enabled) without another visit to the pane.
- The HUD gains a `failed` state so mid-flow errors are named instead of "Transcribing…" vanishing.

**2. One trailing alignment line (`PopupControlChrome`).** AppKit's `NSPopUpButton` draws its bezel inset inside its layout frame, so raw SwiftUI `Picker`s never sat on the same trailing line as the app's buttons. The Audio tabs' self-drawn popup chrome is extracted into a shared component; Dictation's Model/Hotkey pickers move to `Menu` + chrome (cache glyphs and system checkmarks in the menu), and the banner's inset padding is retuned (xs outside + md inside = lg) so its text sits on the rows' content line and its CTA on the control line. Verified by pixel measurement: toggle, buttons, popups, and banner CTA within ±1 device pixel.

**3. Audio is two lanes: Speech to Text and Text to Speech.** The file-upload Transcription workbench is removed — dictation *is* the speech-to-text product, and a second STT surface behind a third tab split one capability across two UIs. Tab labels now name the capability; AX identifiers (`Audio.Mode.Dictation`/`Audio.Mode.Speech`) stay stable. Text to Speech adopts the same setup-card grammar as Speech to Text (label + caption left, controls on the shared trailing line, banner inset under the Model row, hidden once ready), voices auto-load when the model turns ready (Load Voices becomes a quiet refresh with the same identifier), and control width is unified at 260 across both tabs.

## Contract change vs #2053

The old `audio-readiness` golden flow asserted the dictation pane exposes **no** readiness banner, per #2053's "load on demand" contract. That contract covered *loading*, not *downloading*: the flow is rewritten so an uncached dictation model must expose the shared explicit-Download lane, pressing it must invoke `pull`, and finishing the download while dictation is off must **not** start the model.

## Testing

- `swift build -c release` clean; `swift test --filter "DictationTests|AudioCatalogTests|AudioClientTests"` — 39 tests, 3 suites, green (stale pins on the three-tab contract and the deleted file-picker caption updated to the new product decision).
- Live-verified on the Mac Studio with a real 1.4 GB `whisper-medium` pull: needsDownload banner → live bytes/percent/speed/ETA + determinate bar → auto-flip to ready + prewarm, without leaving the pane. Both tabs screenshot-verified for the single trailing line.
- `gui-golden-flows.sh` passes `bash -n`; the `audio-readiness.speech` AX baseline and the golden-flow run need a CI pass to regenerate (`audio-readiness.transcription` baseline deleted with its tab). Dictation's Model/Hotkey AX roles change from pop-up button to menu button, so AX golden baselines will need the usual refresh from the CI artifact.


raullenchai:
## 🤖 Codex review — round 1

- apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:221 — `enable()` suspends while refreshing the catalog but never rechecks `isEnabled` afterward. If the user turns dictation on and then off while the catalog subprocess is running, `disable()` executes first, then the stale `enable()` continuation can install the hotkey and set `phase = .idle` while the UI and persisted preference remain off. No test exercises this lifecycle race.

- apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:506 — The mid-session deletion check relies on cached `catalogByAlias` state, but nothing refreshes it when `DownloadManager.cacheGeneration` changes. If an enabled model is deleted through Model Management, the next hotkey press still records. Later, `catalogFacts` returns the stale `cached == true` entry and passes its non-nil Hugging Face repo to `ensureServing`, allowing an unannounced redownload—the behavior this PR intends to prevent. The new golden flow tests download completion only; they do not exercise deletion after enabling.

VERDICT: CHANGES_REQUESTED

raullenchai:
## 🤖 Codex review — round 1 fixes

Implemented locally in commit `37b639ea`; push follows immediately.

- Prevented an in-flight `enable()` catalog refresh from re-arming the hotkey after Dictation is switched off.
- Revalidated the selected transcription model on disk at the hotkey boundary, and made that check cancellable so a deleted model cannot trigger a silent re-download.
- Added regression coverage for both races.
- Rejected findings: none.

Validation: 41 Audio/Dictation Swift tests passed; 69 AX contract tests passed; `git diff --check` and GUI golden-flow shell syntax passed.

raullenchai:
## 🤖 Codex review — round 3

1. apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:501 — A second hotkey tap is discarded while the new asynchronous catalog refresh is pending. Since `phase` remains `.idle` and `beginRecordingTask` is non-nil, the tap cannot cancel the request. If catalog probing takes several seconds, a user can tap once, tap again to abort, and then have the microphone start unexpectedly when the probe completes. The new tests call only the cache-refresh helper and never exercise this double-tap behavior.

2. apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:360 — Every refresh replaces known-good cache facts with the loader result, but `ModelCatalog.audioEntries` represents any subprocess failure as an empty array. A transient failure of either `rapid-mlx models` or `rapid-mlx ls` therefore makes a cached model appear deleted: launch-time enablement turns the persisted switch off, or a hotkey press refuses to record with “isn't on disk anymore.” Preserve the last successful snapshot or propagate catalog failure separately from a valid empty catalog.

VERDICT: CHANGES_REQUESTED

raullenchai:
## 🤖 Codex review — round 3 fixes

Implemented locally in commit `75133a58`; push follows immediately.

- A second hotkey tap now cancels the pending asynchronous disk check, preventing delayed surprise microphone activation.
- Catalog probes now retain subprocess success/failure; Dictation replaces its cache snapshot only after both `models` and `ls` succeed, so transient CLI failures no longer look like model deletion.
- Added regression tests for double-tap cancellation and preservation of the last successful cache snapshot.
- Rejected findings: none.

Validation: 56 Dictation/Audio/catalog Swift tests passed; 69 AX contract tests passed; `git diff --check` and GUI golden-flow shell syntax passed. The automated review loop has reached its configured three-round cap; this commit proceeds to PR validation.

PR #2189, opened 2026-08-21
feat(desktop): voice co-loads with LLM/VLM on the same server

raullenchai:
## Summary

The desktop previously ran speech (STT/TTS) on its own audio-only process: an audio request swapped the primary chat LLM/VLM away ("This version runs one model at a time"), so voice and text/vision could never coexist.

The server already supports mounting the `/v1/audio/*` lane on a text-mode boot via `--enable-audio`, with the STT/TTS engines loaded lazily on first request. This PR wires the desktop into that so speech runs side-by-side with the primary LLM/VLM in **the same process**.

## Changes

- **`ServerManager.serveArguments`** now always appends `--enable-audio` (placed before `--cors-origins` to keep its last-flag `nargs+` contract), so every spawned server (LLM, VLM, or audio) carries the voice lane and lazy-loads the chosen STT/TTS engine only when asked.
- **New `ServerManager.voiceCoLoadsOnPrimary` + `ensureVoiceLane`**: when a model is already serving, voice traffic targets that same process (no swap); only when nothing is running do we fall back to a voice-only server.
- **`AudioViewModel` and `DictationController`** route through `ensureVoiceLane` instead of forcing `residencyEligible:false` process swaps.
- **`AudioView`**: readiness reports the selected voice model as ready when it co-loads with a live primary; the obsolete "runs one model at a time" `swapNotice` is removed.
- **`DictationController.warmUpEngine`** now also prewarms (lazy-loads the STT engine) when reusing the primary server; stale "process swap" copy removed.

## Tests

- `SpawnArgumentsTests` pins `--enable-audio` before `--cors-origins`.
- New `VoiceCoLoadTests` covers the co-load decision + `ensureVoiceLane` contract (incl. crashed/stopped states and the ready+bearer requirement).
- `AudioCatalogTests` / `MCPConnectorsTests` updated for the new argv/activation call.
- Full desktop suite passes (2681 tests); the only reds are timing/throughput suites that also flake under full-parallel load and pass standalone (environment, unrelated to this change). Server audio tests (80) pass with no regression.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Audio/AudioViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Audio/AudioViewModel.swift
@@ -117,10 +117,9 @@ final class AudioViewModel {
         errorMessage = nil
         transcription = nil
         defer { isTranscribing = false }
-        guard await server.ensureServing(
+        guard await ensureVoiceLane(
             alias: entry.alias,
-            hfPath: entry.hfRepo,
-            residencyEligible: false
+            hfPath: entry.hfRepo
         ) else {
             errorMessage = "The audio model couldn't start. Audio support may be unavailable in this app build."
             return
@@ -145,10 +144,9 @@ final class AudioViewModel {
         isLoadingVoices = true
         errorMessage = nil
         defer { isLoadingVoices = false }
-        guard await server.ensureServing(
+        guard await ensureVoiceLane(
             alias: entry.alias,
-            hfPath: entry.hfRepo,
-            residencyEligible: false
+            hfPath: entry.hfRepo
         ) else {
             errorMessage = "The speech model couldn't start. Audio support may be unavailable in this app build."
             return false
@@ -187,10 +185,9 @@ final class AudioViewModel {
         errorMessage = nil
         synthesizedAudio = nil
         defer { isSynthesizing = false }
-        guard await server.ensureServing(
+        guard await ensureVoiceLane(
             alias: entry.alias,
-            hfPath: entry.hfRepo,
-            residencyEligible: false
+            hfPath: entry.hfRepo
         ) else {
             errorMessage = "The speech model is no longer running. Load its voices and try again."
             return
@@ -219,10 +216,9 @@ final class AudioViewModel {
         errorMessage = nil
         defer { previewingVoice = nil }
 
-        guard await server.ensureServing(
+        guard await ensureVoiceLane(
                   alias: entry.alias,
-                  hfPath: entry.hfRepo,
-                  residencyEligible: false
+                  hfPath: entry.hfRepo
               ),
               !Task.isCancelled else {
             if !Task.isCancelled {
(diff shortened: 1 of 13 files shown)


raullenchai:
## Review finding (self-review round 3, end-to-end smoke)

I ran a real end-to-end smoke test to validate the PR's core assumption that a **text primary server + `--enable-audio` can lazily run STT in the same process**:

- Text server `qwen3-0.6b-4bit --enable-audio` → `/health` 200, `/v1/models` lists the text model, `/v1/audio/*` lane **mounted** (not 404), ASR engine **lazy-loads** (`STT model loaded`), and the text model **stays resident alongside** — i.e. process/route/model **co-existence works**.
- ❌ But the actual STT **transcription** fails in that shared process with `RuntimeError: There is no Stream(gpu, 1) in current thread`.
- ✅ **Control**: the identical transcription on a **pure audio-only server** returns HTTP 200.

Root cause: MLX GPU streams are thread-local. The STT engine executes inline on FastAPI's event-loop thread, which has no GPU stream in the context where the text model's MLX stream was established. The audio-only mode never exercises this because it is single-purpose.

**This is a pre-existing server-side defect** (my PR does not touch server code; `--enable-audio` is existing behaviour), but this PR is what first routes the desktop onto this path, so it directly affects whether "voice actually runs alongside LLM/VLM" today on GPU.

So the mechanism (same-process co-loading) is delivered, but GPU transcription in the reused process needs a server-side stream fix (e.g. run STT/TTS inference on an appropriately-streamed thread/executor) before voice *inference* truly works alongside a text model. Open question for the owner: fix it in this PR, or track as a follow-up?

raullenchai:
## Server-side fix added + verified remotely (commit `04c67e7f`)

Following the earlier finding (audio lane crashed with `There is no Stream(gpu, N) in current thread` inside a text primary process), this PR now includes the **server fix**, so the desktop's same-process co-loading actually works end-to-end on GPU.

### Root cause
`mlx_lm.generate.generation_stream` is a **module-global** the text engine rebinds to its own MLX worker thread's stream at startup (`engine_core._init_mlx_step_thread`). The audio lanes ran inline on the asyncio loop thread, so their `mx.eval` touched that worker-owned stream from a thread that doesn't own it.

### Fix (`vllm_mlx/routes/audio.py`)
- `_resolve_mlx_worker_core()` walks the nested `BatchedEngine` topology (`server._engine._engine.engine`) to find the worker-owning `_mlx_executor` — same path `batched.py` uses.
- `_run_audio_mlx` (async, `run_in_executor` + await) for STT; `_run_audio_mlx_sync` (via `_run_on_step_thread`) for the off-loop TTS/music/alignment lanes.
- Load **and** inference routed onto that worker so weight-array streams and infer-time evals stay on one thread.
- Falls back to inline when no primary engine worker exists (audio-only server) — unchanged there.

### Verification (remote Mac mini, real GPU)
| case | result |
|---|---|
| text primary `--enable-audio` → ASR transcription | **HTTP 200** (was 500) |
| …text chat still works after voice | HTTP 200 (co-exist) |
| audio-only server → ASR transcription (control) | HTTP 200 (no regression) |
| audio route + alias (17) unit tests + new `test_audio_mlx_worker.py` (7) | all pass |

The desktop change (`--enable-audio` + reuse primary) now equals real GPU co-loading, not just same-process residency.

raullenchai:
Closing this PR as superseded by the narrower implementations now on `main`:

- #2282 landed the server-owned audio model-worker and lifecycle contract at `1629a51b9e1a333ca80d615b94569e38bef3746b`.
- #2307 landed the Desktop shared voice-lane behavior and preserved the active conversation model through dictation at `21dee049072b262151fa36a52ea0da5c00e1b02a`.

Both commits are confirmed ancestors of current `main@33dd7a3009bca9ebb72ebca9340ed5dd557c6756`. Together they cover this PR’s server and Desktop intention with stronger lifecycle and regression evidence, so keeping the older conflicting implementation open would duplicate the shipped path.

PR #2186, opened 2026-08-21
fix(doctor): recognize the desktop audio dependency profile

osdodo:
Teach `rapid-mlx doctor` to recognize the managed desktop sidecar and validate its audio installation against the intentional `audio-desktop` dependency profile instead of the full `audio` extra.

  The change:

  - checks desktop audio dependencies against `mlx-audio` and `soundfile`;
  - preserves the full audio-extra validation for ordinary CLI installations;
  - avoids recommending `pip install` inside managed desktop environments;
  - provides repair guidance appropriate to embedded and runtime-override sidecars;
  - treats optional dependencies intentionally excluded from the desktop bundle as not required;
  - adds regression tests locking the doctor contract to the `audio-desktop` extra.

  Fixes #2185.

  The signed/notarized desktop bundle intentionally installs `rapid-mlx[audio-desktop]`, but `rapid-mlx doctor` evaluated it against the full `rapid-mlx[audio]` dependency set. As a result, healthy desktop installations were reported as incomplete and users were told to mutate the app's managed Python environment with `pip install`.

  Doctor should reflect the dependency contract of the environment it is inspecting while retaining the existing full-extra diagnostics for regular CLI installations.

Issue #2185, opened 2026-08-21
fix(doctor): recognize the desktop audio dependency profile


raullenchai:
## Summary

The signed/notarized 0.12.18 RC desktop bundle installs the intentional `audio-desktop` dependency set, but `rapid-mlx doctor` evaluates it as though the full `audio` extra were required and reports an incomplete audio installation.

## RC evidence

Tested the workflow-dispatch artifact from source SHA `e7b3687b1667708b4018da39816930e14a3be709` on an Apple M3 Ultra Studio:

- `rapid-mlx 0.12.18`
- Python 3.12.13
- `mlx-audio 0.4.3` present
- desktop sidecar built through the release workflow and passed signing/notarization

Doctor output:

```text
⚠ mlx-audio (audio extras) 0.4.3 incomplete — missing: f5-tts-mlx, numba,
  tiktoken, misaki, spacy, num2words, loguru, espeakng-loader,
  phonemizer-fork, cn2an (`pip install 'rapid-mlx[audio]'`)
```

The desktop builder intentionally installs `rapid-mlx[audio-desktop]`, not the full general-purpose `audio` extra.

## Expected

When doctor is running from the bundled desktop sidecar, validate the `audio-desktop` contract and avoid recommending mutation of the signed app's embedded Python environment. Keep the existing full-extra check for ordinary CLI installations.

## Impact

Small diagnostic UX issue; no RC blocker observed. Core completion, SSE, forced tool calling, bundle signature, Gatekeeper, and notarization checks passed.

PR #2190, opened 2026-08-21
fix: preserve terminal reasoning source order

### LINKED ISSUE #2182: reasoning_max_tokens: parser content on the final (finished=True) chunk is dropped after cap-hit
Pre-existing on main (verified independent of #2171 — reproduces identically with the qwen3 parser on a #2171-free checkout), found while probing the cap machinery during #2171's review convergence.

## Repro

`StreamingPostProcessor` with `reasoning_max_tokens=1`, any think-tag parser:

```python
chunks = ["<think>delib", "erate</think>4"]   # second chunk has finished=True
# expected: reasoning='deli', content='b' (overflow) + 'erate' (spill) + '4' (answer)
# actual:   reasoning='deli', content='b' — the answer '4' and the spill are DROPPED
```

The same content arriving on a NON-final chunk (with a separate empty finished chunk after it) is delivered correctly — the parser returns `content='erate4'` and it reaches the stream. Only the finished=True chunk's parser content is lost, and only after `_reasoning_cap_hit`.

## Impact

Any request combining `reasoning_max_tokens` (including Anthropic-adapter effort mapping) with a model whose final token batch carries the post-thought answer silently loses that answer. Severity limited by the cap being opt-in per request.

Suspected area: the finished-chunk event assembly path interacting with cap-hit state (the non-final path emits correctly, so the split/flip logic itself is sound).

raullenchai:
## What does this PR do?

- Preserves the original source order of reasoning and content when a per-request reasoning cap spills reasoning bytes into visible content.
- Carries ordered `reasoning` / `content` segments through multi-block `<think>` parsing and in-reasoning tool-call promotion.
- Keeps the legacy content-first fallback for parsers that do not expose ordered segment provenance.
- Retains trailing reasoning prose outside malformed, unclosed tool-call buffers.

Fixes #2182.

## Why is this needed?

A capped Qwen reasoning stream could assemble a terminal chunk out of source order: chunks `"<think>delib"` and `"erate</think>4"` produced `4erate` instead of `erate4`. A simple prepend fixed that shape but reversed other valid interleavings, including promoted tool calls, multiple think blocks, and content-first DeepSeek V4 deltas. The parser now carries explicit ordered segment provenance so the postprocessor can reclassify only the over-budget reasoning suffix without moving it across adjacent content.

## Scope note

The draft originally also addressed #2185. While this PR was in review, #2186 landed on `main` with a more complete managed-sidecar provenance implementation. This branch merged that change and dropped its superseded environment-marker implementation; the remaining diff is only the #2182 reasoning-order fix.

## AI assistance disclosure

Codex inspected the issue report, implemented the production and test changes, and ran iterative adversarial review. Every finding was either fixed with a focused regression or removed when superseded by upstream `main`.

> By submitting this PR I confirm I can explain the intent, risk, and behavior of every non-generated change in this PR. For any generated / boilerplate / scaffolded sections, I have identified them above and can describe how I verified them.

## Test plan

- Mac mini: reasoning/parser suite — 461 passed, 1 skipped.
- Codex review on final semantic head: two consecutive clean rounds, independently on Mac mini and Mac Studio.
- Mac Studio: `python3.12 -m scripts.pr_validate 2190 -v` — **MERGE-SAFE**.
  - targeted + negative control: 350 passed, 2 skipped.
  - full unit: 18,255 passed, 102 skipped, 22 deselected, 6 xfailed, 1 xpassed.
  - Codex adversarial review: PASS.
  - ruff check + format: PASS.

### DIFF
--- a/apps/rapid-mac/Tests/RapidTests/SidecarBuildScriptTests.swift
+++ b/apps/rapid-mac/Tests/RapidTests/SidecarBuildScriptTests.swift
@@ -92,6 +92,7 @@ struct SidecarBuildScriptTests {
     @Test("Desktop sidecar bundles and smokes both audio lanes")
     func audioRuntimeIsBundled() throws {
         let script = try String(contentsOf: Self.scriptURL, encoding: .utf8)
+        let shim = try String(contentsOf: Self.shimURL, encoding: .utf8)
         let pyproject = try String(contentsOf: Self.pyprojectURL, encoding: .utf8)
 
         #expect(pyproject.contains("\naudio-desktop = ["),
@@ -100,6 +101,8 @@ struct SidecarBuildScriptTests {
         #expect(pyproject.contains(#""soundfile>=0.12.0""#))
         #expect(script.contains(#""${RAPID_MLX_SOURCE}[audio-desktop]""#),
                 "The desktop sidecar must install the bounded desktop audio dependency set.")
+        #expect(shim.contains("export RAPID_MLX_DESKTOP_SIDECAR=1"),
+                "Doctor needs sidecar provenance to validate audio-desktop instead of full audio.")
         #expect(script.contains("from mlx_audio.stt.utils import load_model"),
                 "The build smoke must import the transcription loader, not only mlx_audio's package root.")
         #expect(script.contains("from transformers.models.whisper.feature_extraction_whisper import WhisperFeatureExtractor"),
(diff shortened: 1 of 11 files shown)


raullenchai:
Final review and validation record:

- Codex review converged after all findings were addressed; the final two semantic review rounds were clean on `mac-mini` and `mac-studio`.
- Remote focused suite on `mac-mini`: 461 passed, 1 skipped.
- Final `python3.12 -m scripts.pr_validate 2190 -v` on `mac-studio`: **MERGE-SAFE**.
  - targeted/negative-control suite: 350 passed, 2 skipped
  - full unit suite: 18,255 passed, 102 skipped, 22 deselected, 6 xfailed, 1 xpassed
  - adversarial Codex review: PASS
  - Ruff check and format: PASS
- GitHub checks for final HEAD `ecdc44c6` are green.

Scope note: #2185 was superseded on `main` by #2186, so the overlapping doctor/environment-marker changes were dropped while rebasing. This PR's final scope fixes #2182 only.

PR #2193, opened 2026-08-21
fix(mac): surface dictation cold model loading [skip-version-bump]

raullenchai:
## Summary
- keep the global dictation hotkey unregistered while the selected ASR model is loading
- make readiness explicit in the Audio UI: `Loading <model> into memory…` until both the audio sidecar and lazy STT weights are hot
- register the hotkey only after warmup succeeds; model changes, foreground revalidation, disable, and stale async completions cannot bypass that boundary
- if another workload later displaces the audio sidecar, disarm and return to the same loading → ready lifecycle
- add a hermetic Dictation golden journey that observes `Loading → Ready` across a deliberately delayed STT warmup

## User impact
The first hotkey press no longer appears to start recording and then hang in transcription while Rapid secretly loads a model. The UI tells the user that preparation is in progress, and the shortcut becomes active only when dictation is genuinely ready.

## Validation
- `swift test --package-path apps/rapid-mac --filter "DictationTests|AudioCatalogTests|AudioClientTests"` (46 passed)
- readiness regressions: warmup success arms exactly once; warmup failure and disable-during-warmup never arm
- `python3 -m pytest tests/test_gui_control_behavior_contract.py tests/test_gui_preflight_contract.py tests/test_rapid_mac_ax_identifiers.py -q` (86 passed)
- `bash -n apps/rapid-mac/scripts/gui-golden-flows.sh`
- `bash -n apps/rapid-mac/scripts/fake-rapid-mlx.sh`
- `git diff --check`
- local `--flow dictation` attempted; runner stopped at preflight because macOS screen was locked, before app launch. CI reruns the same blocking flow on its GUI runner.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift
+++ b/apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift
@@ -47,6 +47,9 @@ final class DictationController {
     /// one slow" without a log dive.
     private(set) var lastLatencyDetail: String?
     private(set) var elapsed: TimeInterval = 0
+    /// The model is installed but not yet hot. This is user-visible readiness:
+    /// “on disk” alone is not enough to promise an immediate first dictation.
+    private(set) var isPreparingModel = false
     /// Set when the TCC row says Accessibility is granted but this process
     /// still cannot install an event tap — i.e. the grant landed after launch.
     private(set) var accessibilityNeedsRelaunch = false
@@ -305,6 +308,7 @@ final class DictationController {
         beginRecordingRequestID = nil
         prewarmTask?.cancel()
         prewarmTask = nil
+        isPreparingModel = false
         stopTicking()
         recorder.shutdown()
         hud.hide()
@@ -423,6 +427,7 @@ final class DictationController {
             return
         }
         var created: Task<Void, Never>!
+        isPreparingModel = true
         created = Task { [weak self] in
             await self?.performPrewarm()
             // Only the flight that still OWNS the slot may clear it. A
@@ -430,6 +435,7 @@ final class DictationController {
             // task a later enable started, or single-flight breaks.
             if let self, self.prewarmTask == created {
                 self.prewarmTask = nil
+                self.isPreparingModel = false
             }
         }
         prewarmTask = created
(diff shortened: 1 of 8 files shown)


raullenchai:
## 🤖 Codex review — round 1

apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:83 — Changing the selected model while `phase` is `.starting` or `.recording` overwrites the phase with `.preparingModel` without stopping the recorder. The next hotkey press is ignored in `.preparingModel`, so capture can continue indefinitely with no path through `finishRecording()`, leaving the microphone active. Model changes must explicitly terminate/cancel an active capture or be disallowed until it finishes. No test covers this transition.

apps/rapid-mac/Sources/Rapid/UI/DictationView.swift:56 — The download watcher treats any existing `.completed` job as a new completion. `DownloadManager` retains the most-recent completed job, and SwiftUI starts this `.task` whenever the view is recreated. After a model has been downloaded once, every later visit to Dictation calls `modelDownloadDidFinish()`, which calls `enable(replacingCurrentPrewarm: true)`, stops the working hotkey, and performs another warmup. Thus merely reopening the pane temporarily disables an already-ready global dictation service and repeats inference unnecessarily. The watcher must detect the actual running-to-completed transition or make completion handling idempotent. The added tests do not exercise view re-entry with a retained completed job.

VERDICT: CHANGES_REQUESTED

raullenchai:
Addressed Codex review round 1:

- Active `.starting` / `.recording` capture is now explicitly cancelled before a model swap enters preparation, and the model picker is disabled outside `.off` / `.idle`. This prevents a live microphone from being stranded in `.preparingModel`.
- Retained `.completed` download jobs are now idempotent for an already-ready same-alias session, so reopening Dictation preserves the armed hotkey and skips redundant warmup.
- Added regression tests for both transitions.

Validation: `swift test --package-path apps/rapid-mac --filter DictationTests` (27 passed), 86 GUI/AX contract tests passed, and `git diff --check` passed.

raullenchai:
## 🤖 Codex review — round 2

apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:357 — Model changes do not cancel an active `transcribeTask`. If `modelAlias` changes while `transcribe()` is suspended in `ensureModelServing()` or `client.transcribe()`, that task subsequently reads the new mutable alias and can submit the old recording to the newly selected model. Its unconditional `defer` also sets `phase = .idle`, overwriting `.preparingModel`; the concurrent `enable()` then fails its phase guard and leaves enabled dictation unarmed. The new test covers `.recording` cancellation only, not `.transcribing`.

VERDICT: CHANGES_REQUESTED

raullenchai:
Addressed Codex review round 2:

- Every transcription now has a request ID and captures its model alias at creation.
- Model changes cancel and invalidate the active transcription before entering preparation.
- Stale tasks can no longer submit against a new alias, publish errors, clear the new task slot, hide the new HUD, or overwrite `.preparingModel` with `.idle`.
- `ensureModelServing` and retranscription now operate on an explicit captured alias.
- Added a regression test for model changes during `.transcribing`.

Validation: 49 focused Dictation/Audio tests passed, 86 GUI/AX contract tests passed, and `git diff --check` passed.

raullenchai:
## 🤖 Codex review — round 3

apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:547 — Any warm-up transcription error now returns `false`, causing `enable()` to leave dictation unarmed and report an out-of-memory-style “couldn't load” error. A transient HTTP failure, sidecar timeout, or malformed probe response therefore makes an otherwise loaded model unusable until the user manually retries. This also contradicts the nearby documented behavior that a failed probe should merely make the first real dictation pay the loading cost. The hotkey should arm once the model is serving even if the optional weight-warming probe fails, while preserving an accurate diagnostic.

apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift:488 — Replacing a prewarm only cancels its Swift task; the awaited `ServerManager.ensureServing` can continue mutating the shared server after cancellation. For example, while model A is being brought up, selecting model B cancels A and starts B, but A’s non-cancellation-aware stop/start sequence can resume later and replace B. The new enable then either fails its `servingAlias` check and leaves dictation off or the server ends on the wrong model. The request ID protects controller state but not the shared server mutation; replacement needs serialization/latest-request ownership at the server-operation boundary. The added tests replace prewarming with a closure and therefore do not exercise this race.

VERDICT: CHANGES_REQUESTED

raullenchai:
Addressed Codex review round 3:

- Serving readiness is now the hard arm boundary; failure of the optional silence probe records a non-fatal warning and lets a healthy serving model arm, so transient probe errors no longer make Dictation unusable.
- Replacement prewarms now form a serialized task chain. A new model waits for the cancelled predecessor's non-cancellation-aware `ensureServing` stop/start mutation to finish before touching the shared sidecar.
- Added regressions proving probe failure still arms and replacement preparation cannot overlap the prior server mutation.

Validation: 51 focused Dictation/Audio tests passed, 86 GUI/AX contract tests passed, shell syntax checks passed, and `git diff --check` passed.

raullenchai:
## 🤖 Codex review — round 4

VERDICT: LGTM

raullenchai:
## 🤖 Codex review — round 5

VERDICT: LGTM

PR #2280, opened 2026-08-24
fix(mac): smooth streaming Markdown rendering

chen-yijiang:
## What does this PR do?

- Keep completed top-level Markdown blocks stable while recompiling only the mutable tail.
- Decouple transport updates from visual presentation with a frame-paced grapheme buffer.
- Append growing prose through TextKit without replacing the whole rendered block.
- Preserve stable block identity while Markdown structure is still forming.
- Interpolate bottom-follow scrolling instead of applying discrete layout jumps.
- Reconcile authoritative completion content and preserve Dynamic Type, links, reference definitions, and conversation lifecycle behavior.

## Why is this needed?

Long structured replies could visibly flicker, jump, and appear in uneven batches while streaming.

The growing Markdown tail was repeatedly recompiled and replaced as its structure changed. At the same time, transport bursts and discrete scroll corrections caused abrupt text and viewport movement. The effect was most visible with headings, lists, code blocks, inline formatting, and long replies while following the bottom of the transcript.

This change makes streaming presentation continuous while keeping completed Markdown content stable.

## AI assistance disclosure

Codex authored and iteratively reviewed the Swift/AppKit implementation and tests across the 21 files changed under `apps/rapid-mac/Sources/Rapid/UI` and `apps/rapid-mac/Tests/RapidTests`.

I manually reproduced the original streaming behavior and evaluated the updated long-Markdown output and scrolling. The generated changes were additionally verified with focused tests, a release build, boundary review, and removal of unrelated runtime changes from this branch.

## Test plan

- [x] 63 focused streaming, Markdown, TextKit, fade, and scrolling tests passed across 9 suites after rebasing onto current `main`.
- [x] Release app built successfully with `SKIP_SIDECAR=1 BUNDLE_MODEL=0 ./scripts/build.sh`.
- [x] `git diff --check`.
- [x] Manually exercised a long Markdown reply containing headings, short paragraphs, lists, bold text, inline code, and fenced code while scrolling.

## Checklist

- [x] Relevant tests pass locally.
- [x] No Python lint or format changes are applicable to this Swift-only diff.
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 2280`.
- [x] New behavior is covered by focused regression tests.
- [x] README/docs changes are not required.
- [x] No breaking changes to existing APIs.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
@@ -436,6 +436,7 @@ struct ChatView: View {
                         TranscriptScrollPositionProbe(
                             isPinnedToBottom: $isPinnedToBottom,
                             bottomResumeSlack: bottomResumeSlack,
+                            isStreaming: viewModel.isStreaming,
                             scrollToBottomRequest: scrollToBottomRequest
                         )
                     )
--- a/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
@@ -213,11 +213,12 @@ struct PhotoCapabilityNotice: Equatable {
 /// or attachments.
 struct ChatView: View {
     @Bindable var viewModel: ChatViewModel
-    /// Compiled blocks for the streaming message. Keeps the raw streamed
-    /// string out of the streaming view's inputs — see
-    /// `StreamingMarkdownStore`. (`MessageRow` still receives the raw
-    /// `ChatMessage` for non-markdown concerns like tool chips.)
+    /// Incremental documents for assistant messages created in this view.
+    /// A completed row keeps its document so finishing never swaps renderers.
     @State private var streamingMarkdown = StreamingMarkdownStore()
+    /// Messages that started on the incremental renderer keep using it after
+    /// completion. The set changes once per response, not once per token.
+    @State private var incrementalAssistantIDs: Set<UUID> = []
     @Bindable var server: ServerManager
     @Binding var alias: String
     /// The single readiness value for this window, resolved once by
@@ -298,6 +299,20 @@ struct ChatView: View {
         )
     }
 
+    /// Startup failures and app-added grounding sources can change content
+    /// after `streamingBody` disappears. Observe only the retained live row so
+    /// token updates do not compare every completed message in the transcript.
+    private var retainedSettledAssistantBody: ChatViewModel.StreamingBody? {
+        guard let id = streamingMarkdown.documentMessageID,
+              incrementalAssistantIDs.contains(id),
+              let message = messages.first(where: { $0.id == id }),
+              message.role == .assistant,
+              message.status != .streaming else {
+            return nil
+        }
+        return .init(id: message.id, text: message.content)
+    }
+
     var body: some View {
         // The reader exists for one value: the surface's own width, which
         // the lifecycle band needs in order to pick its height on the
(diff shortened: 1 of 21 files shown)


xiaoxiunique (commented):
Reviewed by checking out the head and measuring, because the four concerns in the description are separable and one of them inverts on this app's main content type. Two findings.

---

## 1. The word-by-word fade is switched off, and the description does not say so

`StreamingTextKitMarkdownView` now passes `fadeState: nil` and `fadeConfiguration: .off`. With that, nothing under `Sources/` enables the reveal — `TextFadeAnimationState()` survives only in tests. The `fadeConfiguration` factory went with it, and it carried two gates: `NSWorkspace.accessibilityDisplayShouldReduceMotion` and the `rapid.chat.fade.disabled` default. So did the comment explaining why one state is shared across a whole message ("each block would restart the animation at its own boundary, and a long answer would visibly re-fade at every paragraph").

`TextFadeAnimator.swift` itself is untouched apart from one comment, so this is a disconnection rather than a removal.

The reasonable guess is that the new frame-paced grapheme buffer already reveals text progressively and a second reveal layer would be redundant or fight it. I tested that guess rather than assuming it: hoisting one `TextFadeAnimationState` to the parent and passing it plus the restored `fadeConfiguration` down to each segment is a one-file change, and with both layers running

- there is no visible conflict on a real streamed answer, and
- the suite is unchanged at 2796 tests — the 4 `DMGPresentationScriptTests` `icvp` failures and the NXDOMAIN DNS test fail identically on the unmodified head.

So the architecture does not force this. If it is still the right call, it is a product decision about how streamed text reads, and it deserves a line in the description and its own discussion — the fade was deliberately shaped after ChatGPT's, and the Reduce Motion gate is an accessibility behavior that is now gone either way.

---

## 2. `topLevelStreamingSplit` stops splitting as soon as the answer contains a bracket, and the fallback is slower than the code it replaces

`containsPotentialReferenceLink` runs `(?<!\\)\[[^\]\n]+\](?!\()` over the raw source, which does not know about code spans or fenced blocks. Any `[…]` inside code matches — `arr[0]`, `[pivot]`, `[T]`, a slice, a generic. Because the stable prefix only grows, the first such match disables committing for the rest of the document.

Measured on a real 23 040-character assistant answer from my own history (179 headings, 84 table rows, 60 fences), streamed in 31-character deltas, comparing full-recompile-per-flush against `StreamingMarkdownDocument`:

| | full recompile | incremental |
|---|---|---|
| stable commits | — | **10** (of 744 flushes) |
| compiled bytes | 11,491,691 | 11,060,364 (−3.8%) |
| wall clock | 11.0 s | **17.3 s** |

The same document with only `[`/`]` replaced by `(`/`)`, nothing else changed:

| | with brackets | without |
|---|---|---|
| stable commits | 10 | **358** |
| compiled bytes | 11,060,364 | **78,196** |
| wall clock | 17.3 s | **0.17 s** |

The design is a large win when it engages — 147× less work, 70× faster — and a 1.6× regression when it does not. The regression comes from `topLevelStreamingSplit` calling `Document(parsing:)` over the whole mutable source and then `compile` parsing it again, so a non-splitting append costs two full parses where the previous code cost one.

Cadence makes this heavier than the table suggests. `document.append` is driven from `advancePresentationFrame`, which is the `StreamingPresentationDisplayLink` callback (`ChatView.swift:393`) — once per display frame, 120 Hz on ProMotion, not once per 100 ms coalesce. `StreamingMarkdownStore` is `@MainActor`.

Three things would address it, roughly in order of value:

1. Decide the reference-link question on the parsed tree rather than on raw source, or strip code spans and fenced blocks before matching.
2. Do not let one match disable splitting permanently — the condition should be about the boundary being considered, not about everything before it.
3. When no split is found, skip the second parse; the split probe already produced a document.

---

Everything else here reads well to me. The blank-line boundary rule in `topLevelStreamingSplit` is a nice invariant — it falls out that setext headings, lazy continuations and unterminated fences can never be frozen early, which is more than the table case the comment cites. The move from `id: \.offset` to per-kind stable IDs in `MarkdownBlockStack` is the part that most directly addresses "flicker", and it is 40 lines.

raullenchai (dismissed):
I checked out and reviewed exact head `16996557527dcd7c429b5a27de8de7a59c5c605e`. The incremental Markdown, completion reconciliation, fade/Reduce Motion, link, Dynamic Type, and accessibility paths look well covered; the 63 focused tests passed locally, as did 23 proportional accessibility/table checks.

One scoped behavior regression needs correction before approval. This head removes the existing one-shot release when a streamed answer grows beyond one viewport, and replaces its contract with `A long streaming answer stays pinned without user input`. As a result, a reader who starts at the beginning of a long answer is continuously pulled toward newly arriving text even without asking to follow; the Jump to Bottom control no longer appears at the established automatic release point. Smooth interpolation does not require changing that follow policy.

Please preserve the existing one-shot release/reader-control contract while applying interpolation only while the transcript remains pinned, and cover both sides deterministically: growth beyond one viewport releases once, while an explicit Jump to Bottom re-pins and follows smoothly thereafter. This request is limited to the affected follow lifecycle and its focused tests; no renderer redesign or adjacent scrolling refactor is requested.

The PR is also behind the current base, so the eventual approved head will still need base reconciliation and exact-head gates.

raullenchai (approved):
Approved on exact head `2f7caf9b8` after independent review convergence and focused local validation. The final validator finding about `sourceURL` is a false positive: the first `deletingLastPathComponent()` removes the filename, so three calls correctly land at `apps/rapid-mac`; the test executed and passed. The streaming parity, completion handoff, grapheme buffering, and viewport-release behavior are covered and pass.

PR #2286, opened 2026-08-24
fix(mac): gate residency flags by runtime support

chen-yijiang:
## What does this PR do?

- Probe the selected `rapid-mlx serve` binary before spawning the server.
- Pass `--resident-memory-limit-gb` and `--resident-model-idle-ttl` only when the runtime advertises them.
- Fall back conservatively when the probe cannot run, times out, exits unsuccessfully, or returns invalid output.
- Drain probe output while the child is running so large help text cannot block on pipe backpressure.
- Log when Rapid cannot confirm support and omits the residency flags.

## Why is this needed?

Rapid-MLX Desktop currently passes both residency flags to every server launch.

Older runtime overrides and explicit `RAPID_BIN` installations may not recognize these options. Argument parsing then exits before the server or model starts, while the app only surfaces a generic “Couldn't start” error. This can look like a model-memory failure even when the machine has enough available memory.

The managed-runtime version selection on `main` avoids stale overrides when a newer bundled runtime exists. It does not cover source builds, slim bundles without a bundled sidecar, or an explicitly selected `RAPID_BIN`.

On the reproducing machine, the selected runtime was `rapid-mlx 0.11.1`; its `serve --help` does not advertise either residency flag.

## AI assistance disclosure

Codex authored and iteratively reviewed the Swift implementation and tests in `ServerManager.swift`, `ServerRuntimeCapabilities.swift`, and `ServerRuntimeCapabilitiesTests.swift`.

I reproduced the startup failure with a local `rapid-mlx 0.11.1` runtime override and verified the compatibility behavior with focused tests and manual app testing. The probe's large-output and unsuccessful-exit paths are covered by dedicated regression tests.

## Test plan

- [x] `swift test --filter 'ServerRuntimeCapabilitiesTests|ServerLocatorVersionSelectionTests|SpawnArgumentsTests|FileHandleSafeReadTests'` — 48 tests passed, including the retained-output-pipe deadline regression.
- [x] Verified a help response larger than the pipe buffer is drained without timing out.
- [x] Verified help text from a non-zero exit is ignored.
- [x] Verified missing or unexecutable runtimes fall back conservatively.
- [x] Verified current runtimes retain both residency flags and older runtimes receive neither unsupported flag.
- [x] Verified a successfully exiting parent with a descendant retaining the output pipe returns unsupported within the probe deadline.
- [x] `SKIP_SIDECAR=1 BUNDLE_MODEL=0 ./scripts/build.sh` completed successfully.
- [x] `git diff --check`.

## Checklist

- [x] Relevant Swift tests pass locally.
- [x] Python lint and formatting are not applicable to this Swift-only diff.
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 2286`.
- [x] Critical parser, scheduler, and security mutation checks are not applicable.
- [x] README/docs changes are not required.
- [x] No breaking changes to existing APIs.


raullenchai (changes requested):
Thank you for keeping this compatibility fix capability-based and conservative. I found one boundedness gap on exact head `08c18f1b59fba945277a80810c75e78b5e4c0ac0` that is within the stated timeout contract.

`probeBlocking` bounds the selected process exit, but after a successful exit it calls `outputReader.wait()` without a deadline. If a runtime wrapper exits 0 after spawning a descendant that inherited stdout/stderr, the descendant keeps the pipe open and model startup waits past `timeoutSeconds` before Rapid enters `.starting`.

I reproduced this against the production probe with a deterministic fork helper: the help parent printed both flags and exited 0 immediately, while its child retained stdout for 2 seconds. With `timeoutSeconds: 1`, `probe` returned after 2.0089 seconds and failed a 1.5-second bound. The diagnostic test and helper were removed afterward; the review worktree is clean.

Please bound output-drain completion by the probe deadline as well, conservatively return unsupported when the pipe does not close, and add a focused regression contract where the help parent exits successfully but a descendant retains the output pipe. The fix should also avoid leaving an indefinitely blocked reader behind. No UI redesign or adjacent launch refactor is needed.

Other exact-head evidence is green: `ServerRuntimeCapabilitiesTests|ServerLocatorVersionSelectionTests|SpawnArgumentsTests` passed 39/39, including old/current runtime selection, large help output, non-zero exit, launch argv, and environment contracts. The probe remains off the main actor and the user-facing fallback is a non-blocking log message; no accessibility identifiers or native controls change.

raullenchai (approved):
Approved on exact head `7cfb04f5e9a188fa2d0af6f8c56992b8932035f2`.

The review finding is resolved: the output drain is bounded by the same probe deadline, a retained output pipe now falls back conservatively, and the reader loop is deadline-bounded. The new deterministic regression failed on the prior head (returned `allKnown` after 2.20s) and passes on this head in 1.005s.

Focused evidence: 48/48 tests across `ServerRuntimeCapabilitiesTests|ServerLocatorVersionSelectionTests|SpawnArgumentsTests|FileHandleSafeReadTests`; local Swift build passed. Exact-head CI is green for `tests`, `desktop-tests`, and `version-bump-guard`.

PR #2315, opened 2026-08-25
perf(mac): shrink the release DMG safely with LZMA

osdodo:
## Why

The published 0.13.1 Desktop DMG is 187,505,002 bytes. The exact signed and notarized artifact from this branch is 105,881,983 bytes, saving 81,623,019 bytes (43.53%) per download. The tradeoff is explicit: LZMA makes the one-time Finder copy slower on the measured Studio host (5.94 s to 17.80 s), while reducing network transfer by about 78 MiB.

## What

| Metric | Published 0.13.1 | Exact branch artifact |
|---|---:|---:|
| DMG download | 187,505,002 bytes | 105,881,983 bytes |
| Installed app file bytes | 509,621,417 bytes | 490,569,194 bytes |
| Finder copy on this host | 5.94 s | 17.80 s |

- Convert the staged writable image to an LZMA-compressed `ULMO` DMG instead of zlib `UDZO`.
- Remove dead console launchers and build-time-only sysconfig data from the sidecar.
- Delete `libpython3.12.dylib` only after a fail-closed dependency scan proves every bundled Mach-O is readable and none consumes it.
- Retain the Python interpreter symbol table so native sidecar crashes remain symbol-compatible with existing diagnostics.
- Keep the measured Mach-O baseline at the current 172 binaries.

## Scope and non-goals

This PR changes Desktop release packaging only. It does not remove runtime model/config modules, strip the Swift app or Python interpreter, change product behavior, publish a release, create a tag, or modify updater state.

## Design precedent

The packaging flow keeps the established writable-layout-then-convert sequence and changes only the final compression format. Runtime pruning follows a fail-closed dependency-graph pattern: every bundled Mach-O must enumerate successfully, any consumer blocks deletion, and all byte changes happen before signing.

## Verification

- Exact head: `cf0995188dfb6e234706e31c089be153aeb0a238`.
- Signed/notarized artifact-only dry run: https://github.com/raullenchai/Rapid-MLX/actions/runs/33260624721 — all build and desktop-ready jobs passed; publishing jobs were skipped.
- DMG SHA-256: `23f2dd9b107620ae2ea6bfb47620f831467f2e2d94322554b0711b7dae0a306b`.
- `hdiutil verify`, DMG `codesign --verify`, `stapler validate`, and Gatekeeper assessment passed.
- Mounted and copied app passed deep/strict code-sign verification, stapler validation, and Gatekeeper assessment as Notarized Developer ID.
- Installed sidecar launched from the copied app in an environment-stripped shell and reported `rapid-mlx 0.13.1`.
- Full sidecar build passed with 172 Mach-Os; raw bundle 455 MiB, tarball 145,612,923 bytes.
- Guard fixtures cover no-consumer, consumer-present, incomplete scan, and `otool` failure. Focused Python and Swift packaging tests, Ruff, shell syntax, and diff checks passed.

### DIFF
--- a/apps/rapid-mac/Tests/RapidTests/SidecarBuildScriptTests.swift
+++ b/apps/rapid-mac/Tests/RapidTests/SidecarBuildScriptTests.swift
@@ -150,8 +150,53 @@ struct SidecarBuildScriptTests {
                 "Only model-family directories outside Qwen3 TTS may be removed.")
         #expect(!script.contains(#"rm -rf "$STAGE/site-packages/mlx_audio/tts/models""#),
                 "The trim must never remove the complete TTS model directory.")
-        #expect(script.contains(#"MACHO_BASELINE_COUNT="${MACHO_BASELINE_COUNT:-174}""#),
-                "The signing baseline must match the measured post-audio bundle.")
+        #expect(script.contains(#"MACHO_BASELINE_COUNT="${MACHO_BASELINE_COUNT:-172}""#),
+                "The signing baseline must match the measured post-audio bundle, less the orphaned libpython dylib dropped in step 3.")
+    }
+
+    /// The interpreter strip must stay ABOVE the signing loop.
+    ///
+    /// `strip` rewrites the Mach-O, which invalidates any signature already
+    /// attached to it. Sinking the strip below step 5 would still produce a
+    /// bundle locally — the breakage only surfaces as a `codesign --verify`
+    /// failure on the user's Mac (and a notarisation reject in CI), so pin
+    /// the relative order here rather than trusting a comment.
+    @Test("The interpreter is stripped before the Mach-O signing pass")
+    func stripsInterpreterBeforeSigning() throws {
+        let script = try String(contentsOf: Self.scriptURL, encoding: .utf8)
+
+        guard let strip = script.range(of: #"strip -x "$STAGE/python/bin/python3.12""#) else {
+            Issue.record("The interpreter strip step is missing; the bundle regains ~1.6 MB of symbol table.")
+            return
+        }
+        guard let signing = script.range(of: "step 5: count + sign Mach-Os") else {
+            Issue.record("Could not locate the Mach-O signing step to order against.")
+            return
+        }
+        #expect(strip.lowerBound < signing.lowerBound,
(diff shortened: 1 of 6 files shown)


raullenchai (dismissed):
Exact-head review (`eee39b6c`) found three in-scope blockers before this packaging change can be validated on a release artifact:

1. **Rebase onto current `main` before further measurement.** This head is 120 commits behind. The same files now pin newer MLX/vision dependencies, use the extracted distribution checker, and include a fail-closed real-vision release smoke. The merge tree is syntactically clean, but the published 181 MB → 96.9 MiB and 461 MB → 441 MB numbers no longer describe the bundle we ship. Please rebase without dropping those current-main contracts, then refresh exact byte counts and mount/copy timing.

2. **The libpython deletion must fail closed when a future wheel links it.** The current unconditional `rm -f` turns a dependency update into a bundle that builds and signs successfully but fails when that extension is imported. #2294 already demonstrated and implemented the required shape: scan every bundled Mach-O other than libpython, refuse deletion when a consumer exists, and refuse deletion when the scan itself did not enumerate a credible tree. Please adapt that guarded mechanism and cover the no-consumer, consumer-present, and broken-scan cases; a string assertion that `rm -f` exists does not test the safety contract.

3. **Drop the interpreter `strip -x` unless crash symbolication is preserved end to end.** I ran the exact pinned python-build-standalone executable through this step: it saves 1,677,848 raw bytes but removes 40,224 non-external symbols (42,757 total symbols → 2,533). That is the same diagnostic regression for which this branch correctly reverted stripping the Swift executable: a native sidecar crash loses internal CPython frame names. The 15 MB app symbol table already measured as compressing away under ULMO; there is no evidence that this much smaller 1.6 MB table changes the DMG download materially. Keep the exported-symbol safety, signing-order checks, and meaningful size win by omitting this low-value strip, unless there is a working UUID-based sidecar crash-symbolication path and an exact proof through it.

Positive evidence: the ULMO format is supported by the app's OS floor, the existing writable-layout → convert sequence is retained, focused `SidecarBuildScriptTests` pass 10/10, and independent exact-head review found no additional code defect. After the corrected head, I will rerun focused tests and PR validation, then verify a current artifact through `hdiutil verify`, mount/copy, signatures, stapling, and first sidecar launch.

Issue #2300, opened 2026-08-24
Desktop: allow speech-input models to coexist with active LLM, VLM, and speech-output workloads


raullenchai:
## User impact

Voice input is an input method for an ongoing conversation, not a replacement for the conversation model. If starting speech recognition evicts or stops the active LLM, VLM, or speech-output model, the user cannot complete the natural voice workflow: speak a request, have the active model answer, and optionally hear the response.

Observed as a product-level residency problem while dogfooding the 0.13.0-rc1 Desktop build.

## Expected workflow

1. Keep the selected text or vision conversation model available.
2. Capture and transcribe speech through the configured speech-input model.
3. Send the transcript to the already active conversation model.
4. Optionally synthesize the response with the configured speech-output model.
5. Continue the conversation without repeated model replacement, cold starts, or lost state.

## Actual concern

Speech-input model activation appears to participate in a mutually exclusive model lifecycle. The UI does not clearly distinguish input, conversation, and output roles or explain when memory pressure prevents coexistence.

## Expected behavior

Model residency should be capability- and workload-aware. Speech input, the active LLM or VLM, and speech output should coexist when the Mac has enough memory. Transient speech workloads should not unconditionally replace the active conversation server. When the requested combination is unsafe, the app must predict that before loading and offer a clear, reversible choice rather than silently evicting another model.

## Acceptance criteria

- Represent speech input, conversation, vision, and speech output as distinct runtime roles rather than one globally exclusive selected model.
- Preserve the active conversation model and conversation state across dictation.
- Permit ASR plus LLM, ASR plus VLM, and ASR plus conversation model plus speech output on hardware where the combined validated footprint is safe.
- Account for combined model weights, caches, audio buffers, and runtime headroom before starting another role.
- When capacity is insufficient, identify the exact conflict and offer explicit choices such as stop speech output, choose a smaller speech model, or unload the conversation model.
- Never silently evict or restart a running model as a side effect of pressing the microphone button.
- Repeated dictation does not cause unnecessary reloads or cold starts; define an idle-unload policy for transient speech models.
- Cancellation, transcription failure, and microphone-permission denial leave the conversation model untouched.
- Server and Desktop expose consistent role and residency state.
- Deterministic lifecycle tests cover all role combinations, insufficient-memory decisions, cancellation, app relaunch, and recovery after a model process exits.
- End-to-end GUI coverage proves speak → transcribe → LLM or VLM response → optional speech output in one uninterrupted conversation.

[...]

PR #2320, opened 2026-08-25
fix(server): enforce end-to-end audio residency memory bounds

### LINKED ISSUE #2305: Server: budget auxiliary audio roles in shared residency lifecycle
Follow-up to #2300.

## Goal

Extend the existing process residency control plane so speech-input and speech-output engines participate in the same explicit memory admission and lifecycle policy as the conversation engine.

## Scope

- Account for combined model weights, KV/prefix caches, audio buffers, measured process footprint, and runtime headroom before loading an auxiliary audio role.
- Use catalog/cache metadata and allocator/process measurements; do not infer capacity from model names or hashes.
- Reject unsafe combinations before weight loading with a typed, actionable conflict that identifies the resident roles involved.
- Define and implement an idle-unload policy for transient speech engines without disturbing the active conversation engine.
- Keep role and residency telemetry consistent across admission, load, active work, cancellation, failure, idle eviction, and process recovery.
- Add deterministic tests for safe/unsafe combinations, rollback, cancellation, TTL, relaunch, and process exit.

## Non-goals

- Desktop conflict-choice presentation.
- Changes to the narrow #2300 no-eviction/co-load release fix.
- Silent eviction of the active conversation model.

osdodo:
## Why

Fixes #2305. Speech-input and speech-output engines currently bypass the process-wide residency ceiling, so an audio request can load multi-gigabyte weights or allocate large decoded/generated buffers after the conversation model has already consumed the safe memory budget.

## What

- Put persistent speech and alignment engines into the shared residency ledger.
- Reserve known model weights and request working memory before loading or allocating them.
- Reject unsafe or unmeasurable combinations with typed, actionable capacity errors.
- Keep audio roles charged and visible through loading, active work, cancellation, eviction, and shutdown.
- Reclaim only idle auxiliary roles; never silently evict the active conversation model.
- Bound transcription duration, reference audio, and single-yield TTS generation so the admission estimate remains enforceable.

## Scope and non-goals

This PR owns the server-side memory admission and lifecycle contract for speech input, speech output, and alignment. It does not add Desktop conflict-choice UI, change the narrow #2300 no-eviction/co-load behavior, or permit silent eviction of the active conversation model.

## Design precedent

The design follows an engine-owned lifecycle: one scheduler-visible residency record per role, a single process memory budget, explicit active/idle state, admission before allocation, and terminal unload transitions. Request routes acquire leases from that control plane instead of maintaining independent memory policy.

## Verification

- Deterministic safe/unsafe admission and pre-load rejection tests.
- Combined measured/reserved accounting and request-buffer charge tests.
- Cache completeness and model-capacity metadata tests.
- Cancellation, rollback, replacement, TTL, shutdown, relaunch, and process-exit lifecycle tests.
- Cold/warm error-envelope consistency and route-level audio coverage.
- Ruff, formatting, and full repository validation before merge readiness.


raullenchai (changes requested):
Reviewed exact head `164350dfcc0eb6be073d50a873950fe8d681b579` against #2305 and the #2300 primary-preservation contract. Two in-scope blockers remain:

1. **The revised shared-usage math double-counts an existing secondary reservation after process usage grows.** In `ResidentModelManager._accounted_usage`, `measured_projection` always adds `estimated_bytes - measured_bytes` for every non-primary record. For a lazy secondary whose load-time delta was 0, once the process footprint later grows from 1 GiB to 9 GiB, the method reports 12 GiB (9 GiB footprint + the same 3 GiB estimate) instead of the established 9 GiB aggregate truth. This breaks the existing deterministic contract: `uv run pytest -q tests/test_resident_models.py tests/test_audio_model_worker.py` fails at `tests/test_resident_models.py::test_accounting_reserves_lazy_model_estimates_and_tracks_larger_actual_usage` with `12884901888 != 9 * GiB` (51 passed, 1 failed). It can reject a safe audio/model combination even though the measured process footprint has already overtaken the complete reservation set. Please preserve the aggregate `max(measured, reserved)` semantics while accounting for genuinely pending audio/request reservations without attributing the same bytes twice.

2. **Cancellation during asynchronous role unload leaves non-terminal residency truth.** `_release_role_locked` sets the role to `evicting`, awaits the unload callback, and catches `Exception`, but `CancelledError` escapes. The production lane handle marks itself released before awaiting the worker. Exact-head reproduction: cancel `manager.release_role("speech-output")` while a delayed production-style unload runs; after the worker finishes, `snapshot()["roles"][0]["state"] == "evicting"` indefinitely. A second explicit release then removes the row. Until that extra call, the role remains charged, is not an idle reclamation candidate, and can cause false 507s despite the engine already being unloaded. Please make the unload transition cancellation-safe/terminal and add deterministic cancellation coverage that checks ledger state, capacity reuse, and at-most-once engine unload.

Positive evidence: the PR-focused audio suite passed 256/256; Ruff and `git diff --check` passed. The PR is also behind current main, so the corrected exact head will need current-base reconciliation before landing.

PR #2331, opened 2026-08-25
fix(mac): onboarding 'Before loading' memory verdict re-evaluates live (ONBOARD-MEM-LIVE)

raullenchai:
## Why

rc2 dogfood bug (ONBOARD-MEM-LIVE): onboarding step 4 "Before loading" does not re-evaluate live. On an 18 GB Mac with `lfm2.5-1b-4bit`, after the user closed apps and FREE NOW rose to 5 GB, the verdict/headline/percent/hint/NEEDS colour/top icon all stayed in the original red state. Expected: when free memory becomes sufficient, every visible element updates together, and the reverse when memory drops.

## Root cause

The 3s live memory refresh correctly recomputed `severity` + `freeGB` from one fresh probe snapshot (that data logic is tested), but the refreshed value **never reached SwiftUI**. `pendingMemoryWarning` is a computed property over a plain value-typed `MemoryLoadConfirmationQueue` stored in the `@Observable ServerManager`. In Swift Observation, an in-place mutation of an element inside a stored **value-type** property does NOT fire `withMutation` on the owner — so the card kept rendering the original parked snapshot. (This is exactly why the codebase already makes `DownloadManager`/`Job` `@Observable` classes — see the `@Observable is load-bearing` comment near `Job.progress`.)

## Fix (smallest, no redesign)

Turn `MemoryLoadConfirmationQueue` into an `@Observable` reference type (all method bodies unchanged, `mutating` → `func`). Now replacing the head warning's measured facts fires observation, so the "Before loading" card re-renders live from the same latest probe snapshot — verdict, headline, percent, hint, NEEDS colour and the top icon all flip together on any true red→green / green→red crossing. The 85% headroom rule is unchanged; no other onboarding steps touched.

## Evidence (before/after)

Before (the frozen red state, driven by the bug numbers 18 GB / 5 GB free / 3 GB model): severity `.unsafe` → title "may crash your Mac right now", message "…about 89% of 18 GB…", NEEDS 3 GB red, `exclamationmark.triangle` icon. **Before the fix these are computed but never render after the refresh** — the card is frozen.

After (free ≥ the 7.5 GB crossing threshold on 18 GB): severity `.safe` → title "ready to load", green checkmark, green NEEDS — all elements flip together in place as the live probe updates each 3 s tick.

A focused regression test pins the mechanism: `withObservationTracking` on the exact view-bound property `server.pendingMemoryWarning` proves a live refresh with injected memory facts **invalidates observation** (the verdict would re-render), not merely updates an invisible value.

Note on "screenshots": the card is a rare blocking state reachable only through a real parked unsafe load, so a pixel capture isn't reproducibly captured here; the before/after is instead pinned deterministically (field-level rendering above + the observation-invalidation test). No visual redesign was made, so rendering is byte-identical; the change is purely that the verdict now updates live.

## Tests

- New: `LaunchAutoStartMemoryTests/liveRefreshInvalidatesPendingMemoryWarningObservation` — live refresh invalidates the view-bound property (observation fires).
- Existing, still green: `MemoryLoadConfirmationQueueTests` (10, incl. red→green→red), `LaunchAutoStartMemoryTests` (7, incl. "parked warning re-samples the injected probe in both directions"), `OnboardingWontFitReviewTests` + `OnboardingRecoveryBehaviorTests` (86). 103 tests in the combined surface all pass.

Non-goals respected: no threshold/85% change, no other onboarding steps, no release/rc2 files.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
+++ b/apps/rapid-mac/Sources/Rapid/Server/ServerManager.swift
@@ -5,7 +5,18 @@ import Observation
 /// FIFO state machine for memory-risk confirmations. A request token is
 /// present for ``ensureServing`` callers that must await their own answer;
 /// direct ``start`` calls still queue a prompt but retain no result.
-struct MemoryLoadConfirmationQueue {
+///
+/// Referenced type (``@Observable`` class, not a struct) so that replacing the
+/// head warning's measured facts — e.g. the 3s live memory refresh in
+/// ``ServerManager/refreshPendingMemoryWarning()`` — fires SwiftUI observation
+/// and the "Before loading" verdict re-renders live. With a plain value type,
+/// an in-place mutation of `pending[0].warning` inside a stored, value-typed
+/// property does NOT invalidate the ``@Observable`` owner, so the card kept
+/// showing the original parked snapshot even as free memory changed
+/// (ONBOARD-MEM-LIVE). Mirrors how ``DownloadManager``/``Job`` stay
+/// ``@Observable`` so nested `status`/`progress` mutations re-render.
+@Observable
+final class MemoryLoadConfirmationQueue {
     enum Decision: Equatable {
         case confirmed(sequence: Int)
         case cancelled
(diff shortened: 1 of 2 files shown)


raullenchai (commented):
Pixel technical verdict: **APPROVE** on exact head `d1492875f043376a96214c0a5f397c2e4bad51c1`. GitHub cannot record a formal approval because this reviewer account is also the PR author.

The change stays within the onboarding live-memory contract: the existing confirmation queue becomes an observable reference type; queue transitions, safety thresholds, refresh generation ordering, and UI copy are unchanged. The new test observes the exact computed property bound by onboarding and proves an unsafe → safe refresh emits Observation invalidation.

Local exact-head evidence on Apple Silicon:
- `swift test --filter LaunchAutoStartMemoryTests`: 8/8 passed
- `swift test --filter MemoryLoadConfirmationQueueTests`: 10/10 passed
- `swift test --filter QuickstartViewTests`: 44/44 passed
- release-shaped app build completed
- real AX-driven low-memory onboarding journey passed against that app with the 8 GB hardware fixture; artifacts retained at `/private/tmp/rapid-pr2331-pixel-gui`

The dynamic pressure transition itself is covered deterministically by the injected live snapshot contract (unsafe → safe plus observation firing); the GUI run separately verifies the native onboarding surface and low-memory path. No accessibility or native-behavior regression found.

PR #2346, opened 2026-08-25
fix(mac): preserve the chat model across dictation relaunch

raullenchai:
## Why

The final 0.13.0 candidate could restore a transcription-only model as the chat model after Dictation was enabled and the app relaunched. The composer was then disabled, blocking chat. This was requested by the maintainer during final candidate dogfood.

## What

- Make the persisted last-served alias chat-lane state, written only for a catalog-proven chat model.
- Validate legacy rc2 state through case-insensitive catalog capability metadata, without alias-name or repository heuristics; preserve pending ownership while probing, perform one bounded authoritative retry after an empty shared launch snapshot, then fail closed to an explicit terminal unresolved state if that retry also fails.
- In the terminal unresolved state, preserve the persisted key for the next launch while treating this session as having no restored chat model: no auto-start, onboarding available, Dictation barrier paired.
- Establish persisted-enabled Dictation launch deferral synchronously when the controller is created, so Audio-view revalidation cannot start its model before chat restore.
- Preserve exact-alias catalog provenance through start and ensureServing. An authoritative catalog row takes precedence; a caller-owned exact-alias hint is used only when that row is unavailable.
- Share the validated chat outcome between auto-start and first-run onboarding decisions.
- Start chat restore first, arm the persisted Dictation shortcut without starting its audio model, then prepare the audio lane only after chat restore settles; cancellation releases the barrier without starting audio from a superseded task.
- Cover both fresh selection and legacy rc2 persisted state through quit and relaunch.

## Non-goals

- No server or serve-command changes.
- No audio residency redesign or new settings.
- No model-name, hash, or prefix heuristics.
- No polling, timed retry, background catalog refresh, or second retry.

## Exact-head evidence

Head: `f2cd5e9b7f76cd07c35af38f4f129cd7becf93fb`

- Focused restore, ensureServing lifecycle/residency, catalog-cache retry/exhaustion, onboarding, auto-start, Dictation, voice co-load, picker/Quickstart, and AX inventory: 252 passed across 14 suites.
- Release-config app build with strict local bundle assembly: passed.
- Fresh selection → Dictation enabled → quit → relaunch: original chat restored, Send enabled, Dictation retained; passed.
- Legacy rc2 persisted chat + Dictation state → launch: chat restored, Send enabled, Dictation retained, transcription alias never became process owner; passed.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift
+++ b/apps/rapid-mac/Sources/Rapid/Dictation/DictationController.swift
@@ -114,7 +114,36 @@ final class DictationController {
     private let server: ServerManager
     private let client: AudioClient
     private let hotkey = DictationHotkey()
-    private let recorder = DictationRecorder()
+    /// Constructing `DictationRecorder` initializes AVAudioEngine/CoreAudio.
+    /// Keep that work behind the first real capture so state-only controllers
+    /// (including launch restore and tests) never claim audio resources.
+    private final class CaptureLifecycle: @unchecked Sendable {
+        var recorder: DictationRecorder?
+        var ticker: Timer?
+
+        deinit {
+            ticker?.invalidate()
+            recorder?.shutdown()
+        }
+    }
+
+    private let captureLifecycle = CaptureLifecycle()
+    private var recorderStorage: DictationRecorder? {
+        get { captureLifecycle.recorder }
+        set { captureLifecycle.recorder = newValue }
+    }
+    private var recorder: DictationRecorder {
+        if let recorderStorage { return recorderStorage }
+        let recorder = DictationRecorder()
+        recorder.onLevel = { [weak self] value in
+            Task { @MainActor in self?.level = value }
+        }
+        recorder.onFirstSample = { [weak self] in
+            Task { @MainActor in self?.markRecordingStarted() }
+        }
+        recorderStorage = recorder
+        return recorder
+    }
     private let hud = DictationHUD()
     /// `nil` means the catalog probe itself failed; an array is authoritative.
     /// Keeping those states distinct prevents a transient CLI failure from
@@ -128,7 +157,10 @@ final class DictationController {
     private let testingRecorderCancel: (@MainActor () -> Void)?
     private let testingTranscribeCancel: (@MainActor () -> Void)?
 
-    private var tickTimer: Timer?
+    private var tickTimer: Timer? {
+        get { captureLifecycle.ticker }
+        set { captureLifecycle.ticker = newValue }
+    }
     private var recordingStart: Date?
     private var capturingApp: String?
     private var level: Float = 0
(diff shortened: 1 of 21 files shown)


raullenchai (commented):
## APPROVE — review of head `5fb1d12b`

**Scope reviewed:** the relaunch/alias-restore contract across `ServerManager`, `ContentView`, `DictationController`, new `SessionModelRestore`, and the GUI AX journey.

### Contract holds (systemic, not a band-aid)
- **Chat-lane / audio-lane separated at the persisted-state & restore-path level.** `SessionModelRestore` owns lane-aware persistence; the rc2 key `rapid.serve.lastAlias` is now explicitly the *chat* lane's key (kept at the same key → no data-loss migration). `persistReadyAlias` writes only when `catalogEntry.kind == .chat` (catalog capability metadata — **no model-name/hash/repo heuristics**). `resolve()`/`launchPlan()` validate legacy rc2 state and the dictation/speech lanes strictly through catalog `kind` (`.chat` / `.audio`), and handle a failed catalog probe (`[]` sentinel) fail-closed via `chatAliasResolved` rather than misreading it as "definitely non-chat".
- **The `Send disabled` mechanism is closed and verified.** `ensureVoiceLane → ensureServing(residencyEligible:false) → start(whisper-small) → .ready` no longer writes the audio alias to the chat-lane key (whisper-small is excluded from the primary chat catalog → `catalogEntry?.kind == .chat` fails closed), and restore accepts `rapid.serve.lastAlias` only as a chat-catalog member. On relaunch auto-start restores the real chat model.
- **Restore ordering + concurrency hardening.** `restorePersistedSession` resolves the chat alias first, publishes `restoredChatAlias` independently of the auto-start preference (fixes the c71c1af7 latent bug where auto-start-off suppressed onboarding), arms Dictation's shortcut immediately, defers audio model prep until chat settles, then `finishDeferredBootstrap`. The dictation `modelPreparationDeferred` barrier is established before the first suspension and inherited by re-entrant `enable()` calls, so no audio-only fallback can tear down the still-starting primary child.

### State-layer tests (all required cases present, all green)
Dictation on + relaunch; dictation + TTS; audio-only history; rc2-written persisted state (**rc2ChatAliasUpgrade**); catalog-probe-failure; on-disk regression via the **real fake sidecar** (`ensureVoiceLane` audio ready never overwrites the chat-lane key); and concurrency races (model change during deferred bootstrap / during catalog refresh / hotkey-failure cannot release the barrier).

### My independent verification (not just trusting the PR)
- **59 focused tests pass** on a clean checkout of the exact head (`SessionModelRestoreTests`, `LaunchOnboardingOrderingTests`, `DictationTests`), including the fake-sidecar lifecycle.
- **Release-shaped app built clean** (`scripts/build.sh`), `dictation` AX journey **PASSES on a fresh isolated persona**, with evidence cross-checked directly from artifacts: `server_started` = `fake-alias` twice (0× `fake-whisper-small`), post-relaunch `ModelPickerBar.ModelMenu` = `fake-alias`, `dictation-relaunch-send-ready.json` present (Send ready), `Dictation.Enable = 1` (dictation still enabled after relaunch).
- No regression risk identified in `ServerManager`, `ContentView`, or `DictationController`.

PR #2387, opened 2026-08-26
fix(mac): allow deleting generated images

### LINKED ISSUE #2564: Desktop: delete generated images from the session gallery
## Why

Generated images cannot currently be removed from the Images tab. Unwanted results remain in the in-memory session gallery until the rolling session limit removes them.

## Owner

Echo shepherding community PR #2387; implementation by osdodo.

## Scope / Non-goals

Add a native, confirmed delete action for session-owned image results. Deletion must not remove files the user saved to disk and must not introduce persistent gallery storage.

## Acceptance

- Keep leaves the gallery unchanged.
- Delete removes only the selected session result and selects an adjacent result.
- Deleting the final result restores the empty state.
- Deleting the active edit source exits edit mode without stale state.
- Imported or saved files on disk remain untouched.
- The action has stable accessibility identifiers and focused Swift plus GUI coverage.

osdodo:
## Why

Generated images currently cannot be removed from the Images tab. Users can only keep generating new images, leaving unwanted results in the session gallery and consuming memory until the session limit rolls them off.

This restores a basic gallery-management workflow while preventing accidental deletion and preserving files the user explicitly saved.

## What

Adds a delete action to the Images tab with a native confirmation dialog.

Deleting an image removes it from the in-memory session gallery without affecting copies saved to disk. The view selects an adjacent result after deletion, restores the empty state when the last result is removed, and exits edit mode when the edit source is deleted.

Regression coverage exercises gallery selection, empty-state restoration, edit-source cleanup, imported images, accessibility identifiers, and the native Keep/Delete journey.

## Scope / Non-goals

The change is limited to in-memory gallery state, the destructive confirmation surface, and focused Desktop coverage. It does not delete saved files, add persistent gallery storage, alter image generation or editing behavior, or change the engine/server.

## Design precedent

The interaction follows the app's established native destructive-confirmation pattern: a clearly labeled trigger, a safe default that preserves data, an explicit destructive action, stable accessibility identifiers, and state mutation owned by the gallery view model. Session-owned results are removed independently from user-owned files on disk.

## Verification

- `swift test --no-parallel --filter 'ImageResultDeletionTests|CoreWorkspaceVisualFoundationTests'`: 34 passed on `a31da5d1`.
- `pr_validate 2387`: MERGE-SAFE on `a31da5d1` (19,081 passed; 97 skipped; 22 deselected; 6 xfailed; 1 xpassed).
- Native generated-image journey covers Keep, confirmed deletion, repeated deletion, and final empty-state restoration.

Fixes #2564

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/Images/ImageGenViewModel.swift
+++ b/apps/rapid-mac/Sources/Rapid/Images/ImageGenViewModel.swift
@@ -263,6 +263,35 @@ final class ImageGenViewModel {
         }
     }
 
+    /// Remove an image from the in-memory session gallery. Saved copies are
+    /// independent files and are deliberately untouched.
+    func delete(_ image: GeneratedImage) {
+        let wasActive = activeImage?.id == image.id
+        let wasEditSource = editSource?.id == image.id
+        let removedIndex = results.firstIndex { $0.id == image.id }
+
+        guard removedIndex != nil || wasEditSource else { return }
+        if let removedIndex {
+            results.remove(at: removedIndex)
+        }
+        if wasEditSource {
+            cancelEdit()
+        }
+        guard wasActive else { return }
+
+        let replacement: GeneratedImage?
+        if let removedIndex, results.indices.contains(removedIndex) {
+            // Prefer the next older image, which now occupies the same slot.
+            replacement = results[removedIndex]
+        } else if removedIndex != nil {
+            replacement = results.last
+        } else {
+            // Imported edit sources are not gallery entries.
+            replacement = results.first
+        }
+        activeID = replacement?.id
+    }
+
     /// Load the image-gen alias catalog (safe to call repeatedly).
     func refreshCatalog() async {
         catalogRefreshGeneration &+= 1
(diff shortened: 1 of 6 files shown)


raullenchai (approved):
Approved on exact head `be5dc474`.

Evidence:
- The change is scoped to in-memory gallery deletion plus the native confirmation surface; saved files are never addressed or removed.
- Selection is deterministic after deleting the active result, the last-result path restores the empty state, and deleting either a gallery or imported edit source exits edit mode without stale selection/model state.
- The UI follows the existing native destructive-confirmation pattern, keeps the destructive action unavailable during generation, and provides distinct AX identifiers/labels for trigger, confirm, and cancel.
- Local exact-head focused run: `swift test --filter 'ImageResultDeletionTests|CoreWorkspaceVisualFoundationTests'` — 33/33 passed.\n\nResidual risk is low: this is session-only state and the dialog uses the app’s already-established confirmation pattern. The branch is behind current main, so it still needs normal current-base reconciliation and gates before landing.

PR #2393, opened 2026-08-26
docs: stage 0.13.1 release notes

### LINKED ISSUE #2145: `rapid-mlx pull`: allow selecting a specific quantization bit-width and format (e.g. `--format mlx --bits 4`) for multi-variant repos
### What are you trying to do?

I want `rapid-mlx pull <repo>` to download only the model variant I actually need — a specific format (MLX vs GGUF vs safetensors) and a specific bit-width (2/4/6/8-bit) — instead of pulling every variant bundled in the Hugging Face repo. My workflow is: pick a model, decide on 4-bit MLX for my Apple Silicon box, run `rapid-mlx pull`, and start inference with the smallest footprint that fits my RAM/quality tradeoff.

### What's blocking you today?

Many HF repos now ship multiple formats and multiple quant levels side by side in one repo — e.g. [OBLITERATUS/Qwen3.8-27B-OBLITERATED](https://huggingface.co/OBLITERATUS/Qwen3.8-27B-OBLITERATED), which contains: 

- GGUF: Q8_0 (~27 GB), Q6_K (~21 GB), Q5_K_M (~18 GB), Q4_K_M (~15 GB), IQ4_XS (~14 GB)
- Safetensors bf16 (~50 GB)
- MLX: 4-bit (~14 GB) and 8-bit (~27 GB)

That's roughly 180GB+ total in the repo, but I only ever want one ~14GB slice (MLX 4-bit). `rapid-mlx pull` currently has no flag to filter by format or bit-width, so it appears to fetch the whole repo tree (or at minimum doesn't expose a way to restrict the download), and I end up with 60-180GB+ on disk for a model I use at 10-20GB.

### Current workaround (if any)

I can't do this cleanly today. I've tried manually browsing the HF repo file list and using `huggingface-cli download <repo> --include "mlx-4bit/*"` outside of Rapid-MLX to grab just the folder I want, then pointing Rapid-MLX at the local path — but that bypasses Rapid-MLX's pull/registry flow entirely and I lose whatever bookkeeping (manifest, cache indexing, model naming) `pull` normally sets up.

### Proposed approach (optional)

Add explicit filter flags to `pull`, e.g.:

```
rapid-mlx pull <MODEL> --format mlx --bits 4
```

- `--format` accepts `mlx`, `gguf`, `safetensors` (repo-dependent).
- `--bits` accepts common quant labels (`2`, `4`, `6`, `8`) and maps them to the repo's actual subfolder/quant-suffix naming (e.g. `mlx-4bit/`, `Q4_K_M`), since naming conventions vary per uploader.
- If a repo has ambiguous or non-standard naming, `pull` could print the detected variants and ask the user to pick one (similar to `ollama pull` disambiguation on tags), rather than silently downloading everything.
- Without `--format`/`--bits`, keep current default behavior for backward compatibility, but maybe warn if the repo has multiple variants and estimated total download size is large.

### Alternatives you considered

- `huggingface-cli download --include <pattern>` — works but is a manual, error-prone workaround outside Rapid-MLX's model management (no manifest/cache integration).
- Ollama already supports tag-based selection at pull time (`ollama pull model:4b`), which is the UX I'd like to mirror here since it solves exactly this "one repo, many variants" problem.
- mlx-lm's own download helpers don't filter by quant either, so this isn't solved upstream.

### Hardware (if relevant)

Apple Silicon Mac

### Model (if relevant)

OBLITERATUS/Qwen3.8-27B-OBLITERATED (This pattern is increasingly common across community-quantized/finetuned uploads, not unique to this one model)

### Are you willing to contribute this?

Yes, I can help test

### Additional context

This mainly affects repos that bundle multiple formats/quants under one HF repo namespace (common for community abliterated/finetuned models). Filtering at pull time would cut disk usage from 60-100GB+ down to the 10-20GB actually used per model.

### LINKED ISSUE #2219: Desktop: do not use the 1.2B lightweight model as the first-run best experience
## Summary

The first-run flow currently recommends LFM2.5 1.2B. It is fast and inexpensive to run, but its observed quality is below the bar for the model that represents Rapid to a new user. Users do not attribute these failures to a lightweight model; they conclude that Rapid is not intelligent, does not remember context, and cannot use tools reliably.

This is intentionally deferred from the current release. Keep the issue for the next product-planning cycle.

## Observed first-run failures

During GUI dogfood with the default 1.2B model:

- After the user said `My name is Maya`, the next turn claimed it could not access personal information and did not answer `Maya`.
- When asked whether it could generate images, it said it could not, although Rapid exposes Images in the sidebar.
- A request for exactly two sentences produced one generic sentence.
- The first TXT attachment summary was contaminated by the preceding photography conversation and hallucinated a photography-tips document.
- Only a more explicit follow-up correctly extracted `APAC` and `42` from the attachment.
- A Tokyo weather request used Search instead of the Weather tool and returned approximately 16°C; the actual Weather tool had returned 24.8°C.

## Product impact

This damages the first impression across several core promises at once:

- multi-turn chat memory
- instruction following
- attachment grounding
- tool selection and trust
- awareness of Rapid product capabilities

Severity: high-priority product friction, but not targeted for the current release.

## Proposed scope

1. Keep LFM2.5 1.2B available, but position it as `Fastest` / `Low memory`, with clear limitations for multi-turn chat, files, and tool selection.
2. Remove its `Recommended` / `Best experience` and automatic first-run-default status.
3. Select a validated model around the 4B class as the first-run default, subject to device compatibility.
4. Present a simple onboarding choice:
   - Best experience — more reliable chat, files, and tools
   - Fastest download — smaller and quicker, with limited capability
5. Inject current Rapid capabilities into the chat system context so models can correctly direct users to Images and enabled tools.
6. Separately verify whether attachment contamination is caused only by model quality or also by prompt/context assembly.

## Acceptance criteria

The proposed first-run default must pass a small product regression set on supported baseline hardware:

- remembers the name `Maya` on the following turn
- follows an exactly-two-sentences instruction
- summarizes a newly attached TXT without contamination from prior conversation
- extracts `APAC` and `42` from the test attachment
- routes a Tokyo weather request to the Weather tool when available
- correctly explains that image generation is available through Rapid Images

Additional requirements:

- The lightweight-only model cannot be selected automatically as `Best experience`.
- If no quality-qualified default fits the machine, Rapid must explain the tradeoff instead of silently downgrading and retaining the Recommended badge.
- Validate with a local Release build and a clean first-run GUI journey before landing.

## Out of scope

- Removing support for the 1.2B model
- Building the complete future model-management SSOT redesign
- Releasing from this issue alone

### LINKED ISSUE #2222: Ornith 1.5 9B ignores strict format and explicit Weather tool routing in Desktop
## Summary

Real local Release-build dogfood with `ornith-1.5-9b-bf16` found two repeatable quality/tool-selection failures. The model is responsive and remembers basic multi-turn context, but it does not reliably follow a strict sentence-count constraint and chooses Search even when explicitly told to use the available Weather tool.

This is non-release-blocking dogfood feedback.

## Environment

- Rapid Desktop local Release build: 0.12.18 (162)
- Source: `origin/main` at `4bbca765` plus an unrelated local prefix-cache startup fix
- Model: `ornith-1.5-9b-bf16`
- Hardware: Studio Mac, 256 GB unified memory
- Model state: Resident; approximately 1 second first-token latency after startup blocker was bypassed

## Reproduction 1: strict format

Prompt:

```text
Explain why local AI can be useful. Write exactly two sentences, no heading or bullets.
```

Observed response was one long sentence:

```text
Local AI keeps your data private since processing happens on-device rather than sending it to external servers, and it works reliably offline without needing a constant internet connection or paying ongoing subscription fees.
```

Expected: exactly two sentences.

## Reproduction 2: Weather routing

Prompt:

```text
What is the current weather in Tokyo? Use the Weather tool, not web search, and report the tool result.
```

Observed:

- Tool chip showed `web_search`.
- The answer said no Weather tool was available.
- The same answer then listed `weather — Get current weather for a city or place` among its available tools.
- It asked the user whether it should use the built-in weather tool instead of simply using it.

Expected:

- Route directly to `weather`.
- Report the Weather tool result.
- Do not claim that an advertised tool is unavailable.

## Control result

Multi-turn memory worked correctly:

1. `My name is Maya. Acknowledge in one short sentence.` → `Hi Maya!`
2. `What is my name? Answer with the name only.` → `Maya`

This suggests the server and conversation history were functioning; the failures are narrower instruction-following and tool-selection behavior.

## Investigation areas

- Ornith tool schema / tool-name presentation
- System prompt wording for available Rapid tools
- Whether explicit tool requests are being weakened by the generic web-research policy
- Tool-choice evaluation for Ornith 1.5 aliases
- Strict-format eval coverage for recommended chat models

## Acceptance criteria

- The exact Weather prompt produces a `weather` tool call, not `web_search`.
- The final answer does not contradict the supplied tool inventory.
- The exact-two-sentences prompt produces exactly two sentences.
- Add a reproducible model dogfood/eval case without hard-coding an answer in product code.

### LINKED ISSUE #2239: macOS: “Run setup again” confirmation overstates onboarding-only reset
## Dogfood finding

On current `main` (`dcd6fb81`) in an isolated Rapid Mac profile, Settings → App → **Run setup again…** presents this confirmation title:

> Erase this Mac's Rapid state and restart?

But this entry point calls `ReonboardingReset.perform(scope: .onboarding, ...)`. The body correctly says it only erases the record that Quickstart has run, and the Settings section explicitly promises that settings, conversations, downloaded models, and telemetry choice remain untouched. The generic title contradicts that narrower promise and can make users abandon a safe recovery action because it sounds like a full data wipe.

## Evidence

- Mini: M2 Pro, 32 GB, macOS 26.5.2
- Clean release build and CFPreferences/Application Support-isolated app
- AX exposed `Settings.App.RunSetupAgain`, then `Settings.App.ConfirmRunSetupAgain` / `Settings.App.CancelRunSetupAgain`
- Dialog body: “This erases the record that Quickstart has run. Rapid restarts immediately afterwards.”

## Suggested fix

Make `ReonboardingReset.confirmation(for:)` choose a scope-specific title for onboarding-only, e.g. **Run guided setup again?**, while retaining destructive erase wording when `.preferences`, `.conversations`, or `.telemetry` are selected. Add a copy assertion for the onboarding-only title.

## Severity

Non-release-blocking UX issue: the action works and accurately scopes the body, but the title undermines discoverability/trust of this newly added recovery path.

### LINKED ISSUE #2342: Cancel endpoint cannot resolve the public streaming request ID
## User impact

A client cannot cancel a streaming chat request using the public request ID returned by the API. The documented cancel route returns HTTP 404 while generation continues to completion.

This behavior is pre-existing in the 0.13.0 RC2 source; it is not a regression introduced after RC2.

Suggested owner: Vector  
Target: 0.13.1

## Exact environment

- Host: Apple M2 Pro, 32 GiB, arm64
- macOS: 26.5.2 (25F84)
- RC2 source checked: `33dd7a3009bca9ebb72ebca9340ed5dd557c6756`
- Runtime repro source: `a8f507a04c712ca55276f955330c9e74062f1182`
- Built wheel SHA-256: `c998bbea01f126b5d67f8ab1a6f37708edeb40a98e64cc86eee21376f7aac613`

## Reproduction

1. Start a cached model with the locally built wheel:

```console
HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 rapid-mlx serve mlx-community/LFM2.5-1.2B-Instruct-4bit \
  --host 127.0.0.1 --port 18337 --hybrid-cache-entries 4 \
  --max-tokens 14000 --log-level INFO
```

2. Start a long streaming request and read its first SSE event:

```console
curl -N http://127.0.0.1:18337/v1/chat/completions \
  -d '{"model":"mlx-community/LFM2.5-1.2B-Instruct-4bit","messages":[{"role":"user","content":"Write a very long numbered list."}],"stream":true,"max_tokens":2000}'
```

3. Copy the public `id` from the first event and submit it to the documented cancel route:

```console
curl -i -X POST http://127.0.0.1:18337/v1/requests/chatcmpl-8c8b5e49/cancel
```

## Actual result

- First public SSE ID: `chatcmpl-8c8b5e49`
- Cancel response: HTTP 404
- Scheduler request ID in the same request's logs: `2b8c969e-8b6`
- Generation continued instead of being cancelled.

## Expected result

The public request ID emitted to the client should identify the corresponding scheduler request for cancellation, and the cancel route should abort it successfully.

## Verified contract mismatch

The cancel route in `routes/health.py` documents that clients should use the first public `chatcmpl-*` SSE ID and forwards that ID directly to `abort_request`. However:

- `routes/chat.py` creates the public `chatcmpl-*` response ID independently.
- `engine_core.py` creates a separate UUID for scheduler submission.
- The public ID is not wired to that internal request ID.

The relevant files are byte-for-byte identical between RC2 source `33dd7a30` and runtime source `a8f507a0`, confirming this is pre-existing in RC2:

- `routes/health.py`: blob `d7799dbd1427ec534ed1324976f0941d3cb8e47e`
- `routes/chat.py`: blob `4fe1c68777fa417b44ade86ce83346e1286cd643`
- `engine_core.py`: blob `eacae33393eedaa4640679137d4f63dad14c0b0d`
- `engine/batched.py`: blob `cbe118d8919aea233baf1469b85efd89fb470de4`

### LINKED ISSUE #2343: Desktop model picker can trigger removable-volume permission for a mounted installer DMG
## What happened

With the release DMG still mounted, choosing a different model from the Desktop model picker can trigger the native macOS “access files on a removable volume” permission prompt even though the user did not choose an external model folder or approve a file tool.

## Reproduction

1. Mount the Rapid-MLX Desktop release DMG.
2. Launch an installed/copy-on-disk Desktop app.
3. Open the chat model picker and select another model.
4. Observe the removable-volume access prompt.

## Observed behavior

- The prompt repeated while the DMG remained mounted.
- Choosing **Don’t Allow** did not break the picker; the list remained usable and model switching succeeded.
- After ejecting the DMG, the same picker path no longer prompted.

## Expected behavior

Normal model selection should not touch unrelated mounted installer media or request removable-volume access. External model storage should be accessed only after an explicit user choice.

## Evidence

- Reproduced on the 0.13.0 candidate content at exact commit 23e899c8 in an isolated app identity.
- The installed release metadata already contains the same removable-volume usage description, so this appears to be an existing volume-discovery behavior rather than a candidate-only regression.
- No user data was modified; the DMG was read-only.

### LINKED ISSUE #2348: serve: connection helper ignores the running server port and model
## User impact

After a successful server start on a non-default port, the readiness banner tells the user to run `rapid-mlx connect openai-python`. That command prints a client configured for port 8000 and the placeholder model `<detected-model>`, not the live server port and model. A new user who copies the advertised command cannot connect to the server they just started.

Suggested owner: Vector (engine/server CLI).

## Reproduction

Exact source: `31a1524b611714f0044a296abedb05c31bea3b1a`, installed as a non-editable `rapid_mlx-0.13.0-py3-none-any.whl` in a fresh Python 3.11 venv on Studio.

1. Run `rapid-mlx serve lfm2.5-1b-4bit --port 8123`.
2. Wait for `Ready: http://127.0.0.1:8123`.
3. The banner advertises `Python       rapid-mlx connect openai-python`.
4. Run that exact command.

Observed output starts with:

```
Python (OpenAI SDK)  →  http://localhost:8000/v1
...
model="<detected-model>"
```

`rapid-mlx connect openai-python --base-url http://127.0.0.1:8123/v1` is rejected as an unrecognized argument, so the user cannot repair the advertised workflow through the connect command.

## Expected

The post-start connection path should produce a directly usable example for the live host, port, and served model identity. If `connect` remains a separate process, it needs an explicit supported way to receive or discover that instance context; otherwise the readiness banner should print the correct snippet itself.

### LINKED ISSUE #2349: pull: cached models are reported as newly downloaded
User impact

A fully cached model produces both Already cached and Downloaded in 3.8s. This makes users think a network transfer occurred, obscures whether offline reuse worked, and makes the elapsed time look like download performance.

Suggested owner: Vector (engine/server CLI).

Reproduction

Exact source: 31a1524b611714f0044a296abedb05c31bea3b1a, installed as a non-editable 0.13.0 wheel in a fresh Python 3.11 venv on Studio.

Run rapid-mlx pull lfm2.5-1b-4bit when all ten files are already cached.

Observed terminal tail:

[10/10] tokenizer.json cached (5 MB)
Already cached (10 files, 663 MB)
Cached at: .../snapshots/dee2f8...
Downloaded mlx-community/LFM2.5-1.2B-Instruct-4bit — 632.7 MiB in 3.8s

Expected

The final outcome should say that the existing cache was verified or reused. Reserve Downloaded and transfer timing for a run that actually fetched bytes.

### LINKED ISSUE #2350: large catalog aliases bypass the pre-download size confirmation
User impact

The large-download confirmation gate only inspects a raw model argument containing a slash. A user-facing catalog alias bypasses it before alias resolution, even when the catalog already knows the model is hundreds of GiB. This can start an unintended model download from a simple serve command.

Suggested owner: Vector (CLI download admission).

Reproduction

Exact source: 31a1524b611714f0044a296abedb05c31bea3b1a, installed as a non-editable 0.13.0 wheel in a fresh Python 3.11 venv on Studio.

The model table reports kimi-k2.6 as approximately 438.3 GiB. In an interactive terminal, run HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 rapid-mlx serve kimi-k2.6 --port 8127.

Observed: the alias resolves to its repository; no Continue confirmation appears; output says size is unknown and Proceeding without confirmation; the serve path immediately attempts first-time download twice before the offline failure. The offline variables safely prevent transfer in this reproduction. Without them, the same alias path can begin fetching.

Expected

Resolve a trusted catalog alias before download admission and use its declared footprint when remote metadata is unavailable. Models above the existing threshold should receive the same interactive confirmation as a full repository id. Explicit automatic-pull behavior must remain available for automation.

### LINKED ISSUE #2351: models --cached marks a loadable singleton snapshot as incomplete
User impact

A complete immutable snapshot without refs/main is loadable by the routing and loader contract, but rapid-mlx models --cached labels the same artifact incomplete. The inventory contradicts the serve path and can make a user redownload or remove a valid model.

Suggested owner: Vector (engine/server CLI/cache inventory).

Reproduction

Exact source: 31a1524b611714f0044a296abedb05c31bea3b1a, installed as a non-editable 0.13.0 wheel in a fresh Python 3.11 venv on Studio.

Cache shape: repository mlx-community/Qwen3.5-2B-MLX-4bit; exactly one snapshot at revision 93760be4f1f69842a46bc13dbdc0f19e291392a3; no refs/main; complete config and weights.

Observed human output marks mlx-community/Qwen3.5-2B-MLX-4bit incomplete at 1.6 GiB. The JSON form reports state incomplete, while the installed wheel returns the exact snapshot path from read_model_metadata and detects Qwen3_5ForConditionalGeneration.

Expected

Cache inventory and loading should consume the same completeness/resolution contract. An unambiguous complete immutable snapshot should be reported as available, with a truthful alias or repository identity.

### LINKED ISSUE #2352: doctor validates a different rapid-mlx executable than the active venv
User impact

Running Doctor from a fresh virtual environment can show a green shell-integration check for a different global Rapid-MLX executable. This can hide a PATH or version mismatch and sends troubleshooting toward the wrong installation.

Suggested owner: Vector (CLI diagnostics).

Reproduction

Exact source: 31a1524b611714f0044a296abedb05c31bea3b1a, installed as a non-editable 0.13.0 wheel in /private/tmp/rapid-server-friction.Wwvky8/venv.

Run /private/tmp/rapid-server-friction.Wwvky8/venv/bin/rapid-mlx doctor.

Doctor correctly reports Python 3.11.16 and rapid-mlx 0.13.0 from the venv, but Shell Integration says rapid-mlx in PATH at /Users/raullenstudio/.local/bin/rapid-mlx. That path is not the executable used for the Doctor invocation.

Expected

Doctor should identify the running CLI executable and separately report PATH resolution. If they differ, it should warn with both paths and an actionable activation or PATH correction.

### LINKED ISSUE #2353: served-model-name is hidden by the readiness banner and ps output
User impact

A user sets --served-model-name specifically to choose the API model identity, but the readiness banner and rapid-mlx ps continue to display the requested catalog alias. The user cannot tell which model string to copy into an SDK request.

Suggested owner: Vector (server CLI).

Reproduction

Exact source: 31a1524b611714f0044a296abedb05c31bea3b1a, installed as a non-editable 0.13.0 wheel in a fresh Python 3.11 venv on Studio.

Run rapid-mlx serve lfm2.5-1b-4bit --port 8123 --served-model-name studio-assistant --enable-audio.

Observed: readiness banner Model is lfm2.5-1b-4bit; rapid-mlx ps MODEL is lfm2.5-1b-4bit; GET /v1/models ids are studio-assistant and lfm2.5-1b-4bit; residency primary id is studio-assistant.

Expected

The readiness and process surfaces should clearly identify the served API name, and may separately show the requested alias or checkpoint. The copyable connection example should use the served identity.

### LINKED ISSUE #2357: offline serve errors repeat download attempts and recommend ineffective lane flags
User impact

Serving an uncached model while offline produces a long internal routing explanation, repeats the failed download phase, and recommends --no-mllm or --mllm even though neither flag can supply the missing checkpoint. The actionable fact is buried: the model is not cached and the network is unavailable.

Suggested owner: Vector (server CLI errors).

Reproduction

Exact source: 31a1524b611714f0044a296abedb05c31bea3b1a, installed as a non-editable 0.13.0 wheel in a fresh Python 3.11 venv on Studio.

Run HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 rapid-mlx serve badorg/offline-missing-model --port 8125.

Observed: First-time download and Pre-download skipped are printed twice; internal engine initialization logs appear between attempts; final text discusses MLLM-vs-text routing and GH #352; it recommends --no-mllm or --mllm even though config.json and weights are absent.

Expected

Detect the offline and uncached condition once, state which repository is missing, point to rapid-mlx pull after connectivity is restored or the expected cache location, and exit before server initialization. Only recommend a lane override when the checkpoint exists and the remaining failure is actually lane ambiguity.

### LINKED ISSUE #2358: OpenAI-compatible stop parameter rejects the scalar-string form
## Summary

The OpenAI-compatible chat endpoint rejects the canonical scalar-string form of `stop`. The equivalent one-element array succeeds. This creates avoidable client friction because callers may legally send either a single string or a sequence.

Suggested owner: Vector
Target: 0.13.1 (non-blocking; pre-existing in RC2)

## Environment

- Host: Mac mini, Apple M2 Pro, 32 GB, arm64, macOS 26.5.2
- Current source: `31a1524b611714f0044a296abedb05c31bea3b1a`
- Wheel SHA-256: `c9e83b83a5abf0aa09b503d7b10f5b7bc27236c675fad743c98672ad2411d734`
- Model: cached Qwen3.5 9B 4-bit checkpoint
- Server: loopback-only, base wheel, no optional extras

## Reproduction

```sh
curl -sS -i http://127.0.0.1:18337/v1/chat/completions \
  -d '{
    "model": "<served-model>",
    "messages": [{"role": "user", "content": "Say alpha STOP omega"}],
    "temperature": 0,
    "max_tokens": 32,
    "stop": "STOP"
  }'
```

Observed:

```
HTTP 400
{"error":{"message":"Invalid request body: stop: Input should be a valid list","type":"invalid_request_error","code":"invalid_request","param":"stop"}}
```

Change only the final field to `"stop": ["STOP"]`: HTTP 200, visible content `Alpha`, `finish_reason: "stop"`.

Reproduction rate: 1/1 for each form on the exact wheel.

## Expected

Normalize a scalar string to a one-element stop-sequence list at the request boundary, while preserving existing list validation and generation behavior. The same request shape should be considered for the legacy completions endpoint because both request models currently declare only `list[str]`.

## Regression check

This is not a regression from RC2. Source `33dd7a30` declares `stop: list[str] | None` on both chat and text-completion request models, matching the current behavior.

## User impact

SDKs and compatibility clients that serialize one stop sequence as a scalar string receive a deterministic 400 and must add a Rapid-specific conversion even though the request is otherwise valid.

### LINKED ISSUE #2359: Negative request timeout is accepted and reported as HTTP 504
## Summary

The OpenAI-compatible chat request accepts a negative per-request `timeout`, enqueues the request, and returns HTTP 504 with the nonsensical message `Request timed out after -1.0 seconds`. This is a client-input validation error and should return a clear 4xx before inference admission.

Suggested owner: Vector
Target: 0.13.1 (non-blocking; pre-existing in RC2)

## Environment

- Host: Mac mini, Apple M2 Pro, 32 GB, arm64, macOS 26.5.2
- Current source: `31a1524b611714f0044a296abedb05c31bea3b1a`
- Wheel SHA-256: `c9e83b83a5abf0aa09b503d7b10f5b7bc27236c675fad743c98672ad2411d734`
- Model: cached Qwen3.5 9B 4-bit checkpoint

## Reproduction

```sh
curl -sS -i http://127.0.0.1:18337/v1/chat/completions \
  -d '{
    "model": "<served-model>",
    "messages": [{"role": "user", "content": "Reply exactly TIMEOUT_NEG"}],
    "temperature": 0,
    "max_tokens": 32,
    "timeout": -1
  }'
```

Observed in 0.0594 seconds:

```
HTTP 504
{"error":{"message":"Request timed out after -1.0 seconds","type":"api_error","code":null,"param":null}}
```

Reproduction rate: 1/1. The resident request count returned to zero within one second and an immediate residency query remained healthy, so no slot leak was observed.

## Expected

Reject non-positive or non-finite request timeouts at schema validation with a 400/422 error naming `timeout` and requiring a positive finite number. The legacy completions request exposes the same unbounded field and should share the validation contract.

## Regression check

This is not a regression from RC2. Source `33dd7a30` also declares the chat and completion timeout fields as unconstrained optional floats.

## User impact

A configuration typo is reported as a server timeout rather than invalid input, which sends clients down retry/recovery paths and obscures the actual correction.

### LINKED ISSUE #2360: Failed resident reload restore can split routing from residency state
## User impact\n\nIf a per-model performance reload fails and rebuilding the previous configuration also fails, the residency manager can have no record for the primary model while legacy request routing still retains a primary-engine reference. Health/model-residency state and request behavior can therefore disagree after a rejected Settings change.\n\n## Reproduction contract\n\n1. Start with a primary resident model.\n2. Request reload_if_changed with a configuration whose loader fails.\n3. Make the rollback loader for the previous configuration fail as well.\n4. Observe that the original record was already removed and no replacement record is published; legacy primary routing may still retain its prior reference.\n\n## Expected\n\nReload failure and rollback failure must leave one explicit, consistent lifecycle state across residency, health, and request routing. No stale engine may continue serving outside residency accounting.\n\n## Scope recommendation\n\nTarget 0.13.1. Extend the existing reload transaction/primary handoff failure contract; do not redesign scheduling or capacity policy. The 0.13.0 release fix separately preserves canonical id, model path, and aliases across successful reload and successful rollback reconstruction.

### LINKED ISSUE #2361: Residency load validation leaks `<field>` for schema-owned settings
## Summary

Invalid residency performance settings return an error containing the literal placeholder `<field>` while `error.param` is null. The message explains the mutual-exclusion rule but does not identify the `performance` object that must be corrected.

Suggested owner: Vector
Target: 0.13.1 (non-blocking; pre-existing in RC2)

## Environment

- Host: Mac mini, Apple M2 Pro, 32 GB, arm64, macOS 26.5.2
- Source: `0f303f31ee1295560db9ea3c0087c6d0cff9809b`
- Locally built wheel SHA-256: `999874234fdde72429f63c2cfcf829924249e44006e9f536ff5d5126eb680e91`
- Package version: `0.13.0rc2`
- Cached Qwen3.5 9B 4-bit primary, loopback-only server

## Reproduction

```sh
curl -sS -i http://127.0.0.1:18347/v1/models/load \
  -d '{
    "model": "qwen3.5-9b-4bit",
    "model_path": "<cached-snapshot>",
    "reload_if_changed": true,
    "performance": {
      "kv_cache_dtype": "int8",
      "kv_cache_turboquant": "v4"
    }
  }'
```

Observed, 1/1:

```json
{
  "error": {
    "message": "Invalid request body: <field>: Value error, KV dtype and TurboQuant are mutually exclusive",
    "type": "invalid_request_error",
    "code": "invalid_request",
    "param": null
  }
}
```

HTTP status is 400. The model is not reloaded, its performance setting remains unchanged, and subsequent chat remains healthy.

## Expected

Keep the existing safe validation behavior, but identify the schema-owned location, for example `performance`, in the message and `error.param`. Do not reflect arbitrary client-controlled keys.

## Regression check

The RC2 source uses the same model-level validator and has no residency-load field mapping in the shared safe error-location contract, so this is pre-existing rather than introduced by the current engine fixes.

## User impact

Desktop or API callers that accidentally select mutually exclusive cache modes receive an internal placeholder instead of the setting group to correct, making recovery and support diagnostics unnecessarily difficult.

### LINKED ISSUE #2362: Residency control plane coerces string values for boolean fields
## Summary

The residency load control plane accepts string values such as `"yes"` for the boolean `pin` field. It coerces the string to `true`, loads the model as pinned, and makes it ineligible for normal unload/eviction instead of rejecting the invalid request.

Suggested owner: Vector
Target: 0.13.1 (non-blocking; pre-existing in RC2)

## Environment

- Host: Mac mini, Apple M2 Pro, 32 GB, arm64, macOS 26.5.2
- Source: `0f303f31ee1295560db9ea3c0087c6d0cff9809b`
- Wheel SHA-256: `999874234fdde72429f63c2cfcf829924249e44006e9f536ff5d5126eb680e91`
- Package version: `0.13.0rc2`

## Reproduction

With a healthy pinned primary, load a cached secondary:

```sh
curl -sS http://127.0.0.1:18347/v1/models/load \
  -d '{
    "model": "dogfood-pin-string",
    "model_path": "<cached-secondary-snapshot>",
    "estimated_size_gb": 0.5,
    "pin": "yes"
  }'
```

Observed, 1/1:

- Load returns HTTP 200 with `"pinned": true`.
- Immediate `DELETE /v1/models/dogfood-pin-string` returns HTTP 409 `pinned models cannot be unloaded`.
- Explicitly setting `{"pinned": false}` returns 200; unload then returns 204.
- Final residency returns to the original sole primary and remains healthy.

## Expected

The control-plane schema should accept only JSON booleans for `pin` and `pinned`. Strings, integers, and other truthy/falsy representations should return a 400/422 error naming the field, before any model load or pin-state mutation.

## Regression check

This is not introduced by the current engine fixes. RC2 also declares these control-plane fields as permissive boolean fields, so the same input coercion is pre-existing.

## User impact

A configuration or client serialization typo can unexpectedly make a secondary model non-evictable. On memory-constrained Macs that can block later model loads with capacity errors until the caller discovers and clears the unintended pin.

### LINKED ISSUE #2377: share CLI tests fail ONLY on Python <3.12 (stdlib argparse '--'-before-subcommand divergence), pass on 3.12; hidden from CI
## Summary
Two tests in `tests/test_share_cli.py` fail on pr_validate's `full_unit` (which runs on Python 3.11), and are excluded from CI (`.github/workflows/ci.yml` curated list). **They pass on Python 3.12.** Root cause is a **stdlib argparse behavior change**, not a rapid-mlx parsing bug.

## Failing tests (exact IDs)
- `tests/test_share_cli.py::test_share_native_double_dash_before_model_is_not_passthrough[argv1]` — argv1 = `["--", "share", "hy3-preview-4bit"]`
- `tests/test_share_cli.py::test_non_share_positional_value_share_skips_probe_no_double_convert` — argv = `["diag", "share", "--n", "5", "--", "x"]`

## Actual root cause (verified across interpreters)
argparse's handling of a leading `--` in the `add_subparsers` command slot — and of `--` before a trailing `nargs='*'` positional inside a subcommand — **changed in Python 3.12**. Confirmed with an isolated pure-argparse repro (no rapid-mlx code):

| Python | `parse_args(["--","share","model"])` | `parse_args(["diag","share","--n","5","--","x"])` |
|---|---|---|
| 3.10 | `SystemExit(2)` "invalid choice: '--'" | fails (unrecognized `-- x`) |
| 3.11 | `SystemExit(2)` "invalid choice: '--'" | fails (unrecognized `-- x`) |
| 3.12 | parses `command='share', model='...'` | parses `diag` with `rest=['x']` |

The tests encode the **desired/correct** behavior and pass on Python 3.12 (which the CI test-matrix runs). They fail only on 3.10/3.11 because of this stdlib divergence. This is **NOT a production share-CLI parsing bug**: `rapid-mlx share <model>` and `rapid-mlx share <model> -- <serveflags>` behave identically on all three versions. The exotic `-- share MODEL` / `diag MODEL -- x` forms don't occur in real production invocations.

pr_validate's `full_unit` runs Python 3.11, which is why it surfaces these failures. `test_share_cli.py` was never in ci.yml's curated pytest list, so CI never ran them.

## Not a regression
Verified failing identically at `rapid-mac-v0.12.18` (`291ede85`) and current `main`. No bisectable regression between v0.12.18 and main — it is a Python-version stdlib difference, not code rot.

## Resolution options (awaiting Atlas/goof)
- (A) Make the two assertions version-conditional (assert parse-success on Python >=3.12, assert argparse error on <3.12) so the file can be added to ci.yml without failing the 3.10/3.11 matrix nodes.
- (B) Keep `test_share_cli.py` out of the CI matrix list; accept pr_validate's 3.11 full_unit flagging these two as known pre-existing noise (document).
- (C) Do nothing to tests; leave as-is (they are correct on modern Python).

An earlier revision of this issue mis-attributed this to "tests assert impossible argparse behavior, broken since #1096" — that was based on testing on Python 3.11 only and is INCORRECT. Corrected here.

### LINKED ISSUE #2378: bug(mac): failed photo attachments poison subsequent chat turns
## User impact

A photo request that fails leaves the conversation unusable: later plain-text messages in the same conversation resend the rejected image and fail again. Starting a new conversation works. The composer also offers photo attachment whenever the catalog predicts vision, even when the resident runtime cannot actually process the image, and the resulting server reason is replaced by a generic restart message.

This is P0 for 0.13.1 because photo chat must work on the recommended Desktop model and one failed attachment must never poison later text chat.

## Exact reproduction

Validated with the published 0.13.0 Desktop DMG at tag rapid-mac-v0.13.0 / commit 4f9f9165. DMG SHA-256:

    dccdc00b7dff58db265419e7b8a748ef823a479d126519c14137f921456057d2

1. Serve qwen3.5-9b-4bit from Desktop.
2. Start a new conversation.
3. Attach a PNG and send: Tell me something about this picture.
4. Observe: Rapid could not finish the request and suggests retry/restart.
5. Send hi in the same conversation: it fails with the same message.
6. Start a new conversation and send hi: it succeeds.

## Preserved evidence

Desktop launched the model with the multimodal lane enabled, and the health endpoint reported ready=true, model_loaded=true, model_type=mllm. The photo action remained enabled and accepted the image.

The same image content-part request sent directly to the same published sidecar returned HTTP 500:

    {"error":{"message":"Internal server error"}}

The underlying sidecar stack is:

    mlx_vlm/models/qwen3_vl/vision.py:402
    TypeError: repeat(): incompatible function arguments
    Invoked with types: mlx.core.array, mlx.core.array

The scheduler converts that exception to a stable internal MLLM error, and the Desktop converts all non-capacity HTTP failures to the generic retry/restart copy. The bundled sidecar does include and execute its vision dependency; this is a runtime compatibility failure, not a missing-package error.

The same-thread failure is deterministic. The request builder deliberately inherits the latest image attachment for a plain-text follow-up, so a rejected image turn is sent again. A new conversation succeeds and the model remains healthy.

## Required Desktop behavior

- Read the resident model lane and capabilities from the server contract instead of relying only on catalog prediction.
- Disable or explain photo attachment before send when the resident lane cannot accept images. Tell the user that the current model is running text-only and that photos require a vision-capable model.
- When an image request is rejected or fails, mark that attachment turn failed with the server-provided reason and do not include its image content in later turns.
- Preserve ordinary image follow-up semantics after a successful image turn.
- Show a specific, actionable image-capability or vision-runtime message instead of the generic request failure.
- Add state/request-builder tests for rejected-image then plain-text follow-up, plus a real-image Desktop journey on the recommended model once the server vision contract is available.

## Related scope

The engine-side vision runtime compatibility and explicit residency capability/reason fields are coordinated separately. This issue owns the Desktop admission, failure-state, history, copy, and real-image journey behavior.

Markdown files are currently not document attachments by contract; only PDF, CSV, and TXT are recognized. Dragging a Markdown file can therefore insert its path as text. That is a separate format-support follow-up and does not expand this P0.

### LINKED ISSUE #2397: Reloaded secondary VLM reuses a stale MLX stream from the prior engine
## Summary

After a secondary VLM is unloaded and the same VLM is loaded again in one server process, its next image request can fail with HTTP 500 because the batching generator retains the previous engine worker stream.

Suggested owner: Vector, after #2342.

## Reproduction

Environment: Studio, installed wheel built from `33dd7a30`, cached `qwen3.5-35b-4bit` primary (revision `1e20fd8d`) and cached `gemma-4-26b-4bit` secondary (revision `0d77464e`), offline mode.

1. Start the primary and load the secondary through `POST /v1/models/load`.
2. Exercise the secondary, then unload it through `DELETE /v1/models/gemma-4-26b-4bit` (204).
3. Load the same secondary again (200).
4. Send an image chat request to the reloaded secondary.

The request returns HTTP 500. The engine log reports: `There is no Stream(gpu, 3) in current thread.` The reloaded worker owns a different stream, while `MLLMBatchGenerator` retains the stream created for the prior engine lifetime. The primary remains usable, so this is distinct from #2333.

## Expected

Each loaded VLM engine owns lifecycle-valid execution state. Unload/reload must not reuse an MLX stream belonging to a stopped worker.

## Scope

Reset or instance-own the existing MLLM batching stream at engine lifecycle boundaries, with a focused load → image request → unload → reload → image request regression test.

## Non-goals

No scheduler redesign, model-specific heuristic, capacity policy, GUI change, or model download.

raullenchai:
## Intention

Ship an accurate, user-facing 0.13.1 Desktop CHANGELOG and release-note page before the version bump. Every product claim maps to a change already integrated on main or to an explicitly named train-6c member.

## Inventory

- Direct/main: #2279, #2327, #2338, #2369, #2375, #2376, #2432, #2440.
- Train 1: #2367, #2371, #2372, #2382, #2322, #2384.
- Train 2: #2388, #2394, #2401, #2391, #2398, #2386, #2385, #2379, #2341, #2400.
- Train 3: #2404, #2408.
- Train 4: #2396, with contributor credit preserved.
- Train 5: #2402, #2423, #2392, #2395, #2417, #2424.
- Train 6/6b: #2441, #2443, #2444, #2448, #2433.
- Train 6c (merged as #2477): #2471, #2472, #2467, #2470.
- Train 6d: #2476 Desktop high-resolution photo downscaling and typed-rejection conversation recovery; #2478 explicit over-capacity load confirmation.

Not shipping and intentionally absent: #2374, #2387, #2469. The open suffix-decoding limitation is documented as a known issue with #2463.

Fixes #2239
Fixes #2145
Fixes #2222
Fixes #2358
Fixes #2359
Fixes #2361
Fixes #2362
Fixes #2377
Fixes #2378
Fixes #2360
Fixes #2397
Fixes #2350
Fixes #2349
Fixes #2353
Fixes #2342
Fixes #2348
Fixes #2219
Fixes #2351
Fixes #2343
Fixes #2352
Fixes #2357

## Scope

Documentation only: `apps/rapid-mac/CHANGELOG.md` and `docs/release-notes/v0.13.1.md`. No version bump, tag, publication, workflow, or product behavior change. The branch is reconciled to exact content-complete train-6d main `c730e45f`.

## Verification

- Release/version contract suite: 122 passed.
- `git diff --check`: passed.
- Explicit inventory audit against `v0.13.0..origin/main`, the landed train-6c contracts, and the staged train-6d release item.

### DIFF
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -30,12 +30,13 @@ can actually understand.
 - **Assistant changes are explicit, safe transactions.** API clients can choose
   whether a busy text or vision model switch should reject, wait, or abort
   active assistant work. Desktop rejects a busy switch safely, and auxiliary
-  speech models remain resident. The replacement API accepts
-  `memory_policy=keep_then_commit` or `evict_first_if_needed`; the default
-  projection-gated policy evicts the replaced model only when both cannot fit
-  and the projected replacement fits, otherwise returning a typed 507 with
-  `replacement_projection` before eviction. Use `keep_then_commit` to require
-  a rollback-safe load. ([#2369](https://github.com/raullenchai/Rapid-MLX/pull/2369))
+  speech models remain resident. The `/v1/models/load` API defaults
+  `memory_policy` to `evict_first_if_needed`: it keeps the rollback-safe load
+  when old and new fit together, evicts the replaced assistant first only when
+  keeping both breaches the limit but the replacement alone fits, and otherwise
+  returns a typed 507 with `replacement_projection` before destructive mutation.
+  Set `memory_policy=keep_then_commit` to require rollback-safe loading.
+  ([#2369](https://github.com/raullenchai/Rapid-MLX/pull/2369))
 - **Large multi-variant repositories can download one serving format.** The CLI
   accepts `--bits` or `--format`, so users do not need to fetch every
   quantization in a repository. ([#2145](https://github.com/raullenchai/Rapid-MLX/issues/2145),
@@ -55,13 +56,16 @@ can actually understand.
   validation now requires image-dependent answers from two cached vision
   families before the sidecar can ship. ([#2384](https://github.com/raullenchai/Rapid-MLX/pull/2384),
   [#2380](https://github.com/raullenchai/Rapid-MLX/issues/2380))
-- **Vision-capable Qwen checkpoints choose the working lane by default.** With
-  mlx-vlm 0.6.16 or newer (included in the Desktop sidecar), Qwen3.5, Qwen3.6,
-  and supported Qwen3.8 vision-capable checkpoints serve on the vision lane so
-  photos work directly in chat. The experimental Flash-Next checkpoint remains
-  text-only in this release. Use `--no-mllm` in the CLI or the per-model
-  Performance override in Desktop to force text-only serving; MTP and
-  speculative decoding automatically retain the text lane.
+- **Vision-capable Qwen checkpoints choose the working lane automatically.**
+  Vision-capable Qwen3.5, Qwen3.6, and supported Qwen3.8 checkpoints use the
+  vision lane automatically when mlx-vlm 0.6.16 or newer is installed and the
+  checkpoint meets its measured vision-memory recommendation. If automatic
+  admission cannot use that lane, serving falls back to text with a
+  machine-readable reason. The experimental Flash-Next checkpoint is explicitly
+  text-only in this release. CLI users can force text serving with `--no-mllm`,
+  and Desktop exposes the same per-model Performance choice. Requests that
+  enable MTP or another speculative decoder select the text lane, where that
+  decoder is supported.
 - **API validation fails early with actionable fields.** Requests accept scalar
   or array `stop`, reject invalid timeouts and non-boolean residency controls,
   and identify the exact invalid load field. ([#2367](https://github.com/raullenchai/Rapid-MLX/pull/2367),
@@ -112,10 +116,12 @@ can actually understand.
 - **Speech to Text arms itself after model lifecycle changes.** When speech
   input is enabled, it becomes ready without an extra manual start button.
   ([#2448](https://github.com/raullenchai/Rapid-MLX/pull/2448))
-- **Structured text requests terminate normally on the vision lane.** JSON
-  schema and title-generation requests now carry their output constraints and
-  complete with a terminal stop while photo requests continue using the same
-  multimodal server. ([#2471](https://github.com/raullenchai/Rapid-MLX/pull/2471))
+- **Structured text requests terminate normally on the vision lane.**
+  JSON-schema responses and bounded-thinking or title-generation requests now
+  terminate normally on the vision lane: their request-local output processors
+  and complete configured EOS set reach the multimodal scheduler, while photo
+  requests continue using the same server.
+  ([#2471](https://github.com/raullenchai/Rapid-MLX/pull/2471))
 - **Serving lanes honor the requested workload.** Experimental
   Qwen3.8-Flash-Next remains on its supported text lane, MTP and other
   speculative decoding requests choose the text lane, and text-diffusion
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -22,8 +22,10 @@ can actually understand.
 ### Added
 
 - **Experimental Qwen3.8-Flash-Next text inference.** The M1 lane supports the
-  mixed 4-bit checkpoint on Macs with at least 128 GB of unified memory. MTP
-  and vision support remain planned for a later release.
+  mixed 4-bit checkpoint through the `qwen3.8-flash-next-4bit` alias on Macs
+  with at least 128 GB of unified memory. This release exposes the experimental
+  text lane through the CLI and API; Desktop, MTP, and vision support remain
+  planned for a later release.
   ([#2433](https://github.com/raullenchai/Rapid-MLX/pull/2433))
 - **Assistant changes are explicit, safe transactions.** API clients can choose
   whether a busy text or vision model switch should reject, wait, or abort
@@ -83,6 +85,19 @@ can actually understand.
 - **Dictation remains available across chat-model switches.** The speech model
   stays resident and is restored after switching chat models.
   ([#2400](https://github.com/raullenchai/Rapid-MLX/pull/2400))
+- **MTP assistants survive loading a second model.** Generation remains bound
+  to its worker, so the original assistant continues answering instead of
+  losing its GPU stream. ([#2441](https://github.com/raullenchai/Rapid-MLX/pull/2441))
+- **Model-switch memory projections credit the assistant being replaced.** The
+  estimate uses the same fit thresholds as the replacement transaction and no
+  longer warns about memory that will be freed before the new model starts.
+  ([#2443](https://github.com/raullenchai/Rapid-MLX/pull/2443),
+  [#2444](https://github.com/raullenchai/Rapid-MLX/pull/2444))
+- **Speech to Text arms itself after model lifecycle changes.** When speech
+  input is enabled, it becomes ready without an extra manual start button.
+  ([#2448](https://github.com/raullenchai/Rapid-MLX/pull/2448))
+<!-- TRAIN-6 PLACEHOLDER: add the Claude Code connection correction only after
+the bisect identifies and validates the exact landed fix. -->
 - **Photo rejection and multimodal reloads recover cleanly.** A rejected image
   cannot poison the next text turn, and a reloaded vision model cannot reuse a
   stale generation worker. ([#2379](https://github.com/raullenchai/Rapid-MLX/pull/2379),
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -55,6 +55,12 @@ can actually understand.
   validation now requires image-dependent answers from two cached vision
   families before the sidecar can ship. ([#2384](https://github.com/raullenchai/Rapid-MLX/pull/2384),
   [#2380](https://github.com/raullenchai/Rapid-MLX/issues/2380))
+- **Vision-capable Qwen checkpoints choose the working lane by default.** With
+  mlx-vlm 0.6.16 or newer (included in the Desktop sidecar), Qwen3.5, Qwen3.6,
+  and Qwen3.8 vision-capable checkpoints serve on the vision lane so photos
+  work directly in chat. Use `--no-mllm` in the CLI or the per-model
+  Performance override in Desktop to force text-only serving; MTP and
+  speculative decoding automatically retain the text lane.
 - **API validation fails early with actionable fields.** Requests accept scalar
   or array `stop`, reject invalid timeouts and non-boolean residency controls,
   and identify the exact invalid load field. ([#2367](https://github.com/raullenchai/Rapid-MLX/pull/2367),
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -38,6 +38,21 @@ can actually understand.
   prevents models from selecting general web search or claiming the advertised
   Weather tool is unavailable. ([#2222](https://github.com/raullenchai/Rapid-MLX/issues/2222),
   [#2327](https://github.com/raullenchai/Rapid-MLX/pull/2327))
+- **Desktop photo input uses the updated bundled vision runtime.** Release
+  validation now requires image-dependent answers from two cached vision
+  families before the sidecar can ship. ([#2384](https://github.com/raullenchai/Rapid-MLX/pull/2384),
+  [#2380](https://github.com/raullenchai/Rapid-MLX/issues/2380))
+- **API validation fails early with actionable fields.** Requests accept scalar
+  or array `stop`, reject invalid timeouts and non-boolean residency controls,
+  and identify the exact invalid load field. ([#2367](https://github.com/raullenchai/Rapid-MLX/pull/2367),
+  [#2371](https://github.com/raullenchai/Rapid-MLX/pull/2371),
+  [#2372](https://github.com/raullenchai/Rapid-MLX/pull/2372))
+- **Guided setup accurately describes data preservation.** Its confirmation no
+  longer implies that onboarding reset deletes user data. ([#2239](https://github.com/raullenchai/Rapid-MLX/issues/2239),
+  [#2322](https://github.com/raullenchai/Rapid-MLX/pull/2322))
+- **Share CLI contracts use supported invocation forms and run in CI.**
+  ([#2377](https://github.com/raullenchai/Rapid-MLX/issues/2377),
+  [#2382](https://github.com/raullenchai/Rapid-MLX/pull/2382))
 
 ### Documentation
 
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -55,6 +55,15 @@ can actually understand.
 - **Share CLI contracts use supported invocation forms and run in CI.**
   ([#2377](https://github.com/raullenchai/Rapid-MLX/issues/2377),
   [#2382](https://github.com/raullenchai/Rapid-MLX/pull/2382))
+- **Environment diagnosis identifies the CLI that is actually running.**
+  `rapid-mlx doctor` reports both the active-environment executable and a
+  different installation found on `PATH`, with actionable mismatch guidance.
+  ([#2352](https://github.com/raullenchai/Rapid-MLX/issues/2352),
+  [#2402](https://github.com/raullenchai/Rapid-MLX/pull/2402))
+- **Offline serving failures are concise and actionable.** An uncached model
+  now stops after one cache and connectivity explanation instead of repeating
+  download work or recommending an unrelated serving lane.
+  ([#2357](https://github.com/raullenchai/Rapid-MLX/issues/2357))
 - **First run chooses a hardware-fit starter and photos use the correct lane.**
   Eligible cached models remain preferred, and supported hybrid vision models
   route automatically. ([#2385](https://github.com/raullenchai/Rapid-MLX/pull/2385),
@@ -90,6 +99,10 @@ can actually understand.
   [#2350](https://github.com/raullenchai/Rapid-MLX/issues/2350),
   [#2392](https://github.com/raullenchai/Rapid-MLX/pull/2392),
   [#2349](https://github.com/raullenchai/Rapid-MLX/issues/2349))
+- **Cached speech checkpoints are recognized by their real layouts.** Complete
+  Kokoro and Whisper Turbo snapshots count as runnable when their verified
+  family-specific weights are present. ([#2406](https://github.com/raullenchai/Rapid-MLX/issues/2406),
+  [#2417](https://github.com/raullenchai/Rapid-MLX/pull/2417))
 - **System-prompt settings explain what is actually sent.** Saved and effective
   prompts are distinguished, and optional credentials are read only when
   needed. ([#2341](https://github.com/raullenchai/Rapid-MLX/pull/2341))
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -36,9 +36,9 @@ can actually understand.
   as LFM2.5 2.6B 4-bit no longer fall back to a slow or unreachable upstream
   route merely because the quantization lives in a repository subfolder.
   ([#2279](https://github.com/raullenchai/Rapid-MLX/pull/2279))
-- **Explicit weather requests stay on the Weather tool.** Evaluation coverage
-  prevents models from selecting general web search or claiming the advertised
-  Weather tool is unavailable. ([#2222](https://github.com/raullenchai/Rapid-MLX/issues/2222),
+- **Explicit weather requests have dedicated regression coverage.** Evaluation
+  scenarios detect models selecting general web search or claiming the
+  advertised Weather tool is unavailable. ([#2222](https://github.com/raullenchai/Rapid-MLX/issues/2222),
   [#2327](https://github.com/raullenchai/Rapid-MLX/pull/2327))
 - **Desktop photo input uses the updated bundled vision runtime.** Release
   validation now requires image-dependent answers from two cached vision
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -53,6 +53,44 @@ can actually understand.
 - **Share CLI contracts use supported invocation forms and run in CI.**
   ([#2377](https://github.com/raullenchai/Rapid-MLX/issues/2377),
   [#2382](https://github.com/raullenchai/Rapid-MLX/pull/2382))
+- **First run chooses a hardware-fit starter and photos use the correct lane.**
+  Eligible cached models remain preferred, and supported hybrid vision models
+  route automatically. ([#2385](https://github.com/raullenchai/Rapid-MLX/pull/2385),
+  [#2219](https://github.com/raullenchai/Rapid-MLX/issues/2219),
+  [#2388](https://github.com/raullenchai/Rapid-MLX/pull/2388))
+- **Dictation remains available across chat-model switches.** The speech model
+  stays resident and is restored after switching, while a proven chat selection
+  survives an in-session restart. ([#2400](https://github.com/raullenchai/Rapid-MLX/pull/2400),
+  [#2374](https://github.com/raullenchai/Rapid-MLX/pull/2374),
+  [#2363](https://github.com/raullenchai/Rapid-MLX/issues/2363),
+  [#2364](https://github.com/raullenchai/Rapid-MLX/issues/2364))
+- **Photo rejection and multimodal reloads recover cleanly.** A rejected image
+  cannot poison the next text turn, and a reloaded vision model cannot reuse a
+  stale generation worker. ([#2379](https://github.com/raullenchai/Rapid-MLX/pull/2379),
+  [#2378](https://github.com/raullenchai/Rapid-MLX/issues/2378),
+  [#2401](https://github.com/raullenchai/Rapid-MLX/pull/2401),
+  [#2397](https://github.com/raullenchai/Rapid-MLX/issues/2397))
+- **Live API identity and cancellation are dependable.** Readiness and connect
+  guidance use the live port and served model name, and public streaming IDs
+  are cancellable. ([#2386](https://github.com/raullenchai/Rapid-MLX/pull/2386),
+  [#2348](https://github.com/raullenchai/Rapid-MLX/issues/2348),
+  [#2395](https://github.com/raullenchai/Rapid-MLX/pull/2395),
+  [#2353](https://github.com/raullenchai/Rapid-MLX/issues/2353),
+  [#2398](https://github.com/raullenchai/Rapid-MLX/pull/2398),
+  [#2342](https://github.com/raullenchai/Rapid-MLX/issues/2342))
+- **Reload failures preserve a truthful serving state.** Registry, residency,
+  routing, readiness, and speech handoff remain consistent even when both a
+  primary reload and its restoration fail. ([#2394](https://github.com/raullenchai/Rapid-MLX/pull/2394),
+  [#2360](https://github.com/raullenchai/Rapid-MLX/issues/2360))
+- **Download status and size warnings stay truthful offline.** Catalog size is
+  retained when remote metadata is unavailable, and cached pulls report
+  verification rather than a new download. ([#2391](https://github.com/raullenchai/Rapid-MLX/pull/2391),
+  [#2350](https://github.com/raullenchai/Rapid-MLX/issues/2350),
+  [#2392](https://github.com/raullenchai/Rapid-MLX/pull/2392),
+  [#2349](https://github.com/raullenchai/Rapid-MLX/issues/2349))
+- **System-prompt settings explain what is actually sent.** Saved and effective
+  prompts are distinguished, and optional credentials are read only when
+  needed. ([#2341](https://github.com/raullenchai/Rapid-MLX/pull/2341))
 
 ### Documentation
 
@@ -60,6 +98,8 @@ can actually understand.
   supported text-model path. ([#2376](https://github.com/raullenchai/Rapid-MLX/pull/2376))
 - Published a reproducible M2 Pro comparison of 0.13.0 and 0.12.18.
   ([#2375](https://github.com/raullenchai/Rapid-MLX/pull/2375))
+
+<!-- Train 3 entries are added here only after that train lands. -->
 
 ## [0.13.0] — 2026-08-26
 
--- a/apps/rapid-mac/CHANGELOG.md
+++ b/apps/rapid-mac/CHANGELOG.md
@@ -24,8 +24,8 @@ can actually understand.
 - **Experimental Qwen3.8-Flash-Next text inference.** The M1 lane supports the
   mixed 4-bit checkpoint through the `qwen3.8-flash-next-4bit` alias on Macs
   with at least 128 GB of unified memory. This release exposes the experimental
-  text lane through the CLI and API; Desktop, MTP, and vision support remain
-  planned for a later release.
+  text lane through the CLI and API; release validation used a 256 GB Mac, and
+  Desktop, MTP, and vision support remain planned for a later release.
   ([#2433](https://github.com/raullenchai/Rapid-MLX/pull/2433))
 - **Assistant changes are explicit, safe transactions.** API clients can choose
   whether a busy text or vision model switch should reject, wait, or abort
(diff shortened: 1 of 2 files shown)


raullenchai (commented):
Technical accuracy pass on engine claims at 496cbbb2.

Please apply these exact wording corrections in both apps/rapid-mac/CHANGELOG.md and docs/release-notes/v0.13.1.md:

1. Replace the automatic vision-lane paragraph with:

“Vision-capable Qwen3.5, Qwen3.6, and supported Qwen3.8 checkpoints use the vision lane automatically when mlx-vlm 0.6.16 or newer is installed and the checkpoint meets its measured vision-memory recommendation. If automatic admission cannot use that lane, serving falls back to text with a machine-readable reason. The experimental Flash-Next checkpoint is explicitly text-only in this release. CLI users can force text serving with --no-mllm, and Desktop exposes the same per-model Performance choice. Requests that enable MTP or another speculative decoder select the text lane, where that decoder is supported.”

This avoids implying that every supported checkpoint takes the vision lane on every RAM tier.

2. Replace the replacement-policy sentence with:

“The /v1/models/load API defaults memory_policy to evict_first_if_needed: it keeps the rollback-safe load when old and new fit together, evicts the replaced assistant first only when keeping both breaches the limit but the replacement alone fits, and otherwise returns a typed 507 with replacement_projection before destructive mutation. Set memory_policy=keep_then_commit to require rollback-safe loading.”

This states the actual public default and the three admission outcomes.

3. Replace the structured-request bullet with:

“JSON-schema responses and bounded-thinking or title-generation requests now terminate normally on the vision lane: their request-local output processors and complete configured EOS set reach the multimodal scheduler, while photo requests continue using the same server.”

The current phrase “title-generation constraints” is imprecise because title generation is a workload, not itself an output constraint.

The Flash-Next claims are otherwise accurate: experimental CLI/API text-only support, 128 GB admission floor, physical validation on 256 GB, and Desktop catalog, MTP, and vision deferred. The suffix-decoding known issue and workaround match #2463. I found no prohibited comparison/reference names.

PR #2396, opened 2026-08-26
fix(mac): stabilize chat image and file attachment input

osdodo:
## What does this PR do?

  - Caps MLX at `<0.32.2` so Desktop bundles the validated MLX 0.32.1 runtime.
  - Prevents file URL drops on the native chat text editor from becoming literal file paths.
  - Routes supported drops through the existing attachment importer and consumes rejected file drops safely.
  - Adds regression coverage for MLX dependency resolution and native text-view file drops.

  ## Why is this needed?

  Pasting an image into a Qwen3.5 conversation returned HTTP 500 in locally built Desktop apps. MLX 0.32.2 changed `mx.repeat` to reject the array-valued `repeats` argument used by mlx-vlm 0.6.3's Qwen3.5 vision tower. Pinning the validated MLX 0.32.1 runtime restores multimodal requests; the original screenshot now returns HTTP 200 from the bundled sidecar.

  Dragging a file directly onto the underlying `NSTextView` could also bypass the outer SwiftUI drop handler. AppKit then inserted the file path into the message draft instead of attaching or rejecting the file.

### DIFF
--- a/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
+++ b/apps/rapid-mac/Sources/Rapid/UI/ChatView.swift
@@ -638,6 +638,7 @@ struct ChatView: View {
                     onSubmit: send,
                     onCancel: { viewModel.stop() },
                     onPasteAttachments: pasteAttachmentsFromClipboard,
+                    onDropAttachments: addAttachmentURLs,
                     onRecallLastUser: {
                         messages.last(where: { $0.role == .user })?.content
                     }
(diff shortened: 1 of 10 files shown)


raullenchai (changes requested):
Thanks for narrowing this to the native composer drop path; the MLX dependency cap is no longer in the diff, and the focused attachment suite passes 18/18 on exact head `59b787a6`.

One blocking native-behavior issue remains:

- `AutosizingTextView` overrides the `NSDraggingDestination` callbacks but never registers the view for file-URL drag types. A plain `NSTextView` reports an empty `registeredDraggedTypes` array on macOS, and AppKit only considers a view a drag destination after `registerForDraggedTypes(_:)`. Therefore a real file drag is not guaranteed to call these new overrides; the tests call `consumeFileDrop` directly and do not exercise the registration contract, so the original path-insertion bug can remain. Please register `.fileURL` on the native editor (without broadening accepted content) and add a focused assertion that the constructed text view advertises that drag type.

This stays entirely within the two-file PR intent; no MLX/dependency changes are requested.

PR #2422, opened 2026-08-26
feat: configurable API authentication for embedded engine

### LINKED ISSUE #2599: feat(mac): persist embedded API bearer across model restarts
## User impact

Rapid-MLX Desktop currently rotates its embedded-engine bearer on every model start. Local tools then lose access until users copy the newly generated key again.

## Intended behavior

- Keep per-launch rotation as the secure default.
- Offer Keychain-backed daily and explicit-rotation lifetimes without ever offering an unauthenticated mode.
- Treat missing, malformed, or unavailable persisted credentials as a fail-safe one-time key, never as permission to start without authentication.
- Make rotation timing and persistence degradation explicit in the UI.
- Preserve loopback binding and the existing CORS boundary.

## Acceptance evidence

Focused Desktop contract tests must cover secret shape, storage separation, rotation, corrupted credentials, Keychain-write degradation, spawned environment injection, and client bearer identity. Exact-head Desktop CI and accessibility gates must pass.

CortonKwok-GGD:
## Why

Rapid-MLX Desktop mints a fresh random bearer for every embedded-engine start. That remains the safest default, but it makes local API tools brittle because every model restart invalidates their saved credential. Users need an explicit way to keep a strong local bearer across restarts without weakening the loopback authentication boundary.

Fixes #2599. Addresses the workflow discussed in Discussion #1763.

## What

- Adds Settings → API Auth with three lifetimes: **Per launch** (default), **Daily**, and **Permanent**.
- Stores persistent bearers in the system Keychain; UserDefaults stores only non-secret mode metadata.
- Daily mode replaces the key on the first model start after 24 hours. A running engine keeps the bearer it was spawned with, and the UI states that timing explicitly.
- Rotate creates a new key and restarts a running engine after confirmation so the replacement takes effect immediately.
- The Connect surface reuses a shared masked/copyable key row and links directly to API Auth settings.

## Security boundary

- No unauthenticated mode is introduced. Per-launch random authentication remains the default.
- Loopback binding and the existing CORS allowlist are unchanged.
- Persisted credentials must match the generated 64-character lowercase-hex shape. Missing, malformed, expired, or unreadable entries resolve to a fresh strong bearer, never an unauthenticated child.
- If Keychain persistence fails, the engine still starts authenticated with a one-time bearer and Settings reports that cross-launch reuse degraded.
- Secrets remain outside preferences and process arguments.

## Design precedent

The implementation follows the established credential-management pattern already used by the app: system credential storage for secrets, preferences only for non-secret configuration, secure random generation, explicit rotation, and a fail-closed runtime path. The lifecycle contract has one authoritative spawn resolver so cold starts, model switches, and automatic respawns cannot diverge.

## Scope and non-goals

Swift/Desktop only. No engine API changes, remote-listen changes, CORS changes, user-entered secret path, or live request-time TTL is added. Daily rotation is intentionally enforced at the next model start rather than interrupting an active inference session.

## Verification

- `swift test --no-parallel --filter APIAuthConfigTests` — 24 passed after maintainer hardening.
- Accessibility identifier audit — 0 unlabeled controls.
- Exact-head `pr_validate`, full Desktop tests, and hosted required checks will be recorded before merge authorization.


raullenchai (changes requested):
Thank you for the thoughtful contribution and for preserving the existing random-key path. The configurable-auth direction is useful, but the new security boundaries need a few changes before this can merge.

Please keep the per-launch random bearer as the default, with the server hard-bound to 127.0.0.1 and the current explicit CORS allowlist unchanged.

For fixed mode:
- Store the bearer through the existing Keychain abstraction, not plaintext UserDefaults. UserDefaults may retain non-secret mode metadata only.
- Provide a generated persistent-key path and enforce a meaningful minimum strength for user-entered values.
- Preserve the current fail-safe behavior: a missing, empty, or corrupt fixed credential resolves to random, never unauthenticated.

Please remove the off mode from this PR. If product ownership later retains an unauthenticated mode, it needs a separate explicit downgrade confirmation, wording that accurately covers all local processes and local OS users, and contract tests that pin loopback binding and the CORS boundary.

Please also add integration/contract coverage for the actual spawned environment, Keychain replace/delete and no secret in defaults, restart/automatic-respawn consistency, and the affected client paths with the configured bearer. The focused APIAuthConfigTests pass 14/14 at exact head 209d9bc1; the remaining gap is proving the end-to-end storage and spawn/client contracts.

This remains a conditional 0.13.2 candidate once those scoped changes and exact-head checks are complete. Thanks again for taking on this security-sensitive UX.

raullenchai (changes requested):
Thanks for the substantial hardening since the first review. The unsafe `off` mode and plaintext secret storage are gone, the random per-launch default remains fail-safe, and the loopback/CORS boundary is unchanged. I reproduced the current exact head `18ab0d90`; four blocking items remain before this is ready for another review.

1. `apps/rapid-mac/Tests/RapidTests/APIAuthConfigTests.swift:10` does not compile against the current tree. `KeychainStoring` is `Sendable`, but the new `InMemoryKeychain` has mutable unsynchronized state:

   ```
   error: stored property 'items' of 'Sendable'-conforming class 'InMemoryKeychain' is mutable
   ```

   Reproduction: `swift test --no-parallel --filter APIAuthConfigTests`. The PR is 33 commits behind `main`; please rebase first, then reuse or adapt the synchronized in-memory Keychain test helper now on `main` (including the failure seams) instead of retaining an unsafely mutable duplicate.

2. `pr_validate` completed with `DO NOT MERGE`: 19,082 tests passed, but `test_shipping_tree_has_no_unlabelled_native_controls` found five unlabeled controls in `SettingsAPIAuthPanel.swift` (the Copy button and the confirmation/dialog buttons at current lines 126, 145, 156, 159, and 167). Add stable accessibility identifiers for every native control and rerun the validator.

3. The “24 hours” contract is not currently true for a running engine. `bearerForSpawn(now:)` checks age only during spawn, while the child keeps the old bearer indefinitely. If a model stays up for several days, the UI's stated expiry passes but the bearer remains valid. Please either enforce expiry at the authentication boundary / with a deliberate restart policy, or rename and word this as “rotate on the next model start after 24 hours” so it does not promise a TTL it cannot enforce. Add a test covering the chosen user-visible behavior. Mature credential designs either enforce an expiry on every authenticated request or present a key as long-lived until explicit rotation; they do not display an unenforced expiry.

4. The previously requested spawn/client contract coverage is still absent. The configuration-only suite does not prove that the resolved key is the exact nonempty value injected into `RAPID_MLX_API_KEY`, published as `activeBearer`, reused/rotated across restart and automatic respawn, and consumed by chat/audio/MCP paths. The rebased tree already has focused environment, bearer-lifecycle, chat-auth, voice, and MCP test seams; please extend those rather than requiring a real model or real Keychain in CI. Also cover a denied persistent write: the engine must remain authenticated and the UI/behavior must truthfully report that persistence degraded.

Before the next review, please also bring the PR body to the repository contract: start with `## Why`, remove task-list checkboxes, add `## Design precedent`, and keep the implementation evidence under normal headings. Once the branch is rebased and these items are pushed, tag me directly here; I will rerun `pr_validate`, the focused Swift suite, and exact-head CI.

raullenchai (approved):
Exact head 6c39d84a: Codex review converged at round 9 (LGTM); local pr_validate is MERGE-SAFE (19,532 passed), full Swift suite completed 3,063/3,063, and the reviewed auth/lifecycle contract is satisfied. Hosted Mac build is still in progress; queue authorization remains withheld until it and all exact-head checks are green.

Issue #2499, opened 2026-08-27
Epic: 0.13.2 release-path hardening (0.13.1 retrospective)


raullenchai:
## Why
The 0.13.1 cycle (0.13.0 → 0.13.1 in ~26 h, 9 integration trains, ~22 integration pushes) hit 32 distinct bumps: CI selection lists, host-dependent tests, Swift test hangs, a required-check facade, a release pre-flight that never runs, a Tier-1 gate first executed at tag time, dependency skew between the CLI and the Desktop sidecar, memory-projection regressions, and lane-parity gaps found late by review. Owner's direction: **0.13.2 must not be this bumpy** — fix the root causes, no tech debt, adopt established upstream patterns, and verify every fix from the user's/contributor's side.

### CI / tests
- [x] #2486 `requires_mlx` marker replaces the hand-curated no-MLX roster
- [x] #2487 `scripts/train_gates.sh` reproduces the hosted gate matrix + drift test
- [x] #2489 hermetic HF cache / env fixtures + declared lane dependency sets
- [x] #2490 unlabeled PR gate = pending/skipped, not failure; merge-queue evaluation
- [x] #2493 PR template with contract fields + exact merge-base review
### Desktop tests
- [x] #2488 per-suite time limits, hang watchdog artifact, RAM/process/clock seams (re-enable #2480, fix #2374)
- [x] #2494 behaviour tests instead of source-string contracts; generated journey manifests; `make ax-baselines`
- [x] #2497 memory-projection invariants (table-driven)
- [x] #2481 flaky drag/paste XCUITest race
### Release workflow
- [x] #2491 required bump-PR guard (CHANGELOG section, release notes, subject) + pre-flight split
- [x] #2492 `dry_run` input for auto-release
- [x] #2484 pinned sidecar snapshot preflight + disk-hygiene protection
- [x] #2451 build-once / promote the gated candidate DMG
- [x] #2496 checked-in top-10 model sequence for the Tier-1 smoke + MTP accept-rate assertion (#2421)
### Engine / packaging
- [ ] #2495 lane-parity contract matrix (text vs multimodal)
- [x] #2475 vision extra resolves the validated runtime (+ fresh-venv resolution contract)
### Host operations
- [x] #2498 protected-model source of truth, hygiene script, dogfood host precheck, large-model host lock

## Done when
[...]

Issue #2438, opened 2026-08-26
[MTP] load-second-and-back 503 after /v1/models/load: 'There is no Stream(gpu, 3) in current thread' (pre-existing in 0.13.0)


raullenchai:
## Summary
Serving `qwen3.8-27b-4bit` with MTP spec-decode (`--speculative-config '{"method":"mtp","num_speculative_tokens":3}'`) then running a **load-second-and-back** cycle (`/v1/models/load qwen3.6-27b-4bit` → chat on the 2nd model) returns **HTTP 503** deterministically. The MTP model's own generation (Chinese/EN chat, auto tool call, 4K prompt, /v1/models, /v1/models/residency) all work; the failure is isolated to generating AFTER a second model is loaded.

**Reproduced on shipped `rapid-mlx 0.13.0` (PyPI) and on release sha `62a038c7` — pre-existing, not introduced by 0.13.1/train-5.**

## Exact sequence (Studio, Apple M3 Ultra, 256GB, mlx 0.32.2 + [vision,mtp] extras)
```
rapid-mlx serve qwen3.8-27b-4bit --port 8310 --host 127.0.0.1 \
  --enable-auto-tool-choice --tool-call-parser hermes \
  --speculative-config '{"method":"mtp","num_speculative_tokens":3}'
# 1) POST /v1/chat/completions (works)
# 2) POST /v1/models/load {"model":"qwen3.6-27b-4bit"}  -> 200 OK
# 3) POST /v1/chat/completions on the 2nd model (model="default") -> 503
```

## Error (exact)
Client gets:
```json
{"error":{"message":"Inference aborted after generation error: RuntimeError: [MTP-vendored] uid=6 generator raised mid-stream (RuntimeError: There is no Stream(gpu, 3) in current thread.); cannot fall back to plain decode without corrupting the output stream."}}
```

Server traceback (shipped 0.13.0 wheel path shown):
```
ERROR:rapid_mlx.scheduler:[MTP-vendored] generator raised on uid=6 mid-stream: There is no Stream(gpu, 3) in current thread.. Terminating request — cannot fall back to plain decode because gb._next_tokens is stale relative to the tokens already emitted by the vendored path.
Traceback (most recent call last):
  File ".../vllm_mlx/scheduler.py", line 1465, in _mtp_step
    tok_int, lp_arr, _from_draft = next(gen)
  File ".../vllm_mlx/spec_decode/mtp/generator.py", line 847, in mtp_generate_step
[...]

Issue #2421, opened 2026-08-26
MTP on qwen3.8-27b hybrid: spec-decode attempts counter stays 0 / accept-rate unmeasurable


raullenchai:
## Summary
With \`--speculative-config {\"method\":\"mtp\",\"num_speculative_tokens\":3}\` on alias \`qwen3.8-27b-4bit\`, the serve log confirms MTP is ACTIVE (\`[mtp.inject] Built MTP module\`, \`Spec-decode: mtp (chain)\`, 31/31 MTP weight tensors loaded, max_k=3), yet the Prometheus \`rapid_mlx_spec_decode_attempts_total{method=\"mtp\"}\` counter reads \`0\` and \`rapid_mlx_spec_decode_accept_ratio\` reads \`0.0\` on a canonical 1024-token generation. The MTP accept-rate is therefore UNMEASURABLE via /metrics despite MTP being enabled.

## Evidence (train-2 engine, fresh wheel 0.13.0 from main eb9b3352, Studio offline cached)
- Serve command: \`rapid-mlx serve qwen3.8-27b-4bit --port X --enable-auto-tool-choice --tool-call-parser hermes --speculative-config '{\"method\":\"mtp\",\"num_speculative_tokens\":3}'\`
- Log: \`[mtp.inject] Built MTP module (1 layer(s), hidden_size=5120)\`, \`Spec-decode: mtp (chain) +sidecar=rapid-mlx/Qwen3.8-27B-4bit-MTP-MLX\`, \`Loaded 31/31 expected MTP weight tensors\`, \`MTP: enabled via --speculative-config, max_k=3\`, \`spec decode MTP (explicit)\`
- Canonical long generation (1024 completion tokens, 3020-prompt-class): \`rapid_mlx_spec_decode_attempts_total{family=\"qwen3.8-27b-4bit\",method=\"mtp\"} 0\`, \`rapid_mlx_spec_decode_accept_ratio 0.0\`
- Only a temperature-0.0 4K prompt run recorded \`4\` attempts / \`4\` accepts (ratio 1.0) — sample too thin to be meaningful.
- Note: decode tok/s was healthy (35.07 tok/s at 1024 tokens) — the model serves fine; the issue is the accept-rate observability/engagement, not raw throughput.

## Probable causes to investigate
1. MTP is throttled / not engaging decode-side on this hybrid Mamba/linear-attention backbone (GatedDeltaNet), so no MTP outer-loop attempts are counted on ordinary generations.
[...]

PR #2509, opened 2026-08-27
tests: checked-in top-10 residency load-order sequence spec (#2496)

### LINKED ISSUE #2496: release smoke: checked-in top-10 model sequence file for the Tier-1 gate + MTP accept-rate assertion
## Why
The 0.13.1 release blocker "second `/v1/models/load` after an MTP primary -> 503 no Stream(gpu,3)" (#2438, fixed by #2441) was pre-existing in 0.13.0 and was missed because the top-10 smoke's load order was not fixed ("second model may have differed"). MTP "on but unmeasurable" cannot fail loudly today (#2421).

## Scope
- `tests/integrations/top10_sequences.yaml` with exact A->B->A load orders (including MTP-second) driven by `tests/integrations/agent_smoke.sh` in the Tier-1 gate.
- Assert the MTP accept-rate metric is present and above a floor when spec-decode is on (#2421).

## Acceptance
- The #2438 sequence is reproduced by the smoke on a build without #2441 and passes on main.

## Owner
ds0731 (sequence file), Vector (#2421 metric). Refs #2438 #2441 #2421.

---
Part of the 0.13.2 release-path hardening epic (0.13.1 retrospective, bump ledger in the Atlas desk `.orca/retro-0.13.1-bumps.md`). Principles: no tech debt (no permanent skips), adopt the established upstream pattern instead of a custom mechanism, verify from the user's/contributor's perspective.

raullenchai:
## Why
The Tier-1 smoke currently hand-rolls per-run model orders. A checked-in, versioned sequence spec makes the exact top-10 residency load orders (incl. the MTP-first-then-second repro) a reviewable, testable contract instead of a shell-only convention.

## Scope
- `tests/integrations/top10_sequences.yaml` — the checked-in top-10 residency load-order sequence spec, including the MTP-first load (the #2438 repro shape), wired into the Tier-1 gates (`auto-release.yml` + `agent-gate.yml`). `agent_smoke.sh` gains a `--sequences` mode that drives it.
- `tests/test_top10_sequences_schema.py` — pure-CPU schema test that validates every sequence entry (real `ModelLoadRequest` field names, per #2361 field-mapping), automatically discovered by the hosted no-MLX lane.
- `tests/integrations/agent_smoke.sh` — `--sequences` driver mode, incl. the #2421 exact-family+method pre/post delta contract (Vector-reviewed) and MTP-serve lifecycle.

## #2421 (now implemented, reviewer B1)
Each MTP sequence evaluates the engagement/observability contract as deltas of the exact (family, method) label pair between a pre- and post-canonical-chat /metrics scrape: delta_attempts >= 1; 1 <= delta_accepts <= delta_attempts; delta_tokens_saved >= delta_accepts; per-window ratio > 0.0; and the exported accept_ratio gauge finite + consistent with the exact cumulative accepts/attempts counters. NaN fails. No uncalibrated floor.

## Non-goals
- No `mtp: second` ordering (reviewer B3 — a dedicated MTP serve already boots spec-decode, so the primary can never arrive at a later residency step; /v1/models/load cannot carry spec-decode). `mtp` is `"none" | "first"`.
- MTP-serve teardown only ever stops its own PID (TERM -> wait -> KILL); it never port-sweeps a process we did not start (reviewer B2).
- No model loads in CI (pure-CPU schema validation only; `--sequences` smoke runs on Apple hardware/Studio, not the hosted no-MLX lane).

## Acceptance
- `tests/test_top10_sequences_schema.py` passes offline (11 tests, no MLX, no network).
- Every YAML entry references real `ModelLoadRequest` fields; the sequence file is the single source for the top-10 smoke order.

## Verification
- Local: `test_top10_sequences_schema.py` 11 pass; `test_sidecar_vision_smoke.py` 9 pass (matches widened auto-release timeout).

## Behaviour delta
Deterministic, versioned top-10 load-order spec + a smoke driver that consumes it (incl. the exact-family MTP accept contract); no order drift between docs and execution.

## Design precedent

- **MTP acceptance metrics.** The gate treats exported counters as cumulative and computes an exact-label pre/post request window. It verifies monotonic deltas, finite ratios, and counter/gauge consistency instead of applying an unrelated global threshold.
- **Owned process lifecycle.** The smoke tracks the one server PID it starts, sends TERM, waits for a bounded interval, and uses KILL only for that same PID. An unexpected port listener is reported and never swept.
- **Declarative sequence execution.** A versioned data spec is the single source of load order; one bounded driver validates and executes it so release jobs cannot drift into hand-written variants.

Bounded, no new tech debt: this is a gate harness (spec + driver + schema test); it introduces no new engine mechanism and no durable state. It makes the gate *capable* of expressing the #2421 contract and reproducing #2438 without faking an MTP primary.

Closes #2496. Part of epic #2499 (0.13.2 release-path hardening).

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/agent-gate.yml
+++ b/.github/workflows/agent-gate.yml
@@ -71,11 +71,20 @@ jobs:
     # 90 rather than auto-release's 125 because this manual job does NOT run
     # the release-gate extras (coherence golden + perf) — those are gated on
     # RAPID_MLX_RELEASE_GATE=1, which only the release path sets. 94 min of
-    # pathological-retry tail still exceeds 90, but that path requires every
-    # agent to burn every retry; a genuine hang fails on its own per-agent
-    # budget long before, and this job is operator-triggered rather than
-    # release-blocking, so the tighter cap is the deliberate trade.
-    timeout-minutes: 90
+    # agent pathological-retry tail still exceeds the old 90, but that path
+    # requires every agent to burn every retry; a genuine hang fails on its own
+    # per-agent budget long before, and this job is operator-triggered rather
+    # than release-blocking.
+    #
+    # 140, not 90/94: this job now ALSO drives the --sequences gate
+    # (top10_sequences.yaml) after the agents. Its budgets, documented in
+    # agent_smoke.sh: std block <= 3600s (60 min), MTP 3 serves x (600s boot +
+    # 1500s seq) = 6300s (105 min) — a ~165 min pathological tail, but each
+    # step/sequence fails on its own budget in seconds, so only a slow-but-
+    # passing run reaches the cap. Healthy run: agents ~8-10 min + std ~25 +
+    # MTP ~30 = ~65 min; 140 keeps headroom for cold model loads on the shared
+    # Studio without guillotining a legitimately warm-up-heavy run.
+    timeout-minutes: 140
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
       - name: Tier-1 agent re-verification
@@ -93,4 +102,10 @@ jobs:
           # Reviewed perf floor (tok/s). Blank → perf runs advisory (prints the
           # number to review); set it here to enforce. Never a fabricated baseline.
           RAPID_MLX_PERF_MIN_TPS: ${{ inputs.perf_min_tps }}
+          # #2496: also drive the top-10 residency-load sequences AFTER the
+          # agents — non-MTP loads + post-load completions on the warm serve,
+          # plus a dedicated MTP serve (own port) per MTP sequence whose
+          # /metrics is scraped for the #2421 accept-rate floor. Reproduces the
+          # #2438 "second load after an MTP primary" path.
+          RAPID_MLX_SEQUENCES_YAML: tests/integrations/top10_sequences.yaml
         run: bash tests/integrations/agent_smoke.sh "$MODEL"
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -282,7 +282,22 @@ jobs:
     # This outer cap is not a substitute for the per-request bounds: it only
     # prevents healthy checks near their individual budgets from being
     # guillotined by the job.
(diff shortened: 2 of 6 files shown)


raullenchai (commented):
Vector pre-Codex review of head 67860c00. This is the fixed BLOCK list; no Codex review round is consumed.

1. The #2421 metric contract is not met. The current evaluator filters only method=mtp, takes a maximum across families, reads one cumulative scrape after the whole sequence, checks only attempts plus a 0.4 gauge, and lets NaN evade the comparisons. Replace it with exact family=qwen3.8-27b-4bit and method=mtp pre/post samples around one canonical temperature=0 chat on the freshly booted MTP primary. Require HTTP 200 and non-empty content, then enforce delta_attempts >= 1; delta_accepts >= 1 and <= delta_attempts; delta_tokens_saved >= delta_accepts; and delta_accepts / delta_attempts > 0.0. The exported accept-ratio gauge must be finite and arithmetically consistent with the exact cumulative accepts and attempts counters. Specify family in YAML. Remove the uncalibrated 0.4 floor: this release smoke proves engagement and observability, not calibrated performance. Compute the evaluator once.

2. Teardown may kill a process this lane did not start. MTP_SERVE_PORT is assigned before the occupied-port check, and cleanup later SIGKILLs every listener on the port. Stop only the tracked MTP_SERVE_PID with TERM, bounded wait, then KILL fallback. Never sweep a port. Clear owned PID/port state on boot failure; report an unexpected listener rather than killing it.

3. The mtp:second sequence is not exercising what it claims. The dedicated process already boots qwen3.8-27b-4bit with the MTP configuration before step 1. Loading qwen3.5 as a secondary and then calling models/load for the MTP alias does not make the MTP primary arrive second, and models/load cannot carry the speculative-decode configuration. Remove this sequence and claim unless a real engine path can boot a non-MTP primary and enable MTP on replacement.

Do not run the prepared Studio model proofs without fresh Atlas authorization. Ping Vector after a correction push; the single permitted Codex round will run only on a head that satisfies this list.

Issue #2461, opened 2026-08-26
fix(mac): serving-lane reason and Quickstart copy mismatches (text_lane_forced, LFM 2.6B vs 1.2B, 'about a minute')


raullenchai:
Found by the Atlas static adversarial review of v0.13.0..aaf2e010 (report: atlas-desk/.orca/reviews/review-desktop-0131.md). Not a 0.13.1 release hold; needs a runtime repro before fixing.

1. `Server/ServerModelProfile.swift:190` maps `operator_forced_text`, but the engine emits `text_lane_forced` (`vllm_mlx/api/utils.py:1384`) → users who forced text mode get the generic copy instead of the 'current settings' copy.
2. `UI/QuickstartView.swift:63` doc says 'Below 16 GB, onboarding starts from LFM2.5 2.6B' but `baselineChoice` (:305-307, #2432) picks 1.2B; `seedMessage` (:397-403) still says the baseline starter was 'picked so you can start chatting in about a minute' for the 3.0 GB Qwen3.5-4B download.

Related: #2454 (vision_memory_insufficient copy).

raullenchai:
Item 1 shipped on main in #2514 (squash commit 83aa8b8f): Desktop now matches the real `text_lane_forced` engine reason and gives an in-app model-picker remedy. Keeping this issue open for the remaining Quickstart copy items.

PR #2514, opened 2026-08-27
fix(mac): map the serving-lane reasons the engine actually emits

### LINKED ISSUE #2454: fix(mac): explain when photo input is unavailable because vision does not fit
## Problem

On exact Desktop source `a6b820cffdbdae57d9bb4d9bceb4d26f2f1ecafb`, the server can report `serving_lane_reason: vision_memory_insufficient`, but `ImageInputAvailability.message(for:)` has no case for that stable reason. It falls through to:

> This model is running text-only. Photos need a vision-capable model.

The model is already vision-capable. Its vision runtime was not admitted because it did not fit the current memory budget, so the fallback copy sends the user toward the wrong remedy and hides the actionable reason.

## Expected

Map the typed memory reason to copy that says vision could not fit and offers the truthful recovery: free memory or choose a smaller vision-capable model. Keep the photo affordance blocked while the live lane is text-only.

## Scope

Desktop reason-to-copy mapping and a focused contract test only. No server schema, admission, model-name heuristic, or memory-policy change.

Jevin-F:
## Why

Closes #2454. Part of #2461 (item 1 only).

`ImageInputAvailability.message(for:)` turns the engine's `serving_lane_reason` into the sentence shown when photo input is disabled. That table had drifted from the engine:

- It matched `operator_forced_text`. The engine has never emitted that string — it sends `text_lane_forced` — so the arm was dead code and a registry-pinned text-only alias fell through to the generic sentence.
- Four reasons had no case at all. Each marks a **vision-capable checkpoint serving text-only**, so the generic copy ("Photos need a vision-capable model") named a remedy the user had already applied.

A user running a vision-capable model that was downgraded — most often because its vision runtime did not fit in memory — was told to go find a vision-capable model. That is the model they are already running.

Two test fixtures asserted on strings the engine never sends (`operator_forced_text`, `vision_hybrid_serialized`, both introduced together in eb9b3352). They agreed with the app instead of with the engine, which is why this stayed green.

## Behaviour delta

| Reason | Before | After |
| --- | --- | --- |
| `text_lane_forced` | unreachable, then generic | names the unavailable vision path, points at the model picker |
| `vision_memory_insufficient` | "choose a vision-capable model" | choose a *smaller* vision-capable model |
| `vision_hybrid_runtime_unsupported` | generic | names the unsupported vision runtime |
| `text_lane_speculative_decode` | generic | names speculative decoding and where to turn it off |
| `vision_weights_unavailable` | already covered | unchanged, now guarded by the contract test |

Every remedy has to be one the user can carry out *from this app*. Speculative decoding is reversible in Settings, so that sentence names the switch. The other two are not: `vision_memory_insufficient` gates on physical RAM (`physical_ram_gb()` reads `sysctl hw.memsize`, so freeing memory cannot lift it), and in-app `text_lane_forced` arrives only from a registry `is_text_only` pin the user cannot change. For both, choosing a different model is the only real remedy.

## Scope

- `ServerModelProfile.swift` — reason-to-copy mapping only
- `ServerModelProfileTests.swift` / `ModelResidencyTests.swift` — cover the new arms, replace the ghost-string fixtures
- `tests/test_serving_lane_reason_contract.py` — new cross-language guard
- `.github/workflows/ci.yml` — one roster line so the guard runs on the Linux leg

## Non-goals

- #2461 item 2 (Quickstart LFM 2.6B/1.2B and "about a minute" copy) — different file, different subject
- Engine-side enum SSOT — tracked in #2515
- `server.py`'s `fallback_detail` log dict omits `text_lane_speculative_decode`, so that downgrade logs a slightly wrong cause. Log-only, pre-existing, not user-facing.
- No server schema, admission, memory-policy, or model-name heuristic changes

## Acceptance

Nothing connects the two sides of this contract: the engine emits bare string literals across its tree, the app matches them in a Swift `switch`, and there is no enum, schema, or generated code between them. That is exactly how it drifted. The new test follows the existing `tests/test_ram_tier_recommendations_agree.py` pattern (this repo's precedent for parsing Swift from pytest, and mlx-free so it runs on the Linux leg). It asserts:

1. no Swift case matches a string the engine cannot emit — the shape of the `operator_forced_text` bug;
2. every reason that silently downgrades a vision-capable model has dedicated copy — the shape of the `vision_memory_insufficient` bug.

Reasons where photos are genuinely available, and `not_applicable` (image-gen / video-gen, which have no vision lane to lose), are deliberately exempt from (2).

## Verification

- `swift test --filter "ServerModelProfile|ModelResidency"` — 65 tests, pass
- `pytest tests/test_serving_lane_reason_contract.py` — 3 tests, pass
- `ruff check` / `ruff format --check` — clean
- Full `pytest tests/` run; remaining failures on this machine are pre-existing optional-dependency import errors, identical on a clean checkout of `main`

Every new guard was verified to fail when the production line is broken, then restored:

| Break introduced | Result |
| --- | --- |
| Swift case reverted to `operator_forced_text` | contract test names the ghost string |
| `vision_memory_insufficient` case deleted | contract test names the uncovered downgrade |
| engine reason renamed in `api/utils.py` | contract test fails, points at both sides |
| `vision_weights_unavailable` case deleted | contract test names it |
| `text_lane_forced` copy pointed at a non-existent text-only switch | Swift test fails, demands the model-picker remedy |
| `vision_memory_insufficient` copy reverted to "free up memory" | Swift test fails on the missing "smaller" |

## Checklist

- [x] Tests pass locally
- [x] New guards spot-checked to fail when the corresponding production line is broken (SOP §Step 3)


raullenchai (approved):
Reviewed exact head 85081a78 after the requested copy and contract-heading updates. The workflow delta is limited to enrolling the new MLX-free contract test; the fork-only supply-chain failure is therefore an expected tripwire, not an unreviewed workflow change. Local exact-head checks: contract test 3/3, ServerModelProfileTests 38/38, ModelResidencyTests 27/27, and git diff --check clean. Approval does not waive required tests + desktop-tests or the zero-behind merge gate.

PR #2532, opened 2026-08-27
docs: add optional contributor X-handle field to the PR template

raullenchai:
## Why
When a community PR merges we want to be able to credit the author on our
public X account. Right now there is no place for an external contributor
to volunteer an X handle, so the follow-up thread has to be tracked down
ad-hoc. This is a contributor-credit template change: it asks for an
**optional** X handle on the PR template so crediting is easy when it
merges — with an explicit opt-out for contributors who'd rather not be
credited.

## What
- `.github/PULL_REQUEST_TEMPLATE.md`: add a short `## Author` section at
  the end — `X handle (optional, external contributors): @` — with an
  HTML comment telling contributors "Leave blank if you prefer not to be
  credited publicly when this merges."
- `CONTRIBUTING.md`: one sentence under the pull-request workflow section
  saying the field is optional and only used to thank contributors
  publicly on merge.

## Scope / Non-goals
- Nothing else changes. No other projects are mentioned. No README /
  workflow / code / tests touched.
- The field is strictly optional and never required for merge; leaving it
  blank must not affect any gate.

## Verification
- Docs-only diff (`.github/PULL_REQUEST_TEMPLATE.md` + `CONTRIBUTING.md`).
- Confirmed the `cl_description_quality` description gate has no
  unedited-template detection — it only requires a title >= 3 words, a
  non-empty body, and one rationale signal (e.g. `## Why`), all of which
  remain present when the optional `## Author` field is left blank. The
  added section contains no `- [ ]` checkbox, so `test_plan_check` is not
  triggered either. An empty optional field does not read as an unedited
  template (#2510 context).
- Applied the `skip-version-bump` label (docs-only; no version change).

Issue #2480, opened 2026-08-27
test(mac): re-enable ResidentLoadFeedbackTests RAM-independently (disabled for 0.13.1 CI hang)


raullenchai:
Disabled for 0.13.1 to unblock the train-6d merge (human decision: skip a CI-only test hang, backfill the fix).

Root cause: the pre-existing suite ResidentLoadFeedbackTests (@Suite 'Resident-load rejection feedback') calls await server.ensureServing(...). #2478 changed ServerManager.ensureServing (+131 lines: c19ecf05/a463ed2c/f2e5fdb1) to await a >100% memory-confirmation decision. The tests install no confirmation responder, so on the low-RAM hosted CI runner the same load projects >100% -> confirmation required -> ensureServing awaits forever -> the swift test job hits its 20-min timeout (observed twice: rapid-mac CI run 33033633665 build job 98396428499). On the 256 GB Studio the load fits, no confirmation, full serial suite passes 2,948/2,948 in ~31s -> shipped app behavior is correct; only the test is environment-dependent.

Fix (0.13.2): make the suite RAM-independent — inject an auto-answering memory-confirmation responder (or stub resident-memory-limit/physical-RAM so no confirmation is requested) + a hard per-test timeout, then remove .disabled(...).

Coverage while disabled: MemoryLoadConfirmationQueueTests (12/12), ModelResidencyTests, GUI golden-flow models/low-memory-choice (green); local full serial Swift 2,948/2,948.

Issue #2345, opened 2026-08-25
XCUITest runner fails locally: CODE_SIGNING_ALLOWED=NO produces broken code signature (Gatekeeper kills Runner)


raullenchai:
## Summary (NON-BLOCKER — harness signing defect, not a product defect)

The native XCUITest journey (`apps/rapid-mac/Tests/RapidUITests`, e.g. `ImageGenerationPixelTests`) **fails at the runner bootstrap** on a local Studio because the test Runner is built with a **malformed code signature**.

**Exact defect line:** `apps/rapid-mac/scripts/run-xcui-tests.sh:36`
```bash
CODE_SIGNING_ALLOWED=NO
```
With `CODE_SIGNING_ALLOWED=NO` the built `RapidUITests-Runner.app` gets an incomplete seal that Gatekeeper treats as "damaged", so the Runner is killed (signal kill) at launch — before XCTest ever establishes the automation connection. **0 test cases run.**

**Evidence — codesign verification:**
```
codesign --verify --deep --strict "…/Build/Products/Debug/RapidUITests-Runner.app"
→ code has no resources but signature indicates they must be present
```
The local machine has **no Developer ID signing identity** (`security find-identity -p codesigning` → `0 valid identities found`), so a validly-signed Runner cannot be produced in a local/dev workflow.

**XCUITest run error (from the result bundle):**
```
RapidUITests-Runner encountered an error
(Early unexpected exit, operation never finished bootstrapping - no restart will be attempted.
 Underlying Error: Test crashed with signal kill before establishing connection.)
```
failedTests=1, passedTests=0, result=Failed.

**xcresult paths (evidence):**
- `/tmp/rapid-gui-xcui/rapid-xcui-full.xcresult` (the failing XCUITest result bundle)
- Per-journey AX golden-flow evidence: `/tmp/rapid-gui-golden-<group>-<flow>/result.json` (all 29 AX journeys PASS on the same candidate — GUI-HARNESS accepted)

**Impact:** this blocks only the native pixel-level XCUITest journey in local/dev workflows. The full AX golden-flow suite (the 29 reader-visible journeys across onboarding-settings/models/chat/images/audio/app-lifecycle) runs fine via `rapid-ax` (Accessibility grant alone). Corroborated independently by Pixel's observation of a stale `RapidUITests-Runner` Gatekeeper alert on the same Studio.

## Proposed fix (owner Harbor)

[...]

PR #2539, opened 2026-08-28
fix(pr_validate): reject unedited PR-template bodies in cl_description_quality (#2510)

### LINKED ISSUE #2510: pr_validate: description gate must detect an unedited PR template (strip HTML comments, require non-empty `## Why`)
Follow-up from the #2507 review. `scripts/pr_validate/steps/cl_description_quality.py` (lines ~82-91) passes on heading presence only: an unedited `.github/PULL_REQUEST_TEMPLATE.md` with the checklist boxes ticked is a false green (verified empirically: raw template + ticked boxes → cl_description_quality PASS, test_plan_check PASS). Fix: strip `<!-- -->` comments before scoring, require non-empty prose under `## Why` (and `## Scope`), and add a test that feeds the raw template file. Also: `--body-only` with `PR_VALIDATE_SKIP_DESC=1` should warn, and tests should exercise `FetchStep.run` end-to-end for merge-base resolution (not only `_derive_merge_base`). Owner: ds0731 after the current queue. Part of epic #2499.

raullenchai:
## Why

`scripts/pr_validate/steps/cl_description_quality.py` currently PASSES on heading / rationale-SIGNAL presence only. An unedited `.github/PULL_REQUEST_TEMPLATE.md` (with the `[x]` checklist boxes ticked) is a **FALSE green**: the template's `## Why` / `## Scope` headings match `_RATIONALE_SIGNALS` even though every word of real prose sits inside `<!-- … -->` HTML comments. Empirically verified in the issue: raw template + ticked boxes → `cl_description_quality` PASS, `test_plan_check` PASS.

This weakens the whole merge-readiness gate: a PR body that never says WHY (only the template skeleton, left untouched) is treated as well-described. The description-quality step exists precisely to catch that — with the headings matching the rationale regex, it could not.

## What

- **Strip HTML comments before scoring** (`_strip_comments`): comment-only prose can no longer satisfy the "body exists" or "has rationale" rules. A body whose only content is `<!-- … -->` is effectively empty and FAILS.
- **Require substantive prose under the `## Why` / `## Scope` contract headings** when present (`_sections` + `_is_substantive_prose`). After comment-strip an unedited template has `## Why` immediately followed by `## Scope` with nothing between → empty → FAIL. Only these two headings are gated.
- **Preserve the lenient non-template paths**: `Closes #N` / `Fixes #N` issue links, inline `Why:` lines, and `because`-clauses still satisfy the rationale check for PRs that don't use the template headings.
- **`--body-only` + `PR_VALIDATE_SKIP_DESC=1` now WARNs** instead of silently reporting MERGE-SAFE: in body-only mode the description gate is the ONLY check, so a skip means the run validated nothing.
- New test file `tests/test_pr_validate_cl_description_quality.py` feeds the **RAW template file on disk** and expects FAIL, plus the prose/comment/Author-path cases. Registered in the ci.yml changed-lines coverage list.

## Scope / Non-goals

- Scope: `cl_description_quality` step logic + the body-only runner warning + tests + ci.yml test registration.
- Non-goals: no change to `fetch.py` / merge-base resolution (untouched); no change to `test_plan_check` (already correctly passes the template); no change to the six-field contract headings themselves.
- The `test_plan_check` step intentionally still passes an unedited template — this PR is specifically about the missing-rationale FALSE green, not checkbox discipline.

## Verification

- `pytest tests/test_pr_validate_*.py` → **329 passed** (includes all 9 new step-level tests, the new runner body-only warning test, and the existing merge-base / runner / test_plan_check suites).
- `ruff check` + `ruff format --check` on changed Python files → clean.
- `mypy` on changed `scripts/pr_validate` files → no new errors (the one `runner.py:140` typing nit is pre-existing at HEAD; `scripts/pr_validate` is outside the mypy shrink-only budget gate).
- Manual trace of the RAW template: stripped body shows `## Why ⏎ ## Scope` with no prose → FAIL "`## Why` section is empty (unfilled template)".

## Behaviour delta

- **Default behavior of the description gate changes**: an unedited (comment-only) PR template body now FAILS `cl_description_quality` instead of passing. A real PR with prose under `## Why` / `## Scope` is unaffected and still passes.
- `--body-only` with `PR_VALIDATE_SKIP_DESC=1` now emits a loud WARNING to stderr instead of a silent clean verdict.

## Design precedent

**What I looked at:** the existing step's rationale signals (`_RATIONALE_SIGNALS`) and its Google eng-practices rationale; the adjacent `test_plan_check` step/tests for the step-under-test idiom; `test_pr_validate_runner.py`'s `TestBodyOnly` for the body-only pipeline idiom; `test_pr_validate_merge_base.py` for the pure-CPU, no-network test style.

**What I adopted:** no new abstraction layer — three small module-level helpers (`_strip_comments`, `_sections`, `_is_substantive_prose`) matched to the existing step's style; the same ``_ctx(...)`` context-shell test helper idiom as `test_pr_validate_test_plan_check.py`; the `TestBodyOnly` stub-fetch + real-step runner pattern for the warning test.

**What I rejected and why:** (a) gating *every* heading — the PR #2532 `## Author` field shows a template can gain optional/empty sections; gating only `## Why`/`## Scope` keeps the gate precise (issue requirement). (b) Requiring the rationale to live specifically in the `## Why` section — that would break the long-standing leniency where an issue link / `Why:` line / `because`-clause suffices for non-template PRs. The prose check is layered in *before* (not instead of) that lenient fallback. (c) Treating a skipped desc-gate in body-only as a hard FAIL — the user explicitly asked for the override via env var, so the right contract is a loud WARNING, not an unrequested block.

Part of #2499. Fixes #2510.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -362,6 +362,7 @@ jobs:
             tests/test_pr_validate_merge_base.py \
             tests/test_pr_validate_diff_coverage.py \
             tests/test_pr_validate_codex.py \
+            tests/test_pr_validate_cl_description_quality.py \
             tests/test_release_baselines.py \
             tests/test_sidecar_distribution_constraints.py \
             tests/test_sidecar_vision_smoke.py \
--- a/scripts/pr_validate/runner.py
+++ b/scripts/pr_validate/runner.py
@@ -209,6 +209,25 @@ def run_pipeline(
             )
             break
 
+    # ``--body-only`` exists to reproduce the description-quality verdict
+    # locally. If that single gate got SKIPPED (via
+    # ``PR_VALIDATE_SKIP_DESC=1``) the run validated nothing — a silent
+    # MERGE-SAFE here would mislead the author into thinking their body is
+    # fine. Surface it as a loud WARNING (non-blocking, the user asked for
+    # the override) rather than a clean pass.
+    if body_only:
+        desc_result = next(
+            (r for r in ctx.results if r.name == "cl_description_quality"), None
+        )
+        if desc_result is not None and desc_result.status == "skip":
+            print(
+                "  WARNING: --body-only ran but the description gate was "
+                "SKIPPED via PR_VALIDATE_SKIP_DESC=1 — this run validated "
+                "NOTHING about the body. Unset that env var and re-run to "
+                "get a real description verdict.",
+                file=sys.stderr,
+            )
+
     # Render the scorecard to stdout (so callers can pipe into PR comments).
     print(render_scorecard(ctx))
 
--- a/scripts/pr_validate/steps/cl_description_quality.py
+++ b/scripts/pr_validate/steps/cl_description_quality.py
@@ -90,6 +90,78 @@ _RATIONALE_SIGNALS = (
     re.compile(r"\bbecause\b", re.IGNORECASE),
 )
 
+# Strip HTML comments (``<!-- ... -->``) BEFORE any scoring so template
+# boilerplate / rationale hidden inside ``<!-- -->`` can't satisfy the
+# "body has rationale" rule. Issue #2510: an UNEDITED
+# ``.github/PULL_REQUEST_TEMPLATE.md`` (with ``[x]`` boxes ticked) was a
+# FALSE green because the template's ``## Why`` / ``## Scope`` headings
+# matched the rationale signals even though all the real prose sat inside
+# HTML comments.
+_HTML_COMMENT = re.compile(r"<!--.*?-->", re.DOTALL)
+
+# Markdown headings ``#``..``######``, for slicing the body into sections.
+_HEADING = re.compile(r"^(#{1,6})\s+(\S.*)$", re.MULTILINE)
+
+# The two contract headings whose sections must carry real prose. A PR
+# that uses the template but leaves ``## Why`` / ``## Scope`` empty (their
+# only content being an HTML comment) must FAIL — the whole point of issue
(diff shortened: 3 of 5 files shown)


raullenchai (commented):
CI diagnosis — 1 diagnosed rerun (job: `gui-golden-flows (chat)`, step `Require native XCUITest`).

Root cause: `RapidUITests.ChatAttachmentJourneyTests.testDragPasteAndRemovalPreserveWireIdentity` failed in `RapidUITestHarness.swift:143` (`XCTAssertTrue failed`, 71.5s) — the native XCUITest for drag/paste-and-removal wire-identity preservation. This is the known-desktop-flaky XCUITest drag-timing area (see desktop test hardening #2480/#2481 + #2500 which added drag settle/retry; NON-BLOCKER #2345 records the XCUITest signing/session limits).

This PR #2539 changes ONLY `scripts/pr_validate/steps/cl_description_quality.py`, `scripts/pr_validate/runner.py`, and `tests/test_pr_validate_*.py` (pure-Python pr_validate tooling) — disjoint from the macOS chat-attachment UI path. No causal path. All other golden flows (fresh-install, chat-restore, message-actions, restored-tools, tool-loop-budget, model-crash-recovery, audio-readiness, dictation, etc.), the engine lanes, test-matrix (3.10/3.11/3.12), test-apple-silicon, changed-lines-coverage all PASS on this exact head. Diagnosing the chat XCUITest as a drag-timing flake; single rerun of that job.

PR #2558, opened 2026-08-28
fix(engine): resolve pulled --bits variant at serve/load (#2340)

### LINKED ISSUE #2340: serve-side selection after pull --bits: variant not persisted into model resolution
## Summary

After `rapid-mlx pull --bits 4 <multi-variant-repo>` fetches only `4bit/`, a later `serve <repo>` cannot load that variant: the serve/load path resolves via `resolve_subfolder(repo_id)`, which returns `None` for a raw repo id (no catalog alias), so it points at the repo ROOT — which has no weights (only `LICENSE`/`README`). The pulled `4bit/` checkpoint is unreachable for serving.

This was flagged by pr_validate codex on PR #2338 (#2145) and **documented-deferred** per Atlas READY-2338 as out of the pull-only #2145 contract.

## Evidence

For `LiquidAI/LFM2.5-2.6B-MLX`, `list_repo_tree(recursive=False)`:
- Folders: `4bit 5bit 6bit 8bit bf16 mxfp4 mxfp8 nvfp4`
- Root files: only `README.md`, `LICENSE`, `.gitattributes`

`4bit/` is self-contained (own `config.json`, `tokenizer*`, `generation_config.json`, weights), but the repo ROOT has no weights, so `serve <repo>` (root resolution) finds nothing runnable.

## Scope / non-goals

- This PR #2338 intentionally ships **pull-side** variant selection only.
- Making `serve`/`load` variant-aware is a separate design (persist the pulled variant, or add a serving selector that resolves the same subfolder). GGUF runtime / Desktop UI serving selectors are already non-goals of #2145.

## Suggested direction

One of:
- Persist the variant choice (e.g. a small marker in the HF snapshot) that `resolve_model`/downstream loads consult, OR
- Add an explicit serving selector (`serve --bits <N>` / `--format <name>`) that resolves the same subfolder for a multi-variant repo.

Refs: #2145 (this feature), #2338 (the pull PR).

raullenchai:
## Why

`rapid-mlx pull --bits 4 <multi-variant-repo>` downloads only the selected subfolder, but a later `serve <repo>` previously resolved the repository root or catalog default because the pull choice was discarded. Multi-variant repository roots can contain no loadable weights, so the successfully downloaded checkpoint was unreachable or the wrong quantization could be loaded.

The fix must also preserve an explicitly typed alias, behave identically for startup and dynamic residency, remain compatible with standard cache inspection, and never silently retain an older serving choice when metadata cannot be updated.

## What

- `vllm_mlx/_download_gate.py`
  - Stores the latest narrowed-pull choice atomically at `models--<repo>/.rapid-mlx/variant`, outside the cache manager's `refs/` and `snapshots/` namespaces.
  - Validates relative subfolder names, supports literal glob metacharacters safely, reads the marker fail-soft, and clears it after a successful ordinary pull.
  - Returns marker-update success and best-effort invalidates an older marker when a replacement fails.
- `vllm_mlx/cli.py`
  - Persists a successful `--bits/--format` selection on the Hugging Face fallback path.
  - Clears stale selection after ordinary pulls on both mirror-success and fallback exits.
  - Warns when weights downloaded but serving-choice metadata could not be updated, including the stale/default-checkpoint risk and recovery action.
- `vllm_mlx/utils/tokenizer.py`
  - Applies explicit alias > bare-repo pulled marker > reverse-catalog default precedence.
  - Escapes literal variant folder names before building narrow allow-patterns.
  - Keeps distinct actionable errors for absent and incomplete variant folders.
- `vllm_mlx/server.py`
  - Resolves concrete subfolder identity before metadata and serving-lane selection.
  - Preserves the CLI-saved alias after early alias normalization and applies the same rule to dynamic residency when alias and canonical path identify the same checkpoint.
- `.github/workflows/ci.yml`
  - Registers the new regression file for changed-lines coverage.

## Scope / Non-goals

Chosen approach: persist the successful pull choice and consume it in the shared checkpoint-resolution path.

Non-goals: no new `serve --bits/--format` syntax, no download-transport or mirror-protocol change, no alias-registry schema change, no scheduler or modality-policy change, no Desktop selector UI, and no version bump.

## Design precedent

The existing catalog `subfolder` contract remains authoritative for explicit aliases, but it cannot represent which one of several sibling variants a user selected for a bare repository ID. The pull path already validates and narrows that selection; this change preserves that state instead of inferring it later from an ambiguous snapshot tree.

The cache manager treats every file under `refs/` as a commit reference, so application metadata belongs in a repository-local application namespace. The marker therefore uses `.rapid-mlx/variant`; a real cache-scan regression verifies that the repository remains inspectable. Writes use the codebase's existing atomic replace pattern with unique temporary files.

The shared server checkpoint resolver is the mechanism boundary: it chooses the concrete checkpoint before routing metadata is read, and both startup and residency consume that result. A separate serve-side selector was rejected because it would require users to repeat and synchronize a choice already made at pull time.

## Verification

- `pytest tests/test_pull_variant_serve_resolution.py tests/test_server_load_model_order.py tests/test_resident_models.py tests/test_alias_subfolder.py tests/test_pull_variant_selection.py tests/test_download_gate.py tests/test_chat_template_registry.py` — 385 passed.
- Ruff check and format checks on all changed Python files — clean.
- Real cache scan with marker present — no corruption warning.
- Independent Codex review rounds dispositioned on the PR; the latest static review before the marker-failure fix found no other merge-blocking issue.
- Exact-head required CI and final `pr_validate` must be green before merge.

No local network pull, GUI run, or large-model load is used by these tests.

## Behaviour delta

- Before: a narrowed pull could be unreachable at serve time, a stale marker could select the wrong quantization, and explicit alias intent could be lost after CLI normalization or in residency.
- After: bare-repo serve recovers the latest successfully recorded narrowed choice; explicit aliases retain their declared subfolder in startup and residency; ordinary pulls clear stale narrowing.
- Literal format names containing glob metacharacters remain literal and cannot broaden the download pattern.
- If marker replacement fails, an older marker is invalidated when possible and the CLI always warns that bare-repo serve may select an older/default checkpoint. Downloaded weights are not falsely reported as failed.
- Flat repositories, local paths, and commit-pinned snapshot resolution retain their existing behavior.

Refs #2145, #2338. Fixes #2340.

### DIFF
--- a/tests/test_pull_variant_serve_resolution.py
+++ b/tests/test_pull_variant_serve_resolution.py
@@ -511,6 +511,32 @@ def test_successful_ordinary_hf_pull_clears_previous_variant(marker_path):
     assert _download_gate.pulled_variant(RAW_REPO) is None
 
 
+def test_ordinary_hf_pull_survives_marker_clear_failure(capsys):
+    """Cleanup metadata is best-effort after a valid HF pull completes."""
+    import argparse
+
+    from vllm_mlx import cli
+
+    args = argparse.Namespace(
+        model=RAW_REPO,
+        bits=None,
+        format=None,
+        _original_alias=RAW_REPO,
+    )
+
+    with (
+        patch.object(cli, "_try_mirror_prefetch", return_value=False),
+        patch("huggingface_hub.snapshot_download", return_value="/cache/snapshot"),
+        patch(
+            "vllm_mlx._download_gate.clear_pulled_variant",
+            side_effect=OSError("read-only cache"),
+        ),
+    ):
+        cli.pull_command(args)  # must not raise after a successful download
+
+    assert "Cached at: /cache/snapshot" in capsys.readouterr().out
+
+
 def test_successful_ordinary_mirror_pull_clears_previous_variant(marker_path):
     """The mirror-success early return applies the same marker transition."""
     import argparse
@@ -533,3 +559,32 @@ def test_successful_ordinary_mirror_pull_clears_previous_variant(marker_path):
         cli.pull_command(args)
 
     assert _download_gate.pulled_variant(RAW_REPO) is None
+
+
+def test_ordinary_mirror_pull_survives_marker_clear_failure(capsys):
+    """Cleanup metadata is best-effort on the mirror-success early return."""
+    import argparse
+
+    from vllm_mlx import cli
+
+    args = argparse.Namespace(
+        model=RAW_REPO,
+        bits=None,
+        format=None,
+        _original_alias=RAW_REPO,
+    )
+
+    def mirror_success(repo_id, *, out):
+        out["network_fetch"] = False
+        return True
+
+    with (
+        patch.object(cli, "_try_mirror_prefetch", side_effect=mirror_success),
+        patch(
+            "vllm_mlx._download_gate.clear_pulled_variant",
+            side_effect=OSError("read-only cache"),
+        ),
+    ):
+        cli.pull_command(args)  # must not raise after a successful download
+
+    assert "Already cached" in capsys.readouterr().out
--- a/tests/test_pull_variant_serve_resolution.py
+++ b/tests/test_pull_variant_serve_resolution.py
@@ -483,7 +483,6 @@ def test_pull_persist_failure_still_completes_the_pull(capsys):
     ):
         cli.pull_command(args)  # must not raise
     out = capsys.readouterr().out
-    assert "R2 mirror skipped" in out
     assert "could not record that serving choice" in out
     assert "may select an older or default checkpoint" in out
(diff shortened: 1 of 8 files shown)


raullenchai (commented):
@Vector — PR #2558 (serve-side pulled-variant persistence, #2340) is now fully CI-CLEAN on exact head `90139321` (tests+desktop-tests+validate+version-bump-guard all green; pr_validate codex clean). Awaiting your review round to converge before I label train:boarding. One round, judge `90139321`. (Option A persist-marker decision is in the body; the mypy-budget drift caveat is noted for you to confirm on GH CI.)

PR #2568, opened 2026-08-28
feat(rapid-web): add web client for Rapid-MLX

osdodo:
## Why

  Rapid-MLX does not currently provide a phone-friendly way to chat with models or manage them while they continue running on a Mac.

  This PR adds a standalone web client that users can expose through a tunnel they control. It does not require the Desktop app and does not change the existing Desktop or `rapid-mlx serve` workflows.

  ## Scope

  - Add the standalone `rmlx-web` FastAPI service and CLI.
  - Supervise or attach to `rapid-mlx serve` while keeping the browser-facing port stable.
  - Add a responsive React interface for:
    - Streaming chat, markdown, reasoning output, conversation search, persistence, and message branching.
    - Model selection, startup, downloads, deletion, and live residency management.
    - Speech transcription, dictation, text-to-speech, image generation, and image editing.
  - Add the built-in `weather`, `web_search`, and `browse` tools with server-side execution gates and per-origin browse approval.
  - Add MCP connector configuration, per-tool grants, and per-conversation instructions.
  - Enforce same-origin requests, catalog validation, download limits, and SSRF protections.
  - Package content-hashed frontend assets and add a Homebrew formula.
  - Document installation, operation, security constraints, tools, and MCP connector support.
  - Add Python, Vitest, and Playwright coverage for the new surface.

  ## Non-goals

  - Shipping or operating a public tunnel.
  - Replacing or modifying the Rapid-MLX Desktop app.
  - Synchronizing conversations between browsers or with the Desktop app.
  - Supporting multiple users, server-side conversation storage, or user sessions.
  - Changing existing `rapid-mlx serve` API defaults.

  ## Acceptance

  - `rmlx-web` serves a phone-friendly UI on a stable local port and can be reached through a user-managed tunnel.
  - Users can start, switch, download, inspect, and remove supported models from the web interface.
  - Chat responses stream correctly and searchable, branchable conversation history persists in browser storage.
  - Audio and image lanes remain usable alongside a resident chat model.
  - Disabled built-in or MCP tools are neither advertised to the model nor executable by the server.
  - Browse requests require approval for each new origin.
  - MCP configuration and tool grants persist and are reflected in subsequent model requests.
  - Rebuilding the frontend reproduces the committed static assets and remains within the configured size budget.

  ## Verification

  - `.venv/bin/python -m pytest -q`
    - 444 passed; one upstream Starlette/httpx deprecation warning.
  - `pnpm run test`
    - 31 test files and 560 tests passed.
  - `pnpm run lint && pnpm run typecheck`
    - ESLint and TypeScript checks passed.
  - `pnpm run build`
    - Production build passed.
    - Size gate passed at 945.7 KB across three files.
    - Rebuilt static assets matched the committed output.
  - `pnpm run e2e`
    - 151 Playwright tests passed against the built application.
  - `uvx ruff check . && uvx ruff format --check .`
    - Passed; 23 Python files are formatted.
  - `git diff --check origin/main...HEAD`
    - Passed.
  - PR self-validation was not run because a PR number and authenticated GitHub CLI session were unavailable.

  ## Behaviour delta

  - `rmlx-web`: unavailable → available as a standalone phone-facing client.
  - By default, `rmlx-web` binds to `127.0.0.1:7788`, starts without selecting a model.
  - Existing `rapid-mlx serve` and Desktop app defaults remain unchanged.

  ## AI assistance disclosure

  AI assisted with implementation, tests, documentation, packaging, and generated frontend assets under `apps/rapid-web/**`.

  Claude contributed implementation and test changes in the co-authored commits. Codex reviewed the branch, applied Python formatting and documentation corrections, and ran the Python, Vitest, ESLint, TypeScript, build, size-gate, Playwright, Ruff, and whitespace checks listed above. The generated frontend assets were verified by rebuilding them and confirming that the output matched the committed files.

  ## Checklist

  - [ ] Tests pass locally (`python3 -m pytest tests/ -x`)
  - [ ] Self-validated with `python3 -m scripts.pr_validate.pr_validate <PR#>` — see [CONTRIBUTING.md](../CONTRIBUTING.md#self-validating-your-pr) (opt out heavy steps with `PR_VALIDATE_NO_CODEX=1 PR_VALIDATE_NO_STRESS=1` if you don't have the hardware / a codex login)
  - [ ] If new tests touch a critical code path (parser / scheduler / security), I've spot-checked that they fail when the corresponding production line is broken (see SOP §Step 3)
  - [x] Updated README/docs if applicable


raullenchai (changes requested):
## Echo R1 disposition — changes requested

The product intent is understandable, but this exact head is not safe to run or ready for hosted CI approval yet. Four in-scope blockers need resolution:

1. **Default remote surface permits unauthenticated host mutation and a path to local code execution** (`rmlx_web/auth.py:127`, `rmlx_web/app.py:1099`, `rmlx_web/app.py:1074`). The default is `token=None`, and requests with no `Origin` are intentionally accepted. A non-browser caller that can reach the advertised tunnel can therefore POST JSON without any credential, write an stdio MCP entry, and enable it. The engine's allowlist includes package launchers such as `npx`, `uvx`, and `python`, while the connector accepts arbitrary package/script arguments. On this head, a no-token/no-Origin TestClient request returned 200 for both writes and persisted an enabled `npx -y attacker-controlled-package` launch config. Documentation or a console warning cannot be the security boundary. Require first-party authentication by default for a remotely usable surface (or structurally make all host-mutating/tool/MCP routes unreachable without it), and add a regression proving an unauthenticated non-browser request cannot mutate host state.

2. **The browse SSRF check has a DNS-rebinding TOCTOU** (`rmlx_web/tools.py:673`, `rmlx_web/tools.py:739`). `validate_url()` resolves and validates addresses, then `httpx` opens the URL by hostname and performs an independent resolution. A host can return a public address to the check and a private/loopback address to the connection. Pin the validated destination into the transport while preserving the original Host/SNI identity, revalidate each redirect hop, and add a public-first/private-second DNS regression.

3. **Model switching has a check-to-detached-task race** (`rmlx_web/app.py:425-455`). Two requests can both observe READY and enqueue switches before either task marks the engine STARTING. A stream can also begin after the `streams.active` check but before the task runs, after which the switch may terminate that in-flight generation. Make admission atomic with the pending transition and cover both orderings.

4. **Connector restart can enqueue duplicate, untracked restarts** (`rmlx_web/app.py:1181-1208`). Repeated requests can all observe the same READY model and queue expensive restarts. These tasks are not stored in `app.state.boot`, so lifespan shutdown cannot cancel/join them. Use the same single-flight lifecycle ownership as model load and add double-request plus shutdown coverage.

Evidence on exact head `75e4fdcf`: 133 focused auth/connector/tool tests passed, `git diff --check` passed, and the no-auth connector-write reproduction above succeeded. All seven hosted check suites are currently `action_required`; approval stays withheld until the security boundary is corrected and the branch is rebased from its current 69-commit lag.

Ownership/routing: Echo will track the contribution; architecture/security needs Atlas, server lifecycle and SSRF need Vector, UI/accessibility needs Pixel, and packaging/CI needs Harbor. Please keep fixes within the stated standalone-client contract and reply with the new exact head plus focused regression evidence.

PR #2912, opened 2026-09-01
perf(qwen4): fuse single-token GDN decode

pierre427:
## Why

Qwen4-architecture models currently execute the single-token Gated DeltaNet recurrence as a sequence of small GPU operations. At decode batch size 1, dispatch and intermediate-materialization overhead leave meaningful performance on the table. This PR adds an opt-in fused path because the same recurrence can be computed in one dispatch while preserving the stock path's observable results.

## Scope

- Fuse Qwen4 B=1, M=1 BF16 Gated DeltaNet decode into one Metal dispatch.
- Preserve strict structural admission, a resident `stock|fused` selector, invocation/fallback counters, and fail-closed stock fallback.
- Match the current runtime's BF16 materialization, reduction, sigmoid, normalization, and cache-update boundaries exactly.
- Add a real-Metal 32-step bit-exact regression test, enroll it in the Apple test roster, and add reproducible layer and same-resident-model A/B benchmarks.
- Keep the fast path default-off behind `RAPID_MLX_QWEN4_FUSED_GDN_DECODE=1` while it receives wider hardware qualification.

## Non-goals

- Prefill, batch size greater than 1, masks, ragged cache, training, or non-BF16 execution.
- Speculative/MTP/PLD rollback paths; `record_rollback` deliberately uses the stock implementation.
- Fusing the affine-quantized output projection; measurements showed that variant was slower.
- Enabling the optimization by default.

## Design precedent

The implementation follows the engine's established optimized-kernel pattern: narrowly admit only a proven geometry, retain the stock implementation as the semantic authority, fail closed for every unsupported state, expose resident selection and counters for observability, and require exact output plus cache-state parity rather than tolerance-only acceptance. The fused arithmetic was aligned to the current native runtime's BF16 materialization and reduction boundaries instead of relying on compiler-preserved intermediate precision.

## Acceptance

- An eligible single-token decode uses one fused dispatch and is measurably faster than the stock recurrence.
- Across 32 sequential real-weight steps, output, convolution cache, and FP32 recurrent state are bit-exact with stock.
- An end-to-end 256-token generation produces the same token sequence in stock and fused modes.
- Every unsupported shape or execution mode falls back to stock without changing existing behavior.
- The feature remains off unless the operator explicitly opts in.

## Verification

- Real Qwen3.8-Flash-Next-4bit layer gate on M3 Ultra, 32 sequential steps: exact output and both cache states; 32 fused calls, 0 fallbacks.
- Same layer, interleaved median: stock 0.006521 s vs fused 0.005164 s, a 26.28% improvement.
- Same resident model, 3 interleaved runs per mode at 256 decode tokens: all six token hashes identical; stock 25.43 tok/s vs fused 27.04 tok/s, a 6.35% improvement.
- Focused and adjacent local suites: 186 passed, including bit-exact execution of every threadgroup candidate supported by the test GPU.
- `ruff check`, `ruff format --check`, `actionlint .github/workflows/ci.yml`, Python compilation, and `git diff --check`: passed.
- Environment and exact reproduction commands are recorded in `docs/engineering/performance/2026-09-01-qwen4-fused-gdn-decode.md`.

## Behaviour delta

- Default behavior: unchanged; the stock path remains active.
- With `RAPID_MLX_QWEN4_FUSED_GDN_DECODE=1`: eligible Qwen4 single-token BF16 decode uses the fused recurrence and all other cases retain stock fallback.

## Rollout and risk

The primary risk is arithmetic drift across Metal/runtime variants. The feature is therefore default-off, guarded by exact structural admission, and covered by an Apple-lane bit-exact stateful test. Wider hardware qualification is required before any separate default-on proposal.

## AI assistance disclosure

Codex assisted the maintainer with Rapid-specific debugging, tests, benchmark tooling, documentation, and review of the kernel port. Every generated or modified path was inspected against the stock implementation and verified with the focused suites and real-model bit-exact A/B gates listed above.

## Attribution

Original fused-kernel implementation and performance design by Pierre Lamy. Rapid-specific qualification and port repairs were performed by the maintainer while retaining Pierre's original authored commit. The Metal reduction and precision structure is adapted from mlx-vlm #2105 under its MIT license.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -471,7 +471,6 @@ jobs:
             tests/test_qwen4_exp_reference_parity.py \
             tests/test_routing_groups_0131.py \
             tests/test_qwen4_exp_vendored.py \
-            tests/test_qwen4_fused_gdn_decode.py \
             tests/test_mtp_cli_wiring.py \
             tests/test_continuous_mtp_routing.py \
             tests/test_continuous_self_mtp_mlx_backend.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -471,6 +471,7 @@ jobs:
             tests/test_qwen4_exp_reference_parity.py \
             tests/test_routing_groups_0131.py \
             tests/test_qwen4_exp_vendored.py \
+            tests/test_qwen4_fused_gdn_decode.py \
             tests/test_mtp_cli_wiring.py \
             tests/test_continuous_mtp_routing.py \
             tests/test_continuous_self_mtp_mlx_backend.py \
--- a/docs/engineering/performance/2026-09-01-qwen4-fused-gdn-decode.md
+++ b/docs/engineering/performance/2026-09-01-qwen4-fused-gdn-decode.md
@@ -27,12 +27,30 @@ and is intentionally not included.
 
 ## Rapid port verification
 
-The focused admission, dispatch, resident-selector, cache-update, and fallback
-contracts pass: 7 tests. Python compilation and `git diff --check` also pass.
+Rapid-specific validation used an M3 Ultra with 256 GB unified memory,
+Python 3.12.14, MLX 0.32.2, and
+`rapid-mlx/Qwen3.8-Flash-Next-4bit` at revision
+`dcf657e4acda2aae72da99cde65b6c491cd96998`. No other model was resident.
 
-Rapid-specific Metal and real-model validation is pending because the shared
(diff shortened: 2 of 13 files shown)


raullenchai (approved):
Maintainer approval on exact head `49b1ecabd`: independent Codex review converged at round 7; local `pr_validate` is MERGE-SAFE with 20,548 full-unit passes; real-model token/cache parity and the documented opt-in performance result are accepted. Original implementation and performance design remain credited to Pierre Lamy.

PR #2987, opened 2026-09-03
fix(ci): restore managed macOS GUI queue

### LINKED ISSUE #2986: CI: Manzanita GUI shards fail because python shim is absent
## Problem

The first `main` rapid-mac run after #2968 moved GUI golden-flow shards to `manzanita-standard` failed all six shards before any GUI journey executed.

Run: https://github.com/raullenchai/Rapid-MLX/actions/runs/33762870842

Every shard failed in **Verify and extract GUI app** with `python: command not found`. The managed image provides `python3` (reported as Python 3.14.6), but does not install a `python` compatibility shim.

## Acceptance

Use the portable `python3` executable for commit-bound GUI artifact creation and verification; actionlint and the artifact helper contracts pass; a full Mergify Mac candidate reaches the actual GUI journeys on Manzanita.

### LINKED ISSUE #3008: CI: canonicalize transient update-card AX ordering on managed macOS runners
## Problem

The managed macOS GUI matrix reaches the real Desktop journeys, but `update-busy.app-panel` intermittently reports a structural mismatch because SwiftUI exposes the identified `UpdateCard` overlay on either side of the live Memory footer element. The controls, parent, state, and authored order inside both components are unchanged.

Evidence:
- rapid-mac run 33767105574
- rapid-mac run 33783451159
- rapid-mac run 33788832199

This shared false red causes Mergify to bisect and eject unrelated Mac queue entries.

## Required behavior

Canonicalize only the identified transient update overlay relative to the base window content while preserving all authored order inside the overlay and footer, with focused positive and negative-control tests. The gate must still fail for missing controls, reparenting, changed identifiers, changed state, and unrelated sibling reordering.

## Scope

AX structural-baseline normalization and focused regression coverage only; no Desktop product behavior, update policy, or broad sibling sorting.

raullenchai:
## Why

The managed macOS queue exposed three independent CI portability failures after the runner migration: the image provides `python3` without an unversioned `python`; the GUI catalog assertion could race before all entries arrived; and SwiftUI exposed the identified update overlay on either side of the live Memory footer node across otherwise identical sessions. Because each split PR was tested in isolation, every Mergify candidate lacked another prerequisite and bisection repeatedly ejected healthy source PRs.

This intentionally consolidates the queue prerequisites into one exact-head candidate so the real six-shard GUI matrix can verify the foundation atomically.

Fixes #2986.
Fixes #3008.
Supersedes #2985.

## What

- invoke the GUI artifact helper with the portable `python3` executable at both producer and consumer boundaries;
- wait for the complete GUI model catalog before asserting its contents, with bounded polling and contract coverage;
- isolate the Gemma serve-error mapping test from shared application state;
- canonicalize only the uniquely identified `UpdateCard` overlay immediately before its stable `Footer.DesktopVersionPill` anchor;
- update the measured launch-integrations and update-card AX baselines.

## Design precedent

The AX normalizer already handles session-dependent placement for a uniquely identified SwiftUI overlay while preserving all unrelated authored sibling ordering. This change applies that same narrow, fail-visible pattern to the measured update-card case. It deliberately does not sort generic footer or window children.

## Scope

Allowed: GUI artifact interpreter portability, complete catalog polling, focused test isolation, and narrowly identified AX overlay normalization.

Non-goals: product UI behavior, runner selection, timeout policy, broad AX sorting, release/signing behavior, model catalog content, or user-visible update semantics.

## Verification

- `pytest -q tests/test_ax_baseline.py tests/test_ax_baseline_os_variance.py tests/test_gui_control_behavior_contract.py tests/test_kv_cache_gemma4_gate.py` (123 passed)
- both recorded Manzanita `update-busy` AX order variants normalize to the committed baseline
- `ruff check` and `ruff format --check` on all changed Python files
- `actionlint .github/workflows/rapid-mac-ci.yml`
- `git diff --check`
- exact-head `tests`, `desktop-tests`, and `version-bump-guard`: green
- `pr_validate`: MERGE-SAFE (21,411 passed, 108 skipped, 6 xfailed)
- pending: the Mergify candidate's real six-shard GUI matrix on Manzanita

## Risk and rollback

Risk is limited to test infrastructure. Duplicate/missing update-card anchors remain unnormalized, and unrelated sibling reordering remains regression-sensitive. Revert this PR to restore the prior queue behavior.

### DIFF
--- a/.github/workflows/rapid-mac-ci.yml
+++ b/.github/workflows/rapid-mac-ci.yml
@@ -331,7 +331,7 @@ jobs:
           codesign --verify --deep --strict "$app"
           mkdir -p "$ARTIFACT_DIR"
           ditto -c -k --keepParent "$app" "$ARTIFACT_DIR/Rapid-MLX-Desktop.zip"
-          python ../../scripts/gui_app_artifact.py create \
+          python3 ../../scripts/gui_app_artifact.py create \
             --archive "$ARTIFACT_DIR/Rapid-MLX-Desktop.zip" \
             --source-sha "$GITHUB_SHA" \
             --output "$ARTIFACT_DIR/manifest.json"
@@ -431,7 +431,7 @@ jobs:
           ARTIFACT_DIR: ${{ runner.temp }}/gui-app-artifact
         run: |
           set -euo pipefail
-          python ../../scripts/gui_app_artifact.py verify \
+          python3 ../../scripts/gui_app_artifact.py verify \
             --archive "$ARTIFACT_DIR/Rapid-MLX-Desktop.zip" \
             --manifest "$ARTIFACT_DIR/manifest.json" \
             --expected-source-sha "$GITHUB_SHA"
--- a/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/launch-integrations.complete.txt
+++ b/apps/rapid-mac/Tests/GUIGoldenFlows/__Snapshots__/launch-integrations.complete.txt
@@ -13,6 +13,11 @@ AXApplication title="Rapid-MLX"
         AXGroup subrole=AXHostingView
           AXHeading desc="Connect your agents, Connect any agent or editor that supports a local base URL. It's free and stays on your Mac." enabled=true
           AXScrollArea
+            AXStaticText value=text enabled=true
+            AXTextField id="ConnectTools.ServerPort.Field" value=empty enabled=true
+            AXButton id="ConnectTools.ServerPort.Save" desc="Save" enabled=false
+            AXStaticText value=text enabled=true
+            AXStaticText value=text enabled=true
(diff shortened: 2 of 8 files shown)


raullenchai (commented):
Vector independent review — exact head `3ddf3497490f6099c673b4a1de99878021e7f341`

Contract reviewed: restore commit-bound GUI artifact creation/verification on the managed Mac image; scope limited to the two interpreter invocations; no runner-routing, test-policy, product, timeout, or signing changes.

LGTM. The two-line diff consistently uses the guaranteed versioned `python3` executable at both producer and consumer boundaries, preserves arguments and artifact semantics, and directly matches the observed exit-127 failure. Existing actionlint, 87 focused contracts, helper smoke test, diff check, `pr_validate` MERGE-SAFE result, and exact-head required checks are sufficient for this bounded change. The separate missing Aqua/Dock session is an operator prerequisite, not a defect in this diff.

PR #3085, opened 2026-09-05
fix(qwen4): use the MLX-exact sigmoid forms in fused GDN decode

pierre427:
## Why

The fused single-token GDN decode kernel (#2918) ships under a bit-identity gate against the stock path, and two of its three sigmoid gates do not meet it. `mlx_sigmoid_fast` is used for the beta gate (bf16) and for the float32 output gate, but the stock ops those gates replace are `mx.sigmoid` calls, and on the tested MLX builds `mx.sigmoid` and the compiled `nn.silu` do not agree at the bit level: the runtime-compiled sigmoid resolves its exponential to the fast approximation (ml-explore/mlx#4461 records the mechanism).

An exhaustive sweep over all 65,280 finite bf16 values (MLX 0.32.1 and a 0.32.2 development build, identical counts) shows:

| boundary | stock op | exact form | rejected form |
|---|---|---|---|
| beta gate (bf16) | `mx.sigmoid` | `mlx_sigmoid_precise<T>`: 0 mismatches | `mlx_sigmoid_fast<T>`: 1 mismatch (x ~ -6.85) |
| output gate (float32 of bf16 z) | `mx.sigmoid(z.astype(float32))` | `mlx_sigmoid_precise<float>`: 0 mismatches | `mlx_sigmoid_fast<float>`: 628 mismatches |
| SiLU (bf16) | `nn.silu` | `x * mlx_sigmoid_fast(x)`: 0 mismatches | `x * mlx_sigmoid_precise(x)`: 1 mismatch |

The beta-gate value is reached by real Qwen3.8-Flash-Next activations: while extending this kernel body to a width-3 speculative-verify block (a shape the single-token kernel declines), a full-model gate matched stock for two rounds and then diverged by one bf16 ULP in one GDN layer; bisecting an instrumented copy put it at `shr[3] = float(mlx_sigmoid_fast(beta[hv]))` (`0.0010681152` vs `mx.sigmoid`'s `0.0010604858`). The same beta projections feed the single-token kernel, and the 32-step layer gate and the 1K–64K ladder simply did not sample that input; a single-token reproduction has not been captured, so this PR is best read as exactness hardening on the sweep evidence rather than a reproduced end-to-end defect. It was reported on #2912 (comment 5502937813) after that PR had been re-landed as #2918, so it never reached `main`.

## Scope

- `vllm_mlx/kernels/qwen4_fused_gdn_decode.py`: the beta gate and the float32 output gate use `mlx_sigmoid_precise`. Two lines plus comments; the helpers already existed in the header.
- `tests/test_qwen4_fused_gdn_decode.py` (already in the Apple CI roster, so no workflow edit): source-string contracts that pin which helper each production assignment line uses, and Metal-gated sweeps that compile isolated helper bodies against the kernel header and re-check the whole finite-bf16 domain against the installed MLX for all three boundaries.
- `docs/engineering/performance/2026-09-01-qwen4-fused-gdn-decode.md`: a short section recording why the two stock sigmoids need different forms, which builds were swept, and that the note's timings predate this change.

## Non-goals

- The convolution SiLU keeps `mlx_sigmoid_fast`: `nn.silu` is `mx.compile`d, so the fast form is the one that matches it on every value.
- No change to the decay path (`mlx_softplus_fast` + `metal::precise::exp`), the q/k reduction, or the rsqrt; separate probes over the same domains matched their stock ops and are not touched here.
- No change to admission, probing, or the fused speculative-verify variant.

## Acceptance

- With `RAPID_MLX_QWEN4_FUSED_GDN_DECODE=1`, the fused kernel's beta gate and output gate produce the same bits as `mx.sigmoid` on every finite bf16 input; the kernel's outputs no longer differ from stock at the affected activation values.
- The three source-string contracts in `tests/test_qwen4_fused_gdn_decode.py` fail (two of three) if either gate is switched back to the fast form.

## Verification

- `ruff check` and `ruff format --check` clean on the touched files.
- `python -m pytest tests/test_qwen4_fused_gdn_decode.py -k "not every_finite and not every_bf16_valued"` (everything except the three Metal sweeps): 20 passed. The three source-string contracts against the pre-fix kernel file: 2 failed (beta gate, output gate), 1 passed, as intended.
- The three Metal sweep tests were not executed on the submitting machine while preparing this PR (no Metal execution was permitted during that window). Their bodies are a transcription of the free-standing probes that produced the table above on MLX 0.32.1 and a 0.32.2 development build on 2026-09-01/02; the lab's verify-kernel variant of the same kernel with these two changes was gated bit-exact through 24 engine-rule rounds and 8×256-token generations on Qwen3.8-Flash-Next. Please let the Apple lane run them.
- The full test suite and `pr_validate` were not run for this PR (the full suite exercises Metal); `pr_validate --body-only` was run on the description.

## Behaviour delta

No default changes: the kernel remains opt-in behind `RAPID_MLX_QWEN4_FUSED_GDN_DECODE`. With it enabled, the two gates compute the same function as the stock ops over the whole bf16 domain; the note's layer and end-to-end timings predate the change and were not re-measured here (one precise exponential per output element replaces a fast one; the decay path already used the precise form).

## AI assistance disclosure

AI-assisted throughout. The two-line kernel change, the tests added to the decode test file and the performance-note section were written by an AI coding agent (Claude) and adversarially reviewed by a second agent (Codex) before submission; the bf16 sweep probes that produced the mismatch counts were also AI-written and were run on the lab's machines as free-standing scripts. The per-gate decision (precise for the two `mx.sigmoid` boundaries, fast for `nn.silu`) and this description were reviewed by the submitter, who can explain the intent, risk and behaviour of each change.

## Checklist

- [x] Targeted tests pass locally (see Verification; the full suite was not run for this PR)
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate <PR#> --body-only` (heavy steps left to the maintainer lane)
- [x] New tests spot-checked to fail when the corresponding production line is broken (see Verification)
- [x] Updated docs (performance note; timings labelled as pre-change)

### DIFF
--- a/docs/engineering/performance/2026-09-01-qwen4-fused-gdn-decode.md
+++ b/docs/engineering/performance/2026-09-01-qwen4-fused-gdn-decode.md
@@ -75,6 +75,32 @@ The environment flag remains disabled by default. Prefill, multi-request
 batching, ragged caches, and speculative rollback are explicit stock-path
 fallbacks rather than unqualified extensions of this result.
 
+## Sigmoid forms per boundary
+
+The kernel matches two stock sigmoids that do not agree at the bit level on
+the tested MLX builds. `mx.sigmoid` comes from MLX's prebuilt kernels;
+`nn.silu` is `mx.compile`d at runtime, and the runtime-compiled sigmoid
+resolves its exponential to the fast approximation (ml-explore/mlx#4461
+records the mechanism). An exhaustive sweep over all 65,280 finite bf16
+values on MLX 0.32.1 and on a 0.32.2 development build gave the same counts:
+`mlx_sigmoid_precise` matches `mx.sigmoid` on every value in bf16 and in
+float32, while `mlx_sigmoid_fast` differs on one bf16 value (x ~ -6.85) and on
+628 values in float32; `x * mlx_sigmoid_fast(x)` matches `nn.silu` on every
+value and the precise form differs on one. The beta gate and the float32
+output gate therefore use `mlx_sigmoid_precise`, and the convolution SiLU
+keeps `mlx_sigmoid_fast`.
+
+The x ~ -6.85 beta value was observed in real Qwen3.8-Flash-Next activations
+during a width-3 speculative-verify round of the same projections (a shape
+this single-token kernel declines); the single-token gate sequences above did
+not sample it. The change is exactness hardening on the sweep evidence: it
+makes the kernel's gates the same functions as the stock ops over the whole
+bf16 domain. The layer and end-to-end timings in this note were measured
+before the change; the output gate now evaluates one precise exponential per
+output element, and the decay path already used `metal::precise::exp`.
+`tests/test_qwen4_fused_gdn_decode.py` pins the per-gate choice as a
+source-string contract and re-runs the sweep on Metal.
+
 ## Failure lifecycle
 
 The capability probe compile-and-runs the exact BF16 kernel specialization
--- a/tests/test_qwen4_fused_gdn_decode.py
+++ b/tests/test_qwen4_fused_gdn_decode.py
@@ -724,9 +724,14 @@ def test_precise_bf16_sigmoid_matches_mx_sigmoid_on_every_finite_bf16():
     )
     mx.eval(reference, precise, fast)
     assert _bit_mismatches(precise, reference, mx.uint16) == 0
-    # The fast form is what the beta gate shipped with; keep the evidence
-    # that it is not the same function as the stock op.
-    assert _bit_mismatches(fast, reference, mx.uint16) >= 1
+    # The fast form is what the beta gate shipped with. Whether it differs
+    # from the stock op is GPU-family dependent (metal::exp's fast path is
+    # bit-identical on the GitHub Apple runners but not on M3 Ultra), so it
+    # is reported, not asserted: the contract this test pins is that the
+    # PRECISE form matches everywhere.
+    print(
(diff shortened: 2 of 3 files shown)


raullenchai (approved):
Maintainer review (Atlas, M3 Ultra Studio, mlx 0.32.2).

**Claims verified against source**
- Stock beta gate is `mx.sigmoid(b)` in `mlx_lm.models.gated_delta.gated_delta_update` (not compiled) → `mlx_sigmoid_precise<T>` is the matching form.
- Stock output gate is `mx.sigmoid(gate.astype(float32))` in `SigmoidRMSNormGated.__call__` (not compiled) → `mlx_sigmoid_precise<float>` is the matching form.
- Conv activation is `nn.silu` (`mx.compile`d) → keeping `mlx_sigmoid_fast` there is correct.

**Ran here what you could not run**
- `tests/test_qwen4_fused_gdn_decode.py`: **23 passed** including the three Metal exhaustive-bf16 sweeps (`mx.metal.is_available()` true). Precise/fast mismatch counts behave as your table says.
- `pr_validate` CLI: codex adversarial review — no blocking findings; full suite 22503 passed / 0 failed; lint clean. Verdict MERGE-SAFE. The supply-chain scanner's three `eval()` hits are `mx.eval(...)`, false positives.
- Workflow runs approved so the Apple lane runs on the real commit.

Kernel stays opt-in behind `RAPID_MLX_QWEN4_FUSED_GDN_DECODE`, so the one extra precise `exp` per output element carries no default-path risk; a re-measured timing in the perf note can follow separately. Will label for the merge queue once CI is green.

Issue #2955, opened 2026-09-02
Paged prefix cache silently rejects rotating/hybrid KV layouts and retains duplicate snapshots


raullenchai:
## User-visible problem

With `--use-paged-cache` on a sliding-window model, the server can stay healthy while prefix reuse remains zero. The paged path also retains full per-request KV snapshots even though it cannot rebuild those cache layouts, increasing memory rather than reducing it. One observed Gemma-4 run used about 1.2 GiB more memory than the non-paged path.

This silent failure shape is not Gemma-specific. The current paged rebuild eligibility rejects rotating and recurrent/hybrid cache layers, so GPT-OSS and other sliding-window families can enter the same ineffective path.

## Verified code findings

- Paged block construction/rebuild accepts only its narrow cache-shape allowlist; rotating and recurrent cache layers produce no reusable blocks.
- Request records can still retain full `cache_data` after that incompatibility is known.
- Hit/tokens-saved counters are updated before rebuild success is known, so a failed rebuild can report a hit that did not save prefill.
- Failed rebuild cleanup/ref ownership needs an explicit regression test to rule out retained block references.

## Required behavior

Unsupported cache layouts must not silently enter a paged mode that provides zero reuse. Admission must either use a proven compatible fallback or reject the explicit option with an actionable message. Metrics must count a hit only after a cache state is successfully restored, and rejected layouts must not retain duplicate full snapshots or block references.

## Acceptance

- Capability-driven tests cover full-attention, rotating/sliding-window, and recurrent/hybrid cache layouts; no model-name-only gate.
- Gemma-2/3/4 and GPT-OSS rotating-cache shapes do not enter an ineffective paged path.
- Qwen/Mistral configurations with sliding-window disabled retain the supported full-attention behavior.
- Recurrent Qwen3.5/3.6/Bonsai cache shapes fail closed or use their existing bounded snapshot fallback.
- Hit/tokens-saved metrics increment only after successful restore/rebuild.
- Failure-injection tests prove block refs and retained snapshots are released on rebuild failure.
- A repeated-prefix user-side proof shows nonzero saved tokens on the selected fallback and no memory increase from duplicate snapshots.

## Non-goals

- No new paged representation for rotating or recurrent state in this fix.
- No model-family regex shim.
- No unrelated cache-performance tuning.

Issue #1122, opened 2026-07-17
Prefix cache never stores when using a hybrid model


literally-anything:
### Rapid-MLX version

0.10.10

### Hardware

MacBook Pro M3 Pro, 36 GB

### macOS version

Golden Gate 27.0 Beta 2 (26A5378j)

### Python version

Python 3.14.6

### Model

qwen3.5-9b-4bit, qwen3.6-35b-dwq, etc.

### Full serve command

rapid-mlx serve qwen3.5-9b-4bit --port 8000 --tool-call-parser qwen3 --enable-auto-tool-choice --reasoning-parser qwen3 --kv-cache-turboquant --enable-prefix-cache --log-level DEBUG

### Streaming or non-streaming?

Streaming (`stream: true`)

### What happened? What did you expect?

I expect it to store in the cache properly, but hybrid models seem to always respond with something along the lines of:
```
INFO:rapid_mlx.scheduler:[cache_store] request=2f425bbe-5e6 tokens=33 (13 prompt + 20 output) stored=False time=0.000s cache_entries=0 cache_mem=0MB
```
Even though the cache is empty, I have more than enough memory for the cache, and caching is enabled, it refuses to use the prefix cache on hybrid models.

### Minimal reproduction

Start the server and run:
```bash
curl -X POST http://localhost:8000/v1/chat/completions \
     -H "Content-Type: application/json" \
     -d '{
       "model": "qwen3.5-9b-4bit",
       "messages": [{"role": "user", "content": "Test"}]
     }'
```

### Server logs / error output

```shell
Alias: qwen3.5-9b-4bit → mlx-community/Qwen3.5-9B-4bit
INFO:rapid_mlx.model_auto_config:Resolved alias profile for 'mlx-community/Qwen3.5-9B-4bit' → tool_call_parser=hermes, reasoning_parser=qwen3, is_hybrid=False, supports_spec_decode=False, suffix_tier=unknown, pflash_tier=verified, turboquant_tier=k8v4_verified
INFO:rapid_mlx.pflash:PFlash default: alias 'mlx-community/Qwen3.5-9B-4bit' is pflash_tier=verified — engine defaults to --pflash always. Pass --pflash off to compare against the no-compression baseline.
[...]

PR #3086, opened 2026-09-05
fix(cli): report cache flags not applied under --no-memory-aware-cache

pierre427:
## Why

`--hybrid-cache-entries` and `--prefix-cache-index` configure the prefix-cache backend through `MemoryAwarePrefixCache`. They are parsed, resolved and threaded into `SchedulerConfig`, and then the legacy branch of `Scheduler.__init__` (`--no-memory-aware-cache`) builds a bare `PrefixCacheManager(model, max_entries=prefix_cache_size)` that applies neither (the hybrid count is not dead elsewhere: the batched engine still reads it as the message-boundary snapshot signal). A serve that typed them under `--no-memory-aware-cache` comes up healthy, logs `Prefix cache enabled with max_entries=...`, and silently gets something else than it asked for:

- a typed hybrid quota (the bound on live-stored, evictable non-trimmable entries in the memory-aware cache) is replaced by the global `--prefix-cache-size` limit; the entry-count cache does keep recurrent-state entries and reuses shorter prefixes without trimming, it just has no separate bound for them;
- an explicit `--prefix-cache-index` never builds `RadixPrefixIndex`; the entry-count cache always uses its own token trie.

This is the silent-option class #2955 closed for the paged cache ("fail loudly rather than come up healthy" with a different configuration than requested). Confirmed on the lab's own serve script, where `--no-memory-aware-cache` had been quietly overriding a typed `--hybrid-cache-entries`.

## Scope

- `vllm_mlx/cli.py`: `_legacy_prefix_cache_dropped_flags(args, argv)` returns one line per explicitly typed flag the legacy branch will not apply (either `--flag value` or `--flag=value` spelling, the same `sys.argv` check `serve` already uses for `--hybrid-cache-entries` and `--prefill-step-size`). `serve_command` prints each as `Warning: with --no-memory-aware-cache, ...` right after the hybrid-entry resolution, before the scheduler config is built.
- `tests/test_legacy_prefix_cache_flag_warnings.py`: explicit (including a typed `0`, which the memory-aware cache honours by retaining no hybrid entries while the legacy cache keeps them under `--prefix-cache-size`), `=`-form, auto-derived, defaulted, both-flags, and not-legacy configurations; an AST test pinning that `serve_command` calls the helper with `sys.argv`; and two behavioral tests that drive the real `rapid-mlx serve` path (heavy prologue stubbed as in `tests/test_serve_listen_fd.py`) through to `uvicorn.run` and assert the warning lines are printed under `--no-memory-aware-cache` and absent under the default cache. The file is registered in the Apple CI roster so the changed-lines-coverage gate sees the print site.

## Non-goals

- Making the entry-count cache honour either flag. That would be a feature; this PR makes the substitution visible.
- Warning on defaults. `--prefix-cache-index` defaults to `radix`, and the auto-derived hybrid count (#1122) still gates message-boundary snapshots in `BatchedEngine._needs_prefix_boundary_snapshot`, so neither is a dropped expectation and neither is reported.
- The paged-cache branch (`--use-paged-cache`), which has its own capability gate (#2955), and the scheduler itself; no cache construction or default changes.

## Acceptance

- `rapid-mlx serve <model> --no-memory-aware-cache --hybrid-cache-entries 8` prints `Warning: with --no-memory-aware-cache, --hybrid-cache-entries=8 is a memory-aware cache quota; the legacy entry-count cache keeps hybrid entries under --prefix-cache-size=... instead`.
- `rapid-mlx serve <model> --no-memory-aware-cache --prefix-cache-index hash` prints `Warning: with --no-memory-aware-cache, --prefix-cache-index=hash is not read by the legacy entry-count cache, which always uses its own token trie (RadixPrefixIndex is built only for the memory-aware cache)`.
- A default serve, a `--no-memory-aware-cache` serve with no typed cache-index and no typed hybrid count (including a hybrid model whose count is auto-derived), and any paged or prefix-cache-disabled serve print nothing new. A typed `--hybrid-cache-entries 0` under `--no-memory-aware-cache` is reported: the memory-aware cache honours 0 by not retaining hybrid entries, the legacy cache retains them regardless (maintainer review follow-up).
- Deleting the call site in `serve_command` fails the AST test.

## Verification

- `ruff check` and `ruff format --check` clean on both touched files.
- `python -m pytest tests/test_legacy_prefix_cache_flag_warnings.py`: 13 passed. With the `serve_command` loop deleted, the call-site test fails (1 failed), as intended.
- Neighbouring suites over the same CLI helpers and an existing `serve_command` entry test: `tests/test_hybrid_prefix_cache_auto_default.py`, `tests/test_audio_boot_check.py`: 48 passed.
- Not run for this PR: the full suite and the model-loading steps of `pr_validate` (no Metal execution was permitted on the submitting machine during this work); `pr_validate --body-only` was run on the description.

## Behaviour delta

One or two new `Warning:` lines on the startup banner in the explicit cases above; no change to cache construction, defaults, or admission.

## AI assistance disclosure

AI-assisted throughout. The CLI helper and the tests were written by an AI coding agent (Claude) and adversarially reviewed by a second agent (Codex) in two passes before submission (the first pass corrected an over-claim about the entry-count cache's hybrid handling, which is why the warnings describe a substitution rather than a dead flag); verification is the targeted runs listed above. The diagnosis of where each flag is consumed was checked against the source by the submitter, who can explain the intent, risk and behaviour of each change.

## Checklist

- [x] Targeted tests pass locally (see Verification; the full suite was not run for this PR)
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate <PR#> --body-only` (heavy steps left to the maintainer lane)
- [x] New tests spot-checked to fail when the corresponding production line is broken (see Verification)
- [x] No docs change needed beyond the warning text

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -490,6 +490,7 @@ jobs:
             tests/test_sdxl_alias.py \
             tests/test_bonsai_image_alias.py \
             tests/test_sd35_alias.py \
+            tests/test_legacy_prefix_cache_flag_warnings.py \
             -v --tb=short \
             -m "not slow" \
             -k "not Integration" \
--- a/tests/test_legacy_prefix_cache_flag_warnings.py
+++ b/tests/test_legacy_prefix_cache_flag_warnings.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import ast
 import inspect
+import sys
 from types import SimpleNamespace
 
 import pytest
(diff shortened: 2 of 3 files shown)


raullenchai (approved):
Atlas maintainer review (external contribution, adversarial pass).

Verified: helper semantics against the source (memory-aware cache skips storing hybrid entries when the quota is 0 — memory_cache.py store path; legacy PrefixCacheManager retains everything under --prefix-cache-size; RadixPrefixIndex only built in the memory-aware branch), so both warnings describe real substitutions. Two maintainer commits on top of the contributor's: (1) end-to-end tests driving the real serve path to uvicorn.run and asserting the printed warnings, plus Apple roster registration for the coverage gate; (2) a typed --hybrid-cache-entries 0 is reported too (previously codified as silent). pr_validate: codex MERGE-SAFE after 4 rounds, full_unit 22541 passed, lint clean, patch coverage 100%.

Issue #2561, opened 2026-08-28
feat(engine): lossless draftless n-gram decoding for hybrid Flash-Next


raullenchai:
## User goal

Allow users serving the hybrid Flash-Next text model to opt into draftless n-gram speculative decoding for copy, edit, structured-output, and agent workloads without changing greedy output.

## Scope

- Reuse the existing suffix decoder and scheduler integration.
- Add a minimum suffix-match floor so weak one-token matches never enter verification.
- Add an engine-owned verify transaction for recurrent/QSA cache state: checkpoint before a multi-token verify, restore on rejection, and replay only the committed prefix.
- Make sliding-window append/rollback admission account for verify growth.
- Export proposal, acceptance, rejection/replay, and fallthrough evidence through existing metrics.
- Keep the feature explicit opt-in and preserve baseline behavior by default.

## Acceptance

- Greedy output is token-identical to baseline across the correctness battery, including rejection and QSA compression-boundary cases.
- High-overlap copy/JSON/edit workloads show a material decode win; novel generation stays within the regression floor by falling through when no 24-token match exists.
- Cancellation, EOS/stop, max-token termination, cache extraction, and consecutive requests preserve cache/request lifecycle.
- A real user-side serve request succeeds with the released model checkpoint and documented opt-in config.
- Existing suffix decoding and scheduler tests remain green.

## Non-goals

- No draft model, MTP weights, sampling-mode speculation, release bump, automatic enablement, or CI workflow changes.

Also resolves the existing sliding-window verification boundary failure tracked in #2463 through the same transaction admission layer.

PR #3104, opened 2026-09-05
feat(suffix): opt-in hybrid scratch-verify transaction (#2561 B2.1)

raullenchai:
## Why (Part of #2561 B2 — lossless hybrid n-gram)

The portfolio B2 contract + issue #2561 want draftless n-gram speculative decoding to work losslessly on **hybrid Flash-Next** (recurrent/GatedDeltaNet/PLE/QSA). The pure-attention suffix path rejects hybrid because chunked-batched verify isn't step-equivalent on recurrent layers. This PR implements the **engine verify transaction** for hybrid, default-off and opt-in, mirroring vLLM/SGLang's n-gram worker.

## Change (suffix_hybrid, default-off)

Adds an opt-in `--suffix-hybrid` path to the existing suffix decoder (`_install_suffix_decoding` / `_suffix_step`):
- **Scratch-verify** (`_verify_scratch`): run the tentative multi-token `model([X, d_0..d_{K-1}])` forward on a **deepcopy'd scratch cache** so the persistent recurrent/conv/PLE/QSA state stays untouched (SGLang NGRAMWorker isolation).
- **Commit-only-accepted** (`_commit_scratch_accepted`): replay ONLY the accepted prefix `[X, d_0..d_{n-1}]` stepwise onto a fresh pre-verify copy, then swap it into the live cache — the rejected tail is never advanced (drop-the-tail), so it's lossless.
- **24-token min-match floor** (`suffix_min_match_len=24`): a draft shorter than the floor falls through (`ft_short_match`) without paying the multi-token verify cost — novel text never touches the tentative-verify path.
- **Bit-exactness guard** (`--suffix-hybrid-bit-exact`, default ON; `--no-suffix-hybrid-bit-exact` is an explicit unsafe evaluation opt-out): replays the committed path stepwise and compares greedy preds to the chunked verify; on any mismatch (quantized-hybrid drift signature) it refuses to commit and falls through (`ft_hybrid_drift`) rather than silently corrupting.
- Hybrid gate opens only when `suffix_hybrid=True` AND `profile.is_hybrid`. Pure-attention path untouched. Additive `ft_short_match`/`ft_hybrid_drift` fallthrough buckets + `hybrid_drifts` counter; **no existing metric renamed**.

## Key finding (independently verified)
The #3090 `Qwen4ExpStateCache.can_trim`/`trim_all` contract, while a correct standalone invariant, does NOT by itself unlock the hybrid flip: on the real hybrid capture, `record_slot_snapshots(finalize=True)` publishes **K** boundaries for a K-token verify (not K+1), so `can_trim(K)`=False → `can_advance(c,K)`=False → the attention verify preflight fails on hybrid. This design therefore **sidesteps `trim_all`/`can_advance` entirely** using the scratch+replay (drop-the-tail) pattern — #3090 + B2.1 together deliver the lossless hybrid enablement.

## Reference (raullen rule)
SGLang `python/sglang/srt/speculative/ngram_worker.py` + `spec_utils.py`: hybrid Mamba/linear-attention n-gram spec-decode via scratch verify + commit-only-accepted of the last correct step. This PR mirrors that (isolation + selective commit) rather than reinventing.

## Test plan
- Expanded `tests/test_suffix_hybrid_scratch_verify.py` transaction/regression battery: scratch-verify core (accept-all/partial/reject-first state-equality on the **real tiny Qwen4-exp hybrid** via exact recurrent/QSA/conv equality vs stepwise gold), install-gate opt-in, 24-floor short-match fallthrough, bit-exactness guard (step-exact no-misfire + forced-drift refuses cleanly), token-identity + lifecycle, prefix-cache decodability after commit.
- Current post-main-rebase gate: 171 focused suffix/config tests pass; ruff, format, and diff checks are clean. Real served dogfood and the release hold are recorded in the latest PR comment.

## Non-goals (this slice)
- No default enablement (hybrid stays opt-in). No version bump, no CI changes, no metric renames. Full real Flash-Next serve bakeoff = B2.2 follow-on.

### DIFF
--- a/tests/test_no_mllm_flag.py
+++ b/tests/test_no_mllm_flag.py
@@ -178,6 +178,9 @@ NON_ROUTING_FLAGS_ALLOWLIST: frozenset[str] = frozenset(
         "--enable-kv-cache-turboquant",
         "--enable-prefix-cache",
         "--disable-prefix-cache",
+        # Evaluation-only numerical guard override for the explicitly selected
+        # hybrid suffix feature; it does not auto-detect or select a route.
+        "--no-suffix-hybrid-bit-exact",
         # Chat-template toggle, not engine routing.
         "--no-thinking",
         # `--no-think` is a hidden back-compat alias of `--no-thinking` on
--- a/tests/test_speculative_config.py
+++ b/tests/test_speculative_config.py
@@ -395,18 +395,38 @@ def test_no_speculative_config_fills_suffix_runtime_defaults() -> None:
     assert args.suffix_min_draft_len == 2
 
 
-def test_hybrid_cli_keeps_default_cap_for_pure_attention() -> None:
-    """The CLI must NOT raise ``suffix_max_draft`` when ``--suffix-hybrid`` is
-    set: the hybrid width raise is gated on an actual hybrid model in the
-    installer, so a pure-attention model (where the flag is a no-op) keeps the
-    normal 8-token default instead of silently tripling its verify width."""
+def test_hybrid_below_floor_cap_fails_fast(capsys) -> None:
+    """``--suffix-hybrid`` with a below-floor ``--suffix-max-draft`` is a
+    contradictory config: either the hybrid feature is a silent no-op or the
+    max-width limit is silently violated. Fail fast with an actionable error
+    naming the flag to raise (no silent override, no pure-attention leak)."""
+    import pytest
+
     from vllm_mlx.cli import _normalize_speculative_config_or_exit
 
     args = _spec_config_args(
         suffix_decoding=True, suffix_hybrid=True, suffix_min_match_len=24
     )
+    with pytest.raises(SystemExit) as excinfo:
+        _normalize_speculative_config_or_exit(args)
+    assert excinfo.value.code == 2
+    err = capsys.readouterr().err
+    assert "--suffix-hybrid requires --suffix-max-draft" in err
+    assert "suffix_max_draft=8" in err
+
+
+def test_hybrid_at_floor_cap_ok() -> None:
+    """``--suffix-hybrid`` with ``--suffix-max-draft >= floor`` validates."""
+    from vllm_mlx.cli import _normalize_speculative_config_or_exit
+
+    args = _spec_config_args(
+        suffix_decoding=True,
+        suffix_hybrid=True,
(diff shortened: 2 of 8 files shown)


pierre427:
The -9% and -34% numbers look like copy cost rather than a flaw in the scheme. Each round
deep-copies the full cache, KV included, and commits by stepping n+1 full-model forwards.
In our MLX stack we got rid of both by making the recurrent cache trimmable instead of
scratch-verifying.

During a speculative forward, each GDN layer records a closure over its pre-forward
`(conv_state, state)` and the per-token q/k/v/a/b. Those are MLX array refs, so nothing is
copied. `cache.trim(k)` then re-runs only `gated_delta_update` over the accepted prefix.
KV caches trim by offset:
https://github.com/pierre427/mlx2/blob/41ab59c/src/mlx2/runtime/models/qwen3_5.py#L193-L222
https://github.com/pierre427/mlx2/blob/41ab59c/src/mlx2/runtime/models/cache.py#L2080-L2129

Against `_verify_scratch` and `_commit_scratch_accepted`, it would look roughly like this:
```python
for c in live_cache: c.start_speculation(rollback_window=K + 1)
logits = model(verify_input, cache=live_cache)      # no deepcopy
n = accepted_prefix(mx.argmax(logits, -1).tolist()[0], draft)
if n < K:
    for c in live_cache: c.trim(K - n)               # GDN: replay first n+1 tokens only
for c in live_cache: c.stop_speculation()
```
The terminal repair in `_repair_hybrid_terminal` becomes another `trim`.

Evidence: on Qwen3.8-Flash-Next (M5 Max) a fused variant gives bit-exact restored state and
continuation logits against a snapshot baseline, with 2.9–6.9x lower peak memory
(https://github.com/pierre427/mlx2/blob/41ab59c/docs/experiments/QWEN4-GDN-REPLAY-GPU-2026-09-17.md).
We have not measured end-to-end n-gram speed on hybrids, or anything on Ling. One caveat:
the state matches the chunked verify, not stepwise decode, so your drift guard is still a
separate question. The code is MIT. Happy to help if you try it.


---
<sub>AI-assisted: drafted with Claude from our code, then reviewed. The links are pinned to a public snapshot of the code, and the claims above state the evidence level we actually have.</sub>

PR #3249, opened 2026-09-08
fix(desktop): park DMG .background icon below the install window fold

guo:
## Why

The DMG install page renders a stray `.background` folder icon next to the app and Applications icons on any Mac running Finder with hidden files shown (`AppleShowAllFiles=1`). It sits on top of the branded artwork, so the first thing a user does after downloading the app is look at a broken-looking install page.

A dot prefix hides a name only while Finder keeps its default setting, and no file attribute overrides that. Both `chflags hidden` and the kIsInvisible FinderInfo bit were measured against a live Finder on a mounted volume; neither suppresses the icon.

Root cause: the committed `.DS_Store` template gives `Iloc` records only to the two draggable items, so Finder auto-grids `.background` into the first free cell. At a 96pt icon size that cell is the top-left of the 720x460 window. Finder reported it at `(105,64)` — inside the visible page.

For contrast, a shipped installer that does not exhibit the symptom keeps its equivalent support files at `y=450` against a 400pt window, i.e. below the fold. Its larger 128pt icon size happens to push Finder's auto-placement onto a row outside the window. That is incidental, not a deliberate position — but it is the behaviour this PR makes explicit and deterministic for us.

## Scope

- `Resources/finder-layout.DS_Store` — adds an `Iloc` for `.background` at `(100,560)`. An `Iloc` is the icon's centre, so a 96pt icon starts ~512pt down, below the 460pt window. Edited in place with the `ds_store` module, which preserves the `bwsp`/`icvp` records and the background alias blob byte for byte: 112 bytes differ, file size unchanged at 16388.
- `scripts/verify-dmg-layout.py` — now requires three `Iloc` records, pins the two draggable icons to their exact spots, and gates `.background` on the property that matters: its icon top must sit below the window fold. Computed from the parsed record, not a hardcoded pair, so the parking spot can move without weakening the guarantee.
- `scripts/configure-dmg-layout.sh` — documents why the position exists and the one-liner for moving it, since the template is binary and has no generator script.
- `scripts/validate-dmg.sh` — corrects a comment that claimed dot-prefixed support files "remain hidden from the user's icon view". They do not; the icon positions are what keeps them off the page.
- `tests/test_dmg_layout_verifier.py` — fixtures carry the third `Iloc`; two new tests cover the fold invariant.

## Non-goals

- Does not make `.background` invisible. Enlarging the install window past the fold still reveals it, and that is true of other installers with the same shape. Achieving real invisibility means removing the item from the volume root entirely — relocating the wallpaper into `Contents/Resources/` of the app bundle and retargeting the alias — which touches the signed bundle and the notarisation path and is deliberately out of scope.
- Does not change the wallpaper artwork, window bounds, icon size, or the two draggable icon positions.
- Does not touch the bootstrapper DMG path beyond what it inherits by calling the same `configure-dmg-layout.sh`.
- Does not add a generator script for the binary `.DS_Store` template.

## Acceptance

- A DMG built from this template mounts with `.background` positioned below the window fold; Finder does not draw it on the install page at the shipped window size, with or without hidden files shown.
- `verify-dmg-layout.py` fails a template whose `.background` icon would land on the visible page, with an error naming the fold.
- `verify-dmg-layout.py` accepts any below-fold parking spot, not only `(100,560)`.
- The background alias, window bounds, icon size and the two draggable icon positions are unchanged from `main`.

## Verification

- `pytest tests/test_dmg_layout_verifier.py` — 19 passed, including the fold and duplicate-record regressions.
- Mutation check per SOP §Step 3: disabling the fold comparison in `verify-dmg-layout.py` makes `test_background_inside_window_fails` fail (1 failed, 18 passed); restoring it returns 19 passed.
- `ruff check` and `ruff format --check` on both changed Python files — clean.
- `bash -n` on both changed shell scripts — clean.
- `verify-dmg-layout.py` against the patched template — OK, confirming the alias and window records survived the in-place edit.
- End-to-end Finder readback: built a scratch HFS+ volume from the patched template and queried Finder for icon positions. Result: `.background @ (100,560) | Applications @ (540,228) | Rapid-MLX Desktop.app @ (180,228)`.
- `pr_validate` on exact head `5b998be2d` — MERGE-SAFE: 23,107 passed, 131 skipped, 22 deselected, 6 xfailed, 1 xpassed; adversarial review found no blocking issues.
- `swift test --package-path apps/rapid-mac --no-parallel --filter DMGPresentationScriptTests` — 10 passed.
- Real `scripts/dmg.sh` build with an existing main Desktop app payload — ad-hoc codesign and `hdiutil verify` passed; `validate-dmg.sh` passed on the resulting 128 MB DMG. Live Finder readback: `.background @ (100,560) | Applications @ (540,228) | Rapid-MLX Desktop.app @ (180,228)`. 

## Behaviour delta

DMG install page with Finder set to show hidden files: `.background` drawn at `(105,64)`, overlapping the artwork → parked at `(100,560)`, below the window fold.

## AI assistance disclosure

Claude investigated the cause and wrote all five changed files. Verification was measurement-driven rather than assumed: the claim that a dot prefix or a hidden flag would suffice was tested against a live Finder and disproved before the fix was chosen, the resulting icon position was confirmed by Finder readback on a scratch volume rather than inferred from the template, and the new tests were mutation-checked against the production line they guard.

## Checklist

- [x] Tests pass locally — targeted layout verifier, focused Swift tests, and `pr_validate` full-unit suite passed on exact head `5b998be2d`
- [x] Lint passes (`ruff check && ruff format --check`) — on both changed Python files
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate 3249` — MERGE-SAFE on exact head `5b998be2d`
- [x] If new tests touch a critical code path, I've spot-checked that they fail when the corresponding production line is broken (see SOP §Step 3)
- [x] Updated README/docs if applicable — script header comments updated; no user-facing docs affected


raullenchai (approved):
Approved after scope-locked Harbor review. The DMG layout fix is necessary and correctly limited to deterministic Finder presentation. One blocking duplicate-Iloc false-pass was fixed and regression-tested; the follow-up review converged with no blocking findings. Exact head 5b998be2d passed pr_validate (23,107 unit tests), focused Swift tests, an ad-hoc end-to-end DMG build, hdiutil verification, validate-dmg, and live Finder position readback. No signing, notarization, or release behavior changed.

PR #3275, opened 2026-09-09
fix(serve): preserve alias in profile diagnostics

### LINKED ISSUE #3274: serve verbose profile loses alias capabilities after snapshot resolution
## Problem

Serving a catalog alias resolves it to a local Hugging Face snapshot before the
engine is created. The normal startup summary reports the active speculative
lane correctly, but `RAPID_MLX_PROFILE_VERBOSE=1` then derives the capability
table from the snapshot path. Alias-only facts such as an external MTP sidecar
are lost, so the same startup can claim both `spec decode MTP (active)` and
`MTP path: disabled`.

## Expected behavior

Operator-facing startup diagnostics retain the validated alias identity and
reconcile its capabilities with the actual loaded-checkpoint architecture and
runtime lane. Loading, scheduling, cache keys, and speculative policy must not
change.

## Acceptance criteria

- The verbose profile names the requested alias after snapshot resolution.
- Active/default MTP and suffix copy cannot contradict the runtime lane.
- Existing public positional constructor behavior remains compatible.
- Static checks and focused server/profile tests pass.

raullenchai:
## Why

Fixes #3274. Final 0.14.0 server dogfood found that an alias resolved to a
local snapshot could print two contradictory startup claims: the summary said
`spec decode MTP (active)`, while the verbose table said `MTP path: disabled`.
That makes the release's new runtime-reporting contract misleading to operators.

## Scope

- Preserve the validated alias/profile identity as diagnostic-only metadata
  when `load_model` and dynamic residency construct `BatchedEngine`.
- Use that identity for the startup summary/table while retaining loaded
  checkpoint architecture facts and the actual runtime lane.
- Add regression coverage for alias-to-snapshot startup copy and constructor
  compatibility.

## Non-goals

- No scheduler, speculative-decoding, model-loading, cache-key, residency, or
  multimodal logging behavior changes.
- No policy/default changes and no new model aliases.

## Acceptance

- A resolved `qwen3.5-9b-4bit` startup names the alias and reports its sidecar
  MTP capability without contradicting the active runtime lane.
- The loaded checkpoint remains the engine and scheduler identity.
- Existing positional `EngineConfig(model_name, scheduler_config)` callers keep
  their historical argument mapping.

## Verification

- `uv sync --extra test --locked`
- `uv run pytest -q tests/test_no_mllm_flag.py tests/test_server_load_model_order.py tests/test_model_auto_config.py` — 440 passed
- `ruff check` and `ruff format --check` on all five changed files
- `git diff --check`
- Adversarial review caught and fixed one compatibility risk by appending the
  new dataclass field instead of shifting existing positional arguments.

## Behaviour delta

`serve --verbose` after alias-to-snapshot resolution: contradictory/anonymous
profile copy → alias-preserving copy reconciled with loaded architecture and the
active decoding lane. Runtime behavior is unchanged.

## Design precedent

The implementation follows the established split between a stable user-facing
model identity and a resolved runtime artifact identity: retain both, use the
former for presentation and the latter for execution. Rapid-MLX already applies
this separation to chat-template and serving-lane metadata; this change extends
the same boundary to diagnostics instead of introducing a parallel registry.

## AI assistance disclosure

Codex reproduced the release-dogfood contradiction, wrote and adversarially
reviewed the five-file patch, and verified it with the focused 440-test suite,
lint, formatting, and diff checks. The compatibility finding was fixed and
pinned by a regression test.

## Checklist

- [x] Tests pass locally (focused 440-test affected surface)
- [x] Self-validated with `python3 -m scripts.pr_validate.pr_validate <PR#>`
- [x] New regression tests were spot-checked against the production failure
- [x] README/docs not applicable: operator output now matches existing docs

## Author

X handle (optional, external contributors):

### DIFF
--- a/tests/test_no_mllm_flag.py
+++ b/tests/test_no_mllm_flag.py
@@ -2634,7 +2634,7 @@ def test_routing_override_kwargs_are_keyword_only_in_load_model():
         )
 
 
-def _make_engine_core_for_override_test(monkeypatch, cfg, *, base=None):
+def _make_engine_core_for_override_test(monkeypatch, cfg, *, base=None, detector=None):
     """Build an ``EngineCore`` with heavy dependencies stubbed so the
     routing-override block in ``__init__`` can be exercised in
     isolation. Returns the constructed core (or raises if __init__
@@ -2683,7 +2683,11 @@ def _make_engine_core_for_override_test(monkeypatch, cfg, *, base=None):
     if base is None:
         base = ModelConfig(is_hybrid=True, supports_spec_decode=False)
 
-    monkeypatch.setattr(mac, "detect_model_config", lambda _name: base)
+    monkeypatch.setattr(
+        mac,
+        "detect_model_config",
+        detector if detector is not None else lambda _name: base,
+    )
 
     # Stub respects its argument (not closure). This means pre-enrich
     # mutation of ``base_cfg`` is invisible — matching production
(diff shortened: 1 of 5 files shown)


raullenchai (commented):
Scope-locked adversarial self-review complete. The diff fixes only diagnostic identity propagation. One compatibility defect was found and corrected before push: profile_name now appends to EngineConfig so existing positional scheduler callers cannot be shifted. Focused regression suite, negative control, static checks, exact-head hosted pr_validate, and all required checks are green. No unresolved findings.

PR #3365, opened 2026-09-12
perf(qwen4): submit bounded short forwards per layer

pierre427:
## Why

Short Qwen4 decode forwards can leave device work waiting while Python constructs the next layer. On the composed qualification source, bounded per-layer submission improved median decode from 30.85 to 36.82 tok/s at 16K, 29.75 to 35.22 tok/s at 32K, and 26.73 to 31.32 tok/s at 64K. All 18 fixed-length ABBA cells passed output, memory, exclusivity, engagement, and thermal gates.

## Scope

- Submit each completed Qwen4 layer with `mx.async_eval` when `batch * sequence_length <= 64`.
- Add a process-level setter and `RAPID_MLX_QWEN4_EAGER_DISPATCH=0` opt-out.
- Add four CPU contract tests for defaults, opt-out, admission, and per-layer submission.

## Non-goals

- Server scheduling, compiled replay, cache reuse, and MTP admission.
- Applying eager submission to long prefill forwards.

## Acceptance

- Eligible short forwards submit once per completed layer.
- Ineligible or explicitly disabled forwards retain the existing path.
- Model outputs and cache behavior are unchanged.

## Verification

- `tests/test_qwen4_eager_dispatch.py`: 4 passed on CPU.
- Ruff check and format check pass for both changed files; `git diff --check` passes.
- The isolated PR head has CPU validation. The GPU results above are from composed source `f3e089b5`; executable AST checks match this PR's dispatch gate and model call path to that source.
- Composed-source semantic gates: 8/8 multi-turn pairs and 100/100 domain pairs matched exactly; all 216 responses ended naturally, with no candidate-specific qualitative regression.
- Bounded repository self-validation returned `MERGE-SAFE`: lint passed, and all four targeted failures reproduced on upstream `main`. Automated review, the full unit suite, and stress E2E were explicitly skipped; advisory diff coverage was unavailable because `pytest-cov` is not installed.

## Behaviour delta

Eligible Qwen4 forwards: deferred graph submission → bounded per-layer asynchronous submission by default. Set `RAPID_MLX_QWEN4_EAGER_DISPATCH=0` to retain the previous behavior.

## AI assistance disclosure

Codex generated and reviewed both changed Python files under the contributor's direction. The result was verified with the focused CPU suite, Ruff, diff checks, source-equivalence checks, and the separately identified composed-source GPU and semantic campaign.

## Checklist

- [x] Focused tests pass locally (4 tests)
- [x] Critical gate tests cover disabled, ineligible, and eligible production behavior
- [x] No documentation update is required for the bounded internal scheduling change
- [x] No breaking API change

### DIFF
--- a/tests/test_qwen4_eager_dispatch.py
+++ b/tests/test_qwen4_eager_dispatch.py
@@ -1,0 +1,116 @@
+# SPDX-License-Identifier: Apache-2.0
+"""CPU-only contracts for short-forward Qwen4 eager dispatch."""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+from vllm_mlx.models import qwen4_exp
+
+ROOT = Path(__file__).resolve().parents[1]
+
+
+def test_eager_dispatch_default_and_small_row_gate(monkeypatch):
+    previous = qwen4_exp.set_qwen4_eager_dispatch(True)
+    try:
+        monkeypatch.setattr(qwen4_exp, "_EAGER_DISPATCH_MAX_ROWS", 64)
+        assert qwen4_exp.qwen4_eager_dispatch_admitted(1, 1)
+        assert qwen4_exp.qwen4_eager_dispatch_admitted(4, 16)
+        assert not qwen4_exp.qwen4_eager_dispatch_admitted(1, 65)
+    finally:
+        qwen4_exp.set_qwen4_eager_dispatch(previous)
+
+
+def test_eager_dispatch_runtime_disable_and_validation():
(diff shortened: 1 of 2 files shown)


raullenchai:
I reran the focused tests on this head (4/4 passed) and the implementation is small, but I am holding merge on the stated performance benefit. The 16K/32K/64K numbers in the body compare composed source, while this isolated PR head has CPU-only validation; that does not attribute the reported speedup to per-layer `mx.async_eval` or establish its effect on supported Mac tiers. Since this flag is enabled by default, please provide same-checkpoint/same-host/same-runtime A/B with this flag as the only changed variable, including decode tok/s, output parity, peak memory, and thermal/noise controls. The code also needs a current-main port after the `rapid_mlx/` package rename. I will re-review once the isolated benefit and regression bounds are reproducible.

PR #3366, opened 2026-09-12
fix(mtp): charge resident lanes only for projected cache growth

pierre427:
## Why

Continuous self-MTP can reject a lane whose cache is already resident because admission charges the full cache again even though available memory already accounts for it. This makes warm resident lanes look more expensive than joining lanes.

## Scope

- Track whether the request owns a resident cache through scheduler and routing metadata.
- Charge resident lanes only for projected verification growth.
- Continue charging joining lanes for base cache plus projected growth.
- Preserve positional constructors and the existing single-lane refusal merged in #3350.

## Non-goals

- Expanding supported MTP models or cohort shapes.
- Replacing the conservative full-cache growth estimate with allocator telemetry.
- Claiming a throughput gain from the B1 Qwen4 ladder, which does not exercise multi-lane admission.

## Acceptance

- Resident and joining lanes receive distinct, explicit memory charges.
- Existing callers that construct metadata positionally retain their field mapping.
- #3350's single-lane guard and current routing contracts remain intact.

## Verification

- 187 focused CPU tests passed across admission, routing, scheduler, driver, generation-batch, qualification, and runtime suites.
- Ruff check and format check pass for all five changed files; `git diff --check` passes.
- Tests cover resident versus joining budgets and legacy positional construction.
- Bounded repository self-validation returned `MERGE-SAFE`: lint passed and 213 diff-aware targeted tests passed. Automated review, the full unit suite, and stress E2E were explicitly skipped; advisory diff coverage was unavailable because `pytest-cov` is not installed.

## Behaviour delta

Resident lane admission charge: base cache plus growth → projected growth only. Joining lane admission remains base cache plus projected growth.

## AI assistance disclosure

Codex generated and reviewed all changed Python files under the contributor's direction. The result was verified with 187 focused CPU tests, Ruff, diff checks, and explicit resident/joining accounting cases.

## Checklist

- [x] Focused tests pass locally (187 tests)
- [x] Critical scheduler tests cover resident, joining, and positional-constructor regressions
- [x] No documentation update is required for the internal accounting correction
- [x] No breaking API change

### DIFF
--- a/tests/test_batched_self_mtp_contract.py
+++ b/tests/test_batched_self_mtp_contract.py
@@ -301,7 +301,9 @@ def test_resident_metadata_preserves_preexisting_positional_lane_signature():
     assert lane.terminal is True
     assert lane.resident_cache is False
     assert lane.estimated_bytes(2) == 50
-    resident = LaneAdmission("resident", 10, 20, sampling, False, True, resident_cache=True)
+    resident = LaneAdmission(
+        "resident", 10, 20, sampling, False, True, resident_cache=True
+    )
     assert resident.sampling is sampling
     assert resident.estimated_bytes(2) == 40
 
--- a/tests/test_batched_self_mtp_contract.py
+++ b/tests/test_batched_self_mtp_contract.py
@@ -293,6 +293,19 @@ def test_lane_metadata_contracts_fail_closed(changes, message):
         LaneAdmission(**values)
 
 
+def test_resident_metadata_preserves_preexisting_positional_lane_signature():
+    sampling = SamplingContract(greedy=False)
+    lane = LaneAdmission("old-caller", 10, 20, sampling, False, True)
+    assert lane.sampling is sampling
+    assert lane.cache_ready is False
+    assert lane.terminal is True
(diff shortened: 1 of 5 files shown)


raullenchai:
Thanks for the careful accounting work. I reran the focused suites on this head (71 passed). I am holding merge for one production-path gap: in the current scheduler, `_metadata()` sets `cache_ready = not history and not any(_cache_offset(cache) for cache in caches)`, while this PR sets `resident_cache = any(_cache_offset(cache) for cache in caches)`. Therefore every scheduler-produced `resident_cache=True` lane has `cache_ready=False` and is rejected by `assess_lane()` before the new incremental-memory calculation can affect admission. The added test constructs `resident_cache=True` with the default `cache_ready=True`, but the live scheduler cannot produce that combination. The repository has also moved these modules to `rapid_mlx/`, so the change needs a current-main port. Could you provide an end-to-end reproduction that reaches the intended resident lane (or revise the lifecycle/ownership contract so it can), plus a before/after admission result? I will re-review once that path is demonstrated; no merge until the benefit is reachable.

PR #3367, opened 2026-09-12
feat(qwen4): load validated PLE rows from a bounded sidecar

pierre427:
## Why

Qwen4 PLE shard tensors consume substantial resident memory even though each forward selects only a small set of embedding rows. Rapid lacked a source-bound, bounded file-backed loader that preserves its native model and sharded ABI while refusing unsafe artifacts and inherited resources.

## Scope

- Add an opt-in q4/group32 PLE sidecar format and validator bound to the resolved checkpoint.
- Replace resident PLE tensors before weight evaluation and read only selected rows.
- Bound row dequantization chunks to 4096 rows and the optional packed-row LRU to 512 MiB.
- Refuse descriptor replacement, post-fork lookup, malformed geometry, invalid source binding, and non-finite quantization parameters.
- Close only readers owned by a failed strict load, including traceback-retained partial models.

## Non-goals

- A new n-gram hash, model implementation, worker pool, or MoE expert streaming path.
- Claiming an isolated PLE speedup; both composed benchmark arms used PLE offload.
- Exhaustive whole-file content hashing. Validation checks index binding, geometry, edge rows, and a random sample.

## Acceptance

- A valid sidecar removes resident PLE tensors and returns selected rows with the intended FP32-dequantize/one-BF16-round storage math.
- Cache charge never exceeds its configured bound.
- Forked children refuse inherited readers before taking inherited locks.
- Failed strict loads preserve the original exception and close their scoped readers.

## Verification

- `scripts/test_qwen4_ple_sidecar.py`: 32 passed on CPU, including real held-lock forks and traceback-retained load failures.
- Ruff check and format check pass for all five changed files; `git diff --check` passes.
- Composed-source multi-turn run: 1,302 lookups, 285,952 unique rows, and 65,484-byte maximum charge under a 65,536-byte limit.
- Composed-source 10x10 run: 15,622 lookups, 412,560 unique rows, and 67,108,860-byte maximum charge under a 67,108,864-byte limit.
- The isolated PR head has CPU validation. The composed GPU source was `f3e089b5`; both benchmark arms used PLE, so these are engagement and lifecycle receipts rather than an isolated speed comparison.
- Bounded repository self-validation returned `MERGE-SAFE`: lint passed, and all four targeted failures reproduced on upstream `main`. Automated review, the full unit suite, and stress E2E were explicitly skipped; advisory diff coverage was unavailable because `pytest-cov` is not installed.

## AI assistance disclosure

Codex generated and reviewed all changed Python files under the contributor's direction. The result was verified with 32 focused CPU tests, real fork negative controls, Ruff, diff checks, source-equivalence checks, and separately identified composed-source engagement receipts.

## Checklist

- [x] Focused tests pass locally (32 tests)
- [x] Critical load and resource tests include malformed-input, fork, descriptor-reuse, and failure-cleanup negative controls
- [x] No user-facing default changes; the loader is opt-in
- [x] No breaking API change

### DIFF
--- a/scripts/test_qwen4_ple_sidecar.py
+++ b/scripts/test_qwen4_ple_sidecar.py
@@ -1,14 +1,14 @@
 # SPDX-License-Identifier: Apache-2.0
 """Small synthetic sidecars and CPU-only Rapid load/lookup contracts."""
+# ruff: noqa: B023, SIM117
+
 from __future__ import annotations
 
-from contextlib import ExitStack
 import gc
 import hashlib
 import importlib.util
 import json
 import os
-from pathlib import Path
 import signal
 import struct
 import sys
@@ -17,6 +17,8 @@ import threading
 import time
 import unittest
 import weakref
+from contextlib import ExitStack
+from pathlib import Path
 from types import SimpleNamespace
 from unittest import mock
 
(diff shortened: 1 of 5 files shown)


raullenchai:
I reran the focused sidecar suite on this head (32/32 passed). The opt-in reader has useful validation/lifecycle contracts, but this is a large runtime change and the PR currently has no isolated user-benefit measurement: the composed benchmark used PLE offload in both arms. In particular, every selected-row lookup calls `mx.eval(indices)` and copies indices to NumPy before file reads, so the memory saving must be weighed against host synchronization and decode/TTFT cost. Please provide a same-checkpoint/same-host A/B with sidecar off versus on showing peak resident memory, TTFT, decode tok/s, output parity, and actual table bytes removed, plus the exact artifact/revision and invocation. This must also be ported from `vllm_mlx/` to current `rapid_mlx/` before exact-head CI. I am keeping it out of the merge queue until that tradeoff is demonstrated.

raullenchai:
I ported this branch onto current `main` and the current `rapid_mlx/` package namespace, preserving the original commits/author attribution. I also wired the 32 focused contracts into the Apple-Silicon CI lane; previously the script lived outside pytest discovery and had no workflow invocation.

Mac mini verification on head `79f4d38a1`:

- focused PLE sidecar contracts: 32/32 passed
- Ruff check + format check: passed
- `git diff --check`: passed
- existing Qwen4 vendored suite: 114/115 passed; the sole failure is an unchanged test using `mx.full_like`, which is absent from this host's installed MLX
- same-geometry synthetic lookup smoke: bit-exact between resident q4/group32 `ShardedEmbedding` and file-backed lookup

I also pinned the production geometry from `rapid-mlx/Qwen3.8-Flash-Next-4bit@dcf657e4acda2aae72da99cde65b6c491cd96998`. It has 320,001,536 PLE rows at 100 packed bytes/row, so this removes exactly **32,000,153,600 bytes** of resident PLE tensors. The public model index SHA-256 is `1ffe41e4484e7dc137900b12b14ae29cf92d1737e89fbe72c4fa2dd1b42ee7f1` and reports 104,681,488,408 bytes of model tensors.

The Mac mini has 32 GiB unified memory and about 34 GiB free disk. The resident arm needs roughly 104.7 GB of tensors; the offloaded arm still leaves roughly 72.7 GB before runtime state. Neither arm can run here, and source + 32.0 GB sidecar cannot be staged in the available disk space, so I will not present the synthetic hot-page-cache timings as the requested production A/B.

One additional delivery blocker remains: the immutable public model revision contains no `ple_rows.bin` or manifest, and this PR contains the reader but no artifact builder. The only sidecar receipt I could locate points to a contributor-local 32,000,153,600-byte file. Before queueing, this still needs:

1. separate-process, same-host off/on measurements on a host large enough for the resident baseline (practically 128 GB+): peak memory, TTFT, decode tok/s, and greedy output parity; and
2. either a reproducible sidecar builder or an immutable published sidecar + manifest, with exact hashes and invocation.

The branch is now current and reviewable, but I am keeping it out of the queue until those user-benefit and artifact-delivery gates are satisfied.

raullenchai:
Studio gate is complete on head `26d56f02f`. I found and fixed two checkpoint-binding bugs that prevented the public target artifact from loading: the raw HF source keys use `model.language_model...` before sanitize, and snapshot weights resolve through the same cache repository's sibling `blobs/` directory. The real checkpoint also exposed one trained RMSNorm anchor outlier; detection now uses 95% producer consensus plus a robust median, while the mixed-convention refusal remains covered.

Exact source:

- model: `rapid-mlx/Qwen3.8-Flash-Next-4bit@dcf657e4acda2aae72da99cde65b6c491cd96998`
- index SHA-256: `1ffe41e4484e7dc137900b12b14ae29cf92d1737e89fbe72c4fa2dd1b42ee7f1`
- indexed model tensors: 104,681,488,408 bytes
- sidecar: 128 shards, 320,001,536 rows, 160 dims, exactly **32,000,153,600 bytes**
- host: Apple Silicon Mac Studio, 256 GiB unified memory
- environment: `HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1`; separate process per arm; greedy sampler; 64 MiB bounded PLE row cache

Warm results after one warmup:

| Request | Resident PLE | File-backed PLE | Delta |
| --- | ---: | ---: | ---: |
| Peak MLX memory, short prompt | 102.978 GB | 70.975 GB | **-32.003 GB** |
| TTFT, short prompt / 64 output | 226.3 ms | 305.5 ms | +79.2 ms |
| Decode, short prompt | 25.97 tok/s | 25.54 tok/s | -1.65% |
| Peak MLX memory, 1024-token prompt | 106.025 GB | 74.022 GB | **-32.003 GB** |
| TTFT, 1024-token prompt / 32 output | 1.442 s | 1.487 s | +44.3 ms |
| Decode, 1024-token prompt | 27.31 tok/s | 26.59 tok/s | -2.64% |

Greedy token IDs were identical in every pair. Cold load/TTFT varied significantly in both directions because this shared host had background model load, so I am not using cold timing as a speed claim.

The branch now includes a reproducible builder:

```sh
python -m rapid_mlx.models.qwen4_ple_build \
  --model ~/.cache/huggingface/hub/models--rapid-mlx--Qwen3.8-Flash-Next-4bit/snapshots/dcf657e4acda2aae72da99cde65b6c491cd96998 \
  --output /Volumes/NVMe-4T/pr3367-artifacts/ple_rows.bin
```

It rebuilt all 32,000,153,600 bytes in 25.04 s. The output was byte-identical (`cmp`) to the independently generated A/B artifact and passed edge validation for all 128 shards plus 256 random rows.

Verification: 36/36 focused sidecar contracts, 13/13 Qwen4 norm contracts, Ruff check/format, and `git diff --check` all pass. The measured tradeoff is useful as an opt-in: about 32 GB less peak memory for roughly 1.6–2.6% lower warm decode throughput and 44–79 ms added warm TTFT in these requests.

raullenchai:
The validated branch is now carried by internal PR #3729 because this repository's managed queue explicitly rejects fork heads. Pierre's original commits and authorship are preserved there, along with the current-package port, production fixes, reproducible builder, and Studio A/B evidence. Closing this source PR removes Mergify's duplicate-head conflict so #3729 can run the exact-head Mac queue.

PR #3465, opened 2026-09-14
feat(rapid-mac): selectable UI language, with Simplified Chinese

LXD-8:
Based on `main` @ `3123b05b`.

## Why

The Desktop app compiles a `zh-Hans` table and ships `Contents/Resources/zh-Hans.lproj`, but **nothing can select it**. macOS chooses an app's language from the system setting, and Settings has no Language row — so a user on an English Mac sees English copy with no way to ask for anything else. On a Chinese Mac they get a *half*-Chinese app, which is the worse case.

The root cause repeated itself across the whole app: **`Text(someStringVariable)` renders verbatim and never consults the catalog**, and a literal returned from a `-> String` helper is never *extracted* into it either. The catalog starts from 80 translated keys.

## Scope

- **Settings → Appearance gains a Language picker**: Follow the system / English / 简体中文, persisted under `rapid.language.v1`, applied to a running process.
- **Two mechanisms, split by surface** (rationale in `docs/engineering/decisions/2026-09-15-desktop-language-switch.md`): SwiftUI copy rides the `\.locale` environment; Foundation/AppKit copy (`String(localized:)`, `NSMenuItem`, `NSAlert`, window titles) rides `BundleLanguageOverride`, a one-time swizzle of `Bundle.localizedString(forKey:value:table:)` scoped to `Bundle.main`.
- **Catalog: 1301 keys**, all but the one intentionally-empty key carrying a `zh-Hans` value.
- Three shapes fixed app-wide:
  - **Components typed `String`** (`SectionHeader`, `SettingsSection`, `SettingsRowLabel`, `EmptyState`, `InstructionTextEditor`, `SettingsView.Category.title`, the sidebar's `row(title:)`, the onboarding primitives) are retyped to `LocalizedStringKey`. Typing the parameter is also what makes the compiler *extract* the literals at their call sites.
  - **~180 copy helpers** across onboarding/Quickstart, the model picker, benchmarks, the media views, the About panel, diagnostics and failure diagnosis now resolve through `String(localized:)`, interpolation preserved so a sentence stays one translatable unit.
  - **Literals passed to `String` parameters**, and `Text("a " + "b")` concatenations — the latter builds a `String`, so `Text` picks its verbatim initialiser. Fragments were merged back into single literals.
- **`DevSnapshot.settingsShell`** — the second hand-built environment chain that renders the real `SettingsView` — needed the new observable too, and `SettingsEnvironmentInjectionTests` now covers the harness as well as the scene.

## The catalog

The 80 pre-existing `zh-Hans` values are kept as the authoritative wording (`采样`, `网页搜索`, `已置顶`, `就绪`, `空闲` …) rather than re-translated, and only two of them are edited:

- `License (EULA)` → `授权许可（EULA）` — full-width brackets, matching the 40 full-width `（）` elsewhere in the table.
- `image_input.unavailable.speculative_decode` — its sentence referenced `“Performance”`; the Settings category is now translated to 性能, so the reference reads `“性能”`.

Auditing the merged table normalised the remaining drifts in the new values toward the convention the pre-existing catalog already used: the ellipsis is a single `…` (not `……`), UI names are quoted with `“”` (not `「」`), and the allow/approve concept is unified on 允许/批准 so `授权` means only "grant"/"license".

The catalog is free of all of them: `……` = 0, `「」` = 0, half-width parens = 0, `授权` = 2 (only `Grant…` and `License (EULA)`). Checked and found already consistent: apparent term collisions that are actually distinct senses (session 会话 vs conversation 对话, store 存储 vs save 保存, revoke 撤销 vs cancel 取消, mode 模式 vs model 模型, connect 连接 vs connector 连接器), 27 English concepts cross-checked for alternate renderings, and the punctuation rules (no half-width comma/colon/period inside a Chinese sentence, no missing space before a Latin term).

Data still renders verbatim, by design: model aliases, HF repo ids, sizes, rates, percentages, counts, hardware strings, engine error text, and every accessibility identifier.

## Non-goals

- **No 繁體中文 / other locales.** Only `en` and `zh-Hans` `.lproj` tables are compiled, so only those are offered.
- **Formatting does not follow the picker everywhere.** `Locale.current`-driven date/number formatting still follows the system; `LanguageConfig.resolvedLocale` is the seam for callers that want to opt in.
- **`RelativeTimestamp`** (`"just now"`, `"5m ago"`) is still English-only — it deliberately avoids `RelativeDateTimeFormatter`, and changing it is its own change with its own tests.
- **The exported conversation document keeps English headings** (`## You`). It is a portable artifact with its own pinned format tests, not app chrome; localizing it is a product decision rather than a plumbing gap.
- **Data stays verbatim by design**: model aliases, HF repo ids, byte sizes, rates, percentages, counts, hardware strings, engine-reported error text, and every accessibility *identifier*. `DiagnosticsBundle`'s report body and `DraftPostFlow`'s browser AX matchers were treated as borderline and deliberately left alone.
- No new Settings **category**: the sidebar list is captured in 12 committed AX baselines, so a new row invalidates all 12 for one control.

## Acceptance

- Picking 简体中文 in Settings → Appearance repaints the Settings window (rail included), the main window and the onboarding wizard in Chinese, without a relaunch and without discarding a half-typed message in the composer.
- Picking English on a Chinese Mac produces the same English copy the app shipped before this change.
- Follow the system (the fresh-install default) leaves the macOS choice in charge.

## Verification

- `swift build` — clean.
- The localization suites this change adds or extends:
  `swift test --no-parallel --filter "LanguageConfigTests|LocalizationTests|SettingsEnvironmentInjectionTests|SettingsVisualFoundationTests|Step2ModelSelectionBehaviorTests"`
  — **101 tests in 5 suites passed, 0 failures** (~2.6s).
- Catalog checks over the committed `Localizable.xcstrings`: 1301 keys, 1300 with a non-empty `zh-Hans` value and 1 intentionally empty; exactly 2 pre-existing `zh-Hans` values differ from `main` (the two listed above); `……` = 0, `「」` = 0, half-width parens = 0, `授权` = 2.
- `scripts/check_rapid_mac_ax_identifiers.py --base-ref origin/main` — PASS.
- Net diff: 61 files, +14741/−1007.
- **Copy coverage**: the `.app` was rendered in both languages through `RAPID_DEV_SNAPSHOT_DIR` and the 221 PNGs compared pairwise — 220 differ, the single identical one being `menubar-mark.png`, a glyph with no text. The existing screenshots predate this head, so this comparison does not cover the current revision.
- **Known pre-existing caveats**, not caused by this change and reproducible on clean `main` @ `3123b05b`: `CurrentDateTimeContextTests` (5 issues) and one `ChatMessageActionsGoldenTests` popover timeout (1 issue) fail; and a full `swift test --no-parallel` does not complete on this machine because `LocalWorkspaceToolsTests` hangs partway. Plain parallel `swift test` additionally trips the hang watchdog; `--no-parallel` is the CI invocation.

## Behaviour delta

- **Settings gains a control**: Appearance now has a Language section; a fresh install defaults to Follow the system, so no existing user's language changes on upgrade.
- **The GUI harness now pins English** (`-AppleLanguages (en)`) unless a flow sets `RAPID_GUI_APP_LANGUAGE`; only `flow_localized_photo_hint` does. Before this change the comparison silently depended on the runner's macOS language — harmless while every string was a hardcoded literal, not once the copy is localizable.
- **`MarkdownCodeBlockView`** copy button: `复制`/`已复制` (hardcoded, so it read Chinese in *every* language including English) → catalog-backed `Copy`/`Copied`; `Preview`/`Code` now resolve too. `chat-depth.*` baselines move `复制` → `Copy` accordingly.
- **`AppearanceMode.displayName`** now resolves through `String(localized:)`; the theme picker's three rows had catalog entries that never took effect because the picker renders them with `Text(_: String)`.


LXD-8:
@raullenchai — could you take a look when you get a moment, and let me know what you think of the approach?

**What it does:** adds a Language row to Settings so the Desktop app's UI language can be picked (Simplified Chinese included), and fixes the root cause behind the half-translated app: `Text(someVariable)` never consults the catalog, so most copy stayed English-only. The catalog goes from 80 to 1301 keys; the 80 pre-existing values are byte-identical to `main` apart from two documented edits.

**State:** rebased onto `main` @ `65d4a677`. The last conflict (`SettingsToolsPanel.swift`, from #3538's new Computer Use tools) keeps main's cases and localizes them, with the 10 strings they introduce translated.

**Verification:** `swift build` clean; `scripts/check_rapid_mac_ax_identifiers.py --base-ref origin/main` PASS (51 files); the localization suites this PR adds pass (19 tests); the only test failures are the six pre-existing ones. I also ran `scripts/pr_validate` — fetch / description / supply-chain / review-vocabulary / test-env all pass; `full_unit` cannot complete on this machine (offline HuggingFace, no `torch`/`mlx-vlm`), which is unrelated to a Swift-and-docs-only diff.

No deadline on this — a review or an opinion on whether this is the right shape is all I'm after. Happy to change or split anything.

LXD-8:
This PR is behind main. Could you rebase onto the latest `main` when convenient? The i18n changes are complete and tests pass; just need sync with upstream.

LXD-8:
Closing as I'm no longer contributing to this project.

PR #3502, opened 2026-09-15
perf(mllm): skip singleton cache repack on the serialized lane

raullenchai:
## Why

On the serialized hybrid MLLM lane the structural B=1 policy already
guarantees every batch holds exactly one request, but the prefill path
still merged that request's regular cache leaves into batched wrappers.
Every subsequent decode step then had to detect the one-row batch cache,
extract a regular row, run the forward, and merge the row back — pure
execution-shape repacking with no semantic content.

## Scope

- Eligibility-gated singleton fast path in `MLLMBatchGenerator._process_prompts`:
  when the flag is `auto` (the default), the batch holds exactly one request,
  and every cache leaf is an exact regular `KVCache`/`ArraysCache` class, the
  prefill-written regular leaves are kept and the batch is marked
  `cache_layout="singleton_regular"`.
- Explicit `cache_layout` marker on `MLLMBatch`; downstream code branches on
  the marker, never on array rank.
- Detached terminal extraction for singleton-regular batches: fresh leaves
  built from explicit `mx.contiguous` copies, evaluated on the worker stream
  (upstream `ArraysCache.extract` returns lazy slices that alias live state).
- `MLLMBatch.extend` refuses a singleton-regular batch instead of promoting it
  mid-generation.
- Operator rollback threaded end to end: `--mllm-singleton-fastpath auto|off`
  (serve CLI) → `SchedulerConfig` → `MLLMSchedulerConfig` → generator. Typed
  config value, no environment variable in the hot path.
- Tracked qualification harness: `scripts/bench_qwen36_mllm_singleton.py` +
  21-case manifest `evals/prompts/qwen36_mllm_runtime.json` (20 media cases,
  4 per category, from Desktop snapshot images, plus one text-only fallback
  case). Methodology note and trimmed evidence JSON under `docs/`.

## Non-goals

- No model-name allowlist and no routing changes: text-only requests that
  qualify for the shared-weight native text engine are out of scope (their
  separate shipped evidence is not claimed here).
- No cache-class substitution; the legacy merge path is unchanged and remains
  the fallback for unknown/wrapped/subclassed/quantized leaves.
- No Desktop settings toggle: the rollback is an operator escape hatch.
- No new model families, MTP, processor ownership, or loader changes.

## Acceptance

- Contract suite (`tests/test_mllm_singleton_fastpath.py`, 23 tests): config
  validation at every layer, single qualified request skips `merge`,
  `_process_prompts([r1, r2])` always takes the merge path, unknown/wrapped/
  subclassed leaves fail closed, second request stays queued while a batch is
  active, `extend` refusal, and detached-extraction aliasing tests for both
  qualified leaf types.
- Paired exact-head Studio run (Mac Studio M3 Ultra 256 GB, Qwen3.6-35B-A3B-4bit
  rev `38740b847e4cb78f352aba30aa41c76e08e6eb46`):
  - **Exactness: 21/21 cases bit-identical** between `off` and `auto`
    (SHA-256 of full response text; 3 measured passes per case, all passes
    identical within a phase).
  - **Decode: median +35.4% generation tok/s** across the 21 cases
    (per-case ×1.27–×1.49). TTFT unchanged (median −1.6%, within noise).
  - Memory: indistinguishable (candidate active/cache bytes marginally lower,
    peak identical).
  - Lifecycle: mid-generation cancellation accepted, post-abort recovery
    byte-exact, queued concurrency serializes correctly, and a 50-iteration
    randomized abort/recovery soak passed with zero failures.
  - Evidence: `docs/benchmarks/results/2026-09-15-qwen36-mllm-singleton.json`;
    methodology: `docs/engineering/performance/2026-09-15-mllm-singleton-no-rebatch.md`.
- Server-level dogfood: `--mllm-singleton-fastpath off` boots and serves media
  requests (rollback path), and the default `auto` boot serves greedy, seeded
  (repeatable, stream==non-stream), and `top_k=1` media requests.

## Verification

- `pytest tests/test_mllm_singleton_fastpath.py tests/test_mllm_batch_generator.py
  tests/test_mllm_sampling_parity.py tests/test_lane_parity_contract.py` — 169 passed on exact head `50ee3e063` (including fail-closed import, metadata-copy, invalid-leaf, constructor, and merge-error coverage).
- MLLM regression batch (cache, continuous batching, logprobs, corrupt image,
  hybrid probe): 285 passed.
- `ruff check` / `ruff format --check` clean; repo mypy error budget: no growth.
- Harness commands (offline model resolution):
  `RAPID_MLX_MEDIA_ROOT=<repo> python -m scripts.bench_qwen36_mllm_singleton
  --model <snapshot> --pairs 3` and `--lifecycle --abort-iterations 50`.

## Behaviour delta

Eligible serialized multimodal requests skip singleton cache repacking:
default-on, decode ~+35% median on the measured Qwen3.6 media workload with
bit-identical outputs. Operators can restore the legacy merge/rebatch path
with `--mllm-singleton-fastpath off`. No visible behaviour change for
out-of-scope lanes; release notes should state the serialized-lane cache
repacking avoidance, not a generic model-kernel speedup.

## Review-boundary clarifications

- The rollback setting is wired from `SchedulerConfig` in `BatchedEngine._load_mllm` directly into `MLLMSchedulerConfig`, which validates it and passes it to `MLLMBatchGenerator`. The legacy `MLXMultimodalLM` compatibility wrapper is not constructed on this production batched path and is intentionally unchanged.
- `MLLMBatch.filter([])` is deliberately rejected for `singleton_regular`: both terminal completion in `_next` and cancellation in `remove` set `active_batch = None` when no rows remain and never call the filter with an empty keep-set. The identity filter `[0]` remains legal; any other call would indicate an invariant violation.
- This PR remains scope-locked to the serialized singleton cache lifecycle, its operator rollback, and its qualification harness. Unrelated cache, scheduler, or model-loader improvements are follow-ups.


raullenchai:
## Round-1 adversarial review — dispositions (commit 34718059a)

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1 | major | A/B ran with `enable_prefix_cache=False`; warm exact-prefix resumes feeding singleton leaves never qualified | **Fixed.** New `--apc on` mode runs both phases with the production-default APC and sends every case twice per measured pass; send-2 is the warm hit and exactness compares cold vs warm streams separately. Re-running on Studio. |
| 2 | major | Adopted singleton leaves keep left_padding/lengths that legacy `ArraysCache.merge` drops — rollback not faithful on warm resumes | **Fixed.** Adoption now calls `finalize()` on each leaf, matching merge output semantics (decode masks come out `None` in both modes). |
| 3 | minor | Harness never asserts the fast path actually engaged | **Fixed.** New `MLLMBatchStats.singleton_batches` counter; harness asserts auto > 0 and off == 0 (exit 2 otherwise), recorded as `fastpath_engaged`. |
| 4 | minor | Fast path silently inert on the dense lane (mlx-lm leaves need `allow_arrays_cache`) | **Won't-fix (documented).** Dense-lane exclusion is deliberate scope — the dense merge/extract lifecycle around plain mlx-lm leaves is unqualified. Documented in the eligibility contract and CLI help. |
| 5 | minor | KVCache extraction copies whole slab incl. zero padding | **Fixed.** Trim to `offset` exactly like `KVCache.extract`; slab-padded unit test added. |
| 6 | nit | `MLLMBatch.filter` silently no-ops on KVCache leaves for non-identity keep_idx | **Fixed.** Raises ValueError for a singleton-regular batch with keep_idx != [0]. |

Also in this commit: `SchedulerConfig.mllm_singleton_fastpath` moved to the appended-field block (positional-compat test), harness registered in `scripts/bench_metadata.py` with stamped JSON artifacts, and 4 new contract tests.

CI evidence refresh (methodology hash changed with the harness edits): cold A/B ×3 pairs, APC warm A/B ×2 pairs, lifecycle ×50 abort soak — will be re-recorded in `docs/benchmarks/results/` before merge-ready.

raullenchai:
Final scope-locked adversarial review is MERGE-SAFE on head `154d2daaa` (148 focused tests passed; real 35B cancellation/recovery/queued-concurrency/50-iteration abort soak passed).

Two non-blocking harness notes were dispositioned without expanding this runtime-performance PR:
- APC qualification deliberately fails closed unless the designated warm-prefix case is selected; `--cases` is diagnostic filtering, not a way to weaken the qualification corpus.
- `--summary-only` controls console output; the JSON artifact intentionally retains full per-sample evidence for audit/reproduction.

No product/runtime blocker remains. This PR stays unqueued until its parent #3496 lands and it is cleanly restacked on the resulting main.

PR #3505, opened 2026-09-15
perf(mllm): media-aware prior-turn boundary cache on the serialized lane

raullenchai:
**Why this PR:** on the serialized hybrid MLLM lane (structural B=1), multi-turn media conversations re-encode the same image and re-prefill the same rendered prefix on every turn — because the lane keeps no media-aware prior-turn state, turn-two TTFT pays the full image prefill again. This PR closes that gap with an exact, fail-closed boundary resume.

First milestone of the media prefix cache (design note: `docs/engineering/design/2026-09-15-mllm-media-prefix-cache.md`, commit 55e99dc17).

**What it does.** The generator derives an expanded-space boundary from the engine's rendered-space `prefix_boundary`, snapshots the boundary KV state (detached clone + strict token prefix + MRoPE delta) keyed by content-hashed media identity, and on a later strict token-prefix match resumes: replaces the live cache with a re-clone of the stored leaves, installs the saved delta, and forwards only the suffix with `pixel_values=None`.

**Fails closed, everywhere.** Flag off, text-only, no boundary, no structural rope-deltas consumption, unqualified leaves, unresolvable placeholder ids, placeholder outside the boundary, non-empty `extra_kwargs`, partial attention mask, short boundary (<16 tokens), uncloneable leaves, prefix mismatch, position-override wrapper without a usable LM-direct escape → the exact historical cold single forward. A mismatch is a clean counted miss — never a trim.

**Rollback:** `--mllm-media-prefix-cache off` (default `auto`), threaded through `SchedulerConfig` → `MLLMSchedulerConfig` → generator → CLI, positional-compat preserved.

**Default value (`auto`) — deliberately accepted trade-off:** the storing turn pays a bounded one-time snapshot cost (~+10–25 ms / ~8% TTFT on short first-eligible prompts, cost-neutral on long ones — design note §"Turn-one cost"); every resuming turn wins decisively (turn-two TTFT −54.8% median). The cost lands once per media identity and every subsequent turn in the conversation benefits, which matches this product's chat-shaped workload (screenshots are asked follow-up questions, not one-shot). The flag deliberately mirrors `mllm_singleton_fastpath`'s `auto`-default pattern; `off` restores byte-for-byte today's lane. Single-turn-only media workloads can pin `off` at zero code cost.

**Budget & stats:** media entries share the engine-wide prefix-cache byte ceiling (text exact cache footprint credited both directions); media-preferred eviction drops oldest media snapshots first, then oldest text entries as needed; stats surfaced via `get_stats` (`media_prefix_cache`).

**Evidence (Studio, Qwen3.6-35B-A3B-4bit hybrid lane, 20 conversations × 3 turns, pairs=2):** all hard gates pass (within-phase determinism, dual-phase semantic checkers, engagement); 44/60 turns SHA-exact cross-phase (residual is Metal GEMM batch-size regime ULP drift — same exposure as any warm-prefix resume on this lane, argmax preserved; documented in the design note); turn-two TTFT −54.8% median / elapsed −22.4%; storing turn cost-neutral on long prompts, +10–25 ms on short. Tracked harness: `scripts/bench_qwen36_media_prefix.py` + manifest `evals/prompts/qwen36_media_prefix_conversations.json`.

**Tests:** 93 contract tests in `tests/test_mllm_media_prefix_cache.py` (config layers, plan gates incl. structural rope detection + placeholder invariant + talker-wrapper bypass refusal, store/resume incl. stored-state equality, MRoPE transaction incl. real `nn.Module` sentinel restore, budget incl. media-over-ceiling eviction, detach, stats). mypy budget: no growth. Round-1 findings fixed in 0589886f0 (resume cloned the wrong cache; signature-only rope gate never opened on real mlx-vlm; missing placeholder-position invariant); round-2 in 1cd0761f2 (qwen3-vl wrapper position corruption → LM-direct bypass; 64-token recurrent-tile boundary alignment; harness gate semantics); round-3 in b34b7f468 (checker late binding; text-eviction freeze; probe robustness/memoization). Dispositions in PR comments.

Restacked onto main after #3502 landed. Exact-head review, validation, and managed Mac queue checks are required before merge.

**Harbor release refinement:** enrolled the MLX-only cache contract suite in the Apple coverage union; restored two budget tests hidden by a duplicate class name; added fail-closed branch coverage; removed one unreachable AST-walker return path and one redundant LM-direct gate. Local focused result: 269 passed. Predicted hosted Linux+Apple diff coverage from the previous run artifacts plus the enrolled suite: 100% (598\/598 changed executable lines).


raullenchai:
## Adversarial review disposition (rounds 1–2)

All findings from the two self-adversarial review rounds are dispositioned in `0589886f0` (round 1) and `1cd0761f2` (round 2).

**Round 1 — fixed in `0589886f0`:**
1. **Resume cloned the live cache, not the stored leaves** — a resume installed empty state. Fixed: resume re-clones `entry.leaves` (detached) and falls back to a counted cold miss if the clone fails.
2. **`rope_deltas` gate was source-text based** — a comment mentioning `rope_deltas` opened the gate. Fixed: structural probe requires `**kwargs` in the wrapper signature plus a real reference in source; unit-tested with the comment false-positive.
3. **Placeholder position invariant** — resume assumed placeholders sit below the boundary. Fixed: plan verifies every model placeholder token is inside the prefix before store, and refuses resume into a placeholder-bearing suffix; models without resolvable placeholder ids never store.

**Round 2 — fixed in `1cd0761f2`:**
4. **qwen3-vl-family wrappers corrupt split suffix positions** — `qwen3_vl`, `qwen3_vl_moe`, `qwen3_5`, `qwen3_5_moe`, `minimax_m3_vl` merge `InputEmbeddingsFeatures.to_dict()` (fresh 0-based `position_ids`/`rope_deltas`) into the LM call, so any suffix forward is positionally wrong (measured: 9-logit garbage). Fixed: structural probe + LM-direct bypass (`get_input_embeddings(..., pixel_values=None)` → `language_model(inputs_embeds=..., rope_deltas=entry.rope_delta)`); `qwen2_vl`/`qwen2_5_vl` verified safe (inputs_embeds only). Fail-closed when the bypass is unavailable.
5. **Boundary inside a recurrent-scan tile drifts** — GatedDeltaNet's chunked scan (C=64) tiles from each forward's start, so a split boundary re-tiles the tail and produces bf16-ULP state drift compounding to ~1.5 logits. Fixed: boundary floored to the 64-token grid (`_MEDIA_BOUNDARY_ALIGN_TOKENS`); proven bit-exact at a forced split with four resumes.
6. **Singleton/batch gate** — store/resume against a shared batch cache would corrupt neighbours; plan now silently requires the singleton turn (structural gate, not a modeling miss).
7. **`clear_prefix_cache` leaked media entries** — media entries are now cleared unconditionally, with and without a text cache.
8. **Budget asymmetry** — the text exact-cache budget now subtracts media entry bytes symmetrically; footprint walk takes the entry lock.
9. **Digest truncation** — media identity digests use the full stamped key / image hash, no truncation collisions.
10. **Resume placeholders / unresolvable ids** — covered by (3)'s resume-side gate (counted miss, never partial resume).

**Deliberately not changed:** the media plan does not gate on text-APC liveness. Media entries carry their own flag, byte accounting, and budget credit; the split is verified against the live request tokens at plan time. Tying it to the text cache would silently disable media resume whenever the text lane evicts — a worse failure than a counted miss, and a miss degrades to a cold single forward, never to wrong state.

**Determinism contract change (documented in the design note):** cross-phase byte equality is reported, not gated. The suffix forward runs its input-projection GEMMs at the suffix batch size while the cold lane runs them at full length; Metal's tiling regimes make per-row results coincide for some (suffix, full) pairs and differ by one bf16 ULP for others (measured: M=35 vs 163 exact, M=20 vs 276 divergent at layer 0). That ULP compounds to ~1–2 logits with identical argmax — the same exposure any warm-prefix resume has on this lane. Hard gates are within-phase determinism, semantic checkers in both phases, and engagement.

**Studio qualification** (20 conversations × 3 turns, pairs=2, serialized hybrid lane): all hard gates pass; 44/60 turns SHA-exact cross-phase; turn-two TTFT −54.8% median / elapsed −22.4%; storing turn cost-neutral on long prompts, +10–25 ms on short; 47 unit tests green; mypy at the 12-error file baseline.

raullenchai:
## Round-3 review disposition (`b34b7f468`)

1. **MAJOR — harness checker gate was vacuous (fixed).** The per-turn checker aggregates were generator expressions closed over the loop's `turn_index`, so every aggregate evaluated with the final turn: only the last turn was actually checked and `checkers_pass` could report True with earlier turns failing. Fixed by building the per-turn aggregates eagerly; conversation turn counts are now taken per conversation instead of from `conversations[0]`.
2. **Text eviction freeze (fixed).** `_enforce_exact_cache_budget` returned early when media bytes met the shared ceiling, permanently disabling text eviction and contradicting the budget docstring. Now: the zero/disabled config still returns early (previous behavior preserved), and a positive ceiling whose media credit drives the effective budget ≤ 0 evicts the text side down to its newest entry — mirroring the media store's own newest-entry guarantee. Unit-tested both directions.
3. **Probe robustness (fixed).** `inspect.getsource` can raise `SyntaxError` from a stale linecache; both probes now catch it and fail closed. Probes are memoized per model class (previously re-ran `getsource` up to ~4× per media prefill). The LM-direct bypass additionally refuses thinker/talker wrappers (qwen3_omni_moe shape): their wrapper-side bookkeeping would be skipped by a direct LM call, so those wrappers fail closed to the cold single forward. Unit-tested (talker refusal, memoization, source-read failure).
4. **`clear_prefix_cache` return (fixed).** A media-only generator that dropped real MiB now reports True; False only when nothing was held.
5. **Not changed:** the LM-direct availability probe remains a source-pattern heuristic verified against every currently vendored matching family (cohere_compass, falcon_ocr, minimax_m3_vl, qwen3_vl family all route equivalently; omni excluded by the talker guard). A fully structural signature probe is future work — noted in the design note's open questions.

Verified-clean by the reviewer: store/resume boundary alignment consistency, attention-mask drop soundness, budget lock ordering (no nesting, each side charges the other once), singleton gate, qwen2_vl/qwen2_5_vl safety on the default path, resume position algebra.

raullenchai:
## Codex round-4 (pr_validate) disposition (`63c98d2b8`, `b9967df90`)

Context: pr_validate run 2 came back with full_unit PASS (24,946 tests), stress/bench PASS, lint PASS, description PASS — the only FAIL was codex_review's 5 blocking findings. All dispositioned:

1. **Multi-image conversations ran single-image (fixed).** `_conversation_messages` attached only `conversation["images"][0]`; every declared image now rides on the first user turn, so the four two-image comparison cases actually exercise multi-image cache behavior.
2. **`--pairs < 2` vacuous gates (fixed).** Argument validation now rejects `--pairs < 2` outright — one measured pass cannot demonstrate within-phase determinism.
3. **Harness presented as the whole qualification gate (fixed).** Two parts: (a) the harness now enforces a **resume-turn no-regression gate** — a turn that actually served a media resume (auto `cached_tokens > 0`) must not exceed the baseline median TTFT by more than a 15% stall margin (exit 3); (b) the docstring and design note state the harness's exact scope (determinism, semantics, engagement, resume-turn latency) and point to where the remaining documented gates (bounded-memory soak, cancellation recovery, sampled fixed-seed A/B) get their evidence. The gate was initially mis-scoped to all warm turns — conv-09's storing turn tripped at +16.9% (+29 ms), which is the documented bounded snapshot cost, not a resume regression; the gate now keys on engagement evidence (`cached_tokens > 0`) and re-evaluates clean against the recorded pairs=3 run (18 resume turns, zero regressions, 47/60 SHA-exact cross-phase, turn-2 TTFT −56.4% median).
4. **Two oversized snapshots could coexist (fixed).** Coordinated budget: a media insertion whose own entries are exhausted now evicts TEXT exact entries oldest-first; both sides refuse admission of any single entry larger than the whole ceiling (the snapshot is discarded and the prefill continues cold); both sides share one eviction primitive. Combined worst case is bounded at one newest entry per store past the ceiling, documented in the docstrings.
5. **Digest omitted semantic inputs (fixed).** `_media_identity_digest` folds a semantic salt: the text APC's `semantic_extra_hash` (model/processor dependency set) when the manager exists, a structural fallback otherwise, plus the engine's vision pixel bounds. Process-fixed settings (template, model revision, adapter) die with the process and need no keying; a binding per-request pixel cap changes the expanded token stream and is caught by the strict token-prefix check.
6. **NIT — FIFO not LRU (fixed).** Verified hits promote their entry (`move_to_end`).

54 unit tests green (budget coordination, admission cap, talker-wrapper bypass refusal, memoization, source-read failure, digest salt all covered); mypy at the 12-error file baseline; ruff clean. pr_validate re-running on head `b9967df90`.

raullenchai:
## Codex round-5 (pr_validate run 3) disposition (`5389f4ee7`)

Run 3: full_unit PASS (24,948 tests), stress/bench PASS, lint PASS, coverage 88.9% — again only codex_review failed, with 5 new blocking findings. All dispositioned:

1. **conv-07 checker required only "n" (fixed).** The shortcut checker now uses the new `required_any` alternatives — `⌘N`, `Cmd+N`, or `Command+N` — so it actually identifies the shortcut while accepting legitimate renderings.
2. **Follow-up turns used bare "any" (partially fixed, rest documented).** Every open-ended turn now carries `min_words` (3 for detail turns, 5 for summaries), which rejects the degenerate outputs positional corruption actually produces (empty / single looping token). Fully semantic term checkers on "name a detail you have not mentioned yet" turns would just re-encode one build's specific wording — they would prove nothing beyond the determinism gate, which already pins those outputs. The dual-phase determinism gate plus the strengthened turn-0 checkers carry the semantic load.
3. **json_shape verified only key presence (fixed).** The checker now supports `list_keys` (non-empty list), `item_keys` (required keys on every object item — the `bars` case), and `required` value terms; all four json_shape checkers in the manifest were strengthened accordingly.
4. **Stats read raced the store (fixed).** `get_media_prefix_stats` snapshots under a dedicated media-store lock; every mutation site (insert, budget eviction, hit promotion, clear) and the text budget's media-credit read take the same guard. Lock order is media → text throughout. Covered by a concurrent store/stats smoke test (500 reads against a storing worker).
5. **MRoPE state leaked on exceptions (fixed).** The save transaction now restores prior model-global position state when the prefix forward, snapshot, or suffix forward raises, on both the store and resume paths. Unit-tested for both paths against a real prior state.

Checker validation: every strengthened checker passes the recorded ground-truth outputs from the pairs=3 qualification run (0 failures over 180 turn outputs) and fails the negative controls (wrong shortcut, missing values, empty lists, degenerate text).

57 unit tests green; mypy at the 12-error file baseline; ruff clean. pr_validate re-running on head `5389f4ee7`.

raullenchai:
### Round-6 disposition (codex review on `5389f4ee7`) — all 4 blocking findings fixed in `1dc7d776d`

**1. `enable_prefix_cache=False` not respected by the media plan** — fixed. `__init__` now records `self._prefix_cache_enabled = bool(enable_prefix_cache)` and the `_media_boundary_plan` gate adds `or not getattr(self, "_prefix_cache_enabled", True)`: the explicit text-APC opt-out covers every form of prefix reuse on this lane, media boundary resume included. Test: `test_plan_gated_by_prefix_cache_opt_out`.

**2. Rope probe false-positive on comment-only mentions** — fixed. `_media_model_supports_rope_kwarg` now matches *consumption* shapes (`kwargs.pop("rope_deltas"` / `kwargs.get("rope_deltas"` / `rope_deltas = kwargs...`) instead of any quoted mention. A `**kwargs`-accepting call that only mentions `"rope_deltas"` in a comment or log line fails closed — routing a resume through it would silently drop the installed delta. Tests: `test_comment_mention_with_kwargs_fails_closed` (new `_KwargsCommentRopeModel`), with the existing positive (`test_kwargs_popped_rope_deltas_opens_the_gate`) and no-kwargs (`test_comment_mention_without_kwargs_fails_closed`) cases unchanged and passing.

**3. `any` checker accepts degenerate output by word count** — fixed. With `min_words` set, the checker now also requires the vocabulary to spread: at least `max(2, min_words // 2)` distinct words, so `foo foo foo foo foo` fails. With no `min_words` floor the checker stays pass-through (any output qualifies, as before). Validated against the pairs=3 qualification record: **360/360 samples still pass** their checkers, and the degenerate output fails **40/40** `min_words` checkers. New test module: `tests/test_bench_qwen36_media_prefix_checkers.py` (5 tests).

**4. Two individually admissible entries can jointly exceed the ceiling** — fixed with a contract refinement. `_evict_text_exact_to_fit` gains `allow_empty`; when media entries hold room, both enforcement paths may empty the text store so the combined footprint fits the shared ceiling **strictly** (the survivor is the media newest, which the `_media_store` admission cap bounds at the ceiling). In the text-only regime PR#2's newest-survives contract is unchanged — `tests/test_mllm_batch_generator.py::test_exact_cache_byte_budget_evicts_oldest_entries_first` passes without modification (a lone oversize text entry still beats store/evict churn every turn). Tests updated: `test_text_budget_credits_media_bytes`, `test_text_budget_media_over_ceiling_empties_text`, `test_media_insertion_evicts_text_oldest_first`; design note §4 documents the strict shared-ceiling contract.

**Gates on `1dc7d776d`:** 147/147 tests green (media prefix + batch generator + checker contracts), ruff format/check clean, mypy holds the 12-error shrink-only baseline. No qualification numbers changed — the fixes touch gating/eviction/checker paths only; the pairs=3 Studio evidence (`qualification-r4.json`: 47/60 cross-phase SHA-exact, turn-2 TTFT −56.4%, engagement + resume-turn gates green) remains valid.

raullenchai:
### Round-7 disposition (codex review on `1dc7d776d`) — all blocking findings + nits fixed in `8f839305b`

**1. [BLOCKING] No qualified-family allowlist** — fixed. Added `_media_family_qualified()` + `_MEDIA_QUALIFIED_MODEL_TYPES` (`qwen3_5_moe` / `qwen3_5_moe_text` — the qualified Qwen3.6 hybrid family) and added the gate to `_media_boundary_plan`'s first fail-closed block. Structural gates prove a model *can* be split (rope-deltas consumption, cloneable leaves, singleton turn); only qualification proves the split is *correct* for it — the 64-token boundary alignment is tuned to this family's hybrid recurrent scan, and position/MRoPE handling differs per family. Every unqualified family (or model without a resolvable `model_type`) stays on the cold single forward until separately qualified on hardware, matching the design note's non-goal. Memoized per model-class pair like the other probes. Tests: `test_plan_gated_by_unqualified_family`, `test_family_probe_memoizes_per_model`; all engaging stubs carry the qualified `model_type`.

**2. [BLOCKING] Engagement gate can be colored by warmup** — fixed. `_run_phase` now snapshots `media_prefix_cache` counters immediately after warmup and reports **deltas over the measured passes only** (`hits`/`misses`/`stores`/`budget_evictions` subtracted; gauges `entries`/`bytes`/`budget_bytes` reported end-of-phase). `stores > 0 and hits > 0` can now only be satisfied by activity inside the measured passes, so the engagement and warm-regression gates cannot pass while every measured pass misses.

**3. [BLOCKING] `any` checker accepts fluent-but-unrelated output** — fixed, two layers. (a) The harness `any` checker now honors `required_any` (at least one alternative term list fully present). (b) The manifest's 40 open-ended turns are grounded: every "name one additional visible detail" turn anchors on that screen's visible elements (the same fixture content the turn-0 term checkers pin — buttons, chips, icons, hints), and every "summarize your previous two answers" turn anchors on prior-answer content terms. Validated against the pairs=3 qualification record in **both** phases: **360/360 samples pass**; negative controls (fluent unrelated prose, generic non-answers, degenerate `foo foo foo foo foo`) fail **40/40** grounded checkers with 0 leaks. Tests: `test_any_required_any_requires_an_alternative`, `test_any_required_any_alternative_must_match_completely`, `test_any_without_required_any_stays_word_floor_only`.

**4. [NIT] `clear_prefix_cache(reset_stats=True)` left media counters** — fixed. Media `hits`/`misses`/`stores`/`budget_evictions` now zero together with the text counters when `reset_stats=True`, and are preserved as lifetime totals when `reset_stats=False`. Test: `test_clear_reset_stats_zeroes_media_counters`.

**5. [NIT] Eviction order documented as a global LRU though it is media-preferred** — fixed by documenting the actual policy: `_media_enforce_budget`'s docstring now states the media-preferred order (own entries oldest-first, then text oldest-first once one media entry remains) and why — media snapshots are the far larger objects, and a true global LRU would need comparable recency tracking across two independent stores for no measured benefit.

**Gates on `8f839305b`:** 159 tests green (media prefix, batch generator, checker contracts, text-only gate), ruff clean, mypy holds the 12-error baseline. The grounding change touches manifest checkers + harness only; production qualification numbers on `qualification-r4.json` remain valid (re-validated: all 360 recorded samples pass the grounded checkers).

raullenchai:
### Round-8 disposition (codex review on `8f839305b`) — all 3 blocking findings fixed in `ff13e83d3`

All three findings target the harness; the core implementation passed review this round.

**1. [BLOCKING] `terms` checker ignores per-turn formatting constraints** — fixed. The checker now honors explicit `min_lines`/`max_lines`/`min_words`/`max_words` fields (on `terms`, `json_shape`, and `any`), and the manifest transcribes each prompt's own wording: "Two short lines, nothing else" → `min_lines: 2, max_lines: 2` (conv-01/02), "Three short lines" → `max_lines: 3`, "one per line" → `min_lines: 3, max_lines: 3`, "One short line" / "one short sentence" → `max_lines: 1` (t0/t1/t2 follow-ups), "Reply with only its exact label" → `max_words: 3` / `max_words: 6` / `max_words: 3` (conv-05/06/07), "at least 200 words" → `min_words: 200` (conv-13..16). A terse keyword blob no longer passes on substrings alone. Test: `test_structural_constraints_enforce_prompt_wording`.

**2. [BLOCKING] Substring matching admits wrong answers** — fixed. `_term_matches` now matches terms as contiguous **normalized token runs** (casefolded, edge punctuation stripped, slash/hyphen compounds split consistently on both sides), with a **negation guard on required terms**: a negator (`not`, `no`, `never`, …) directly before the run does not satisfy the requirement. `"bright side"` no longer matches `"right"`; `"not ready"` no longer matches `"ready"`; `"dark-themed"` / `"skip"/"next"` / `"qwen3.6-27b"` tokenize identically on term and text sides. `forbidden` stays negation-blind by design (claiming "not ready" for a chip that says otherwise is still wrong content) — documented in the checker. Negative-control tests added per the finding: `test_required_terms_match_on_token_boundaries`, `test_required_terms_reject_negated_mentions`, `test_compound_forms_tokenize_consistently`.

**3. [BLOCKING] `engine_for()` leaks on partial start** — fixed. `engine.start()` is wrapped: on any exception the partially-initialized engine is stopped (`contextlib.suppress(Exception)` around `stop()`) before re-raising, since the phase-level `finally` is not installed yet.

**Validation.** Against the pairs=3 qualification record, both phases: **360/360 samples pass** the hardened checkers (the token-boundary matcher required two tokenizer fixes — curly-quote edge stripping and slash/hyphen compound splitting — which are covered by the negative-control tests). Negative controls all fail as required: substring collision, negation collision, one-line keyword blob, wrong-content two-line output, unrelated prose on grounded turns, short answer to the 200-word ask, and 240-word fluent filler without the anchor terms — **0 leaks**. Gates: 163 tests green, ruff clean, mypy holds the 12-error baseline.

raullenchai:
### Round-9 disposition (codex review on `ff13e83d3`) — all 5 blocking findings fixed in `6717bed55`, re-qualified on hardware

**1. [BLOCKING] `json_shape` accepts prose around an all-null payload** — fixed. The checker now requires the **whole stripped response** to parse as one JSON object (fence handling kept); the first-`{` search is gone, so `qwen3.6 ready stop\n{"model":null,...}` fails. Test: `test_json_shape_requires_whole_response_to_parse`.

**2. [BLOCKING] Negation detected only at the immediately preceding token** — fixed. The guard scans the **three tokens** preceding a required phrase: "the chip is not currently ready" and "it does not say ready anywhere" now fail; four tokens away reads as a new clause and still passes. Test: `test_negation_window_catches_split_negation`.

**3. [BLOCKING] Engagement needs only one aggregate hit; missed resumes invisible** — fixed with a new **resume-coverage gate**. Turns that are supposed to resume (turn 2+; turn 1 is the documented store turn — turn-0 prompts fall below the boundary min-tokens floor, so turn 1 snapshots instead of resuming) must resume in ≥75% of their slots across the measured passes; every unexpected miss is reported in `resume_misses`, and a shortfall exits 4. This closes both halves of the finding: a declared minimum hit rate, and miss reporting.

**4. [BLOCKING] conv-10 `buttons` accepts `["garbage"]`** — fixed. `list_expect` requires exactly the three visible button labels (`new chat`, `search chats`, `start chatting`), in order, with the exact item count. Test: `test_json_shape_list_expect_validates_items_in_order`.

**5. [BLOCKING] conv-12 `bars` accepts one bar / null / wrong statuses** — fixed. `list_len: 2` plus `list_expect` validates each bar's `model` and `status` in order (`qwen3.6-27b`/`ready`, `fake-alias`/`idle`); nulls, a missing second bar, or a wrong status all fail. Test: `test_json_shape_list_len_and_bars_in_order`.

**Re-qualification (Studio, Qwen3.6-35B-A3B-4bit rev `38740b84…`, pairs=3, final shipped checkers — not an offline replay):** all gates green — within-phase determinism ✓, semantic checkers ✓ (all 360 measured samples across both phases pass the hardened checkers), engagement ✓ (auto stores/hits, baseline zero), zero warm-turn regressions ✓, **resume coverage 85% (17/20 turn-2 slots)** — the three misses (conv-04/07/18) are strict-prefix template mismatches, reported as clean misses in the payload — 47/60 cross-phase SHA-exact, turn-2 TTFT −56% class resume wins reproduced. Also note the checker hardening changed nothing about the model's measured behavior: outputs are byte-identical to the prior qualification record.

Gates on `6717bed55`: 168 tests green, ruff clean, mypy error set byte-identical to the branch base under a rebuilt pinned env (newer stubs surfaced 4 real `union-attr`/`not-callable` spots in the MRoPE transaction and LM-direct resume path — now explicitly cast-guarded with the contract documented).

raullenchai:
## Round-10 disposition (head `fcfd82bc2`)

Run 8 (`6717bed55`): full_unit 24973 PASS · stress matrix 4×3 PASS · lint PASS · diff coverage 89.1% (advisory). codex_review: 4 blocking — all harness/manifest, none in the core implementation. All four fixed:

**1. Per-sample resume gating (`scripts/bench_qwen36_media_prefix.py`)** — coverage and warm-regression no longer read `median_cached_tokens` slot medians:
- `_resume_gate_samples()` counts expected/resumed per measured sample (turn 2+ per pass); a slot where only one of two passes resumed is now a miss for the cold pass instead of being declared covered by a positive median.
- `_resume_regressions()` compares **each** resumed sample's (`cached_tokens > 0`) TTFT against the off-phase median TTFT for its (conversation, turn); a slot-level median mixing cold and resumed executions can no longer hide a slow resume, and cold samples are never compared. Store-turn samples stay excluded (documented bounded snapshot cost). Both helpers are pure and unit-tested, including the exact hide-by-median scenario from the finding.
- Margins lifted to module constants (`WARM_REGRESSION_MARGIN = 1.15`, `RESUME_COVERAGE_FLOOR = 0.75`).

**2. conv-03 t0** — the prompt says "Three short lines, nothing else": checker now pins `min_lines: 3` + `max_lines: 3` and requires the exact subtitle term `spinning up rapid-mlx` alongside `fake-alias`/`starting`.

**3. conv-06 t0** — requires the corner term `bottom right` in addition to `next` (a bare "Next" now fails), keeping the `max_words: 6` label-only shape.

**4. conv-14 t0** — requires the control/position labels the accessibility description claims to cover: `dark`, `skip`, `next`, `dots`, `checkmark`, `button` on top of `pick a model`. conv-15/16 t0 got the same grounded-label treatment preemptively (`search chats`/`empty`/`rapid-mlx desktop`; `qwen3.6-27b`/`stop`).

**Offline validation against the qualification record** (20 conversations × 3 turns × 3 passes, both phases = 360 samples): 360/360 pass with the tightened checkers — grounded, not overfit. Negative controls (200-word filler without labels, bare "Next", wrong corner, missing subtitle, one-line conv-03 answer) are all rejected. 19 checker tests + 34 implementation tests + media-gate tests green; ruff clean; mypy error multiset unchanged vs the branch base.

pr_validate run 9 launched on `fcfd82bc2`.

raullenchai:
## Round-11 disposition (head `a922ba194`)

Run 9 (`fcfd82bc2`) codex: 4 blocking + 1 nit. All addressed — two of them turned out to expose real harness-integrity issues beyond their literal asks:

**1. json_shape fences — resolved as documented behavior, not loophole.** The suggested fix (reject any fenced response) would reject the model's own deterministic correct output: under the old prompt, all recorded JSON answers are fenced pure JSON with zero prose, and a probe prompt that demanded a fence made conv-10 drop a visible button (wrong content). The fence-strip passes only when the fence wraps exactly one JSON object and nothing else — prose outside the fence still fails the whole-payload parse. The checker comment now states this contract explicitly.

**2. Exact-label turn (conv-05 t0)** — `max_words: 3` → `1`; a bare non-label answer can no longer pass.

**3. Image ordering (conv-18/19/20 t0)** — new `line_expect` checker capability: per-line required terms (line 1 = first bar, line 2 = second). Swapping the two answers now fails; missing lines fail; terms on the wrong line fail.

**4. Summary anchors (all 20 t2 turns)** — the per-answer group partition suggested by the finding proved **ungroundable**: qualification on a second host (M4 Pro) shows the detail answers themselves vary at temperature 0 (conv-04 t1 names the mascot icon vs the shortcut hint; conv-14 t1 flips likewise), so a partition by answer cannot be grounded stably. Replaced with `required_any_min: 2` over a pool spanning both answers' screen-grounded subjects — a summary citing one anchor or none fails, any faithful two-element summary passes on either host. Also: t2 `max_tokens` 32→64 (summaries were truncating mid-clause — conv-12/16 ended mid-word, which is itself why they couldn't cover answer 2).

**5. Nit (warmup warms the measured window) — fixed and it paid off.** `_run_phase` now clears the prefix cache after warmup; the measured passes are self-contained. That honesty check exposed a real gate flaw: `cached_tokens` is stamped by **both** the text exact-cache path and the media resume path, so the resume-coverage and warm-regression gates could credit a text warm hit as a media resume (pass-1/2 "resumes" on multi-image conversations were text-cache hits). The harness now snapshots the engine's media hit counter around each request (one request per turn on the serialized lane) and gates on that per-sample `media_hit` boolean.

**Re-qualification on MZR-3 (M4 Pro, pairs=3, final manifest — head `a922ba194`): all gates green.** Within-phase determinism ✓ · checkers 360/360 both phases ✓ · engagement (stores/hits candidate-only) ✓ · resume coverage **0.80** (48/60 turn-2 samples, honest media attribution; ≥0.75 floor) ✓ · zero warm-resume TTFT regressions ✓ · cross-phase SHA 47/60 (report-only, known GEMM-regime divergence). Known documented misses: conv-04/07/18 (strict-prefix template mismatches), conv-17/19/20 pass-0 only (under investigation — 3/60 samples, coverage still green). Checker tests 23 green, implementation tests green, ruff clean, mypy error multiset unchanged vs branch base.

pr_validate run 12 queued on `a922ba194` (waiting for a low-memory window on the Studio host).

raullenchai:
Round-13 disposition (all findings addressed at 098dd7c34):

**1. Per-pass cache clear (blocking)** — `_run_phase` now clears the prefix cache and resets stats before **each** measured pass and sums the per-pass counter deltas; gauges report end-of-phase state. This exposed a real finding: 6 of 20 conversations (conv-04/07/17/18/19/20) were structurally incapable of in-pass resume — the boundary aligns DOWN to the 64-token recurrent-scan tile grid, and their post-image question text left no aligned tile above the last image placeholder run (traced on the qualification host: conv-19 placeholders end at token 220, turn-1 prompt ends at 259 → raw boundary 244 → aligned 192 < 220 → fail-closed), so only the final turn could ever store and nothing was left to resume in-pass. The r13 "coverage 0.80" for these slots was cross-pass self-resume leakage. The manifest now sizes those conversations' follow-up prompts with neutral procedural wording so the first follow-up crosses the tile boundary and turn 2 resumes a same-pass store.

**2. Store-turn exclusion (blocking)** — `_resume_regressions` now skips `turn < 2`, consistently with `_resume_gate_samples` (turn 1 is the store turn; a turn-1 "resume" can only come from an earlier pass's same-prompt entry).

**3. Per-answer groups (blocking)** — `required_any_groups` reinstated (AND over groups, OR within a group) and all 20 summary-turn checkers rewritten to one group per prior answer, so a summary dropping an entire answer fails. Because detail answers vary at temperature 0 across hosts *and* between phases (the resume changes the prefill partition), each group's alternatives are anchored in recorded outputs — across hosts and manifest revisions — and the current records satisfy every group (validated offline against all recorded runs before commit).

**4. NIT (checker_pass_both zip)** — verified before changing: `checker_flags` is turn-major, so `zip(*...)` transposes to phase-major columns and each `all(pair)` is one phase's all-turns verdict. The gate semantics were already correct; a comment now documents the transposition.

Re-qualification after the fixes (same qualified model, 3 measured passes, per-pass clear): within-phase determinism ✓, checkers ✓, engaged ✓, resume coverage **60/60 = 1.0** ✓, zero warm-turn regressions ✓, cross-phase per-turn SHA equality 5/20 (reported, not gated). Unit tests, ruff check/format clean.

gatmgang-web (approved):

gatmgang-web (approved):

raullenchai:
Queue recovery: Mergify closed candidate #3537 at 20:32Z while its `desktop-tests` aggregator was still pending, reporting “manually updated,” although source head `cc5ab6517` and base `cf2585d0` did not change. Candidate #3537 subsequently completed with `tests`, `desktop-tests`, `version-bump-guard`, both GUI golden-flow matrix jobs, Apple MLX, and changed-lines coverage all green. Performing one controlled same-head `merge-ready-mac` label re-entry; no CI rerun and no code change.

raullenchai:
@mergifyio queue

raullenchai:
## Wrap-up: convergence record + post-merge qualification

**Merge**: merged at 21:06 via the queue. Final reviewed head `311fe1294` (round 34 could not run — the PR was already merged; the last full validation that ran to completion was round 33 on `c75708e58` + the suffix-install fix).

**Review rounds 28–33** (codex adversarial loop, all findings fixed and pushed):

| Round | Findings | Resolution |
|---|---|---|
| 28 | 2 blocking + 1 nit | digest carried through store + discard (one immutable identity per request); conv-10 t0 captures the on-screen shortcut so the summary ask is grounded; fixed `json_shape` field_terms tokenization (default `json.dumps` ASCII-escaped non-ASCII values so an on-screen term could never match its own field) |
| 29 | 2 blocking | unresolved byte budget now **disables** media admission (fail closed — a failed memory-limit discovery can no longer grow the store unbounded); enforcement sheds the media side entirely when the text store is un-introspectable and cannot be evicted |
| 30 | 1 blocking | pre-clone admission estimate (live footprint × clone capacity ratio) — an ineligible long-context request no longer transiently materializes the whole snapshot before being rejected |
| 31 | 1 blocking | pushed back with evidence: the digest/decode TOCTOU is closed upstream (`process_image_input` materializes every input kind once — URL download, base64 save, local path via O_NOFOLLOW fd-copy); invariant documented at the hash stamp + contract tests added |
| 32 | 2 blocking | the store plan carries the generation captured at plan time — a `clear_prefix_cache` landing between planning and the snapshot can no longer be repopulated; conv-09 summary checker split into one AND-group per fact |
| 33 | 1 blocking | the store path now runs `_media_mrope_install` before the suffix (same as resume) — the prefix's `_position_ids` can no longer override suffix positioning |

**Parallel-session commits**: the branch also carries ~15 commits from a parallel working session (cancellation cleanup, direct-suffix fail-closed, singleton structural proof, manifest tightening, etc.) — all merged and green before merge.

**Post-merge qualification on merged main** surfaced that two manifest checkers had been tightened past what the deterministic model produces (conv-10 t2 mascot, conv-16 t0 question icon) — `checkers_pass: false` in r18. Fixed in **#3541** with the prompts regrounded to the checkers (the checkers are correct about the screen).

**Final qualification (r20, MZR-3, Qwen3.6-35B-A3B-4bit, pairs=3, on #3541):**
- within-phase determinism PASS · 720/720 checker passes · media engagement PASS
- resume coverage **60/60** turn-2+ samples (100%, floor 75%) · zero warm-turn regressions
- resumed turn-2+ median TTFT **0.610 s → 0.225 s (−63.2%)** · store turn-1 median TTFT −3.3% (bounded snapshot cost neutral)
- cross-phase SHA-exact 35/60 turns (reported, not gated — the documented GEMM-batch-size regime drift, same exposure as text APC warm resume)

Artifacts: `/private/tmp/rapid-mlx-media-prefix-gate/qualification-r18..r20.json` (Studio).

PR #3506, opened 2026-09-15
fix(benchmark): route architecture-owned text models

raullenchai:
## Why

Rapid-MLX 0.14.2 advertised GLM-5.3-Flash as Community Benchmark-ready, but both benchmark entry points sent the checkpoint through the generic text loader. The production server correctly owns this architecture in its serving runtime, so benchmarking failed with `Model type glm5_next not supported` before a result could be submitted.

DeepSeek V4.1 has a related catalog-honesty problem: it requires a dedicated serial runtime that the benchmark does not yet adapt. Without an early guard, a user could download more than 200 GB before the generic loader failed.

## What changed

- Route architecture-owned text checkpoints through the same loader and scheduler used by Server/Desktop.
- Preserve the registered benchmark's exact token IDs, exact output length, sampling settings, counters, timeout, and cleanup contracts through a small adapter.
- Carry request-local `ignore_eos` through the serving stack, including the repetition guard, without changing ordinary chat behavior.
- Retire the obsolete GLM quantized forget-gate compatibility rewrite when the installed runtime has native support.
- Mark the dedicated DeepSeek V4.1 runtime unavailable before disk checks or downloads; keep DeepSeek V4 on its native text path.
- Apply the same behavior to the legacy `bench --submit` entry point.

## Design precedent

Mature serving systems select the model loader from checkpoint architecture and benchmark the actual serving backend rather than maintaining a second, narrower loader. This change adopts that architecture-owned routing pattern while retaining one benchmark protocol and one exact-token accounting contract.

## User walk-through

- `benchmark plan glm5.3-flash-4bit` reports ready and cached.
- A full offline registered run on Apple M3 Ultra / 256 GB completed all 10 measured rounds and produced exactly 128/512 output tokens per round.
- Median observed decode throughput was about 29.0 tok/s for pp512/tg128 and 27.2 tok/s for pp2048/tg512; peak active memory was 175,891 MiB.
- The result was accepted by Community Benchmark as submission `9c4fae1b-8c35-4886-99b3-329460faeeb6`.
- `benchmark plan deepseek-v4-flash-2bit` remains ready.
- `benchmark plan deepseek-v41-flash-reap-2bit` now explains that the dedicated runtime is not benchmarkable and that no model data will be downloaded.

## Verification

- 466 focused tests passed; 3 platform-gated tests deselected.
- Changed-line coverage is 100% (243/243 executable lines), including startup,
  timeout, cancellation, fallback, and cleanup failure paths.
- Ruff format/check passed for all changed files.
- `git diff --check` passed.
- Real cached 169 GB GLM checkpoint strict-load passed with the current native runtime.
- Real GLM exact-token adapter smoke passed.
- Full GLM registered benchmark and explicit upload passed end to end.

## Scope

This PR fixes Community Benchmark runtime selection and the minimum request plumbing required for exact registered measurements. It does not add a DeepSeek V4.1 benchmark adapter, change ordinary chat stop behavior, change benchmark workloads, or alter the website ingestion schema.


raullenchai:
Review disposition for the serving-context finding:

- The reported fallback risk does not occur. `get_model_max_context` explicitly probes `engine._model.args.max_position_embeddings` before tokenizer metadata, and `BatchedEngine` populates `_model` during startup.
- The real GLM-5.3-Flash run on this exact implementation recorded `context_length=1048576` and completed the 2048-token prompt bucket, so this is verified on the production wrapper rather than inferred.
- Added a focused regression asserting a serving wrapper's `_model` context takes precedence over its tokenizer fallback.

No loader change was made for this finding because passing the wrapper is the helper's intended public contract and is also how request admission calls it elsewhere.

raullenchai (commented):
Adversarial Codex self-review on exact head a3bf97dcb: no blocking findings. Scope stayed locked to benchmark runtime selection and the minimum exact-token/sampling/lifecycle plumbing. I specifically re-checked early DeepSeek V4.1 refusal before download, DeepSeek V4 native routing, GLM serving-loader parity, exact token accounting, total timeout semantics, cancellation/abort behavior, and cleanup on startup/stream/shutdown failures. Changed-line coverage is 100% (243/243); unrelated improvements remain out of scope.

raullenchai:
Diagnosed the queue-candidate failure on `7b804f9bca9b172c3cef66c080e195eaddce9093`: the only failed leaf was `l1-smoke (llama3-3b-4bit, 0)` during model prefetch, where Hugging Face returned HTTP 429 with an explicit 100-second retry-after. All code/static/unit/coverage/Apple jobs and the other four L1 smoke cells passed. This is external rate limiting, not a deterministic product failure. Per the one-diagnosed-rerun policy, I am rerunning the failed jobs once; no further rerun will be attempted.

raullenchai:
@mergifyio queue

PR #3508, opened 2026-09-15
fix(benchmark): support DeepSeek V4.1 serial runtime

raullenchai:
## Why

Community Benchmark lists `deepseek-v41-flash-reap-2bit`, but its text executor previously rejected the alias before download because the model uses Rapid's qualified DSpark K4 serial runtime rather than the generic text engine. Users with a supported 256 GiB Mac could serve the model but could not produce a comparable Community Benchmark result.

## What

- route the V4.1 product alias through the same pinned target and DSpark sidecar loader used by `serve`
- feed the registered synthetic workload as exact token IDs without a decode/re-tokenize round trip
- add benchmark-only ignore-EOS behavior while preserving normal server EOS termination
- adapt the synchronous serial stream token-by-token so TTFT and decode windows retain the registered protocol semantics
- record the active DSpark method, K4 block size, exact sidecar identity, and the runtime's qualified 8,192-token input limit
- keep unsupported sampling shapes fail-closed and retain the 224 GiB memory precheck before any download

## Scope

Allowed: Community Benchmark text execution and the existing V4.1 serial generation boundary.

Non-goals: changing V4.1 chat/server behavior, sampling support, continuous batching, prefix caching, model weights, sidecar weights, catalog recommendation policy, or support below 224 GiB.

Depends on #3506, which establishes architecture-owned benchmark routing and must land first.

## Design precedent

The adapter follows the established serving-engine boundary: benchmark orchestration owns workload identity and timing, while the model runtime owns loading, cache state, speculative acceptance/rollback, and token production. The external speculative component is pinned and represented in provenance rather than hidden behind the primary model identity. Unsupported execution shapes fail before model data is downloaded.

## Acceptance criteria

- exact registered prompt token IDs reach the V4.1 runtime unchanged
- greedy ignore-EOS runs produce exactly 128/512 output tokens
- normal OpenAI serving still terminates on EOS
- every MLX step stays on one executor thread
- target and sidecar revisions remain pinned to the production artifact contract
- completed and failed benchmark envelopes report DSpark rather than `none`
- planning rejects insufficient-memory hosts before download

## Verification

- `python -m pytest -q tests/test_cli_bench.py tests/test_community_benchmark_proto.py tests/test_community_benchmark_workspace.py tests/headless_mlx/test_community_bench_text_lane.py tests/test_deepseek_v41_artifacts.py` — 370 passed
- `ruff check` on all changed Python files — passed
- `git diff --check` — passed

A full 256 GiB real-model run is pending storage capacity: Studio's HF cache currently has 12 GiB free, below both the V4.0 benchmark download and the pinned V4.1 product artifact. No alternate cache or unpinned artifact was used.


raullenchai:
Adversarial self-review disposition for exact head `7785d6ce` (scope locked to making the registered DeepSeek V4.1 Community Benchmark executable and provenance-correct):

- Fixed benchmark `ignore_eos` handling so an EOS prediction still advances target/DSpark state instead of repeating a stale token; the normal server keeps EOS termination unchanged.
- Records DSpark K4 (`max_draft_tokens=5`), the pinned sidecar identity, and the qualified 8,192-token runtime limit rather than claiming theoretical checkpoint context.
- Fixed Linux/no-MLX isolation: contract tests now inject through lazy product-runtime seams and never import the large-model implementation. This removes the `mlx_lm.models.switch_layers` CI failure without weakening the real runtime path.
- Cleared all 11 new mypy errors in `local_runner.py`; ruff check/format and diff-check are clean.
- Focused evidence: 29 Linux headless contracts and 370 benchmark/artifact/workspace tests pass. Previous full suite on the pre-seam head passed 24,896 tests; exact-head full validation and CI are running now.

Non-goals remain unchanged: no default recommendation change, no ordinary Server sampling/EOS policy change, and no synthetic performance claim without the pinned real artifacts.

raullenchai:
@mergifyio queue

raullenchai:
@mergifyio queue mac-batch

raullenchai:
@mergifyio queue mac-batch

PR #3513, opened 2026-09-15
fix(engine): load heterogeneous DeepSeek V4 quantization

### LINKED ISSUE #3512: DeepSeek V4 2-bit DQ fails to load heterogeneous routed projections
## Problem

`deepseek-v4-flash-2bit` fails before inference while loading the catalogued checkpoint. Routed MoE gate and up projections can use different affine quantization group sizes, but the runtime always concatenates their packed tensors into one fused projection. Their packed dimensions differ, so model loading raises a shape mismatch.

A second homogeneous layer exposes an orphaned `up_proj.biases` parameter because the fused path currently combines weight/scales but not quantization biases.

## User impact

The catalog advertises the alias, but Server, CLI benchmark, and downstream GUI launch cannot load it.

## Done when

- heterogeneous gate/up quantization remains as separate projections without changing checkpoint values
- homogeneous layouts retain the fused fast path
- weight, scales, and quantization biases follow the same fusion decision
- focused synthetic tests cover both layouts
- the pinned full checkpoint loads and completes the registered community benchmark protocol

raullenchai:
## Why

The catalogued `deepseek-v4-flash-2bit` checkpoint could not load. Its routed MoE gate and up projections intentionally use different affine quantization group sizes in most layers, so their packed tensor dimensions cannot be concatenated into one fused projection. Homogeneous layers also carry quantization biases that must follow the same fusion decision as weights and scales.

Fixes #3512.

## What

- retain separate routed gate/up modules when their effective quantization layouts differ
- preserve the existing fused fast path when group size, bit width, and mode match
- fuse weight, scales, and quantization biases together on compatible layers
- add negative-control contracts for heterogeneous and homogeneous layouts

## Scope

Allowed: the vendored DeepSeek V4 routed-expert construction/sanitization and focused tests.

Non-goals: changing checkpoint values, quantization quality, aliases, benchmark protocol, scheduling, or other model architectures.

## Design precedent

Optimized MoE loaders treat projection fusion as conditional on the quantization backend and packed layout. This change applies that established boundary: optimization is retained only when the serialized representations are compatible; otherwise the architecture preserves the checkpoint’s independent projections.

## Acceptance criteria

- mixed group-size gate/up layers load without repacking or requantization
- compatible layers retain the fused implementation
- no orphaned quantization-bias parameter remains
- the full pinned 2-bit checkpoint completes the registered benchmark protocol

## Verification

- `pytest -q tests/test_deepseek_v4_vendored.py` — 25 passed
- `ruff check vllm_mlx/models/deepseek_v4.py tests/test_deepseek_v4_vendored.py` — passed
- `git diff --check` — passed
- full pinned `mlx-community/DeepSeek-V4-Flash-2bit-DQ@722bf559` load — passed
- registered community benchmark, M3 Ultra 256GB — completed all 10 measured rounds; 94,845 MiB peak active memory; median decode approximately 31.09 tok/s (pp512/tg128) and 29.42 tok/s (pp2048/tg512)


raullenchai:
Adversarial review disposition (scope locked to DeepSeek V4 checkpoint loading):

- merged projection overrides with global quantization defaults before comparing effective layouts
- upgraded heterogeneous coverage to quantize and load both routed projections, including distinct bias metadata
- upgraded homogeneous coverage to quantize/load the fused module and verify gate/up sentinel ordering for weight, scales, and biases
- made fusion fail clearly on unpaired metadata and atomic across later shape failures
- latest review: no blocking findings

Verification on current head: focused DeepSeek V4 suite 27/27 passed, lint and diff check clean. The full pinned 2-bit DQ checkpoint also loaded and completed all 10 registered benchmark rounds; the benchmark will be rerun from the landed revision before publication.

PR #3520, opened 2026-09-16
fix(speculative): bound GLM native MTP prefill memory

raullenchai:
## Why\n\nGLM-5.3 advertises a 1,048,576-token native window, but Rapid native MTP previously disabled bounded prefill and materialized a long prompt in one target forward. A roughly 32K-token request could therefore exhaust Metal memory on a 256 GB M3 Ultra even though the prompt was far below the model metadata limit.\n\n## What\n\n- Preserve target hidden states across bounded prefill chunks for native MTP.\n- Restore the original full prompt token IDs after the generation shell trims input_ids to the final chunk.\n- Pass the resolved model-profile prefill width into the serial native-MTP server.\n- Qualify GLM-5.3-Flash for a 1,024-token prefill chunk.\n- Tolerate iterator implementations that do not expose close().\n- Keep DFlash disabled; this PR is independent of the earlier DFlash experiment.\n\n## Design precedent\n\nLong-context admission must be based on bounded prefill and retained cache state, not checkpoint metadata alone. The implementation follows the existing chunked-prefill contract and keeps one request-owned speculative transaction. It does not add another scheduler or cache lifecycle.\n\n## Scope and non-goals\n\nThis PR fixes the reproducible native-MTP prefill OOM at the qualified 32K workload. It does not claim that the full advertised 1M context is usable, enable DFlash, add SSD offload, or change context-window UI/API advertising. Larger context tiers remain separately qualified work.\n\n## Verification\n\n- Restacked onto current main after the rapid_mlx package rename; the two fix commits are independent of the former parent branch.\n- python3.12 -m pytest -q tests/test_native_mtp.py: 35 passed.\n- Ruff check and format checks passed on every changed Python file.\n- git diff --check passed.\n- Real 256 GB M3 Ultra: two consecutive roughly 32K OpenAI-compatible requests returned HTTP 200 without Metal OOM.\n- The short-request fallback completed normally in 2.13 seconds.\n- Regression tests cover hidden-state accumulation, full-token restoration, resolved prefill propagation, and non-generator iterators.\n\n## Risk and rollback\n\nThe behavior change is limited to native MTP prefill. Removing the recommended_prefill_step_size profile entry and reverting these two commits restores the previous path.


raullenchai:
Post-restart real-model validation on a 256 GB M3 Ultra (exact head `c74e539be`):

- Target: `Vontra/GLM-5.3-Flash-MLX-4bit-MTP@76add2a341a1cd90ad0e86bb69839ea9c35827c6`
- Native MTP sidecar: `rapid-mlx/GLM-5.3-Flash-MTP-4bit@e9d...`
- Server: `--prefill-step-size 1024`, native MTP mode
- Focused gates: Rapid 3/3 passed; upstream GLM policy 1/1 passed; ruff clean
- 32K run 1: HTTP 200, 32,012 prompt tokens, 109.84 s, no Metal OOM
- Post-run short request: HTTP 200 in 2.13 s
- 32K run 2: HTTP 200, 32,008 prompt tokens, 107.96 s; end-of-context marker retained, no Metal OOM
- Final short request: HTTP 200 in 2.79 s, correct answer; service remained healthy
- Memory remained at pressure level 1 throughout; system-wide free/reclaimable reached a low of ~20% and recovered to 31% after the final request

This qualifies 32K repeated use on the tested host. It does not yet qualify 64K/128K/250K; those remain separate memory-scaling work.

PR #3534, opened 2026-09-17
refactor(mllm): deprecate the legacy mlx-vlm generation surface

raullenchai:
## Why

The native serialized MLLM lane is now the production generation path. The remaining direct `MLXMultimodalLM.generate` family and first-party benchmark calls kept a second generation runtime alive, preventing a safe dependency-retirement cycle and making benchmark behavior differ from serving.

## What

- Deprecates `generate`, `stream_generate`, `chat`, and `stream_chat` with one notice per model instance. Convenience image/video wrappers inherit the notice. A warning promoted to an exception does not consume the notice, because generation never started.
- Migrates image and video benchmarks onto `MLLMBatchGenerator`, with per-request sampling, terminal-token accounting, fail-loud idle handling, cleanup after drain failures, and response correlation by both UID and unique request ID.
- Loads benchmark models through `MLXMultimodalLM.load()`, preserving the same runtime patches, tokenizer repair, qualified fusions, and remote-code policy as serving.
- Shares the full stop-token collector between serving and benchmarks.
- Keeps deprecated benchmark call shapes working for one compatibility window.

## Scope

This PR changes the legacy generation deprecation surface and first-party image/video benchmark execution only. It does not remove model loading or prompt templating dependencies. The speculative server follow-up remains separate.

## Stop-token contract

`MLLMScheduler` derives stop tokens from the live underlying `model.config` plus the processor tokenizer. `MLXMultimodalLM.load()` has already augmented that tokenizer from `generation_config.json` before the scheduler is created. The wrapper's separate `load_config()` result is template metadata used by `apply_chat_template`; it is deliberately not a second stop-token source. Benchmarks now follow that exact serving contract.

## Integration review ledger

Harbor's scope-locked review fixed:

- Linux no-MLX test routing and Apple test-roster coverage;
- production-loader bypass in the benchmark path;
- incomplete benchmark EOS collection;
- structured image/video prompt rendering coverage;
- warnings-as-errors retry behavior;
- stale responses sharing a recycled UID;
- exact reuse of the loaded generator, processor, and template config.

One review suggestion to feed wrapper template config into stop-token collection was rejected with source evidence: production `MLLMScheduler` does not use that object, so doing so only in benchmarks would create the divergence this PR removes.

## Verification

- Focused MLLM/benchmark and CI-contract suites: 86 passed.
- Changed executable lines: 100% covered locally (144/144 at the integration-review checkpoint; later deltas are directly regression-tested).
- Exact-head GitHub Linux matrix, Apple Silicon lane, lint, type-check, and Desktop aggregation are required before queue entry.
- Local full-unit run: 25,673 passed. The remaining failures were the documented environment baseline for unavailable optional `torch`/`mflux`, plus two unrelated port-8000 collisions with another active Studio process; GitHub's isolated unit matrix is authoritative for those cases.


raullenchai:
## Harbor scope-locked review disposition (exact head `e6887c74c`)

**Intention:** retire the legacy mlx-vlm generation surface while moving first-party image/video benchmarks onto the same native serialized lane and production loading contract used by serving.

**Fixes made during integration review**
- Marked the seven MLX-bound tests correctly for Linux no-MLX CI and enrolled the file in the Apple test roster.
- Restored benchmark loading through `MLXMultimodalLM.load()`, preserving production runtime patches, tokenizer repair, qualified fusions, and remote-code policy.
- Extracted one stop-token collector shared by serving and benchmarks so plural/wrapper/config EOS sources cannot diverge.
- Added exact regression coverage for production loading, complete stop-token shapes, structured prompt rendering, and both image/video warmup + measured loops.

**Codex finding disposition**
The prior finding claimed `get_chat_template(processor, messages, add_generation_prompt=...)` does not render structured messages. Under the pinned runtime, that is the function's actual public signature and behavior. Replacing it with `processor.apply_chat_template` would regress processors that do not expose that method. The new test drives structured output through both benchmark paths and asserts the current helper receives and renders those messages.

**Evidence**
- Focused: 86 passed.
- Diff coverage: 100% (144/144 changed executable lines).
- Exact-head GitHub unit matrix and Apple Silicon lane: green on the preceding head; the new head changes tests only and is re-running.
- Local full-unit baseline: 25,673 passed; 17 known missing-`torch`/`mflux` environment failures plus 2 unrelated port-8000 collisions from a concurrently running server.

PR #3538, opened 2026-09-17
feat(mac): let Personal Intelligence work with local files and code

raullenchai:
## Summary

Personal Intelligence can now complete bounded work on the user's Mac from the existing chat experience. This PR adds five Desktop-owned tools for searching and reading local text, writing a file, moving a file to Trash, and running a small development command. It deliberately does not add Computer Use, arbitrary shell access, folder deletion, or permanent deletion.

The work is optimized and physically dogfooded with `minicpm5-2b-4bit`, while using the same qualified Personal Intelligence contract for every supported model/build.

## Necessity

Personal Intelligence could answer web/weather questions but could not complete the local tasks users expect from an on-device assistant. In mini dogfood it either claimed it had no filesystem access or stopped after writing code without compiling and running it. The feature is not useful as a default-on experience until those common actions work reliably with explicit consent and bounded permissions.

## Scope

- Desktop chat tools for bounded local search, read, write, Trash, and development-command execution.
- Server harness routing, staging, small-model argument normalization, and one bounded malformed-call retry for those Desktop tools.
- Consent UI, tool settings presentation, regression tests, and physical MiniCPM5-2B mini dogfood.

## Non-goals

- Computer Use, GUI automation, general shell access, background autonomy, permanent deletion, folder deletion, or access outside the user's home directory.
- Widening the existing per-model Personal Intelligence qualification gate.
- Qualifying new model/build pairings in this PR.

## AI assistance disclosure

Codex wrote and reviewed the changed Swift and Python implementation and tests under maintainer direction. The maintainer can explain every non-generated line on demand. Verification included focused Swift tests, the complete agent-server test file, Ruff, diff/secret review, a release-mode Desktop build, and physical GUI dogfood on the M2 Pro mini with real approval interactions and filesystem/output checks.

## User experience

Turn on the sparkle-shaped **Personal Intelligence** toggle beside the chat composer, then ask naturally. Examples:

- `Search the folder /Users/me/Documents for Project Orchid and tell me when the meeting is.`
- `Read /Users/me/Documents/notes.txt and summarize it.`
- `Draft a proposal and save it to /Users/me/Documents/proposal.md.`
- `Write a small C program that prints RAPID_C_OK, compile it, and run it.`
- `Move /Users/me/Downloads/old-report.txt to Trash.`

When the user does not name a destination for generated work, Rapid uses `~/Rapid Workspace`. Local requests are routed only to local tools; a request to search local files is never sent to web search.

Before an action touches local data, Desktop shows the exact JSON arguments—including the path, complete file content, command, and argv. Search/read access can be allowed once or for the current app session. Write, run, and Trash actions always require **Allow once**.

## What ships

- `local_search`: case-insensitive filename and UTF-8 content search, bounded to 500 files and 20 matches, with short snippets.
- `local_read`: reads one regular UTF-8 text file, capped at 512 KB.
- `local_write`: atomically writes one UTF-8 file, capped at 512 KB; replacing an existing file requires model-authored `overwrite=true` and user approval.
- `local_trash`: moves one regular file to macOS Trash; folders are never accepted and the file remains recoverable.
- `local_run`: executes an approved executable plus argv without a shell. It supports `clang`, `cc`, `gcc`, `go`, `swift`, `python3`, `node`, `ruby`, plus an executable inside the user's home directory. The child runs under a macOS sandbox with networking and Mach service lookup denied and file access limited to the approved working directory (plus system toolchain reads and temporary compiler storage). Runtime is capped at 30 seconds and stdout/stderr at 64 KB each.

The small-model harness also:

- stages one local tool at a time, in the order search/read/write/run/Trash;
- gives a malformed pinned tool turn one bounded format retry;
- normalizes omitted destinations to `~/Rapid Workspace`;
- converts common model-authored shell recipes such as `cd ... && gcc ... && ./app` into separate no-shell argv actions;
- supplies a deterministic compiler output name when a small model omits `-o`;
- permits a second `local_run` turn so compile-then-run can complete.

## Safety and privacy contract

- All paths must remain under the current user's home directory.
- `.ssh`, `.gnupg`, and `Library/Keychains` are unavailable.
- Hidden files, package descendants, symlinks, non-UTF-8 content, oversized files, and directory deletion are excluded where applicable.
- Invalid, out-of-home, or non-executable actions fail before an approval sheet is shown.
- Commands do not run through a shell, receive a fixed minimal `PATH` and environment, and execute in a macOS sandbox that denies networking, Mach service lookup, reads elsewhere in the home folder, and writes outside the approved working/temp directories.
- Common shell launch paths, `osascript`, and `open` are denied inside the command sandbox.
- Mutation approvals never persist. Closing an approval sheet denies the action.
- Tool outputs are treated as untrusted data by the agent prompt.
- No Computer Use, mouse/keyboard automation, background autonomy, or unrestricted terminal is added.

## Mini dogfood

Physical host: M2 Pro Mac mini, 32 GB unified memory. App: locally built `0.14.2` from this branch. Model: `minicpm5-2b-4bit` (`openbmb/MiniCPM5-2B-MLX`, approximately 3.8 GB resident).

Passed through the actual GUI and approval sheets:

1. Local search found a seeded Project Orchid note and returned Thursday, 3:30 PM, Redwood, Maya Chen, and the Cedar backup without invoking web search.
2. Proposal generation wrote the exact reviewed Markdown file to disk.
3. C workflow wrote a complete `main.c`, normalized the compile action to `gcc main.c -o main`, ran the resulting local executable, and returned the verified stdout `RAPID_C_OK`.
4. Trash workflow showed only **Don't allow** and **Allow once**, moved exactly the requested file, and confirmed it was recoverable from Trash.
5. A natural `Move /path/file to Trash` phrasing was regression-tested after dogfood found that the earlier router only recognized wording containing `the file`.

## Model rollout

This PR does not widen the existing qualification gate. The local tools become available to the exact model/build pairings already qualified for Personal Intelligence, including MiniCPM5-2B, Qwen3.5 4B/9B qualified builds, Qwen3.8 27B qualified builds, Qwen3.6 qualified builds, Bonsai 27B 2-bit, Ling 3.0 Tiny, and LFM2.5 1.2B.

Follow-up qualification is still required before enabling the exact public builds that are present in runtime profiles but absent from the qualification table:

- `qwen3-coder-30b-4bit`
- `gpt-oss-20b`
- the Rapid alias/build for Qwen3.6 35B-A3B 4-bit, if it does not resolve to the already-qualified upstream artifact pairing

That follow-up should run the same local search, write, compile/run, Trash-denial, malformed-arguments, and final-synthesis scenarios. It should update the qualification receipt/table rather than weakening the gate.

## Verification

- `swift test --filter 'BuiltinToolsTests|LocalWorkspaceToolsTests'` — 34 passed
- `uv run --extra dev pytest tests/test_agent_server.py -q` — 108 passed
- `uv run --extra dev ruff check rapid_mlx/agent_runtime/server.py tests/test_agent_server.py` — passed
- `git diff --check` — passed
- Release-mode Swift build and ad-hoc signed app launch on the mini — passed

## Website docs notes

The **User experience**, **What ships**, and **Safety and privacy contract** sections are intended to be lifted into the Personal Intelligence documentation. Public docs should lead with natural-language examples, explain that generated files default to `~/Rapid Workspace`, distinguish session grants from per-action mutation approval, and explicitly state that this is conversational local work rather than Computer Use or a general-purpose shell.


raullenchai:
Codex review round 1 disposition (5 blocking findings, all fixed in fe385805):

1. General-purpose interpreters/home executables now run under a macOS sandbox: network and Mach lookup denied, home reads restricted to the approved working directory, writes restricted to working/temp directories, and common shell/osascript/open launches denied.
2. local_search now skips symlink/protected/out-of-scope descendants and re-validates every file before reading.
3. stdout/stderr now drain through independent in-memory buffers capped at 64 KB each; no unbounded output files are created.
4. timeout now terminates, grants 500 ms, then SIGKILLs the private process group with a bounded final wait.
5. non-overwrite writes now use Foundation withoutOverwriting (exclusive-create semantics); overwrite=true remains atomic replacement.

Regression tests cover each contract. Negative controls reproduced locally: changing the home read deny to allow made commandCannotReadOutsideWorkingDirectory expose RAPID_SANDBOX_SECRET and fail; raising the output buffer cap to 1 MB made commandOutputIsBounded capture 200,030 bytes and fail.

raullenchai:
Additional physical-dogfood fix in 65213843: the first sandbox profile correctly blocked external writes but also blocked clang temporary files, so the mini compile retried and exhausted the action budget. Rapid now creates an ephemeral compiler temp directory inside the approved workspace, points TMPDIR there, includes it in the sandbox, and removes it after execution. A real gcc compile is now a named Swift regression test; focused total is 34 passing.

raullenchai:
Physical mini dogfood found one additional cross-machine compiler issue and fixed it in 05def1c0: on the Xcode-equipped mini, the `gcc` shim tries to launch `sh -c xcodebuild`, which the sandbox correctly denies. Rapid now resolves gcc/cc/clang directly to the installed Xcode or CommandLineTools clang binary and supplies the matching macOS SDK as `-isysroot`; no shell exception was added. The same real C compile regression now passes on both the local Mac and the M2 Pro mini.

raullenchai:
Harbor final scope-locked adversarial review on exact head d997e90bb: converged with no remaining blocker. The final fix closes four approval-boundary escapes found during review: protected Keychains descendants during broad search, mutable package-manager executable replacement, detached fork/setsid children (including Swift), and compiler in-process plugin/response/config loading. Evidence: LocalWorkspaceToolsTests 36/36, focused Python agent-server regression, and git diff --check pass; exact-head CI is running.

raullenchai:
Diagnosed exact-head Mac failure from run 35306393181 (job 105479387926), not rerun blindly. macOS 26.6.2 hung in the sandboxed-Python denied-child path because SIGKILL was followed by an unbounded blocking waitpid. Head now bounds foreground reaping to 1s and delegates only exceptional slow kernel teardown to one eventual background reaper; sandbox denial, process-group kill, descendant kill, and timeout semantics stay intact. Focused LocalWorkspaceTools suite: 36/36 passed in 5.64s, including forced timeout, denied unrelated exec, detached Python child, and detached Swift child. Formatting/diff checks clean.

raullenchai:
Adversarial follow-up on the hang fix: replaced one detached waiter thread per exceptional timeout with a single serial utility reaper queue, so repeated slow kernel teardowns cannot create unbounded threads. The foreground timeout remains bounded and all sandbox/kill semantics remain unchanged. Focused forced-timeout and denied-exec tests pass; diff check clean. New exact head is the authoritative one.

raullenchai:
The repeated macOS 26.6 hang narrowed the blocking point further: the test never reached the bounded `waitpid` fallback. The denied Python fork/exec was still unwinding while the parent repeatedly called libproc to discover descendants; on this OS that child enumeration can block in the kernel and therefore bypass the action deadline.

New exact head skips descendant enumeration only for sandbox profiles that already carry `deny process-fork` (interpreters and Swift). Compiler/Go profiles, which intentionally permit trusted helpers, retain the full descendant tracking and termination path. The sandbox denial and process-group kill remain unchanged.

Local exact-head evidence:
- `swift test --no-parallel --filter LocalWorkspaceToolsTests`: 36/36 pass in 5.75s
- denied unrelated executable: 0.074s
- detached Python child: pass
- detached Swift child: pass
- forced timeout: 1.56s
- `git diff --check`: clean

raullenchai:
Third exact-head diagnosis from run 35309364879: skipping libproc enumeration was insufficient; the same denied fork/exec also wedges `waitpid(..., WNOHANG)` on macOS 26.6 despite its nominally non-blocking contract. The watchdog again stopped on the exact same test.

The command runner now follows the repository's established process-lifecycle precedent: profiles that forbid forks observe their leader with `DispatchSourceProcess(.exit)` and only call blocking `waitpid` after the kernel has delivered `NOTE_EXIT`. A deadline sends TERM/KILL and hands an exceptionally slow teardown to the bounded background reaper; compiler/Go profiles retain descendant discovery because those profiles intentionally permit trusted helpers.

This preserves the security boundary while removing both kernel probes observed to wedge on the provider OS.

Local exact-head evidence:
- `swift test --no-parallel --filter LocalWorkspaceToolsTests`: 36/36, 5.76s
- denied unrelated executable: 0.083s
- short successful commands prove the process-source arm/exit race closes
- forced timeout: 1.52s
- detached Python/Swift child cases pass
- compile clean; `git diff --check` clean

raullenchai:
Exact-head fix pushed: `477c676d5`.

The third Manzanita hang artifact showed 91% free memory and the suite advancing normally until `approved interpreter cannot launch an unrelated executable`; the macOS 26.6.2 sandbox-denied fork/exec path then held the in-thread process observer until the 15-minute suite watchdog. This was not a slow suite or capacity failure.

The command runner now gives one detached waiter sole ownership of the child reap and makes the caller wait only on a bounded semaphore. If kernel teardown stalls, the approved action still reaches its hard deadline, signals the process group, and returns without blocking Desktop or the test runner. Compiler/Go descendant tracking remains unchanged.

Local exact-head evidence:
- `swift test --no-parallel --filter LocalWorkspaceToolsTests`: 36/36 passed (including unrelated-exec denial, forced timeout, and detached Python/Swift child containment)
- `swift build`: passed with no warning from the changed source
- `git diff --check`: passed

No CI rerun was requested; the push created a fresh exact-head run.

raullenchai:
Codex adversarial review on the stacked follow-up PR #3550 (base = main + this branch) reviewed this branch's sandbox code and reported two findings that belong here rather than in #3550:

1. `LocalWorkspaceTools.swift` sandbox profile — starts with `(allow default)` and never restricts `signal`, so approved code can signal unrelated same-user processes despite the workspace confinement. Suggested: deny signaling by default; allow only the sandboxed process group.
2. `terminateProcesses` / descendant tracking — bare PIDs are recorded and later signaled without re-verifying they still belong to the launched tree, so a reused PID could get an unrelated process killed. Suggested: retain a verifiable identity (start time / ancestry) and re-check before each signal, or signal only the owned process group.

Full text: `/tmp/pr_validate/pr-3550/run-20260918T*/codex-review.md` on Studio. Not blocking #3550; noting here so they are tracked with the code they concern.

raullenchai:
Codex round 5 on the stacked #3550 (base = main + this branch, head `0a89c3d5f`) reported two more findings whose lines blame to this branch's `046489a83`, so tracking them here as well:

1. `LocalWorkspaceTools.swift:48` `PinnedPath` — opens the unverified path with a blocking `O_RDONLY` before checking `S_IFREG`, so a `local_read` / `local_trash` aimed at a FIFO can block the main actor before the approval sheet appears. Suggested: `fstatat(..., AT_SYMLINK_NOFOLLOW)` first and/or `O_NONBLOCK`, then reject every non-regular file.
2. `LocalWorkspaceTools.swift:1137` `isImmutableSystemExecutable` — treats every executable beneath a root-owned developer-directory root as immutable, so a writable nested toolchain binary can change after approval and run without the digest check applied to package-manager binaries. Suggested: verify ownership / write bits along the path, or digest and stage every non-SIP-protected executable.

Full text: `/tmp/pr_validate/pr-3550/run-20260918T090335Z-*/codex-review.md` on Studio. Not blocking #3550.

raullenchai:
Root cause of the desktop `build` 15-minute hang, found while stacking #3550 on this branch:

- `LocalWorkspaceToolsTests.runApproved` waited with `while store.pendingRequest == nil { await Task.yield() }`. When `LocalWorkspaceTools.run` answers **before** requesting approval (any preflight refusal), that loop never ends, so the whole suite sat idle until the watchdog killed it and the sample only reached `swift-test`.
- On the desktop runner the early answer is `local_run could not find an installed swift executable`: `installedDeveloperDirectories()` rejects its Xcode because it is not root-owned, so "approved Swift scripts cannot fork detached background work" can never exercise the sandbox there. Local Xcode 26.6 (root-owned) passes the same test in 2 s.

#3550 carries both fixes (`052e2916a` bounded helper that returns the early result or fails after 30 s with a named tool; `24e00ea75` exposes `LocalWorkspaceTools.swiftScriptToolchainIsAvailable()` and gates that one test with `.enabled(if:)`). If this PR merges first, the same two commits apply cleanly here.

raullenchai:
New exact head `a5f6df75a` after the macOS 26.6.2 evidence narrowed the remaining block to timeout-path pipe teardown. The denied descendant can retain inherited stdout/stderr while stuck in kernel teardown; the timeout path now returns bounded output snapshots without waiting for pipe EOF. Normal exits still fully drain both streams.

Local exact-head evidence: `LocalWorkspaceToolsTests` 36/36 passed in 5.71s; `git diff --check` passed. Fresh CI was triggered by the new head; no failed run was rerun.

raullenchai:
Exact-head correction `fe6b2b0e0`: the definitive evidence showed the product process lifecycle was not the hang. The test helper spun forever when a preflight refusal completed before creating an approval request; on this runner the Swift toolchain is intentionally rejected because its Xcode is not root-owned.

I reverted all speculative process-reaping changes from the failed diagnosis. The final delta instead (1) bounds the helper, returns early preflight results, and names a 30-second approval failure; and (2) skips only the Swift-script sandbox case when no trusted Swift toolchain is resolvable. Product sandbox behavior is unchanged from the converged reviewed head.

Local exact-head: LocalWorkspaceToolsTests 36/36 in 5.84s; diff check clean. Fresh CI only; no rerun.

PR #3543, opened 2026-09-18
fix(dmg): write Finder layout records as flat plists so the install window opens (#3468)

### LINKED ISSUE #3468: Latest (v0.14.1) Desktop DMG code-signature validation
### App version

v0.14.1

### Last version where it worked (if this is a regression)

_No response_

### Mac model and memory

Studio M4 Max 64gb

### macOS version

Tahoe 26.6.2

### Where in the app does this happen?

First-run setup wizard

### What happened? What did you expect instead?

Double click DMG after download. It mounts. Click on mounted volume in Finder. Finder auto-closes. I should be able to view the app in Finder and copy it to applications. Threw codex at it and it reported:

The DMG itself verifies cleanly, so it is not a damaged download.  - Its Safari download metadata points to the project’s official GitHub Release delivery path.  - But the bundled Rapid-MLX Desktop.app fails macOS strict code-signature validation: “code or signature have been modified.” Its entitlements blob is also malformed.  - Finder closing when it opens the image is consistent with a bad Finder-view .DS_Store  file or another packaging bug. You can navigate the mounted volume in the terminal

### Steps to reproduce

double click DMG
select mounted volume in finder
finder immediately closes rather than let you view/copy desktop app

### Screenshot or screen recording

_No response_

### Diagnostics report

_No response_

### Which model was loaded?

_No response_

raullenchai:
## Problem

Double-clicking the downloaded DMG and selecting the mounted volume makes Finder blank out and close the window — and on macOS 14 (Sonoma) quit and relaunch outright — so users cannot drag the app into Applications. Reported in #3468 by three people across macOS 14.8.7, 26.6.2 and an M5.

The reporter's Finder crash reports show:

```
-[__NSCFData count]: unrecognized selector sent to instance …
-[__NSCFData setObject:forKeyedSubscript:]: unrecognized selector …
```

Finder is sending `NSDictionary` selectors to an `NSData`.

## Root cause

The committed DMG layout template (`finder-layout.DS_Store`, shipped in 0.14.0–0.14.2) stored its `bwsp` and `icvp` records as a bplist **nested one level deep inside an outer bplist `<data>` object**, instead of the flat dictionaries Finder writes. Finder decodes each record into an `NSData` and then treats it as a dictionary → crash on macOS 14, silent window close on macOS 26. Terminal installs never touch these records, which is why the shipped bytes verified cleanly everywhere except a real Finder open.

The template's records were generated by handing pre-serialised bytes to `ds_store`'s `PlistCodec` (which wraps them again), and the release gate `verify-dmg-layout.py` **deliberately unwrapped** the nested form before checking it — so the defect passed validation and shipped.

## Fix

- **Regenerate the template** with flat, dict-rooted `bwsp`/`icvp` records in Finder's own schema.
- **Add `scripts/make-dmg-layout.py`** — the template is now produced from declared constants (deterministic, reproducible, no by-hand binary edits). Building it against a live Finder + dmgbuild's reference output surfaced two latent defects the old template also had: the branded background never rendered because the record lacked the `backgroundColor` triple Finder requires even in image mode, and the alias lacked the volume mount-point tag (0x0013) Finder needs to resolve it by path. Both fixed.
- **`verify-dmg-layout.py`** now *rejects* the nested `<data>` encoding (the exact ship-blocker) and additionally gates the colour triple, the alias mount point, and the `icvl` icon-view record.
- Extend the hermetic verifier tests (Finder-crashing encodings, render-critical records, committed-template regression). 31 pass.

## Verification (macOS 26.5, real Finder)

- **A/B, deterministic across 3 rounds each:** old template opens **0** Finder windows (the reported symptom); new template opens the branded install page.
- The real `configure-dmg-layout.sh` release path emits a `.DS_Store` **byte-identical** to the committed template and passes the gate.
- **Suffixed-mount check** (codex self-review raised this): with a decoy same-named volume forcing our DMG to `/Volumes/Rapid-MLX Desktop 1`, the background still renders correctly — Finder resolves it from the current volume via the relative path, same as dmgbuild.
- The literal SIGABRT crash reproduces only on older macOS (Sonoma 14.x, per the reporter's `Finder-*.ips`); macOS 26 exhibits the same defect as a silent window close.

Fixes #3468


raullenchai:
Harbor scope-locked adversarial review on exact head b1a89653b: LGTM. Verified the generator is deterministic and reproduces the committed 16,388-byte template byte-for-byte; bwsp/icvp decode as dictionaries; the new verifier rejects the 0.14.0–0.14.2 nested-data template; 31/31 focused tests, ruff check/format, and git diff --check pass. Formatting-only CI fix is pushed, and Fixes #3468 is now in the PR body.

raullenchai:
The branch was subsequently squash/force-pushed to 690d5a3d1, which dropped the prior formatting-only fix and recreated the deterministic lint failure. Re-applied formatting on top of that exact head; no behavior or generated template bytes changed. Evidence on the new head: ruff format/check clean, 36/36 verifier tests pass, diff check clean.

raullenchai:
Reviewed the new force-squashed exact head `b99cf7dba7331938c4124afedc4151a2bfffb33c`. It preserves the formatting fix and adds two in-scope hardenings: atomic temp-file replacement for generator output, and finite/range validation for Finder background color channels. No new blocking issue. Exact-head local evidence: ruff check/format clean, verifier 37/37, diff check clean. This SHA is now authoritative; no further local changes.

raullenchai:
Reviewed final delta on exact head `8cbc13a6268200d4c048ac59ac8439483c44b9c9`: it adds only an in-scope deterministic-generator contract that compares two generated layouts byte-for-byte with the committed template. No blocking issue. Exact-head evidence: ruff check/format clean; verifier 37 passed plus dependency-optional skip; the generator determinism test passes with its declared ds_store/mac_alias dependencies; diff check clean. Superseded b99 runs were cancelled to free CI capacity.

raullenchai:
Queue recovery: Mergify dequeued exact head `8cbc13a6` as “manually updated” even though the source SHA did not change after boarding. The speculative candidate #3546 was closed while its checks were still running. Per the label-only recovery policy, I performed one controlled remove/add of `merge-ready-mac` on the same already-green head; no rebase, push, or CI rerun was requested.

raullenchai:
Reviewed the post-queue force-squashed exact head `92716950d`. Delta from the prior reviewed head is strictly in scope: the verifier now also rejects an enabled Finder preview pane and nonzero sidebar width, matching fields already emitted by the deterministic generator. No product/runtime behavior outside DMG verification changed. Exact-head local evidence: ruff check/format clean; verifier 40/40; committed-template byte-for-byte regeneration passes; diff check clean. Label remains off until this exact head’s required checks settle.

PR #3545, opened 2026-09-18
refactor(mllm): vendor mlx-vlm cache types as the phase-B types+seam slice (step 2a)

raullenchai:
## Why

The native serialized multimodal lane still depends on cache class objects from
an optional runtime. Owning the cache types is the first bounded step toward
runtime independence, but simply copying the classes would create a third Python
type namespace and silently disable exact-type fast paths. This PR therefore
vendors the types and widens the recognition seams together, before any producer
starts emitting the vendored classes.

## What

- Vendors `mlx_vlm/models/cache.py` from the pinned 0.7.1 source.
  - Upstream SHA-256:
    `b736c299bc576f4bdf8d6edf3e1ac6fa3019ca888a2f3cba115d4252f84d5b29`.
  - The vendored copy differs only in three documented, negative-controlled
    bug fixes: rotating-cache merge, pooling-mask scalar offsets, and
    `rotated` metadata round-tripping.
- Makes cache recognition symmetric across the vendored, optional upstream,
  and core text-runtime namespaces at the batch-merge, singleton extraction,
  recurrent-checkpoint, and quantized-cache seams.
- Recursively validates `CacheList` wrappers from all three namespaces while
  preserving fail-closed behavior for unsupported leaves.
- Ships the upstream MIT license in wheels and records the vendored source in
  `NOTICE`.
- Adds provenance, namespace-parity, packaging, and upstream-bug regression
  contracts.

## Non-goals

- No producer switches to vendored cache objects in this PR.
- No APC engine/adapters, vision preprocessing, model implementation, or
  speculative-decoding changes.
- No performance-baseline refresh or unrelated cache refactor.

## Acceptance

- The vendored file differs from the pinned upstream source only in the three
  documented bug-fix hunks.
- All three cache namespaces receive equivalent decisions at every modified
  seam, including compound wrappers.
- Base/text-only installs continue to work without the optional multimodal
  runtime.
- Unsupported cache shapes remain rejected.
- The upstream license is present in source distributions and wheels.

## Verification

- Provenance: upstream 0.7.1 hash matches the recorded digest; mechanical diff
  contains exactly the three documented hunks.
- Focused regression suite: **500 passed** across batching, prefix cache,
  hybrid checkpoints, quantized cache, namespace compatibility, and packaging.
- Ruff check + format: clean on every changed first-party/test file.
- Adversarial review found and fixed:
  1. missing MIT license redistribution;
  2. omitted core-runtime `CacheList` namespace;
  3. stale body/docs claiming byte identity and listing a deferred module.
- Scope-locked convergence review of the exact fixes: **no blocking issues**.

## Known baseline environment failures

The repository-wide strict scorecard still reports the established 17
environment-dependent baseline failures (missing optional image/test runtimes).
They reproduce on the base revision and are not caused by this diff. The
focused surface above is fully green.

## Design precedent

The transition follows the established compatibility pattern for independently
defined but structurally equivalent cache types: widen recognition before
changing producers, keep exact-type dispatch fail-closed, and remove the
temporary multi-namespace bridge only after all producers share one owner.

## AI assistance disclosure

AI-assisted implementation and adversarial review were used. Every finding was
verified against the source, and the final diff was validated with the focused
test suite above.


raullenchai:
**pr_validate r4 disposition** (head `0167bdb8b` → restructured in the follow-up commit):

**codex r4 — 1 blocking: `hybrid_state_checkpoints.py:185` "vendored import remains inside the outer mlx_lm import path".**

This was a misreading of the nesting: the vendored import was already at function scope, *after* the `mlx_lm` `try/except` block — not inside it. The behavior codex was worried about is pinned by `tests/test_hybrid_state_checkpoints.py::test_without_mlx_lm_only_vendored_namespace_is_recurrent`, which monkeypatches both optional distributions to `None` and asserts `_recurrent_cache_types() == (VendoredArraysCache,)` — passing since r2.

That said, the code now reads unambiguously: the vendored import is hoisted *above* every `try` block with a comment stating it must never be guarded (commit on top of `0167bdb8b`). Behavior identical; hybrid + contract tests green.

**Environment notes for this run:**
- `full_unit`: 17 failures byte-identical to clean base `e1144388a` (dspark precision + image-weight + no-mlx-marker files) — pre-existing, unrelated to this diff.
- `stress_e2e_bench`: the 4-model bench failures are **baseline drift**, not this diff. The committed baselines (captured 2026-09-05 on rapid_mlx 0.13.4 / python 3.11.14 / mlx_vlm 0.6.17) measure every current build +30-43% cold, uniformly across unrelated model families; drift on `origin/main` was independently reproduced (Qwen3.5 manual capture 364.5/335.8 ms vs recorded 249.0). Because this diff touches `pyproject.toml` (lint exemptions for the vendored copy), the harness correctly refuses a source-only base A/B, leaving no appeal path. The fix is the separate, independently reviewable baseline refresh **#3551**; this PR should be re-validated after that lands.

raullenchai:
**pr_validate r7 — codex_review PASS (0 blocking, 0 nit)** · head `232a3254f`

Rounds 3-7 summary:
- **r3 (3 blocking)**: baseline-refresh files had entered this diff via the temporary stack — de-stacked; the refresh is independently reviewable in #3551 and must land first.
- **r4 (1 blocking)**: misread nesting of the vendored `ArraysCache` import in `_recurrent_cache_types` — was already function-scope and pinned by the base-install contract test; restructured anyway so the unguarded import sits above every `try` block.
- **r5 (2 blocking) / r6 (1 blocking)**: three **real latent upstream bugs** in the vendored byte-verbatim code, each reproduced against upstream 0.7.1 (first and third also identically in mlx-lm 0.31.3) and fixed as in-source `VENDOR-DEVIATION(upstream-bugfix)` hunks with regression tests that were negative-controlled against the upstream source:
  1. `BatchRotatingKVCache.merge` called the zero-arg in-place `_temporal_order` with an argument → TypeError on every merge with content; the lane's per-leaf merge dispatch reaches this for sliding-window models.
  2. `BatchPoolingCache.make_mask` scalar-offset branch added `offset` twice → pooled tokens admitted earlier than the causal contract allows.
  3. `BatchRotatingKVCache.meta_state` serialized `rotated` via `str()` but parsed with `bool()` → `"False"` restored as `True`, corrupting unrotated caches through the memory-cache restore path.

**Remaining red in this run — both pre-existing/attributed:**
- `full_unit`: 17 failures byte-identical to clean base `e1144388a` (dspark precision / image-weight / no-mlx-marker files).
- `stress_e2e_bench`: bench measures +30-43% cold against baselines captured 2026-09-05 — machine-wide software drift reproduced on `origin/main` (Qwen3.5 manual capture 364.5/335.8 ms vs recorded 249.0). Not appeasable for a `pyproject.toml`-touching diff by design. Fix = **#3551** (baseline refresh + documented diffusiongemma noise-floor thresholds); this PR should be re-validated on fresh baselines after #3551 merges.

raullenchai:
Harbor convergence disposition on exact head `7c1ef1205`:

- The prior pr_validate claim that mlx-lm `KVCache` was omitted was factually incorrect: both `LMArraysCache` and `LMKVCache` were already appended together, and the three-namespace test passed. The code now states this contract explicitly and retains the existing degraded-import behavior proven by `test_missing_optional_mlx_lm_cache_module_keeps_vlm_support`.
- The pinned mlx-lm 0.31.3 namespace does not define `PoolingCache`, so there is no current class to admit. The resolver now uses `getattr` and will admit it if a future coherence-qualified mlx-lm release provides it; the parity test exercises PoolingCache for every namespace that actually defines it.

Focused convergence evidence: 167 passed, 1 expected skip (mlx-lm has no PoolingCache); broader focused suite before this delta: 500 passed; ruff clean. Both original real findings (MIT redistribution and mlx-lm CacheList recursion) remain fixed.

raullenchai:
Resolved the exact-head changed-lines-coverage blocker without weakening Rapid-owned coverage: `.coveragerc` now omits only the provenance-pinned `rapid_mlx/models/mlx_vlm_vendored/cache.py`. A parsed regression contract proves no other current/future module in that namespace and no `mllm_cache_compat.py` path matches any omit glob. Local evidence: 14/14 CI coverage-contract tests, Ruff, diff-check, and adversarial re-review with no blocking issues.

raullenchai:
Final adversarial review found and fixed one real upstream cache bug before queueing: an unrotated single-token decode cache preallocates unused capacity, and merge selected the zero-filled tail rather than the live prefix. The merge now copies the first `length` temporal entries; a regression test exercises the preallocated path. Focused evidence: 46 passed, 1 expected skip; Ruff and diff-check clean; scope-locked adversarial re-review reports no blocking issues.

raullenchai:
Resolved the remaining exact-head coverage failure with behavioral tests rather than weakening the gate. New tests cover (1) singleton extraction when optional mlx-vlm is absent and (2) forward-compatible mlx-lm PoolingCache recognition. Focused coverage confirms the previously missing `mllm_batch_generator.py:605-606` and `mllm_cache_compat.py:51` are now executed; 48 passed, 1 expected skip; Ruff and diff-check clean.

raullenchai:
Root cause of the repeated coverage red is now confirmed: the new behavioral tests are MLX-bound, but their file was absent from the Apple coverage roster, so CI never executed them despite local coverage proving the lines. Added `tests/test_mllm_cache_namespace_compat.py` to the existing Apple coverage lane plus a roster contract. Local evidence: 44 passed, 1 expected skip; Ruff and diff-check clean.

PR #3550, opened 2026-09-18
fix(desktop,agent): make Personal Intelligence and Community Benchmark survive small-model dogfood

raullenchai:
## Why

Release dogfood of Personal Intelligence and Community Benchmark on the small models most users can run exposed several end-to-end failures: local write/compile/run calls could lose their arguments, declined actions could be reported as completed, local follow-ups could route to the web, and the Community Benchmark publish sheet could hide its action row or fail to restore the prior chat model.

This PR is now independently based on the merged #3538 head. It contains only the dogfood fixes; the discarded process-reaping experiments and the two test-runner fixes already landed by #3538 are not present.

## What changed

- Makes the `local_run` wire shape unambiguous with `argv`, while accepting legacy spellings.
- Recovers only schema-valid malformed forced-tool shapes observed from the supported small models.
- Preserves truthful decline/not-executed semantics and gives final synthesis one bounded prose correction.
- Keeps explicit local follow-up requests on-device unless the user explicitly asks for online search.
- Normalizes safe mechanical details such as invented home paths, compiler output/run steps, source working directories, and multi-word local search.
- Rejects fractional `timeout_seconds` before the Desktop's integer decoder sees them.
- Keeps Community Benchmark publish controls visible and individually accessible.
- Restores the chat model that was serving before a Community Benchmark run or stop.

## Scope

Allowed scope: Personal Intelligence local workspace reliability and Community Benchmark release-dogfood regressions.

Non-goals: expanding the tool allowlist, changing consent policy, adding shell execution, changing benchmark methodology, or modifying model/runtime lifecycle policy outside restoring the previously serving model.

## Verification

- `uv run pytest -q tests/test_agent_server.py tests/test_tool_choice_enforcement.py` — 219 passed.
- `swift test --package-path apps/rapid-mac --no-parallel --filter 'LocalWorkspaceToolsTests|CommunitySharePreviewTests|ModelResidencyTests'` — 91 passed.
- `uv run ruff check` on all changed Python implementation/test files — passed.
- Previous release dogfood evidence on qwen3.5 4B/9B and MiniCPM 2B covered write → compile → run, Python run, allow/decline trash and write, local follow-up search, benchmark completion/stop, model restoration, and the expanded publish sheet.
- Exact-head CI and the managed Mac merge candidate remain required before merge.

## Risk and rollback

Risk is concentrated in small-model tool-call normalization. Recovery is fail-closed to declared schema keys, Desktop still presents approval for mutations, and no shell is introduced. Rollback is the PR revert; no data migration or persistent format change is involved.


raullenchai:
Scope-locked review disposition for the latest exact head `5543af783`:

- Fixed the in-scope benchmark lifecycle race: the displaced server model is now restored inside the reservation handoff, and a new benchmark cannot acquire ownership until restoration completes.
- Added a focused regression test proving the next owner remains blocked throughout restoration.
- Kept the change limited to Community Benchmark lifecycle ownership and the existing ContentView restoration call site; no product-policy or unrelated server lifecycle changes.

Verification: `swift test --package-path apps/rapid-mac --no-parallel --filter ServerRuntimeCapabilitiesTests` (21/21 passed), plus `git diff --check`.

raullenchai:
Addressed the latest in-scope adversarial finding: a benchmark that queues during model restoration now inherits the restored model identity, so the final owner in that serialized chain restores the user’s model instead of leaving Desktop idle. The regression test now verifies both exclusion during restoration and alias preservation through the next owner.

Verification: `swift test --package-path apps/rapid-mac --no-parallel --filter ServerRuntimeCapabilitiesTests` (21/21 passed), `git diff --check`.

raullenchai:
Addressed both latest scope-locked findings:

- A present malformed `argv` is no longer rewritten to an empty list; only a genuinely absent `argv` receives the empty default, so the runtime schema rejects invalid model output instead of approving a bare interpreter.
- Desktop now expands `~/` inside the exact joined compiler path options already emitted by server normalization (`-I`, `-F`, `-isystem`, `-iquote`, `-iframework`, and `-o`).

Verification: focused Python normalizer tests 2/2; focused Swift home-expansion test 1/1; `git diff --check`.

raullenchai:
Addressed both latest in-scope routing/execution findings:

- A local path plus a local-search verb remains private/local when “online” is merely part of the query; only a separate web/internet destination or “online instead” overrides local routing.
- “Compile and run <source path>” now counts as explicit compile-and-run intent, so compile-only `-c` is removed before the execution sequence.
- Applied the required formatter cleanup.

Verification: focused routing/normalization tests 4/4; Ruff check/format clean; `git diff --check`.

raullenchai:
Addressed both latest scope-locked execution findings:

- Node/Ruby syntax-check flags now suppress “script completed” only while they are interpreter options; identical strings after the script operand remain ordinary script arguments.
- Compiler `-c` removal now uses an option-aware scan, so `-c` used as the operand of `-include` (or after `--`) is preserved.

Verification: focused history and compile-normalization tests 2/2; Ruff check/format clean; `git diff --check`.

raullenchai:
Addressed both latest scope-locked routing findings:

- “online” remains a web override through the canonical online-intent matcher, with one narrow local-location exception (`find … online … in ~/…`) so a local destination cannot leak to web search; “search online for references to ~/…” stays web-only.
- Follow-up routing no longer parses user intent out of the quoted conversation string. Desktop now sends a bounded structured list containing only prior user-authored messages; assistant content containing forged `user:` delimiters cannot grant a local tool. The human/model context remains unchanged.

Verification: focused Python routing/schema tests 2/2; Swift AgentRuntimeClient 11/11, AgentChatTurn 11/11, AgentSessionController 16/16; Ruff check/format clean; `git diff --check`.

raullenchai:
Adversarial review disposition on the latest head:

- Fixed the valid legacy-compatibility finding: Desktop now strips `declined` only for a non-executed decline when the 422 specifically identifies `declined` as an unknown/extra field. Contradictory `declined=true, executed=true` and unrelated 422 responses are no longer retried; both have regression tests.
- The compound local+online routing finding was already covered by the current implementation and its negative controls: `Read ~/Documents/report.md and search the web for updates` routes local read + web, and `Search online instead, then write ...` routes local write + web. `explicit_online_override` only prevents local intent from suppressing web; it does not suppress independent local tools. No code change was appropriate for that finding.

Focused Swift transport suite: 13/13 passed.

raullenchai:
Latest adversarial review disposition:

- Fixed the valid benchmark lifecycle finding. The restoration callback now reports success; a failed model restart retains the displaced alias, and the next serialized benchmark release retries it instead of silently losing the prior chat model. Added a lifecycle regression test; ServerRuntimeCapabilitiesTests 22/22 passed.
- The follow-up routing finding conflicts with the intentional, tested action-specific recency contract: `Search again` after a newer read correctly finds the most recent prior search, while destructive/write/run actions already stop at the newest mismatched local turn. Changing it as suggested breaks the existing regression test and the user’s explicit verb. Disposition: no code change.

raullenchai:
Adversarial review round disposition:

- Duplicate XML parameter names now fail closed instead of last-value-wins; regression test added.
- `go run ~/path/to/package` now expands the package-directory operand without relying on a shell; regression test added.
- The local_run prompt now names `argv` as canonical and `arguments` only as a compatibility alias.

Focused Python salvage test, Swift home-expansion test, ruff, and diff check all pass.

raullenchai:
Adversarial review disposition: the reported `recent.reversed()` compile error is false. Swift accepts the contextually typed return here, and the exact current head successfully compiled `ChatViewModel.swift`; `swift test --filter AgentChatTurnTests` passed 11/11, including `Personal Intelligence routing history contains only user-authored turns`. No code change is warranted.

PR #3554, opened 2026-09-18
refactor(mllm): vendor the mlx-vlm APC support surface (step 2b-1)

raullenchai:
## Why

The native MLLM lane still reaches into `mlx_vlm.apc_adapters` /
`mlx_vlm.vision_cache` for prefix-cache cloning and vision-feature reuse,
so lane-level behavior depends on code this repo cannot review or patch
independently of upstream's release cadence. Vendoring the support surface
brings it under this repo's tests and review while keeping the upstream pin
for the engine itself (apc.py follows in the next PR of the stack). Because
the transition runs lane caches in two namespaces, the vendored adapters
are rewired to recognize both and to keep results in the producer's
namespace — behavior-neutral today by construction.

Step 2b-1 of the mlx-vlm dependency retirement (Phase B step 2). #3545 has landed; this PR is now based directly on `main` and follows the standard `merge-ready-mac` queue policy.

## What lands

Six upstream `mlx-vlm==0.7.1` modules vendored into `rapid_mlx/models/mlx_vlm_vendored/`:

| File | Provenance |
|---|---|
| `apc_storage.py`, `_stream_cleanup.py`, `vision_cache.py` | byte-identical (sha256 in the package inventory) |
| `kv_quant.py` | one documented redirect: lazy `.turboquant` → pinned upstream `mlx_vlm.turboquant` (7k lines, can never fit a reviewable diff) |
| `apc_coordinator.py` | module-level import resolves the vendored adapters; lazy `.apc` engine calls + `make_prompt_cache` fallback resolve upstream until 2b-2 / step 3; `no-any-return` locals annotated |
| `apc_adapters.py` | dispatch rewired to cover both cache namespaces (vendored + upstream) — the transition mechanism from the design note |

## The duality mechanism (same contract as #3545)

Every exact-type table, capability registration, contract probe and constructor covers **both** namespaces (`_cache_namespaces` / `_cache_namespace_of` helpers, each site marked `# VENDOR-DEVIATION(dual-namespace):`). Constructors are **namespace-faithful**: cloned/merged results keep the producer's cache types, so upstream-typed inputs get byte-identical behavior today and correct typing once producers flip in step 3. A mechanical revert at step 3 drops the helpers.

Lane changes: the four `mlx_vlm.apc_adapters` / `mlx_vlm.vision_cache` imports in `mllm_batch_generator.py` redirect to the vendored modules; the four test modules that stub `clone_cache_entry` were re-pointed at the vendored adapter (required — they patch the module the lane now imports). Producers still emit upstream-typed caches, so behavior is unchanged by construction.

The APC engine (`apc.py`, ~188 KB) follows in the next PR of this stack; `prepare_inputs` in 2c.

## Verification

- New tests: `tests/test_mlx_vlm_vendored_apc_adapters.py` (27 — clone/capability/merge namespace-faithfulness for both namespaces) and `tests/test_mlx_vlm_vendored_apc_support.py` (26 — coordinator wiring, kv_quant upstream-parity probe, storage, vision-cache LRU).
- Focused APC and media suites: 240 passed; changed-line coverage: 815/815 (100%); CI workflow contract tests: 56 passed..
- Full unit suite on the pre-coverage head: 25,778 passed, 120 skipped, 24 deselected, 6 xfailed, 1 xpassed; two port-8000 tests were excluded because the users Desktop dogfood server was intentionally left running.
- Design note updated: `docs/engineering/design/2026-09-18-vendor-mllm-primitives.md`.


raullenchai:
Codex r2 dispositions (items 1 and 4 are verified false positives; items 2 and 3 fixed in c46818cea):

**1. `from_legacy(3.5, 'uniform', 64)` raising ValueError — not reproducible.** Fractional bits auto-route to TurboQuant: `turboquant_enabled(3.5, 'uniform')` returns True, so `from_legacy` takes the `resolve_kv_bits` branch and returns a `turboquant`-scheme policy (verified: `bits=3.5, scheme=turboquant`, no exception). The uniform-requires-integer-bits `ValueError` in `_bits_for_scheme` is only reachable when the resolved scheme is uniform, which fractional nominal bits never produce here. The parity test exercises exactly that path against upstream, which behaves identically.

**4. `merge_cache_entries` tuple branch returning `lm.CacheList(*merged)` — byte-verbatim upstream.** Upstream 0.7.1 `mlx_vlm/apc_adapters.py:654-659`:
```python
if all(isinstance(c, tuple) for c in entries):
    merged = [...]
    return None if any(m is None for m in merged) else lm.CacheList(*merged)
```
The clone branch's tuple→tuple / merge branch's tuple→CacheList asymmetry is upstream's own design; changing it would deviate from the pin. The namespace/shape-preservation contract in this PR covers the constructions the rewiring *added* (namespace routing via `_cache_namespace_of`), and the tuple branch constructs nothing namespace-dependent — `lm` there is only used for `CacheList`, exactly as upstream uses it. Left byte-verbatim per the provenance contract.

**2 (Path sources) and 3 (nesting-boundary collision) — valid, fixed** in c46818cea: `_make_key` now emits self-delimiting type-tagged encodings (`s`/`l`/`p`), injective at every nesting level, with `os.PathLike` normalized via `os.fspath`.

raullenchai:
Scope-locked review disposition for exact head after restacking onto main:\n\n- Fixed the two valid optional-dependency findings: cache cloning now converts only a missing transitional `mlx_vlm` / `mlx_vlm.apc` module into a cold cache miss. Unrelated `ModuleNotFoundError`s from cache implementations propagate. Added positive fallback and negative-control propagation coverage for both media and exact-text clone paths.\n- The reported fractional-uniform parity-test failure was not reproducible: `test_kv_quant_fingerprint_parity_with_upstream` passes against the pinned dependency because that 3.5-bit case selects its supported quantization policy in both implementations. No production or test change was made for that false positive.\n- Focused evidence on the resulting head: 221 passed; ruff and diff check clean. A fresh full `pr_validate` will run on this head.

raullenchai:
Next scope-locked review disposition:\n\n- Fixed both valid pluggability findings. Third-party snapshot/state contracts now clone without a vendored/upstream cache namespace, and `prefix_cache_merge` executes before namespace-dependent built-in merge routing.\n- Added stripped-install negative controls by forcing namespace resolution to `None`; both third-party clone and merge contracts are verified.\n- Focused evidence: 223 passed; ruff and diff check clean. Continuing review until convergence on the new exact head.

raullenchai:
Review disposition:\n\n- Fixed the valid palette-transparency collision: vision feature keys now include type-tagged Pillow transparency metadata in addition to mode, size, palette, and pixel indices. Added a cross-serve negative control for identical palette/index images with different transparent entries.\n- Reconfirmed the fractional quantization parity case directly: `test_kv_quant_fingerprint_parity_with_upstream` passes against the pinned upstream implementation; its 3.5-bit input is routed consistently by both implementations, so the repeated static finding is a false positive.\n- Focused evidence on the resulting diff: 224 passed; ruff and diff check clean.

raullenchai:
Fixed the latest valid namespace finding: tuple composites now derive the result namespace from their first cache child before any generic fallback. Added a dual-namespace regression proving vendored tuples stay vendored and upstream tuples stay upstream while both implementations are installed. Focused suite: 225 passed; lint/format/diff clean.

raullenchai:
Fixed the latest valid APC finding: `ArraysCacheCloneAdapter.merge_rows()` now merges effective `left_padding` and `lengths` metadata alongside recurrent state arrays, zero-filling only rows where that metadata is absent. Added a dual-namespace regression with asymmetric metadata. Focused suite: 226 passed; lint/format/diff clean.

raullenchai:
Clarified the repeatedly misread fractional-KV parity contract without changing production behavior: the 3.5-bit legacy case is now a dedicated assertion proving both vendored and pinned upstream implementations intentionally select TurboQuant even when an old caller supplies the nominal `uniform` string. Both produce `kv3.5-64-turboquant-16`; focused suite remains 226 passed. The 16-hex cache digest note is non-blocking: with an in-memory LRU capped at 20 entries, a 64-bit digest has negligible collision risk and changing the public key format is outside this vendoring slice.

raullenchai:
Fixed the one PR-owned full-suite failure: the empty rotating-cache conversion now uses the repository scanner-approved `getattr(source, "keep", 0)` re-wrap form, preserving behavior while keeping the admission-memory upper bound statically provable. The exact scanner plus the focused APC suite pass (227 tests). The remaining two local failures are both the pre-existing CLI test helper binding hard-coded port 8000 while the user Desktop server is actively listening there; that process will not be killed. Isolated exact-head CI is the authoritative evidence for those tests.

raullenchai:
Coverage failure fixed at exact head `8e87e8ca4`. Root cause: the new vendored APC contract tests were not in the Apple coverage roster, and several defensive/cache-shape branches lacked executable assertions. Added the two suites to the roster and behavior tests for coordinator block/checkpoint flows, clone/merge fallback contracts, quantization policy edges, storage accounting, and vision-cache metadata.

Local evidence:
- focused APC/media suite: 240 passed
- changed-line coverage vs `origin/main`: 815/815 = 100%
- CI/workflow contract tests: 56 passed
- ruff + format + diff check: clean

No production behavior was changed by this final patch.

raullenchai:
Final exact-head review disposition (`8e87e8ca4`): scope-locked Codex review reports no blocking issues; targeted validation is 292 passed / 1 deselected; local changed-line coverage is 815/815 (100%); GitHub required `tests`, `desktop-tests`, and `version-bump-guard` are green. The full local suite produced only the two documented port-8000 environment conflicts while the users Desktop dogfood server remains active (25,792 passed); the server was not interrupted. Added `merge-ready-mac`.

PR #3562, opened 2026-09-18
fix(mac): show Bonsai 2 on 18 GB Macs

raullenchai:
## Why

Rapid-MLX 0.14.3 correctly ships `bonsai2-27b-2bit`, but Desktop estimates it like a dense 27B model. That reserves 6 GB for KV state and reports about 15 GB, pushing Bonsai 2 into the bottom **Not fit for this Mac** section on an 18 GB Mac. Users reasonably read that as the headline model being absent.

Bonsai 2 is a hybrid model with one full-attention layer in four and four KV heads, so the dense-model estimate is not representative.

## What

- Give `bonsai2-27b-2bit` a narrowly scoped 10.5 GB runtime footprint derived from its 8.0 GiB pack, measured 8.8 GiB loaded footprint, and ordinary chat/cache headroom.
- Keep the live-memory admission guard unchanged; it remains the final authority under real host pressure or long prompts.
- Pin the 18 GB picker behavior: Bonsai 2 appears in **All models** as borderline, while genuinely oversized models remain separated.

## Design precedent

Use model-specific measured/runtime metadata when architecture makes parameter-count sizing materially inaccurate, while retaining a conservative generic fallback and a live admission check. This follows the repository's existing measured-footprint path for curated aliases instead of weakening the global fit threshold.

## Verification

- `swift test --no-parallel --filter ModelSizingTests` — 22 passed
- `swift test --no-parallel --filter ModelPickerBarSectionOrderTests` — 26 passed
- Real Studio load: 8.8 GiB after load; 6,316-token request completed successfully; live guard remains unchanged for pressure-sensitive starts.
- `git diff --check`

## Scope

This hotfix changes only Bonsai 2 Desktop sizing and picker placement. It does not change model loading, model files, the global fit thresholds, recommendations, or other aliases.


raullenchai:
Closing after real-device validation on an 18 GB Mac. The existing picker classification is correct: Bonsai 2 can map/load its weights but exceeds the host model budget and fails generation with a Metal allocation OOM. Relaxing the sizing gate would turn an accurate warning into a reproducible HTTP 500. No release change is warranted; a separate future UX fix may translate persistent Metal OOM into actionable memory guidance instead of suggesting a retry.

PR #3558, opened 2026-09-18
refactor(mllm): vendor the mlx-vlm APC engine (step 2b-2)

raullenchai:
## Why

Step 2b-2 of the MLLM runtime dependency-retirement plan. The native MLLM lane's APC engine was still imported directly from the pinned runtime package; bringing the source into Rapid-MLX makes its exact behavior, provenance, packaging, typing budget, and tests reviewable in this repository without changing cache policy or public behavior.

This PR vendors `apc.py` from the pinned `mlx-vlm==0.7.1` source. It is byte-identical after normalizing exactly 18 documented one-line import redirects. Direct APC imports in `MLLMBatchGenerator` now resolve to the vendored engine. Cache producers, lazy coordinator helpers, type-dispatch widening, and the broader producer cutover remain intentionally outside this slice.

## Scope

- Add the 4,994-line pinned APC engine source.
- Redirect its four package-local imports to already-vendored siblings.
- Preserve 15 lazy model-cache imports and three TurboQuant imports on the pinned runtime during the transition; every redirect carries `# VENDOR-DEVIATION`.
- Switch the two direct production APC imports and their test patch points.
- Add a fail-closed source-equivalence contract and file-scoped coverage treatment.
- Update transition comments and the existing mypy baseline for the verbatim vendored files.

Non-goals: no dispatch widening, cache-producer switch, APC redesign, performance-policy change, or unrelated upstream bug fix. Those belong to later, independently reviewed slices.

## Design precedent

This follows a staged pinned-source migration: first establish a byte-for-byte source contract, then move direct ownership, then widen compatibility and finally switch producers. The large imported implementation remains exempt from changed-line coverage only while an Apple-CI test proves its pinned hash and rejects every difference except the 18 explicit redirects. Rapid-owned adapters and all future modules remain measured.

## Review disposition

Earlier review findings inside the verbatim engine were confirmed as pre-existing behavior in the pinned source and are intentionally outside this source-transfer PR. The final scope-locked adversarial review found no blocking issues after the provenance contract and coverage boundary were added.

## Verification

- Pinned upstream SHA-256: `5b2b940852f11f34f7b4daf627bc31fc701f8abffc72d40189bc3e5ac57f878c`.
- Source contract: after normalizing exactly 18 marked redirects, the vendored source equals the pinned upstream source byte-for-byte.
- Focused tests: 262 passed.
- Changed-line coverage for measured Rapid-owned production lines: 100%.
- Ruff: clean for changed Rapid-owned Python files.
[...]

PR #3563, opened 2026-09-18
refactor(mllm): dual-namespace recognition for the vendored APC engine (step 2b-3)

raullenchai:
## Why

Step 2b-3 of the mlx-vlm dependency-retirement plan (design note: `docs/engineering/design/2026-09-18-vendor-mllm-primitives.md`), now based directly on `main` after #3558 landed.

The engine vendored in #3558 dispatches on upstream-typed caches only: its exact-type tables, snapshot/restore constructors, and checkpoint-class resolver would silently drop fast paths when step 3 starts emitting vendored-typed caches. This PR widens recognition while producers still emit upstream caches, so the transition itself does not change request behavior.

## What lands

- Exact-type recognition across the vendored and upstream cache namespaces.
- Namespace-faithful exact snapshot/restore. Vendored entries record `_ns="v"`; an absent key remains the backward-compatible upstream format.
- Fail-closed metadata handling: unknown namespace markers are rejected rather than silently restored as upstream.
- Checkpoint resolution permits the exact vendored cache module, not the wider vendored package tree; local disk metadata cannot select unrelated Rapid-MLX classes.
- Dense checkpoint clones preserve the source cache namespace.
- Three repro-tested engine fixes:
  - unreadable shards no longer inflate disk-byte accounting;
  - overlapping writers only release their own in-flight event;
  - failed layer-major writes clean up their temporary file.
- A source-integrity contract pins both the upstream provenance hash and the reviewed vendored engine hash, plus the 22-deviation sentinel inventory.

## Scope and non-goals

This PR changes cache namespace recognition and the three bounded engine defects above. It does **not** switch cache producers, change cache policy, alter public APIs, change model behavior, or remove the upstream dependency. `make_prompt_cache` and quantize/batch constructors continue producing upstream-typed caches; that producer cutover remains step 3.

## Design precedent

The transition uses the repository's existing dual-namespace adapter seam: consumers recognize both old and new types before producers flip. Disk formats keep an explicit version marker, retain backward compatibility for marker-less entries, and fail closed on unknown future markers. Class restoration stays allowlisted to the one module required by this transition.

## Review disposition

The scope-locked adversarial loop converged with no blocking findings after hardening the exact boundaries introduced by this PR:

- unknown namespace markers now fail closed;
- checkpoint metadata can resolve only reviewed, top-level cache classes from an explicit module/class allowlist; nested attribute traversal and unrelated module imports are rejected;
- the allowlist includes every public `_BaseCache` subclass shipped by the pinned cache and TurboQuant modules, and a test forces review if that inventory changes;
- malformed shards are indexed only after kind-specific metadata validates; shards that cannot be deleted remain counted against the disk budget;
- the temp-file cleanup test now creates a real partial file before injecting failure, eliminating its prior false-positive path;
- source-integrity tests pin upstream provenance, the reviewed vendored source, and the deviation inventory.

The remaining review nit concerned hypothetical legacy model-specific checkpoint classes. A static inventory of pinned 0.7.1 found only the known TurboQuant classes outside the cache module, and those are explicitly preserved. Ring caches use their dedicated snapshot format. Unknown future classes intentionally fail closed and cause cache recomputation rather than arbitrary import traversal.

## Verification

- Focused suites: **295 passed** across engine, support, KV-estimation, batch generator, media-prefix-cache, and adapter coverage.
- Focused namespace/security suite: **23 passed**.
- Pinned Python 3.11 mypy budget: **736 grandfathered errors; no growth**.
- Ruff and `git diff --check`: clean.
- Restacked diff against current `main`: 4 files, limited to the intended step-2b-3 surface.

Co-Authored-By: Claude Code <noreply@anthropic.com>


raullenchai:
## pr_validate r1 — converged (codex clean on first pass)

All PR-controllable gates green at r1: **codex found no blocking issues**, lint PASS, supply-chain clean.

**Full-unit note:** 19 failed = the 17 pre-existing baseline failures (byte-identical to the #3545/#3554/#3558 baseline set, env-related) **+ 2 port-collision flakes**: the two `test_legacy_prefix_cache_flag_warnings` serve-command e2e tests failed with `OSError: [Errno 48] Address already in use` — an unrelated long-lived process holds port 8000 on this host (not a validation-loop process). The same tests passed in the #3558 loop's full_unit runs. Not a regression of this diff.

**targeted_tests** reported "no python files mapping to tests" (diff→test mapping quirk for stacked-PR diffs); the affected suites were verified locally instead: engine + support 23 passed, batch generator + media prefix + adapters 196 passed; mypy pinned-env budget: zero growth; `ruff check` clean.

Letter-level verdict remains DO NOT MERGE solely due to the strict verdict × baseline reds — same accepted stack state. **Not enqueued** (manual queueing only).

# PR #3563 validation scorecard

**Title**: refactor(mllm): dual-namespace recognition for the vendored APC engine (step 2b-3)
**Author**: raullenchai
**Diff**: 4 file(s), +328/-33 LOC, blast radius: **medium**

## Verdict: **DO NOT MERGE**

| step | status | summary | time |
|---|---|---|---:|
| `fetch` | PASS | 4 files, +328/-33 LOC, blast=medium | 1.4s |
| `test_plan_check` | PASS | no checkbox-style test plan found | 0.0s |
| `cl_description_quality` | PASS | title OK + body has rationale (3701 chars) | 0.0s |
| `supply_chain` | PASS | no hooks touched, no suspicious patterns, deps clean | 0.0s |
| `test_env_check` | PASS | all 6 required test packages importable | 0.9s |
| `review_vocabulary` | PASS | no directives | 0.0s |
| `codex_review` | PASS | codex found no blocking issues | 18.4s |
| `lint` | PASS | clean (1 file(s)) | 0.1s |
| `targeted_tests` | skip | diff has no python files mapping to tests | 0.0s |
| `full_unit` | FAIL | 19 failed, 25547 passed, 126 skipped, 22 deselected, 6 xfailed, 1 xpassed in 763.40s (0:12:43) | 765.4s |
| `stress_e2e_bench` | skip | skipped (gating predicate returned False) | 0.0s |
| `diff_coverage` | skip | advisory coverage skipped (no measurable production lines in diff) | 874.4s |

## Details

raullenchai:
**Heads-up from the stack rebase onto the post-#3545/#3554 main:** the queue-added provenance gate on #3558 (`64beaf1de` → `test_engine_matches_pinned_upstream_except_declared_import_redirects`) does a whole-file byte compare after normalizing the 18 sentinel-marked import redirects. The step-2b-3 dual-namespace hunks in `apc.py` (+118/−28: namespace-faithful clones at :479, constructor resolution at :1765, ladder recognition at :2797, checkpoint records at :991, transition gate at :377) carry sentinel-marked COMMENT lines and real behavioral edits, so the probe now fails two ways — `assert "from mlx_vlm." in line` trips on the marker comments, and the behavioral hunks themselves can't survive a byte compare.

This replay broke at exactly this link (passes at 919091a97, fails at 42a886529), and it cascades through #3566/#3575/#3578's vendored suites. Suggested repair shape (in this PR): port the walker pattern from `tests/test_mlx_vlm_vendored_speculative.py` — per-function comparison with (a) import-redirect normalization counted at module level, (b) a documented set for the dual-ns hunk functions (each already reviewed in this PR's loop and inventoried in the package `__init__.py`), (c) module-level constants compared after stripping marker comments. Happy to take it here if you'd rather land it downstream — say the word.

PR #3568, opened 2026-09-18
fix(mac): retire idle animation loop, keep straight quotes, gate tool caption on roster, single instance

raullenchai:
## Summary

GUI dogfood of the 0.14.3 release on an M2 Pro mini (32 GB) with `bonsai2-27b-2bit`, driven through the real app (`dogfood-findings-2026-09-18-mini-0143.md`, local). Inference was correct on every turn; four Desktop defects came out of it, two more surfaced while verifying the fixes, and all six are fixed here. All are pre-existing, none is a 0.14.3 regression.

### Idle CPU (F-1)

The footer status dot's breathing animation was started with `.animation(.repeatForever…, value:)` and "stopped" by switching that modifier to `.default` on the same view. SwiftUI does not retire a repeating animation that way: the window committed a frame every refresh forever after the first Starting → Ready. Cost with the model resident and nothing happening:

| machine | before | after |
|---|---|---|
| M3 Ultra Studio, default pointer (sample of the shipped 0.14.3) | 23–28 % | — |
| M2 Pro mini, no accessibility client (`top -l 7 -s 5`) | 58–60 % | 1.2–3.6 % |
| M2 Pro mini, 2.44× accessibility pointer + AX client attached | 83–105 % | 20–50 % bursts (AX tree rebuilds on footer ticks, see below) |

Fix: `BreathingLoop` (RapidMotion.swift) owns the `repeatForever` loop and is instantiated inside an `if` that is false at rest, so the loop dies with the view; `PulsingStateDot` and the Community Benchmark mascot use it. A source tripwire test keeps `repeatForever` confined to removable owners started with `withAnimation`. A standalone 40-line repro on the Studio showed the shipped pattern at 1.0 % CPU forever after the toggle and the removed-view pattern at 0.0 %.

Note for future measurements: any accessibility client (VoiceOver, window managers such as Magnet, `System Events`, our own AX-driving tools) switches on SwiftUI's accessibility-tree maintenance for the process lifetime and multiplies the cost of every view update. Measure idle CPU before touching the app.

### Curly quotes in the transcript (F-2)

swift-markdown enables cmark's smart typography by default; the compiler passed only `.parseBlockDirectives`. Bare JSON rendered as `{ “city”: “Tokyo” }` and pasted as invalid JSON. `MarkdownCompiler.parseOptions` now includes `.disableSmartOpts` at both compile entry points. Newline collapsing inside unfenced JSON is ordinary Markdown soft-break behaviour and is left alone.

### "Answered without calling any of the available tools" under `391` (F-4)

`ChatMessage.shouldFlagToolNotCalled` gains a roster gate (`advertisedToolCouldServe`): the caption only fires when an advertised tool could have served the prompt's lane. Built-in tools are classified by exact name; the local workspace tools (including `local_run`) serve none of the caption's lanes — `local_run` is "Run development command" behind an approval sheet, not a calculator, and a model that multiplies `17 * 23` in its head instead of asking permission to run a shell did the right thing. Other names (MCP) are classified by the words in the name with prefix matching (`calculator`, `execute_python`, `fetchURL`), and a name with no recognised word keeps the caption, so an opaque connector never silences it. `ChatViewModel` passes `request.tools?.map(\.function.name)`; a `nil` roster keeps the previous behaviour.

### Second app instance (F-6)

`open -n`, exec'ing `Contents/MacOS/Rapid`, or a second copy of the bundle started a second engine on the same port; the first window fell to "Couldn't start bonsai2-27b-2bit — check the model files" (Retry works once the duplicate quits, but nothing says so). Fix in two layers: `LSMultipleInstancesProhibited` in Info.plist for the Finder / `open -a` path, and `SingleInstanceGuard` at the very top of `RapidApp.init` — before the crash marker and before `PortSweep.startLaunchSweep`, which trusts the on-disk sidecar ownership record and would have reaped the running instance's engine as an orphan (codex round 1). Seniority is launch order (`NSRunningApplication.launchDate`, PID as tiebreak), so a concurrent cold launch keeps exactly one instance from both viewpoints. The hand-off goes through `NSWorkspace.openApplication(at:)` so a survivor whose window was closed reopens it (`applicationShouldHandleReopen`), with a bounded wait and `activate()` fallback.

A third layer came out of the pr_validate codex review: an advisory `flock(2)` on `Application Support/Rapid/desktop-instance.lock`, taken before LaunchServices registers the process and held for the process lifetime. It is authoritative when available (two cold launches milliseconds apart cannot both miss each other in a `runningApplications` snapshot); the snapshot rule then only picks whom to hand off to, and an instance that has finished launching always wins it. After handing off, the yielder waits up to 5 s for the lock (`SingleInstanceGuard.yieldsLaunch()`): quit-and-relaunch within a second used to hand the launch to the quitting process and leave **no app at all** (codex; reproduced on the mini with `quit` + exec 0.2 s later → zero Rapid processes). The kernel releases the lock when the old process dies, so the relaunch takes over.

Verified on the mini with the final binary (`b9b2a3be…`) in the installed bundle: cold double direct exec 0.2 s apart → one PID, `Rapid-MLX` window, one sidecar; `open -n` against the running app → still one PID; quit + exec 0.2 s later → a new PID with window and sidecar (was: none); a duplicate launched against a healthy Desktop exits after the 5 s grace; idle CPU 1.7–3.1 % with the model resident.

### Settings-only launch with no engine (F-7)

While driving the fix build, a launch came up with a single 900×720 "Settings" window, no chat window, no sidecar — on every launch of that bundle, by `open -a` and direct exec alike. Unified log: `_reopenWindowsAsNecessaryIncludingRestorableState … hasPersistentStateToRestore=1 … savedIdentifier=settings`. AppKit window restoration had persisted a session whose only restorable window was Settings (main closed with ⌘W, Settings open at quit); SwiftUI restored exactly that and never created the main `Window`, and `ContentView` is what starts the sidecar. `-ApplePersistenceIgnoreState YES` on the same bundle brought main + engine back at once. Fix: the Settings `NSWindow` is `isRestorable = false` (the AppKit flag behind SwiftUI's macOS-15-only `restorationBehavior(.disabled)`; the app floor is 14), and the Settings scene heals an already-persisted session — if Settings is the first window this process has shown, it opens the main window. Pre-existing, not a 0.14.3 regression.

### Out of scope

F-3 (every new chat pays a ~2.3k-token system prompt; ~30 s TTFT for the first turn on Bonsai 27B on this machine, and the prefix cache does not survive an app relaunch) is documented in the findings file and not addressed here.

## Verification

- `swift test` targeted suites (new + neighbouring): 155 tests / 13 suites green; after the guard rework 73 tests / 5 suites and the 17-test guard + init-order pair green again; `ToolNotCalledCaptionTests` 52 green after the roster rewrite.
- Full `swift test` on the Studio: 4 issues in `SSEDeltaCoalescer` (`hops < 12`) and `MermaidTests` (2 s timeout budgets), none touching changed code; they flap between runs on this loaded host (load avg ~4.8) and the same suites were run on the untouched base commit as a control — see the last comment on this PR.
- Codex adversarial review: round 1 raised 2 BLOCKING / 3 MAJOR / 4 MINOR, all addressed as described above; round 2 result posted as a comment.
- Codex round 3 (after the lock + restoration changes) raised 3 BLOCKING: the blind hand-off outcome (documented — the mid-launch holder shows its own window regardless), activation failure (cosmetic; the survivor's liveness decides), and the lock-unavailable case where a Desktop with no reportable launch date ranked as newest — fixed by letting an instance that finished launching win outright, with a test. Nits: monotonic clocks for the waits (done); source-scan tests (kept, they pin ordering that has no runtime seam).
- pr_validate run 2 (b44315cc) codex: 2 BLOCKING — quit-and-relaunch leaving no app (fixed + reproduced/verified above) and the captured-var completion result (now `OSAllocatedUnfairLock`).
- Real-machine checks above (idle CPU, single instance, quit-relaunch race, window reopen, Settings-only launch).
- GUI dogfood of F-2 / F-4 with the final build on the mini (Bonsai 27B, real turns through the app): bare-JSON prompt → transcript shows `{"city":"Tokyo","country":"Japan","population":37400064}` with straight quotes (the only typographic quotes left on screen are in the telemetry-consent banner copy); `What is 17 * 23? Answer with just the number.` → `391` with no "answered without calling any of the available tools" caption under it.

## Test plan

Covered by the suites above plus the mini verification; post-merge the release smoke should include one idle-CPU reading with the model resident and no AX client attached.


raullenchai:
**Full `swift test` on the Studio: the 4 red tests are host-load flakes, not this branch.**

`SSEDeltaCoalescer` ("#1743 … hops < 12", "flush after content+reasoning …") and `MermaidTests` ("A timed-out render/readiness probe …", 2 s budgets) were run alternately on this branch and on the untouched base commit `023b906c2` in a second worktree, same host, load average 4.9–9.5 (other desks building):

| cycle | base `023b906c2` | this branch |
|---|---|---|
| 1 | pass | pass |
| 2 | pass | 3 issues |
| 3 | 1 issue | 3 issues |
| 4 | 3 issues | 1 issue |
| 5 | 3 issues | pass |

Same tests, same assertions on both sides (`hops → 15–32`, `2.5 s > 2 s`, `readinessAttempts → 1`). None of them references code touched here. This is the known "full_unit red on a loaded dev machine" shape (#3340); CI runs them on an idle runner.

raullenchai:
**Codex adversarial self-review (gpt-5.6-sol), two rounds before pr_validate.**

Round 1 (2 BLOCKING / 3 MAJOR / 4 MINOR) — all addressed in the first commit:
- guard ran after `PortSweep.startLaunchSweep`, which would have reaped the running instance's sidecar → moved to the top of `RapidApp.init` (`SingleInstanceGuard`), ordering pinned by test;
- min-other-PID rule could make both racers yield → seniority by launch date;
- `activate()` cannot reopen a closed window → LaunchServices reopen via `openApplication(at:)`;
- roster gate: `local_run` is advertised by default, substring matching invented capabilities, opaque MCP names were silenced → exact-name table for built-ins, word-based classification otherwise, unclassified keeps the caption;
- tripwire test, race test, CHANGELOG wording.

Round 2 (2 BLOCKING / 3 MAJOR) — second commit:
- mixed known/unknown launch dates made `outranks` non-transitive → total order (unknown date ranks last), test from all three viewpoints;
- hand-off discarded the `openApplication` error and exited even if the survivor had just quit → `handOff` returns false when the survivor is terminated and the launch continues; error with a live survivor falls back to `activate()`;
- `URLCalculator` split as one word → acronym-aware splitter;
- `read_document` only reads the user's attachments → serves no retrieval lane.
- One round-2 BLOCKING was a false positive: it claimed the init-order test matches `PortSweep.startLaunchSweep` inside a comment before the guard; the comment does not contain that literal and the test passes.

raullenchai:
pr_validate run 2 (`b44315cc`) codex findings, and what changed:

1. **[BLOCKING] `.yieldToUnregistered` exits even if the holder dies during the window** — real, and wider than stated: any yielder that hands off to a survivor which is in the middle of quitting exits too, and the survivor then finishes quitting. Reproduced on the dogfood mini: `quit` + exec of the binary 0.2 s later → zero Rapid processes. Fix in `4ac1cd95`: `SingleInstanceGuard.yieldsLaunch()` hands off, then waits up to 5 s for the instance lock (kernel-released when the holder dies) and carries on as the Desktop if it gets it. Verified: same sequence now ends with one new PID, main window, sidecar.
2. **[BLOCKING] completion handler mutates a captured `var`** — now an `OSAllocatedUnfairLock<Error?>`.
3. **[NIT] `repeatForever` tripwire is a same-line substring test** — kept as is; it is a tripwire for new spellings, and the two allowed sites are the `BreathingLoop` view and the mascot, both reviewed here.

Two more launch defects fell out of verifying this on the mini and are in the same PR (F-7 Settings-only restoration launch with no engine; F-8 is item 1). Details in the PR body.

raullenchai:
pr_validate run 3 (`801def07`) codex findings:

1. **[BLOCKING] roster gate on attachment-grounded turns** — taken. On a turn that carried a document the tool that should have run is `read_document`, which none of the three capability lanes describes, so the roster gate could hide a caption Gate 1c had deliberately kept (live-data / retrieval prompt beside a PDF). Fix in `02cf2bd6`: Gate 6 is not consulted when `promptHadAttachment` is set; attachment turns keep exactly the pre-gate behaviour. Test added (live-data + retrieval prompts beside an attachment stay captioned under the production roster; literal arithmetic beside one gives the pre-gate answer).
2. **[NIT] `filesystem_search` / `code_search` read as retrieval/network** — this is the safe direction, not the risky one: a tool classified as able to serve a lane makes the caption **appear**, not disappear. A local search tool wrongly counted as "network" means a live-web question under it still wears the caution, which is the pre-PR behaviour. Left as is; the doc comment on `advertisedToolCouldServe` spells out that erring towards "servable" keeps the caption.

Codex self-review round 3 on the guard rework (before this run) is summarised in the PR body under Verification; the one substantive item — an established Desktop with no reportable launch date losing the snapshot rule when the lock is unavailable — is fixed in `801def07` (an instance that has finished launching wins outright, with a test).

raullenchai:
pr_validate run 4 on `02cf2bd6` (`--base 023b906c2`): every gate PASS including codex ("no blocking issues"), except `full_unit`: 3 failed / 25690 passed.

The three failures are host state on the Studio, not this PR — which touches only `apps/rapid-mac`:

- `test_lazy_package_init::test_version_fallback_when_package_metadata_is_missing` — `rapid_mlx.__version__ == "0.0.0"` in this interpreter (no installed dist metadata for the worktree checkout).
- `test_legacy_prefix_cache_flag_warnings` ×2 — `serve` bails with "Try a different port … --port 8001": another operator's process (pid 9097) is listening on 127.0.0.1:8000 on this host.

Control: the same three tests fail identically on a detached worktree of the untouched base commit `023b906c2` (3 failed, 13 passed). GitHub `validate`, `build`, `desktop-tests`, `version-bump-guard`, `accessibility-identifier-tests` and `gui-harness-contracts` are all green on this head.

raullenchai:
Queue failure diagnosis from candidate #3572 / run 35411529281: this is deterministic contract drift within the PR, not a product failure and not a hosted-Mac flake.

The live AX snapshot after Quickstart completion contains exactly one welcome line with the intended straight apostrophe:

`You're chatting with fake-alias, running entirely on your Mac.`

The golden-flow assertion still counts the pre-change smart-typography spelling `You’re` (curly apostrophe), so it reports zero and fails with “Quickstart welcome was not seeded exactly once after confirmation.” That assertion must move with the PR's explicit straight-quotes behavior. I will include the scoped harness correction while restacking this conflicting branch onto current main; no product behavior expansion is required.

raullenchai:
Exact-head update after diagnosing the prior candidate failure:

- Rebased the scoped fix onto main `c776acb91745360f2f6e7e07ee4d04d487598905`; new head: `c81a8dd6d65ae0b24daf498345a55d0db0f11568`.
- Root cause: the product intentionally emits a straight apostrophe after smart typography was disabled, while the golden assertion still searched for the curly form. The candidate artifact contained the correct message exactly once.
- Fix: update only that assertion to match the product contract (`You\u0027re`) and document why.
- Verification: `bash -n`, direct `jq` proof against the failed candidate artifact (= 1), `git diff --check`, and 72 focused Swift tests across the four affected suites all pass.
- Scope-locked adversarial self-review found no P0/P1 issue. No product behavior or policy changed.

I am waiting for the new exact-head required checks before authorizing this head for the Mac queue.

raullenchai:
@mergifyio queue mac-batch

Issue #3564, opened 2026-09-18
Server-side inference errors (esp. OOM) not faithfully surfaced to the GUI


raullenchai:
## Summary

Major server-side inference errors do not reach the Desktop GUI faithfully. A generation-time out-of-memory abort on the MLLM (vision) serialized lane surfaces to the user as the generic *"Rapid couldn't finish that request. Try again, or restart the model."* — the exact wrong recovery for a persistent OOM (retry fails identically), and it hides the real cause.

Reproduced on an 18 GB Mac running `bonsai2-27b-2bit` (0.14.3): the model loads (Resident panel shows 14.76 / 14 GB — over budget), then `POST /v1/chat/completions` returns **500 Internal Server Error** and the failure card shows the generic message.

## Root causes

1. **Engine — streaming preflight leaks a raw 500.** `routes/chat.py` `_preflight_mllm_chat_stream` runs before `StreamingResponse` commits, so an `InferenceAbortedError` raised during priming propagates out of the route. The `try/except` around the preflight catches only `ClientRequestError` (→400); `InferenceAbortedError` escapes to the generic `Exception` handler → `{"error":{"message":"Internal server error"}}` (500). The **non-streaming** path already maps `InferenceAbortedError`→503, so the two lanes disagree.
2. **Engine — the abort cause is not machine-readable.** `mllm_scheduler._fail_all_inflight` collapses every underlying step exception (including Metal/MLX allocation failures) into one opaque, deliberately-sanitized string with `error_kind="lifecycle"`. The real category (OOM vs generic transient) is logged server-side but never carried to the client — by design the raw message is withheld, so **the client has no signal to classify on.**
3. **GUI — classifies by substring, not by code.** `FailureDiagnoser.chatFailureKind` keyword-matches the (already-sanitized) body against a 3-entry OOM list. Since the engine replaced the body with `"Internal server error"`, no keyword can match — the OOM is unrecoverable at the GUI. Falls through to `.requestFailed` (the generic card).

## Fix — structured error-code contract, engine → GUI

The GUI deliberately renders only curated per-`Kind` copy (never raw server text), so faithfulness must come from a **stable, machine-readable `error.code`**, not from leaking server strings.

[...]

PR #3571, opened 2026-09-19
fix(engine,mac): carry stable error codes for the remaining major server failures

raullenchai:
## Why

#3564 established the engine→GUI error-code contract — the engine stamps a stable `error.code` in an OpenAI-shaped `{message, type, code, param}` envelope, and the Desktop maps `code → FailureDiagnosis.Kind → curated card`. But only the generation-time OOM / engine-abort path was actually wired.

A live dogfood (a serve built from the #3564 branch + real error probes, then feeding the captured payloads through the real Swift `FailureDiagnoser`) showed the **other** major failures still reaching the chat UI as the generic **"Rapid couldn't finish that request"** card with a misleading **Retry** — because their bodies carried no code (or, for model-replacement, a bare-string 503).

## What now reflects faithfully

| Failure | Engine code | GUI kind | Recovery copy |
|---|---|---|---|
| Prompt / conversation past the context window | `context_length_exceeded` | **promptTooLong** | shorten / new chat / larger-context model — never a retry that fails identically |
| Unknown model on the chat route | `model_not_found` | **modelUnavailable** | open Model Management |
| Load-time memory shortfall | `insufficient_memory` | **modelOutOfMemory** | free memory / smaller model |
| Any other load failure | `model_load_failed` | **modelLoadFailed** | check files / choose another |
| Request cancelled by a model swap | `model_replacement` | **requestSuperseded** (a calm *notice*, not a red fault) | ask again |

## Engine

- Factor the model-replacement envelope into a single `lifecycle_cancel_error_payload()` shared by the abort lane, the `asyncio.CancelledError` translation, and the guided-chat cancel (these had drifted into two shapes — one with the code, one a bare string).
- Add `model_load_error_payload()` — a load-time memory shortfall reuses the OOM card; everything else is `model_load_failed`.
- Harden `classify_engine_abort()` so a canonical `MemoryError` (host RAM exhausted materialising weights) and a POSIX `OSError(ENOMEM)` (an mmap/allocation failure while loading) — neither of which carries allocation-failure *wording* to substring-match — are recognised by type/errno as `insufficient_memory`, not misfiled as `model_load_failed`.
- The load-failure `503` keeps its standard, uniform `Retry-After` (a `503` means "temporarily unavailable", and a load fault can be transient — a backend-init race, a passing I/O error — as readily as permanent, which we cannot tell apart from the exception). What steers the user is the **code** (OOM card vs "check files / choose another"), not the header.
- Give the chat unknown-model 404 a dict envelope carrying the code (was a bare-string detail, so no code survived to the GUI). It echoes **only the client's own requested id** (OpenAI-standard) — deliberately **not** the served-model list, which in single-model mode is `cfg.model_name` and can be a local filesystem path (clients enumerate models via `/v1/models`).

Every user-facing message is a fixed, safe string — the raw exception is inspected inside the trust boundary to pick the code but is never returned, so no paths or prompt fragments leak.

## GUI

- Add `promptTooLong` / `modelUnavailable` / `requestSuperseded` kinds with curated copy, severity (`requestSuperseded` is a notice), icons, and `code → kind` mappings.
- Complete every exhaustive switch over `Kind` (`legacyPersistedKind`, `severity`, `diagnosis(for:)`, `iconName`) and admit `requestSuperseded` to the enumerated notice lane.
- Make the chat transcript's failure caption severity-aware: a `.notice` (a model-swap supersede) is tinted secondary instead of the red error lane — mirroring how the tool-failure chip already treats a notice — so only genuine faults read red.

## Tests

- Python: the new payloads, the classified load-failure 503, the canonical `MemoryError` / `OSError(ENOMEM)` → `insufficient_memory` classification (with the non-`ENOMEM` `OSError` negative control), the chat model-replacement lane, and the no-raw-leak invariant (including the server-path leak guard on the 404).
- Swift: extends the faithful-diagnosis suite with the three new `code → kind → card` mappings; `DeclinedToolDiagnosisTests` admits `requestSuperseded` to the notice allow-list.

## Follow-up (out of scope here)

The chat transcript surfaces a failure as its curated **message** plus the shared Copy / Retry / Delete controls; it does **not** render a per-kind *action button* (e.g. "Open Model Management" for `promptTooLong` / `modelUnavailable`, in place of the generic Retry). That is true for **every** kind today, including the already-shipped `modelOutOfMemory` from #3564 — a per-kind action surface in the chat transcript is a separate UX change that touches shared rendering for all kinds, so it is deliberately left as a follow-up rather than expanded into this codes-and-messages PR. The curated message already names the correct recovery, and `FailureDiagnosis.action` is defined for the new kinds so the follow-up can consume it.

## Verification

`ruff check` + `ruff format` clean; 100% diff coverage on changed `rapid_mlx` lines (CI `changed-lines-coverage` gate green); `swift build` + the touched Swift suites green; full GitHub CI green (all `test-matrix` shards, `build`, `desktop-tests`, `test-apple-silicon`, `changed-lines-coverage`); standalone `codex exec` adversarial review returned no in-scope blocking findings.

Refs #3564


raullenchai:
Queue failure diagnosis: the merge candidate did not expose a product regression in this PR. Candidate run 35414713995 failed only `Cancelling a gh star request terminates and reaps its child` because the test waited for the PID file to exist, while shell redirection creates that file before its `echo` payload is written. The observed `pids.count == 0` is exactly that TOCTOU window.

The scoped CI fix is #3615 (issue #3614). This PR's exact head has otherwise been adversarially re-reviewed with no P0/P1 findings, and 233 focused engine/server error-path tests pass locally. I will re-authorize this same head after #3615 lands so the candidate cannot be dequeued by the stale test contract again.

raullenchai:
@mergifyio queue mac-batch

PR #3582, opened 2026-09-19
refactor(mllm): vendor the prepare_inputs surface (step 2c)

raullenchai:
## Why

Replacement for #3566, which GitHub auto-closed when its stacked base and head
temporarily became identical during the #3563 restack. This PR preserves that
PR's authorship, review history, and exact intended delta on current `main`.

Step 2c of the mlx-vlm dependency-retirement plan, following #3563. The two
lane-facing call sites still imported `mlx_vlm.utils.prepare_inputs` for every
request, leaving request preprocessing outside Rapid-MLX's reviewed runtime
boundary. This PR vendors that pinned helper closure and redirects only those
two call sites.

## What lands

- `mlx_vlm_vendored/inputs.py`, based on the unbroken `load_image` through
  `prepare_inputs` region from pinned mlx-vlm 0.7.1.
- Mechanical provenance checks for byte-identical functions and module-level
  constants, plus fail-closed hashes for both the upstream region and the
  complete reviewed vendored module.
- Four documented, behavior-tested deviations:
  - normalize an upstream `VideoSampling` result into the vendored namespace;
  - always release the native video capture handle;
  - decode bytes filesystem paths instead of stringifying them;
  - close streamed audio responses.
- Redirect `MultimodalProcessor` and `MLLMBatchGenerator` to the vendored
  `prepare_inputs` implementation.
- Keep the Qwen3.6 transitional text cache recognizable as MLX-LM, upstream
  VLM, and vendored VLM `ArraysCache` simultaneously. This is necessary while
  the already-loaded language module still performs upstream nominal type
  checks and the new engine performs vendored checks.

## Scope and non-goals

This PR changes only the ownership and namespace of the existing preprocessing
surface, plus the four bounded resource/namespace fixes above. It does not
change request schemas, sampling policy, model outputs, public APIs, cache
policy, or remove the remaining `mlx_vlm.models.base` dependency.

The vendored module is private and is not a new raw-media ingress surface.
Both Rapid call sites resolve user image/video URLs through the existing
SSRF-guarded, size-bounded media layer and pass immutable local snapshots to
`prepare_inputs`; neither supplies `audio=`. The copied upstream URL/audio
helper branches remain unreachable from Rapid's public request paths in this
migration. Changing their network or ffmpeg semantics belongs to the later
audio/processor cutover, not this namespace-only step.

## Design precedent

The migration follows the existing staged namespace cutover: consumers accept
both old and new identities before producers switch. Vendored source is pinned
to an exact reviewed upstream region; intentional deviations are sentinel
marked, behavior tested, and covered by a complete-source hash so unrelated
edits cannot hide inside an excluded large function.

## Adversarial review disposition

The scope-locked review found and fixed material gaps beyond the original
branch, and explicitly dispositioned adjacent findings against the pinned
call graph:

1. The four functions containing intentional deviations were excluded whole
   from per-function parity checks, so unrelated drift inside their large
   bodies could pass. The test now pins the upstream region, the complete
   vendored file, and the exact deviation-sentinel count. Direct tests also
   prove response and capture cleanup on failure paths.
2. Replacing Qwen3.6's upstream VLM cache parent with only the vendored parent
   broke two established nominal-type contracts: the upstream singleton-row
   extractor converted the cache to a KV cache. The bridge now retains both
   VLM namespace identities; the previous contract and the new vendored
   recognition are asserted together.
3. A later static review flagged unrestricted URL fetches and ffmpeg fallback
   behavior inside copied upstream helpers. Call-site tracing confirmed that
   Rapid's two consumers pre-resolve remote image/video inputs through the
   guarded media layer, pass only local snapshots, and never invoke the audio
   branch. These are inherited, unreachable helper semantics rather than a
   regression or public surface added by this PR, so they are explicitly held
   out instead of expanding this namespace migration into a media-security
   redesign.
4. The deviation inventory originally counted sentinel text inside the module
   docstring while the response-close hunk lacked a sentinel at its actual
   location. It now tokenizes Python comments only, asserts the exact label
   distribution, and pins the resulting complete-source hash.
5. A nominal-identity concern for `VideoMetadata` was checked against every
   processor in pinned mlx-vlm 0.7.1. Unlike `VideoSampling`, no processor
   imports or nominally checks that class; the consumers use attributes or
   dictionaries. A dual-namespace inheritance bridge would therefore invent
   a constraint rather than preserve one, so no such bridge is added.

No adjacent request-security or processor redesign was added.

## Verification

- Scope-selected preprocessing, image, corruption, batch-generator, Qwen3.6
  cache, provenance, and CI-contract suites: **307 passed, 3 deselected**.
- Continuous batching, media-prefix-cache, and headless engine lifecycle:
  **228 passed, 3 deselected**.
- Ruff and `git diff --check`: clean.
- Exact-head GitHub `tests`, `desktop-tests`, `version-bump-guard`, type-check,
  and Apple Silicon checks will be required before queue admission.

Co-Authored-By: Claude Code <noreply@anthropic.com>


raullenchai:
## Scope-locked review disposition

Replacement restack review is converged on exact head `20837edf`.

- Fixed a real provenance gap: the four deviated large functions could previously drift outside the per-function parity probe. The gate now pins the upstream region, complete reviewed vendored source, and exact deviation count; resource-cleanup failure paths are directly tested.
- Fixed a real Qwen3.6 compatibility regression: replacing the upstream cache identity outright made 2/31 established cache tests fail. The transition bridge now preserves MLX-LM + upstream VLM + vendored VLM nominal identities simultaneously.
- Latest adversarial `pr_validate`: **MERGE-SAFE**, Codex **no blocking issues**, 224 targeted tests passed.
- Post-restack focused verification: **391 passed, 3 deselected**; Ruff and `git diff --check` clean.

Scope remains locked to vendoring `prepare_inputs`, redirecting its two lane call sites, and the four documented resource/namespace deviations. No adjacent processor, request-security, or cache-policy redesign was added.

raullenchai:
## Exact-head CI failure disposition (`20837edf`)

The first exact-head run failed only `changed-lines-coverage` (and therefore the `tests` aggregator): all Linux shards and Apple Silicon tests passed. This was a real PR defect, not a flake.

Root cause: the 888-line pinned upstream input closure was measured as newly owned production code even though its integrity is enforced by upstream-region + complete-vendored-source hashes; the second redirected call site in `MultimodalProcessor.process` also lacked a direct wiring test.

Fixed on new head `8792ee0a`:
- add the exact vendored `inputs.py` file to the same narrow coverage exclusion used for pinned `cache.py`/`apc.py`, with the source-hash and deviation-test invariant documented;
- add a direct `MultimodalProcessor.process` contract proving it resolves and calls the vendored `prepare_inputs` closure.

Focused verification: 110 passed, 3 deselected; Ruff and `git diff --check` clean. No failed job was rerun; the fix produced a new head.

raullenchai:
Exact-head follow-up: CI exposed one deterministic contract omission after adding the file-scoped `inputs.py` coverage exclusion. Commit `978577652` updates `test_vendored_mllm_coverage_omit_is_file_scoped` to require exactly `cache.py`, `apc.py`, and `inputs.py`; future vendored modules and Rapid-owned seams remain forbidden. Focused result: 15/15, Ruff clean, diff clean. No failed job was rerun; the new head is receiving fresh exact-head checks.\n\nThe latest scope-locked static review also flagged inherited URL/audio helpers. Call-site tracing shows both Rapid consumers pre-resolve public image/video URLs via the SSRF-guarded, size-bounded media layer and pass local snapshots; neither supplies `audio=`. Those branches are not a new or reachable public ingress in this namespace migration, so the findings are dispositioned rather than expanding this PR into the later media-security/audio cutover.

raullenchai:
Review follow-up on `VideoMetadata`: inspected every pinned mlx-vlm 0.7.1 processor consumer. None imports or performs a nominal `VideoMetadata` check; GLM5-Next, MiniMax M3 VL, Mage-VL, and Nemotron consume attributes/dicts. This differs from `VideoSampling`, where a real nominal boundary exists and is already bridged. The finding is therefore dispositioned as unsupported by the pinned call graph rather than adding an unnecessary multi-namespace inheritance constraint. The provenance finding was real and is fixed in `2c4476743`: sentinel inventory now tokenizes comment tokens, validates the exact label distribution, and marks the response-close hunk at its actual location.

PR #3578, opened 2026-09-19
refactor(mllm): vendor the speculative-decode coordinator core (step 3b)

raullenchai:
## Rationale

Step 3b of the mlx-vlm dependency retirement plan (`docs/engineering/design/2026-09-18-vendor-mllm-primitives.md`): vendor the speculative-decode coordinator core — the engine behind the dflash and native-MTP lanes — plus the model foundations (`models/{base,linear}`, `fp8`, `quant_utils`) that the drafter/model-class slices (3c) build on. 3a landed the text-AR core in #3575; this slice flips `ar`'s speculative import from pinned upstream to the vendored package.

Stack: #3534 → #3545 → #3554 → #3558 → #3563 → #3566 → #3575 → **this PR**.

## What lands (upstream 0.7.1, verbatim unless noted)

| Vendored file | Upstream source | Deviations |
|---|---|---|
| `speculative/cache_state.py` (299) | `mlx_vlm/speculative/cache_state.py` | 1 redirect: `..models.cache` → vendored root |
| `speculative/common.py` (438) | `mlx_vlm/speculative/common.py` | none — byte-identical |
| `speculative/ddtree.py` (164) | `mlx_vlm/speculative/ddtree.py` | none — byte-identical |
| `speculative/dflash.py` (670) | `mlx_vlm/speculative/dflash.py` | none — byte-identical |
| `speculative/mtp.py` (1,208) | `mlx_vlm/speculative/mtp.py` | 1 redirect: `..models.quantized_verifier` → pinned upstream (2k-line verifier not vendored; `decode_quantized_argmax` is a pure array function) |
| `speculative/utils.py` (439) | `mlx_vlm/speculative/utils.py` | 1 redirect: `.eagle3` → pinned upstream (eagle3 backend not vendored — Rapid serves dflash/mtp kinds; round helpers are pure-array) |
| `speculative/__init__.py` | (not verbatim) | `VENDOR-DEVIATION(subset-exports)`: reduced shim — upstream re-exports `load_drafter`; the drafter registry + concrete drafters land in step 3c |
| `models/base.py` (663) | `mlx_vlm/models/base.py` | redirects: `..turboquant` → pinned upstream (kv_quant precedent), `.cache` → vendored root |
| `models/linear.py` (74) | `mlx_vlm/models/linear.py` | none — byte-identical |
| `fp8.py` (131), `quant_utils.py` (222) | same-named upstream files | none — byte-identical (mlx + stdlib only) |

[...]

PR #3587, opened 2026-09-20
refactor(mllm): vendor the drafter registry and served drafters (step 3c-1)

raullenchai:
# Step 3c-1 — vendor the drafter registry and served drafters

Stacked on #3578 (step 3b, head `72cff6488`). Base branch:
`refactor/vendor-mllm-spec-core`.

Why now: the drafter registry is the load-path entry the speculative
lanes resolve through, and vendoring it unlocks the 3d dispatch flip —
because without a vendored `load_drafter` the generate/dispatch modules
have no in-repo seam to bind.

## Scope

Verbatim from `mlx_vlm/speculative/drafters` @ 0.7.1 (step-3c slice 1 of 3):

- `__init__.py` — the registry: `load_drafter`, `resolve_drafter_kind`,
  `validate_drafter_compatibility`, kind tables
- `mtp_base.py` / `mtp_split.py` / `compatibility.py`
- served MTP families: `glm5_next_mtp/`, `qwen3_5_mtp/`
- served DFlash families: `qwen3_dflash/`, plus the generic `dflash2/`
  (the algorithm the Rapid DFlash lane reports)

Not vendored here (documented pinned redirects, flipped by later slices):

- `dspark` / `laguna_dflash` / `muse_glimmer_assistant` classes (registry
  imports → pinned; unserved families)
- `mlx_vlm.utils` `get_model_path`/`load_model` (step 3e)
- `models.{cache,activations,rope_utils}` (cache slice)
- `models.glm5_next.{config,language}` → step 3c-2
- `models.qwen3_5{,_moe}.{config,language}` → step 3c-3

## Binding

`speculative/native_mtp/runtime.py` and `glm5_compat.py` now bind the
vendored registry (vendored-first). The availability guard still requires
the qualified pinned runtime — the load path (`get_model_path`/`load_model`
and the model-family classes) stays pinned during the transition. The GLM
compat shim's class swap uses `setattr` (noqa B010): the vendored package
is typed, and the shim intentionally swaps the class object.

## Test plan

- [x] `tests/test_mlx_vlm_vendored_drafters.py` — byte-parity probes with a
  redirect-reverting normalizer (fail closed on any undocumented edit),
  registry table equality with pinned, served-family auto-detection,
  unknown-kind rejection, runtime-binding probe
- [x] fake seams in `test_native_mtp.py` / `test_glm5_mtp_compat.py` follow
  the 3a precedent: fakes shadow the vendored module names now
- [x] mypy error budget — no growth (9 new vendored files grandfathered:
  47 errors across the drafter sources; `glm5_compat`/`runtime` typing
  fixed in place rather than baselined)
- [x] ruff check + format clean
- [x] codex r1/r2 findings fixed: dual-registry glm5 adapter swap, served
  DFlash kind mappings (documented upstream-bugfix), safetensors-index
  shard confinement (documented upstream-bugfix), hermetic compat fakes
- [x] mypy budget re-verified after the r1 binding fix (866, no growth)
- [ ] `pr_validate` codex loop

## Slicing context

Measured closure sizes vs the 200 KB codex diff cap: glm5_next family
98 KB (3c-2), qwen3_5 + qwen3_5_moe 133 KB (3c-3), cache.py ~125 KB
(later slice). Each later slice flips this PR's redirects to vendored.


raullenchai:
**Loop r4 response — finding 1 (scalar `_next_position` skipped at qwen3_5_mtp.py:360): addressed in `4f13b9f14`.**

The cited site now promotes the scalar before the correction (vendored `qwen3_5_mtp.py`, replay-finalize block):

```python
padding = mx.array(right_padding, dtype=mx.int32)
if isinstance(self._next_position, mx.array):
    self._next_position = self._next_position - padding
elif int(padding.min()) == int(padding.max()):
    self._next_position = self._next_position - int(padding.min())
else:
    self._next_position = mx.full(
        (len(right_padding),), self._next_position, dtype=mx.int32
    ) - padding
```

- rows with **different `right_padding`** → the `else` branch promotes the scalar to a per-row array and subtracts the padding vector — exactly the reviewer's scenario (uniform initial trims leave the scalar untouched, so the promotion happens here, at the only site the scenario reaches);
- uniform padding → exact scalar/int correction in place;
- array positions → subtract unconditionally (as before);
- all-zero padding → no correction needed (skipped by the enclosing `if any(right_padding)`).

Regression probe: `tests/test_mlx_vlm_vendored_drafters.py::test_qwen_mtp_batch_replay_corrects_scalar_position_for_ragged_rows` builds accepted=[2,1] → lengths [3,2] → `right_padding` [0,1] with a scalar `_next_position=7` and asserts the post-call value is `[7, 6]` (the old code would leave the scalar `7`). The deviation is ledgered in the package inventory as an upstream-bugfix, and the byte-parity probe reverts the hunk and still matches pinned 0.7.1.

If the reviewer still sees a gap, please cite a concrete interleaving (accepted/keep_appended/new_tokens/`_next_position` type) that reaches the cited site with heterogeneous padding and remains uncorrected.

raullenchai:
**Loop r9 response — all four findings are the r8 list verbatim and are already fixed in `07ed0d9ab`.**

The reviewed diff contains the fixes (the review's line numbers match the pre-fix file):

1. **`qwen3_dflash/dflash.py` `bind()`** — the vendored method now force-clears `self.embed_tokens = None` before the resolution guard (with the documented-upstream-bugfix comment), so every `bind()` re-resolves the embedding module and scale from the supplied target. Regression probe: `test_dflash_bind_re_resolves_target_embeddings` binds a stale-embed drafter to a new target and asserts `drafter.embed_tokens is new_embed`.

2. **`mtp_base.py` `accept_verified_tokens_batch`** — mixed bonus presence now raises `ValueError("mixed MTP bonus-token presence across replay rows is unsupported; ...")` before any cache mutation (the reviewer's option A). Probe: `test_mtp_base_rejects_mixed_bonus_presence`.

3. **`mtp_base.py` `draft_block`** — `block_size <= 1` now returns the DFlash2-shaped empty proposal (`mx.zeros((batch, 0), token_dtype)`), matching `dflash2.py:174-176`, before any concatenate. Probe: `test_mtp_base_block_size_one_returns_empty_proposal` asserts shape `(1, 0)`.

4. **runtime-binding test** — replaced with a functional probe: distinct fake `load_drafter` callables are installed on the vendored registry and the pinned `mlx_vlm.speculative.drafters`, and `load_runtime()` must hit the vendored seam (`VENDORED-SEAM-CALLED`) and never the pinned one. See `test_runtime_binds_vendored_registry`.

All four changes carry documented upstream-bugfix ledger entries in the package inventory, and the byte-parity probe reverts each deviation hunk and still matches pinned 0.7.1. If a finding still stands, please cite the current head's line and a concrete interleaving the probe misses.

raullenchai:
**Loop r11 response — all three findings repeat r10 verbatim and are already fixed in `1ac91f47f` (the reviewed diff contains the fixes; the quoted line numbers are from the pre-fix file).**

1. **`qwen3_5_mtp/config.py` `TextConfig.from_dict`** — the vendored method now routes the Qwen3-Next family to `MoeTextConfig`:

```python
model_type = params.get("model_type", "")
is_moe = "moe" in model_type or model_type.startswith("qwen3_next")
text_config_cls = MoeTextConfig if is_moe else DenseTextConfig
```

Probe: `test_qwen35_text_config_routes_qwen3_next_to_moe` asserts `qwen3_next` → `MoeTextConfig` and `qwen3_5` → `DenseTextConfig`.

2. **`mtp_base.py` `accept_verified_tokens_batch`** — the mixed-presence `ValueError` now sits immediately after the uniform-acceptance check and **before** `accepted_i`/`keep_appended`/trim/`_next_position` mutation. Probe updated per the review's spec: `test_mtp_base_rejects_mixed_bonus_presence` uses `_round_appended = 1`, a recording cache, and asserts `cache.trimmed is None`, `_next_position == 5`, `_round_appended == 1` after the raise.

3. **`qwen3_5_mtp/qwen3_5_mtp.py` `draft_block`** — the served drafter now returns the DFlash2-shaped empty proposal for `block_size <= 1` before its loop (`mx.zeros((batch, 0), token_dtype)`), mirroring the guarded base drafter. Probe: `test_qwen_mtp_draft_block_one_returns_empty_proposal` asserts shape `(1, 0)`.

All three are documented upstream-bugfix deviations in the package inventory; the byte-parity probe reverts each hunk and still matches pinned 0.7.1. If a finding still stands against head `1ac91f47f`, please cite the current line and a concrete failing interleaving.

raullenchai:
## Review ran against a stale diff (head 990a5c66f, not d30272b8f)

Both findings are verbatim repeats of the round-14 findings and were already fixed in commit **d30272b8f**, now the PR head.

**Evidence the findings no longer apply to the current head:**

1. **Stale-binding skip (drafters/`__init__.py:69`)** — the current `install_served_architecture_bindings()` no longer skips existing entries; it re-binds whenever `Model`/`ModelConfig` do not match the vendored classes (lines 67-81 of the head):

```python
package = importlib.import_module(f"{__name__}.{model_type}")
existing = sys.modules.get(target)
if (
    existing is not None
    and getattr(existing, "Model", None) is package.Model
    and getattr(existing, "ModelConfig", None) is package.ModelConfig
):
    continue
shim = ModuleType(target)
setattr(shim, "Model", package.Model)  # noqa: B010
setattr(shim, "ModelConfig", package.ModelConfig)  # noqa: B010
sys.modules[target] = shim
```

A pinned module imported earlier — or a GLM shim bound before the compatibility swap — is replaced on the next `load_drafter` call. Covered by `test_binding_hook_replaces_stale_architecture_bindings` (pre-registers a stale `mlx_vlm.models.dflash2` with `Model = object`, then asserts the re-bound module's `Model`/`ModelConfig` modules are vendored).

2. **GLM probe bypassing the hook (test_native_mtp.py:963)** — the current test deletes any cached `mlx_vlm.models.glm5_next_mtp` entry (`monkeypatch.delitem(..., raising=False)`) so the production hook runs, and the fake loader resolves the class through the hook-installed shim exactly like the documented pinned dispatch:

```python
utils.load_model = lambda path, **kwargs: sys.modules[
    "mlx_vlm.models.glm5_next_mtp"
].Model.__new__(sys.modules["mlx_vlm.models.glm5_next_mtp"].Model)
```

It then asserts the dispatch produced the Rapid adapter: `sys.modules["mlx_vlm.models.glm5_next_mtp"].Model is type(drafter)` and its `__module__` starts with `rapid_mlx.`.

**Freshness check:** `git ls-remote origin refactor/vendor-mllm-drafters` → `d30272b8f2a7`; the reviewed `pr.diff` in run `run-20260920T094629Z` contains neither `existing = sys.modules.get` nor the new probe asserts — it is the pre-r14 diff. Please re-review against the current head. Local gates at d30272b8f: 105 tests passed, ruff clean, mypy budget 866/163 with no growth.

Issue #2252, opened 2026-08-23
CI throughput and quality: risk-routed PR gates, merge validation, and stronger coverage


raullenchai:
## Objective

Make agent-authored PRs reach required-green and land materially faster without weakening correctness, compatibility, release confidence, or regression coverage. Add more risk-relevant tests and coverage while removing duplicated or irrelevant work.

Success is not “run fewer tests.” It is:

- fast feedback on every push;
- focused blocking coverage for the change being reviewed;
- full validation of the actual merge candidate;
- broad model/hardware/product coverage on main, nightly, and release;
- explicit evidence explaining why each CI lane ran;
- red main and flaky tests having owners and operational consequences.

## Current bottlenecks and gaps

- [ ] Remove the mandatory `full-ci` label round-trip. Today a PR can pass its relevant gates, then be deliberately failed until labeling retriggers full engine/Desktop/model/GUI workflows against the same PR SHA.
- [ ] Use the existing `merge_group` path as the full composition gate so validation covers the PR combined with current main, not a stale PR head.
- [x] Split the current universal `lint` job: docs-only changes still run engine, parser, installer, release, and microbenchmark contracts.
- [ ] Expand CI classification beyond `engine / desktop / docs_only`; current broad buckets still trigger unrelated expensive lanes.
- [x] Make new type errors blocking with a shrink-only baseline ratchet (#2271).
- [ ] Make coverage actionable. Codecov upload is currently non-blocking and does not enforce changed-line coverage.
- [ ] Eliminate duplicate runs caused by label/unlabel events and standardize per-PR concurrency keys.
- [ ] Add explicit source-to-test and source-to-user-journey mappings, with fail-closed handling for unknown/shared paths.
- [ ] Define red-main admission control and flaky-test ownership.

### L0 — every PR push: fast universal contracts

- [x] Tested merge-base classifier
- [x] Format/lint appropriate to changed content
- [x] workflow expression and immutable-action checks
- [ ] secrets/generated/root-directory guards
- [x] stable aggregate checks that always exist
- [x] per-PR stale-run cancellation

Target: docs-only required-green in <=2 minutes.

### L1 — PR risk-selected blocking lanes

- [x] Python core/API/parser/model-profile/runtime lanes
- [x] MLX/Apple Silicon lane only when relevant
- [ ] selected real-model smoke by affected architecture/contract
- [x] Desktop build/unit/contract lanes when relevant
- [x] selected GUI user journeys when relevant
[...]

PR #3589, opened 2026-09-20
fix(ci): validate installed test dependency versions

### LINKED ISSUE #3422: ci: pr_validate accepts installed test dependencies below declared versions
## What happened

A local `pr_validate` run for CI-only PR #3418 spent 21m51s on the full suite and then failed `test_installed_audio_runtime_contains_silero_vad`. The Studio interpreter had `mlx-audio 0.4.3`, while `pyproject.toml` and `scripts/pr_validate/_test_env.py` require `mlx-audio>=0.5.3,<0.6`.

`test_env_check` nevertheless reported all packages importable because it verifies imports only, not whether the installed distributions satisfy the declared requirements. The exact-head hosted CI installed canonical dependencies and passed. An isolated 2.1 MB install of 0.5.3 made the focused test pass.

## Why this matters

The validator can spend 20+ minutes producing a false project failure even though it already has enough metadata to reject or repair the stale environment before tests start. This wastes operator time and obscures real regressions.

## Acceptance

- Parse and check the canonical test requirement specifiers, including environment markers, in the same interpreter used by `pr_validate`.
- If an installed distribution is missing or outside its declared range, use the existing safe auto-install path before tests; preserve the dependency-file supply-chain guard.
- With auto-install disabled, fail in `test_env_check` with the package, installed version, required range, and remediation.
- Add focused tests for below-range, above-range, marker-skipped, valid, and distribution-name/import-name cases.
- Do not loosen or skip the downstream full suite.

Part of #2252.

raullenchai:
## Why

`pr_validate` previously treated an importable package as healthy even when its installed distribution was outside the version range declared by the canonical `.[test]` extra. A Studio run therefore spent more than 20 minutes in the full suite before failing because `mlx-audio 0.4.3` was importable but did not satisfy `mlx-audio>=0.5.3,<0.6`.

## Intention and scope

User-visible goal: fail or safely repair a stale validation environment before expensive tests start.

Allowed scope:
- parse PEP 508 requirements and environment markers from `[project.optional-dependencies].test`;
- read installed distribution versions and marker inputs from the exact Python interpreter that will run pytest;
- retain the explicit import-to-distribution identity mapping needed for names such as `PIL` / `pillow`;
- route missing and out-of-range distributions through the existing trusted-pin-first recovery flow;
- improve operator diagnostics and executable contracts.

Non-goals:
- no downstream test skips or relaxed full-suite policy;
- no project dependency-range changes;
- no change to external-PR supply-chain fences or install ordering;
- no product/runtime behavior changes.

## Design precedent

The validator separates import packages from installed distributions, reads versions through standard distribution metadata in the target interpreter, and evaluates the project's canonical requirement objects against that interpreter's marker environment. Version ranges are no longer copied into the import roster. The existing trusted-pin-first recovery and dependency-file guard remain fail-closed.

## Acceptance criteria

- Below-range and above-range installed distributions fail before tests.
- Requirements whose environment markers do not apply are skipped.
- Valid versions pass, including import/distribution name mismatches.
- Auto-install uses the existing safe recovery path for version mismatches.
- With auto-install disabled, the result names the distribution, installed version, required range, target interpreter, and remediation.

## Verification

- `python3.12 -m pytest tests/test_pr_validate_test_env.py tests/test_pr_validate_runner.py -q` — 85 passed
- `python3.12 -m ruff check scripts/pr_validate/_test_env.py scripts/pr_validate/steps/test_env_check.py tests/test_pr_validate_test_env.py` — passed
- `git diff --check` — passed
- Live target-interpreter probe — 6/6 imports and canonical versions satisfied

Fixes #3422
Part of #2252


raullenchai:
Review disposition for the latest platform-roster finding:

The reported Linux regression was not present. `required_test_packages_for_platform()` has always filtered `mlx_vlm` and `mlx_audio` explicitly through `DARWIN_ONLY_TEST_IMPORTS`; it does not parse the distribution-name strings. The bare names are intentional because the exact target interpreter passes them to `importlib.metadata.version()`, while canonical PEP 508 markers are evaluated separately by `_active_test_packages()`.

I strengthened the inline contract and regression test on `44b3f90c2`: Linux excludes both Apple-only imports, Darwin includes both, and their metadata lookup names remain exactly `mlx-vlm` / `mlx-audio`. Focused suite: 54/54 passing. No production behavior was changed for this disposition.

